PR #2, opened 2017-04-14
Add initial code

Random-Liu:
This PR added the `cri-containerd` binary which connects containerd and start grpc server.
This PR also added the Godeps, note that for all dependencies shared by Kubernetes and containerd, we are using the containerd version.

Currently the CRI service is empty, we need to implement these functions one by one.

/cc @kubernetes-incubator/maintainers-cri-containerd @feiskyer @resouer


cmd/cri-containerd/cri_containerd.go
```diff
@@ -0,0 +1,44 @@
...
+limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/golang/glog"
+	"github.com/spf13/pflag"
+
+	"github.com/random-liu/cri-containerd/cmd/cri-containerd/options"
```
Random-Liu: This is my private repo, will fix this soon.

pkg/server/container_portforward.go
```diff
@@ -0,0 +1,30 @@
+/*
```
yujuhong: Technically this belongs to the sandbox operations.

Random-Liu: True. Will update.

Random-Liu: Done.

Random-Liu:
Merge based on the LGTMs.

Issue #3, opened 2017-04-14
Create a Makefile


Random-Liu:
We need a Makefile to manage the build, test, package and release process.

Random-Liu:
Close this one since #7 is merged.

PR #7, opened 2017-04-15
initial makefile

mikebrow:
Adds an initial makefile. Addresses issue #3 

Patterned after kubernetes-incubator/cri-o


.tool/lint
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+for d in $(find . -type d -not -iwholename '*.git*' -a -not -iname '.tool' -a -not -iwholename '*vendor*'); do
+	${GOPATH}/bin/gometalinter \
```
AkihiroSuda: This does not work when GOPATH contains multiple entries.

I think it is ok to assume PATH is set to contain gometalinter

mikebrow: works for me.. done

xlgao-zju:
How about using  https://github.com/kubernetes/repo-infra/tree/master/verify which can avoid code duplication.

mikebrow:
@xlgao-zju ok.. took a look at it. Here are my thoughts:
1) I prefer the options selected for gometaliner in this PR over the ones in https://github.com/kubernetes/repo-infra/blob/master/verify/go-tools/verify-gometalinter.sh. 
2) Not sure why we should run gometalinter with vet enabled and govet...
3) I see the need for boilerplate checking but not sure if we should be running that on every file. For example adding a copyright to a gofmt.sh that is verbatim from the examples might be overkill.

Maybe after repo-infra has aged a bit?

pkg/server/service.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
 
-	_ "github.com/containerd/containerd/api/services/content"
+	_ "github.com/containerd/containerd/api/services/content" // TODO
```
cmluciano: Missing context for TODO

mikebrow: context was described below.. none of the functions are implemented :-)

mikebrow: IOW "github.com/containerd/containerd/api/services/content" is the ToDo :)

Random-Liu:
> I prefer the options selected for gometaliner in this PR over the ones in https://github.com/kubernetes/repo-infra/blob/master/verify/go-tools/verify-gometalinter.sh.

> Not sure why we should run gometalinter with vet enabled and govet...

> I see the need for boilerplate checking but not sure if we should be running that on every file. For example adding a copyright to a gofmt.sh that is verbatim from the examples might be overkill.

That's how kubernetes repo is doing it, every source file has a boilerplate. Let's follow the same pattern in this repo. :)

.tool/lint
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+for d in $(find . -type d -not -iwholename '*.git*' -a -not -iname '.tool' -a -not -iwholename '*vendor*'); do
```
Random-Liu: Why use not? If we only care about go files, I believe most of our go files will be in `cmd` and `pkg`.

How about `find . -type d -a \( -iwholename './pkg*' -o -iwholename './cmd*' \)`?

.tool/lint
```diff
@@ -0,0 +1,20 @@
...
+	gometalinter \
+		 --exclude='error return value not checked.*(Close|Log|Print).*\(errcheck\)$' \
+		 --exclude='.*_test\.go:.*error return value not checked.*\(errcheck\)$' \
+		 --exclude='duplicate of.*_test.go.*\(dupl\)$' \
```
Random-Liu: Same here. For my education, what kind of error are we trying to exclude here?

.tool/lint
```diff
@@ -0,0 +1,20 @@
...
+		 --exclude='vendor\/.*' \
```
Random-Liu: What kind of error are we trying to exclude here? Just for my education.

mikebrow:
@Random-Liu yes, if vet is on for gometalinter it runs go vet.

Ok will follow the same pattern. Will get to that this evening.

Makefile
```diff
@@ -0,0 +1,84 @@
+GO ?= go
+EPOCH_TEST_COMMIT ?= f2925f58acc259c4b894353f5fc404bdeb40028e
+PROJECT := github.com/kubernetes-incubator/cri-containerd
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
+GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
+PREFIX ?= ${DESTDIR}/usr/local
+BINDIR ?= ${PREFIX}/bin
```
Random-Liu: For now, can we use `BINDIR` directly?

The name `PREFIX` is confusing.

Makefile
```diff
@@ -0,0 +1,84 @@
+GO ?= go
+EPOCH_TEST_COMMIT ?= f2925f58acc259c4b894353f5fc404bdeb40028e
+PROJECT := github.com/kubernetes-incubator/cri-containerd
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
+GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
```
Random-Liu: The branch is not used. Let's remove it and add it back when we add the `version` support.

.tool/lint
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
Random-Liu: Don't understand, why do we need to put this in `.tool`, but put `verify-gofmt` in `hack/`?
Why can't we all move them into `hack`? Is there any problem?

hack/verify-gofmt.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
Random-Liu: Why we use `/bin/bash` here, but use `/usr/bin/env bash` above?

Based on http://stackoverflow.com/questions/16365130/the-difference-between-usr-bin-env-bash-and-usr-bin-bash, there are pros and cons. We should at least make it consistent.

Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+default: help
+
+help:
+	@echo "Usage: make <target>"
+	@echo
+	@echo " * 'install' - Install binaries to system locations"
+	@echo " * 'binaries' - Build cri-containerd"
+	@echo " * 'clean' - Clean artifacts"
+	@echo " * 'lint' - Execute the source code linter"
```
Random-Liu: Can we replace this with a `verify` command to do all the verification all together?

Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+.PHONY: check-gopath
+
+check-gopath:
+ifndef GOPATH
+	$(error GOPATH is not set)
+endif
+
+lint: check-gopath
+	@echo "checking lint"
```
Random-Liu: Why only echo information for `lint`?

Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+gofmt:
+	@./hack/verify-gofmt.sh
+
+cri-containerd: check-gopath
+	$(GO) build -o $@ \
+		$(PROJECT)/cmd/cri-containerd
+
+clean:
+	find . -name \*~ -delete
```
Random-Liu: For my education, what file do we want to cleanup with these 2 commands?

Makefile
```diff
@@ -0,0 +1,84 @@
...
+.PHONY: .gitvalidation
+# When this is running in travis, it will only check the travis commit range
+.gitvalidation: check-gopath
+ifeq ($(TRAVIS),true)
+	git-validation -q -run DCO,short-subject
+else
+	git-validation -v -run DCO,short-subject -range $(EPOCH_TEST_COMMIT)..HEAD
+endif
+
+.PHONY: install.tools
```
Random-Liu: Why `install.tools` is not in `help`?

Random-Liu: Why not PHONY following commands `.install.gitvalidation` etc. ?

Makefile
```diff
@@ -0,0 +1,84 @@
...
+install.tools: .install.gitvalidation .install.gometalinter .install.md2man
+
+.install.gitvalidation:
+	go get -u github.com/vbatts/git-validation
+
+.install.gometalinter:
+	go get -u github.com/alecthomas/gometalinter
+	gometalinter --install
+
+.install.md2man:
```
Random-Liu: Add this later when we need it. We can file an issue for this.

.tool/lint
mikebrow: no reason was following the cri-o pattern

mikebrow: sure will move and explain these "_" will be removed when used/implemented

.tool/lint
mikebrow: no reason just following the pattern from cri-o yours works just as well but would need to be updated if we add any code somewhere else...

.tool/lint
mikebrow: https://github.com/mibk/dupl
gometalinter uses a tool called dupl ^ 
This filters the _test.go results.

.tool/lint
mikebrow: This is to skip lint on vendor code.

Makefile
mikebrow: I suppose.  We can always add back in later.

Makefile
mikebrow: sure

Makefile
mikebrow: no reason will modify

Makefile
mikebrow: just temp files that we don't have atm so can be removed

.gitignore
```diff
@@ -0,0 +1 @@
+cri-containerd
```
mikebrow: sure

Makefile
mikebrow: sure

Makefile
mikebrow: No reason, usually one likes to PHONY the commands with no output

mikebrow: no reason.. usually one should only phony the commands with no output

Makefile
mikebrow: kk

hack/verify-gofmt.sh
mikebrow: picked up from cri-o... don't have any druthers here

pkg/server/service.go
mikebrow: sure

.tool/lint
Random-Liu: I'm not sure whether the cri-o pattern is intended. To me, if makes more sense to move them all into `hack/` if there is no particular reason that we should separate them.

.tool/lint
Random-Liu: Theirs also need to be updated when we include new non-go directories, right?

The only golang package I could image we'll add is `test` directory. WDYT?

.tool/lint
Random-Liu: I don't understand, we should have skip vendor in `while`, right?

.tool/lint
Random-Liu: From the document it's not clear to me what the tool is doing, and why we should exclude the failure here...

Makefile
Random-Liu: I see.

hack/verify-gofmt.sh
Random-Liu: Then let's make them consistent. :)

mikebrow: `hack/*` and normalize to `#!/bin/bash`

.tool/lint
mikebrow: Don't have druthers here.. will try yours out :-)

.tool/lint
mikebrow: It's looking for possible duplicated code. Typically unit test cases have some duplicated code. We could remove it and see what it does when we add unit tests :-)

.tool/lint
mikebrow: We really really don't want vendor. But you're correct it probably does not need to be on both the loop filter and the --exclude filter since we are passing single files. I imagine a possibility for one of the lint tools to pull in an import and test it as well but that's a guess. I'd have to remove it to see for sure.

mikebrow:
Have addressed all review comments.  Will add boilerplate testing to 'make verify' in a future PR.

hack/verify-gofmt.sh
mikebrow: Done..

mikebrow: done..

cmluciano:
Tested a few tasks locally with success. 

One nit: A build will get triggered with the default task via ``make``. It would be great to have an explicit build task so that ``make build`` works.

mikebrow:
make binaries is the current target for that.. but that's only until we have an official package / distribution plan.  Till then.. either make binaries or all is our build target.. All is the default make target thus why make by itself targets all then binaries...

hack/lint.sh
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+for d in $(find . -type d -a \( -iwholename './pkg*' -o -iwholename './cmd*' \)); do
+	echo ${d}
```
Random-Liu: nit: either remove this `echo` or add some context.

.tool/lint
Random-Liu: I saw you removed this. So we don't need this?

hack/lint.sh
mikebrow: when you run make verify it outputs the context of "checking lint" which is a little slow so I figured let it show what directories lint is checking.  Other than checking lint what additional context are you thinking?

Random-Liu: I mean something like: "checking directory ${d} ...". :)

Random-Liu:
@mikebrow Will merge this after you address comments and squash commits.

Issue #6, opened 2017-04-15
CRI Containerd Integration TODOs


Random-Liu:
Required Internal Services:
- [x] [P0] Metadata store. A metadata store is required to store metadata and system state #9.
  - [x] [P0] In-memory metadata store #9.
  - [ ] [P1] On-disk metadata store. (file-based or db-based)
- [ ] [P1] Container manager. A container manager keeps containerd running in `runtime` cgroup. (See [dockershim container manager](https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/dockershim/cm))

Container Runtime Interface Functions:
- [x] Sandbox Lifecycle
  - [x] [P0] [RunPodSandbox](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_run.go)
  - [x] [P0] [StopPodSandbox](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_stop.go)
  - [x] [P0] [ListPodSandbox](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_list.go)
  - [x] [P0] [PodSandboxStatus](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_status.go)
  - [x] [P0] [RemovePodSandbox](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_remove.go)
- [x] Container Lifecycle
  - [x] [P0] [CreateContainer](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/container_create.go)
  - [x] [P0] [StartContainer](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/container_start.go)
  - [x] [P0] [ListContainers](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/container_list.go)
  - [x] [P0] [ContainerStatus](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/container_status.go)
  - [x] [P0] [StopContainer](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/container_stop.go)
  - [x] [P0] [RemoveContainer](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/container_remove.go)
- [x] Image Management
[...]

PR #9, opened 2017-04-18
Add the metadata store interface and an in-memory implementation.

Random-Liu:
For https://github.com/kubernetes-incubator/cri-containerd/issues/6.

This PR added:
* The metadata store interface;
* An in-memory implementation;
* A sandbox metadata store based on the metadata store.

 In the future, we need to have an on-disk implementation. The `TODOs` in `memory_store.go` describes how to implement a [file-based checkpoint](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/checkpoint_store.go).

The metadata store interface is defined as follows:
```go
// All byte arrays are expected to be read-only. User MUST NOT modify byte
// array element directly!!

// UpdateFunc is function used to update a specific metadata. The value
// passed in is the current value. The updated value should be returned.
// If there is an error, the update will be rolled back.
type UpdateFunc func([]byte) ([]byte, error)

// Metadata is a metadata in the store. It supports transactional update.
type Metadata interface {
	// Get a snapshot of the metadata.
	Get() ([]byte, error)
	// Update the value.
	Update(UpdateFunc) error
}

// MetadataStore is the interface for storing metadata.
type MetadataStore interface {
	// Create the metadata with the specified id.
	// Note:
	// * Create MUST return error if the id already exists.
	// * The id and metadata MUST be added in one transaction.
	Create(string, []byte) (Metadata, error)
	// Get the metadata by id.
	Get(string) (Metadata, error)
	// List all the metadatas.
	List() ([]Metadata, error)
	// Delete the metadata by id.
	// Note:
	// * Delete MUST not return error if the id doesn't exist.
	// * The id and metadata MUST be deleted in one transaction.
	Delete(string) error
}
```

/cc @kubernetes-incubator/maintainers-cri-containerd @feiskyer @resouer

### DIFF
--- a/pkg/metadata/sandbox.go
+++ b/pkg/metadata/sandbox.go
@@ -1,0 +1,157 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	"encoding/json"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
+
+	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+)
+
+// TODO(random-liu): Handle versioning around all marshal/unmarshal.
+// version and versionedSandboxMetadata is not used now, but should be used
+// in the future:
+// 1) Only versioned metadata should be written into the store;
+// 2) Only unversioned metadata should be returned to the user;
+// 3) A conversion function is needed to convert any supported versioned
+// metadata into unversioned metadata.
+
+// sandboxMetadataVersion is current version of sandbox metadata.
+const sandboxMetadataVersion = "v1" // nolint
+
+// versionedSandboxMetadata is the internal struct representing the versioned
+// sandbox metadata
+// nolint
+type versionedSandboxMetadata struct {
+	// Version indicates the version of the versioned sandbox metadata.
+	Version string
+	SandboxMetadata
+}
+
+// SandboxMetadata is the unversioned sandbox metadata.
+type SandboxMetadata struct {
+	// ID is the sandbox id.
+	ID string
+	// Name is the sandbox name.
+	Name string
+	// Config is the CRI sandbox config.
+	Config *runtime.PodSandboxConfig
+	// CreatedAt is the created timestamp.
+	CreatedAt int64
+	// NetNS is the network namespace used by the sandbox.
+	NetNS string
+}
+
+// SandboxUpdateFunc is the function used to update SandboxMetadata.
+type SandboxUpdateFunc func(SandboxMetadata) (SandboxMetadata, error)
+
+// sandboxToStoreUpdateFunc generates a metadata store UpdateFunc from SandboxUpdateFunc.
+func sandboxToStoreUpdateFunc(u SandboxUpdateFunc) store.UpdateFunc {
+	return func(data []byte) ([]byte, error) {
+		meta := &SandboxMetadata{}
+		if err := json.Unmarshal(data, meta); err != nil {
+			return nil, err
+		}
+		newMeta, err := u(*meta)
+		if err != nil {
+			return nil, err
+		}
+		return json.Marshal(newMeta)
+	}
+}
+
+// SandboxStore is the store for metadata of all sandboxes.
+type SandboxStore interface {
+	// Create creates a sandbox from SandboxMetadata in the store.
+	Create(SandboxMetadata) error
+	// Get gets the specified sandbox.
+	Get(string) (*SandboxMetadata, error)
+	// Update updates a specified sandbox.
+	Update(string, SandboxUpdateFunc) error
+	// List lists all sandboxes.
+	List() ([]*SandboxMetadata, error)
+	// Delete deletes the sandbox from the store.
+	Delete(string) error
+}
+
+// sandboxStore is an implmentation of SandboxStore.
+type sandboxStore struct {
+	store store.MetadataStore
+}
+
+// NewSandboxStore creates a SandboxStore from a basic MetadataStore.
+func NewSandboxStore(store store.MetadataStore) SandboxStore {
+	return &sandboxStore{store: store}
+}
+
+// Create creates a sandbox from SandboxMetadata in the store.
+func (s *sandboxStore) Create(metadata SandboxMetadata) error {
+	data, err := json.Marshal(&metadata)
+	if err != nil {
+		return err
+	}
+	return s.store.Create(metadata.ID, data)
+}
+
+// Get gets the specified sandbox.
+func (s *sandboxStore) Get(sandboxID string) (*SandboxMetadata, error) {
+	data, err := s.store.Get(sandboxID)
+	if err != nil {
+		return nil, err
+	}
+	// Return nil without error if the corresponding metadata
+	// does not exist.
+	if data == nil {
+		return nil, nil
+	}
+	sandbox := &SandboxMetadata{}
+	if err := json.Unmarshal(data, sandbox); err != nil {
+		return nil, err
+	}
+	return sandbox, nil
+}
+
+// Update updates a specified sandbox. The function is running in a
+// transaction. Update will not be applied when the update function
+// returns error.
+func (s *sandboxStore) Update(sandboxID string, u SandboxUpdateFunc) error {
+	return s.store.Update(sandboxID, sandboxToStoreUpdateFunc(u))
+}
+
+// List lists all sandboxes.
+func (s *sandboxStore) List() ([]*SandboxMetadata, error) {
+	allData, err := s.store.List()
+	if err != nil {
+		return nil, err
+	}
+	var sandboxes []*SandboxMetadata
+	for _, data := range allData {
+		sandbox := &SandboxMetadata{}
+		if err := json.Unmarshal(data, sandbox); err != nil {
+			return nil, err
+		}
+		sandboxes = append(sandboxes, sandbox)
+	}
+	return sandboxes, nil
+}
+
+// Delete deletes the sandbox from the store.
+func (s *sandboxStore) Delete(sandboxID string) error {
+	return s.store.Delete(sandboxID)
+}
--- a/pkg/metadata/sandbox_test.go
+++ b/pkg/metadata/sandbox_test.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	"testing"
+	"time"
+
+	assertlib "github.com/stretchr/testify/assert"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
+
+	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+)
+
+func TestSandboxStore(t *testing.T) {
+	sandboxes := map[string]*SandboxMetadata{
+		"1": {
+			ID:   "1",
+			Name: "Sandbox-1",
+			Config: &runtime.PodSandboxConfig{
+				Metadata: &runtime.PodSandboxMetadata{
+					Name:      "TestPod-1",
+					Uid:       "TestUid-1",
+					Namespace: "TestNamespace-1",
+					Attempt:   1,
+				},
+			},
+			CreatedAt: time.Now().UnixNano(),
+			NetNS:     "TestNetNS-1",
+		},
+		"2": {
+			ID:   "2",
+			Name: "Sandbox-2",
+			Config: &runtime.PodSandboxConfig{
+				Metadata: &runtime.PodSandboxMetadata{
+					Name:      "TestPod-2",
+					Uid:       "TestUid-2",
+					Namespace: "TestNamespace-2",
+					Attempt:   2,
+				},
+			},
+			CreatedAt: time.Now().UnixNano(),
+			NetNS:     "TestNetNS-2",
+		},
+		"3": {
+			ID:   "3",
+			Name: "Sandbox-3",
+			Config: &runtime.PodSandboxConfig{
+				Metadata: &runtime.PodSandboxMetadata{
+					Name:      "TestPod-3",
+					Uid:       "TestUid-3",
+					Namespace: "TestNamespace-3",
+					Attempt:   3,
+				},
+			},
+			CreatedAt: time.Now().UnixNano(),
+			NetNS:     "TestNetNS-3",
+		},
+	}
+	assert := assertlib.New(t)
+
+	s := NewSandboxStore(store.NewMetadataStore())
+
+	t.Logf("should be able to create sandbox metadata")
+	for _, meta := range sandboxes {
+		assert.NoError(s.Create(*meta))
+	}
+
+	t.Logf("should be able to get sandbox metadata")
+	for id, expectMeta := range sandboxes {
+		meta, err := s.Get(id)
+		assert.NoError(err)
+		assert.Equal(expectMeta, meta)
+	}
+
+	t.Logf("should be able to list sandbox metadata")
+	sbs, err := s.List()
+	assert.NoError(err)
+	assert.Len(sbs, 3)
+
+	t.Logf("should be able to update sandbox metadata")
+	testID := "2"
+	newCreatedAt := time.Now().UnixNano()
+	expectMeta := *sandboxes[testID]
+	expectMeta.CreatedAt = newCreatedAt
+	err = s.Update(testID, func(o SandboxMetadata) (SandboxMetadata, error) {
+		o.CreatedAt = newCreatedAt
+		return o, nil
+	})
+	assert.NoError(err)
+	newMeta, err := s.Get(testID)
+	assert.NoError(err)
+	assert.Equal(&expectMeta, newMeta)
+
+	t.Logf("should be able to delete sandbox metadata")
+	assert.NoError(s.Delete(testID))
+	sbs, err = s.List()
+	assert.NoError(err)
+	assert.Len(sbs, 2)
+
+	t.Logf("get should return nil without error after deletion")
+	meta, err := s.Get(testID)
+	assert.NoError(err)
+	assert.Nil(meta)
+}
--- a/pkg/metadata/store/metadata_store.go
+++ b/pkg/metadata/store/metadata_store.go
@@ -1,0 +1,236 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package store
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/golang/glog"
+)
+
+// All byte arrays are expected to be read-only. User MUST NOT modify byte
+// array element directly!!
+
+// UpdateFunc is function used to update a specific metadata. The value
+// passed in is the old value, it MUST NOT be changed in the function.
+// The function should make a copy of the old value and apply update on
+// the copy. The updated value should be returned. If there is an error,
+// the update will be rolled back.
+type UpdateFunc func([]byte) ([]byte, error)
+
+// MetadataStore is the interface for storing metadata. All methods should
+// be thread-safe.
+// TODO(random-liu): Initialize the metadata store with a type, and replace
+// []byte with interface{}, so as to avoid extra marshal/unmarshal on the
+// user side.
+type MetadataStore interface {
+	// Create the metadata containing the passed in data with the
+	// specified id.
+	// Note:
+	// * Create MUST return error if the id already exists.
+	// * The id and data MUST be added in one transaction to the store.
+	Create(string, []byte) error
+	// Get the data by id.
+	// Note that Get MUST return nil without error if the id
+	// doesn't exist.
+	Get(string) ([]byte, error)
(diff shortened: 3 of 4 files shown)


Random-Liu:
Change to `WIP`, I'll firstly try to use the metadata store in the sandbox lifecycle management to see whether the interface is sufficient.

Random-Liu:
@yujuhong @dchen1107 The metadata store will be very important for the stability of `cri-containerd`, please take a look at this PR to make sure the design makes sense to you.

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
+
+	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+)
+
+// TODO(random-liu): Handle versioning around all marshal/unmarshal.
+// version and versionedSandboxMetadata is not used now, but should be used
+// in the future:
+// 1) Only versioned metadata should be written into the store;
+// 2) Only unversioned metadata should be returned to the user;
```
Random-Liu: Because I don't think the caller want to care about the version.

All version stuff should be handled inside the metadata store.

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+*/
+
+package store
+
+// All byte arrays are expected to be read-only. User MUST NOT modify byte
+// array element directly!!
+
+// UpdateFunc is function used to update a specific metadata. The value
+// passed in is the current value. The updated value should be returned.
```
Random-Liu: Nope, it's the old (current) value

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+	// * Create MUST return error if the id already exists.
+	// * The id and metadata MUST be added in one transaction.
+	Create(string, []byte) (Metadata, error)
+	// Get the metadata by id.
+	Get(string) (Metadata, error)
+	// List all the metadatas.
+	List() ([]Metadata, error)
+	// Delete the metadata by id.
+	// Note:
+	// * Delete MUST not return error if the id doesn't exist.
```
Random-Liu: I just mean Delete should be idempotent, it should not return error when corresponding metadata doesn't exist.

mikebrow: I see so:
// Delete(id) returns error = nil when there is no metadata for the id or when the metadata has been successfully deleted. Otherwise Delete(id) returns an error !=nil when there is metadata but the delete operation has failed...

Random-Liu:
Added unit test, PR ready for review.

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+
+// MetadataStore is the interface for storing metadata.
+// TODO(random-liu): Initialize the metadata store with a type, and replace
+// []byte with interface{}, so as to avoid extra marshal/unmarshal on the
+// user side.
+type MetadataStore interface {
+	// Create the metadata with the specified id.
+	// Note:
+	// * Create MUST return error if the id already exists.
+	// * The id and metadata MUST be added in one transaction.
```
Random-Liu: Done.

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+	Create(string, []byte) (Metadata, error)
+	// Get the metadata by id.
```
Random-Liu: Done.

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+	Get(string) (Metadata, error)
+	// List all the metadatas.
```
Random-Liu: Done.

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+	List() ([]Metadata, error)
+	// Delete the metadata by id.
```
Random-Liu: Done.

pkg/metadata/store/store.go
```diff
@@ -0,0 +1,54 @@
...
+	// Update the value.
+	Update(UpdateFunc) error
+}
+
+// MetadataStore is the interface for storing metadata.
+// TODO(random-liu): Initialize the metadata store with a type, and replace
+// []byte with interface{}, so as to avoid extra marshal/unmarshal on the
+// user side.
+type MetadataStore interface {
+	// Create the metadata with the specified id.
```
Random-Liu: Done

pkg/metadata/sandbox_test.go
```diff
@@ -0,0 +1,127 @@
...
+	assert.NoError(err)
+	assert.Len(sbs, 3)
+
+	t.Logf("should be able to delete sandbox metadata")
+	testID := "2"
+	assert.NoError(s.Delete(testID))
+	sbs, err = s.List()
+	assert.NoError(err)
+	assert.Len(sbs, 2)
+	meta, err := s.Get(testID)
```
mikebrow: t.Logf("after deleting id metadata Get should return nil for metadata and with no error")

pkg/metadata/sandbox_test.go
```diff
@@ -0,0 +1,127 @@
...
+	assert.NoError(err)
+	assert.True(meta == nil)
+}
```
mikebrow: I'd add another test that creates the metadata again for id..  is that the expected pattern? or are we only deleting the metadata as we delete a sandbox?

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,160 @@
...
+	// disk atomically.
+	// Only holding read lock makes sure that:
+	// 1) The in-memory cache is available for reading.
+	// 2) Even though, there is another on-going update, the read lock
+	// makes sure that they must be updating the same version of data.
+}
+
+// delete deletes the data on disk atomically.
+func (m *memoryMeta) delete() error {
+	// TODO(random-liu): Hold write lock, rename the data on the disk.
```
mikebrow: m.data = nil?

pkg/metadata/store/metadata_store_test.go
```diff
@@ -0,0 +1,144 @@
...
+	assert.NoError(err)
+	succeed, err := meta.Get()
+	assert.NoError(err)
+	assert.Equal(testData[1], succeed)
+
+	t.Logf("successful update should not affect existing snapshot")
+	assert.Equal(testData[0], old)
+
+	// TODO(random-liu): Test delete and cleanup after disk based
+	// implementation is added.
```
mikebrow: TODO for delete and cleanup?

pkg/metadata/store/metadata_store_test.go
```diff
@@ -0,0 +1,144 @@
...
+	meta, err = m.Get(testIds[1])
+	assert.NoError(err)
+	assert.True(meta == nil)
+
+	t.Logf("existing reference should not be affected by delete")
+	existingMeta := newMeta
+	existingGot, err := existingMeta.Get()
+	assert.NoError(err)
+	assert.Equal([]byte("updated-metadata-1"), existingGot)
+}
```
mikebrow: create again after delete?

pkg/metadata/sandbox_test.go
Random-Liu: Done.

pkg/metadata/sandbox_test.go
Random-Liu: Will add the case into the basic metadata store.

pkg/metadata/store/metadata_store_test.go
Random-Liu: The TODO is just for this.

pkg/metadata/store/metadata_store_test.go
Random-Liu: Done.

pkg/metadata/store/metadata_store.go
Random-Liu: Option 1: If we set `m.data=nil`, other operation (e.g. ListPodSandbox, PodSandboxStatus) which has already got the metadata reference will fail, and return error.

Option 2: If we don't set `m.data=nil`, other operation which has got the reference will success:
1) If that operation is read only, it will return stale data.
2) If that operation updates the metadata, it will only change the in-memory state.
3) Any future operations won't be able to get the metadata because it is removed from the list already.

Either option 1 or 2 is acceptable to me, because this is unavoidable race unless we lock the whole store for each single metadata operation, which is not what we want. :)

I prefer 2.


We don't necessarily need to set `m.data=nil`.

In cri-containerd, no long running process should hold a specific metadata reference,

mikebrow: Is there a door number 3? I suppose we should talk about this. Maybe defer calls to delete?

pkg/metadata/store/metadata_store_test.go
mikebrow: Not sure how I missed the TODO text.. odd :-)

pkg/metadata/store/metadata_store.go
Random-Liu: What do you mean by "defer calls to delete"?

mikebrow: If deleting metadata can't be done due to dangling references one could treat the request as a deferred request to delete.. or perhaps direct that delete should only be called like this defer meta.delete() after said dangling references have been removed.

mikebrow: guess I don't get the part about the call is only for deleting data on disk.. while leaving it around for further use.. Maybe a quick phone call.

Random-Liu: @mikebrow Hm, after thinking for a while, I feel like option 1 becomes better.

For `ListPodSandbox` we could return nil to indicate the metadata is deleted, so that `ListPodSandbox` could continue.
For `PodSandboxStatus` it should return error anyway, because the sandbox is deleted.

Random-Liu: > guess I don't get the part about the call is only for deleting data on disk.. while leaving it around for further use.. Maybe a quick phone call.

Actually the data already is removed from the store, that is where the delete happens. It's already not accessible by future operations.
The only problem is for the operations which have already go the reference, they can either fail or return/update staled data, because there is no time series guarantee for 2 concurrent operations.

Maybe I should change the name to `deleteCheckpoint` or something to make it more clear.

mikebrow: right by further I meant.. continued use by existing references.. 

yes deleteCheckpoint works !

Random-Liu: @mikebrow OK. Will update the PR. :)

Random-Liu:
@yujuhong Addressed comments. The code does become cleaner after combine `Metadata` and `MetadataStore`.

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+// Create creates a sandbox from SandboxMetadata in the store.
+func (s *sandboxStore) Create(metadata SandboxMetadata) error {
+	data, err := json.Marshal(&metadata)
+	if err != nil {
+		return err
+	}
+	return s.store.Create(metadata.ID, data)
+}
+
+// Get gets a specified sandbox.
```
yujuhong: s/a/the

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+	return nil
+}
+
+// delete deletes the data on disk atomically.
+func (m *metadata) delete() error {
+	// TODO(random-liu): Hold write lock, rename the data on the disk.
+	return nil
+}
+
+// cleanup cleans up all temporary files left-over.
```
yujuhong: Why would there be multiple temporary files for a single piece of metadata?

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+func (m *metadata) cleanup() error {
+	// TODO(random-liu): Hold write lock, Cleanup temporary files generated
+	// in atomic file operations. The write lock makes sure there is no on-going
+	// update, so any temporary files could be removed.
+	return nil
+}
+
+// metadataStore is the implementation of MetadataStore
```
yujuhong: `metadataStore implements MetadataStore` or `metadataStore is an implementation of MetadataStore`

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+		newMeta, err := u(*meta)
+		if err != nil {
+			return nil, err
+		}
+		return json.Marshal(newMeta)
+	}
+}
+
+// SandboxStore is the store for metadata of all sandboxes.
+type SandboxStore interface {
```
yujuhong: Do we need the interface?

pkg/metadata/sandbox_test.go
```diff
@@ -0,0 +1,120 @@
...
+	t.Logf("should be able to delete sandbox metadata")
+	assert.NoError(s.Delete(testID))
+	sbs, err = s.List()
+	assert.NoError(err)
+	assert.Len(sbs, 2)
+
+	t.Logf("get should return nil without error after deletion")
+	meta, err := s.Get(testID)
+	assert.NoError(err)
+	assert.True(meta == nil)
```
yujuhong: Can use assert.Nil()

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"sync"
+)
+
+// TODO(random-liu) Rename memoryMeta to metadata, and add checkpoint. When
+// checkpoint is enabled, it should cache data in memory and checkpoint metadata
+// into files during update. Metadata should serve from memory, but any modification
+// should be checkpointed, so that memory could be recovered after restart. It
+// should be possible to disable the checkpoint for testing.
+// Note that checkpoint update may fail, so the recovery logic should tolerant that.
```
Random-Liu: Will do.

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+	"sync"
+)
+
+// All byte arrays are expected to be read-only. User MUST NOT modify byte
+// array element directly!!
+
+// UpdateFunc is function used to update a specific metadata. The value
+// passed in is the old value. The updated value should be returned.
+// If there is an error, the update will be rolled back.
+type UpdateFunc func([]byte) ([]byte, error)
```
Random-Liu: As is comment above:
```
// All byte arrays are expected to be read-only. User MUST NOT modify byte
// array element directly!!
```

Actually, we could pass in a copy of the current data. However, that introduces extra overhead and actually all user will just marshal the byte array without changing it directly.

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+
+// MetadataStore is the interface for storing metadata.
+// TODO(random-liu): Initialize the metadata store with a type, and replace
+// []byte with interface{}, so as to avoid extra marshal/unmarshal on the
+// user side.
```
Random-Liu: Will do.

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+	// doesn't exist or has been removed.
+	// * The id and data MUST be deleted in one transaction.
+	Delete(string) error
+}
+
+// TODO(random-liu) Add checkpoint. When checkpoint is enabled, it should cache data
+// in memory and checkpoint metadata into files during update. Metadata should serve
+// from memory, but any modification should be checkpointed, so that memory could be
+// recovered after restart. It should be possible to disable the checkpoint for testing.
+// Note that checkpoint update may fail, so the recovery logic should tolerant that.
```
Random-Liu: Will do.

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+
+// An atomic file operator is needed.
```
Random-Liu: This is a part of TODO. We need atomic file writer to do checkpoint. Could be removed now.

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+
+// metadata is unit stored in metadataStore.
```
Random-Liu: Will do.

pkg/metadata/store/metadata_store.go
Random-Liu: There may be temporary files left over from previous failed update operations.

pkg/metadata/store/metadata_store.go
Random-Liu: Will do.

pkg/metadata/store/metadata_store.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+		delete(m.metas, id)
+		return meta, nil
+	}()
+	if err != nil {
+		return err
+	}
+	// The metadata is removed from the store at this point.
+	if meta != nil {
+		// Do not return error for cleanup.
```
Random-Liu: Will do.

pkg/metadata/store/metadata_store_test.go
```diff
@@ -0,0 +1,133 @@
...
+		testIds[1]: []byte("metadata-1"),
+	}
+	assert := assertlib.New(t)
+
+	m := NewMetadataStore()
+
+	t.Logf("should be empty initially")
+	metas, err := m.List()
+	assert.NoError(err)
+	assert.Empty(metas)
```
Random-Liu: I just replace the comment with logging, or else still need to comment for each test case. :)

pkg/metadata/store/metadata_store_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	t.Logf("should be able to delete metadata")
+	assert.NoError(m.Delete(testIds[1]))
+	metas, err = m.List()
+	assert.NoError(err)
+	assert.Len(metas, 1)
+	assert.Equal(testMeta[testIds[0]], metas[0])
+	meta, err = m.Get(testIds[1])
+	assert.NoError(err)
+	assert.True(meta == nil)
```
Random-Liu: Will do.

pkg/metadata/store/metadata_store_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	t.Logf("existing reference should not be affected by delete")
+	assert.Equal([]byte("updated-metadata-1"), newMeta)
+
+	t.Logf("should be able to reuse the same id after deletion")
+	err = m.Create(testIds[1], testMeta[testIds[1]])
+	assert.NoError(err)
+}
+
```
Random-Liu: Will do.

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
+
+	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+)
+
+// TODO(random-liu): Uncomment following code and handle versioning around
```
Random-Liu: OK. Left over from rebase, I've already uncommented them...

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+// all marshal/unmarshal.
+// version and versionedSandboxMetadata is not used now, but should be used
+// in the future:
+// 1) Only versioned metadata should be written into the store;
+// 2) Only unversioned metadata should be returned to the user;
+// 3) A conversion function is needed to convert any supported versioned
+// metadata into unversioned metadata.
+
+// version is current version of sandbox metadata.
```
Random-Liu: Will do.

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+const sandboxMetadataVersion = "v1" // nolint
```
Random-Liu: Yeah, because no one is using it, verify-lint will report error.

pkg/metadata/sandbox.go
```diff
@@ -0,0 +1,157 @@
...
+
+// versionedSandboxMetadata is the internal versioned sandbox metadata.
```
Random-Liu: Will do.

pkg/metadata/sandbox.go
Random-Liu: Will do.

pkg/metadata/sandbox.go
Random-Liu: Why not an interface? Should we avoid interface?

pkg/metadata/store/metadata_store.go
Random-Liu: Done.

pkg/metadata/store/metadata_store.go
Random-Liu: Updated the comments.

pkg/metadata/store/metadata_store.go
Random-Liu: Done.

pkg/metadata/store/metadata_store.go
Random-Liu: Done.

pkg/metadata/store/metadata_store.go
Random-Liu: Removed.

pkg/metadata/store/metadata_store.go
Random-Liu: Done.

pkg/metadata/store/metadata_store.go
Random-Liu: Done.

pkg/metadata/store/metadata_store.go
Random-Liu: Done.

pkg/metadata/store/metadata_store_test.go
```diff
@@ -0,0 +1,133 @@
...
+	err = m.Create(testIds[0], testMeta[testIds[0]])
+	assert.Error(err)
+
+	t.Logf("should be able to list metadata")
+	err = m.Create(testIds[1], testMeta[testIds[1]])
+	assert.NoError(err)
+	metas, err = m.List()
+	assert.NoError(err)
+	assert.Len(metas, 2)
+	for _, expect := range testMeta {
```
Random-Liu: Done.

pkg/metadata/store/metadata_store_test.go
Random-Liu: Done.

pkg/metadata/store/metadata_store_test.go
Random-Liu: Done.

pkg/metadata/sandbox.go
Random-Liu: Done.

pkg/metadata/sandbox.go
Random-Liu: Done.

pkg/metadata/sandbox.go
Random-Liu: Done.

pkg/metadata/sandbox.go
Random-Liu: Done.

pkg/metadata/sandbox_test.go
Random-Liu: Done.

Random-Liu:
@yujuhong Addressed comments.

pkg/metadata/sandbox.go
yujuhong: Not saying we should avoid interfaces, but just asking to see why interface is needed here. Do we expect other implmentations or fakes/mocks for testing?

yujuhong:
LGTM

pkg/metadata/sandbox.go
Random-Liu: I don't think we'll have other implementations at this level. But it seems more clear to me to have an interface to tell the caller what operations are supported.

Random-Liu:
@yujuhong @mikebrow Thanks for reviewing! Will squash and merge this one, and rebase the others! :)

PR #10, opened 2017-04-18
Add version flag

### LINKED ISSUE #8: Add `--version` flag.
Add cri-containerd version.

We should version cri-containerd based on the repo tag and commit number. The version will be used:
1) As part of the name when building a release package or any docker image in the future.
2) As part of `--version` output.

Refer https://github.com/kubernetes/node-problem-detector/pull/71.

cmluciano:
Closes #8 

- ~~Will need a rebase after #7 is merged~~


Random-Liu:
@cmluciano Thanks a lot! Will review this soon after it's out of `WIP`.

mikebrow:
@cmluciano Makefile is all your's merge away..

cmluciano:
Thanks @mikebrow . I will add a new test case for this to try it out.

Makefile
```diff
@@ -78,4 +86,5 @@ install.tools: .install.gitvalidation .install.gometalinter
 	help \
 	install \
 	lint \
-	uninstall
+	uninstall \
+        version
```
Random-Liu: extra space

cmd/cri-containerd/cri_containerd.go
```diff
@@ -38,6 +40,11 @@ func main() {
 	glog.V(2).Infof("Run cri-containerd grpc server on socket %q", o.CRIContainerdSocketPath)
 	service := server.NewCRIContainerdService(conn)
 	s := server.NewCRIContainerdServer(o.CRIContainerdSocketPath, service, service)
+
+	if o.ContainerdVersion {
```
Random-Liu: We should at the beginning. `cri-containerd --version` should not do anything else.

cmd/cri-containerd/options/options.go
```diff
@@ -31,6 +31,8 @@ type CRIContainerdOptions struct {
 	ContainerdSocketPath string
 	// ContainerdConnectionTimeout is the connection timeout for containerd client.
 	ContainerdConnectionTimeout time.Duration
+	// ContainerdVersion is the git release version of cri-containerd
+	ContainerdVersion bool
```
Random-Liu: Actually, this is not containerd version, it's the version of `cri-containerd`.

pkg/server/version.go
```diff
@@ -18,13 +18,21 @@ package server
...
 import (
 	"errors"
+	"fmt"
 
 	"golang.org/x/net/context"
 
 	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
 )
 
+var containerdVersion = "UNKNOWN"
```
Random-Liu: ditto. This is not containerdVersion.

And I prefer putting the logic into `pkg/version/version.go`. This file is to implement the `Version` function defined in CRI, which returns the runtime version.

Makefile
```diff
@@ -3,6 +3,9 @@ EPOCH_TEST_COMMIT ?= f2925f58acc259c4b894353f5fc404bdeb40028e
 PROJECT := github.com/kubernetes-incubator/cri-containerd
 BINDIR ?= ${DESTDIR}/usr/local/bin
 BUILD_DIR ?= _output
+# VERSION is the version of the binary.
+# TODO: push a tag so this will not print a fatal error
+VERSION:=$(shell git describe --tags --dirty)
```
mikebrow: I sort of like getting the version from a flat file rather than a git tag and the commit from git... see opencontainers/runc for a easy to use pattern.  @Random-Liu wdyt?

Random-Liu: Kubernetes is using similar solution, but more complex https://github.com/kubernetes/kubernetes/blob/master/hack/lib/version.sh.

Actually, once we have a tag, this will work well. The problem is that I'm not sure what tag to create given that we don't have the first version.

Makefile
cmluciano: Odd. I don't see this in my editor and when I go to the file in Github, I don't see the space.

cmluciano:
I swapped around the variable naming and corrected a few spacing issues. My editor should be running gofmt automatically, but I ran it manually too.

cmluciano:
Let me know if this feature should have a test case. I couldn't think of anything too useful to add since this functionality just prints a string.

mikebrow:
Just a go unit test for the API. Id presume semver. So you could use a semver test. I saw one the other day somewhere in kublet I'm thinking? I can look for it l8r.

mikebrow:
Just a go unit test for the API. Id presume semver. So you could use a semver test. I saw one the other day somewhere in kublet I'm thinking? I can look for it l8r. We can do integration tests later on.

mikebrow:
Fat fingered on my phone.

cmluciano:
NP. Since I have the version hardcoded at UNKNOWN, the test will continually fail.

@mikebrow @Random-Liu Can we push a pre-alpha version like ```"0.0.1-alpha.preview+123.456``` so that I can revise the PR with a proper tag?

Makefile
```diff
@@ -17,6 +20,7 @@ help:
 	@echo " * 'verify'        - Execute the source code verification tools"
 	@echo " * 'install.tools' - Installs tools used by verify"
 	@echo " * 'uninstall'     - Remove installed binaries from system locations"
+	@echo " * 'version'       - Print git tag version"
```
Random-Liu: "git tag version" sounds a little bit weird to me. What about `Print current cri-containerd release version`?

cmd/cri-containerd/options/options.go
```diff
@@ -31,6 +31,8 @@ type CRIContainerdOptions struct {
 	ContainerdSocketPath string
 	// ContainerdConnectionTimeout is the connection timeout for containerd client.
 	ContainerdConnectionTimeout time.Duration
+	// CRIContainerdVersion is the git release version of cri-containerd
+	CRIContainerdVersion bool
```
Random-Liu: Can we move this up to under `CRIContainerdSocketPath`?

It would be better to group Containerd options together, and CRIContainerd options together. :)

pkg/version/version.go
```diff
@@ -0,0 +1,26 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+import "fmt"
+
+var criContainerdVersion = "UNKNOWN"
```
Random-Liu: Should build version into he binary, see https://github.com/kubernetes/node-problem-detector/pull/71/files#diff-b67911656ef5d18c4ae36cb6741b7965R26.

pkg/version/version.go
```diff
@@ -0,0 +1,26 @@
...
+
+// PrintVersion outputs the release version of cricontainerd
```
Random-Liu: s/cricontainerd/cri-containerd

Random-Liu:
I'll push a alpha tag soon.

Random-Liu:
We have a `0.0.0` now...

Makefile
mikebrow: yeah.. bit of overkill to use tags over a flat file but either works...

mikebrow:
@cmluciano see the semver from string to version struct api... https://github.com/kubernetes/kubernetes/blob/master/pkg/util/version/version.go#L111

We can worry about a CLI integration test at some future time I think.  I just have to remember to add the -ldflags -x option when I add the unit tests to the makefile :-)

pkg/version/version.go
cmluciano: Done

cmluciano:
I addressed the latest comment. I was considering the use of the [semver/blang](https://github.com/blang/semver) library so that we do not have to deal with the regex ourselves. 

[Example in the kubeadm code](https://github.com/kubernetes/kubernetes/blob/2d66f7bd28ad5332a5b892b0cb4944cbd679f639/cmd/kubeadm/app/cmd/defaults.go#L52)

cmluciano:
I noted a couple of bugs with my latest push. Fixing and will push an example for semver/blang

Makefile
```diff
@@ -3,6 +3,9 @@ EPOCH_TEST_COMMIT ?= f2925f58acc259c4b894353f5fc404bdeb40028e
 PROJECT := github.com/kubernetes-incubator/cri-containerd
 BINDIR ?= ${DESTDIR}/usr/local/bin
 BUILD_DIR ?= _output
+# VERSION is the version of the binary.
+# TODO: push a tag so this will not print a fatal error
```
Random-Liu: Remove this `TODO`, since we have a tag now.

Makefile
```diff
@@ -36,8 +43,9 @@ gofmt:
 	@./hack/verify-gofmt.sh
 
 cri-containerd: check-gopath
-	$(GO) build -o $(BUILD_DIR)/$@ \
-		$(PROJECT)/cmd/cri-containerd
+	$(GO) -o $(BUILD_DIR)/$@ \
```
Random-Liu: `build` is missing.

Makefile
```diff
@@ -36,8 +43,9 @@ gofmt:
...
+	-ldflags '-w -extldflags "-static" -X $(PKG)/pkg/version.version=$(VERSION)' \
```
Random-Liu: We don't need other compile flags for now.

I believe `-w` can be kept, but "-static" can be removed.

Makefile
```diff
@@ -50,6 +58,7 @@ install: check-gopath
 uninstall:
 	rm -f $(BINDIR)/cri-containerd
 
+
```
Random-Liu: Extra empty line.

Random-Liu:
Close and reopen to trigger the travis, because this PR was sent before `.travis.yaml` is merged.

mikebrow:
Hey @cmluciano your commit seems to be missing the DCO signature..

cmluciano:
I need to add update the Godeps

Makefile
```diff
@@ -27,6 +27,7 @@ help:
 	@echo
 	@echo " * 'install'       - Install binaries to system locations"
 	@echo " * 'binaries'      - Build cri-containerd"
+	@echo " * 'test'      - Test cri-containerd"
```
mikebrow: shift right 4 sp :-)

Makefile
```diff
@@ -57,6 +58,9 @@ cri-containerd: check-gopath
 	$(GO) build -o $(BUILD_DIR)/$@ \
 		$(PROJECT)/cmd/cri-containerd
 
+test: 
+	go test -timeout=1m -v -race ./pkg/... $(BUILD_TAGS)
```
mikebrow: maybe a comment or two for the build_tags usage

vendor/github.com/blang/semver/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License
```
mikebrow: also in k8s.io/kubernetes/vendor

vendor/github.com/pmezard/go-difflib/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2013, Patrick Mezard
```
mikebrow: not sure about this license but I see it's already included in the k8s.io vendor tree

Makefile
cmluciano: Done

Godeps/Godeps.json
```diff
@@ -1,12 +1,17 @@
 {
 	"ImportPath": "github.com/kubernetes-incubator/cri-containerd",
 	"GoVersion": "go1.7",
-	"GodepVersion": "v74",
+	"GodepVersion": "v79",
```
mikebrow: k8s.io/kubernetes has also moved up to v79

Makefile
```diff
@@ -55,7 +62,11 @@ boiler:
 
 cri-containerd: check-gopath
 	$(GO) build -o $(BUILD_DIR)/$@ \
-		$(PROJECT)/cmd/cri-containerd
+	-ldflags '-X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)' \
+	$(PROJECT)/cmd/cri-containerd
+
+test: 
+	go test -timeout=1m -v -race ./pkg/... $(BUILD_TAGS)
```
mikebrow: I believe go test also supports -ldflags so if you add it your unit test should be able to check ver on the tag..

Maybe make a variable for this since you would be using it twice.

cmluciano: That was a great call. I tried it locally and it revealed a bug.

pkg/version/version_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidateSemver(t *testing.T) {
+	err := validateSemver("UNKNOWN")
+	assert := assert.New(t)
+	assert.NotNil(err)
+	err = validateSemver("0.0.0-1-gdf6a1cc-dirty")
+	assert.Nil(err)
+}
```
mikebrow: err = validateSemver(criContainerdVersion)  :-)

Makefile
Random-Liu: `BUILD_TAGS` is not set and not needed for now.

pkg/version/version_test.go
Random-Liu: I think we should keep the test case.

We need this test case to test that `validateSemver` is working properly.
And we need another test case to test `criContainerdVersion` is valid.

pkg/version/version.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/blang/semver"
+)
+
+var criContainerdVersion = "UNKNOWN"
+
+func validateSemver(sv string) error {
+	badVersion, err := semver.Parse(sv)
+	if err != nil {
+		return fmt.Errorf("couldn't parse kubernetes version %q: %v", badVersion, err)
```
Random-Liu: s/kubernetes/cri-containerd

Random-Liu: I think we should discard the output `badVersion`, and log `sv` here.

pkg/version/version_test.go
mikebrow: Yes that's what I meant/why I added the comment to the end after the dirty check :-)

pkg/version/version.go
mikebrow: yes .Parse() returns a "valid" version struct so badVersion is a misnomer. Since we don't need the version struct "_" is fine. Agree with @Random-Liu that it would be helpful if the err message included the failing version.

cmluciano: I can't discard badVersion since semver.Parse returns two variables. There is a [semver.Validate](https://godoc.org/github.com/blang/semver#Version.Validate), but it require passing a semver.Version variable.

vendor/github.com/blang/semver/LICENSE
cmluciano: It was not found in the [Travis build(]https://travis-ci.org/kubernetes-incubator/cri-containerd/jobs/225271496#L222)

cmluciano:
Ok I think I have addressed all comments in my latest patch set.

Makefile
```diff
@@ -56,7 +63,8 @@ boiler:
 
 cri-containerd: check-gopath
 	$(GO) build -o $(BUILD_DIR)/$@ \
-		$(PROJECT)/cmd/cri-containerd
+        $(BUILD_TAGS) \
```
Random-Liu: Something wrong with your editor I believe. :p

Random-Liu:
I'm fine with fixing the indent in a following PR.

PR #16, opened 2017-04-22
Move boilerplate verification into hack

Random-Liu:
This PR moves the boilerplate verification into `hack/` and let the git validation skip the invalid commits.


Random-Liu:
Apply LGTM based on @mikebrow's approval.

PR #17, opened 2017-04-22
Add travis build and golang report icons.

Random-Liu:
Add build status and go report cards.


Random-Liu:
Will merge after the test passed.

PR #19, opened 2017-04-25
Add sandbox implementation

Random-Liu:
Based on https://github.com/kubernetes-incubator/cri-containerd/pull/9.

The first 3 commits are from https://github.com/kubernetes-incubator/cri-containerd/pull/9. Please only look at the 3 new commits (including one godep update commit).

This PR added the sandbox implementation, including:
* RunPodSandbox
* ListPodSandbox
* PodSandboxStatus
* StopPodSandbox
* RemovePodSandbox

The PR is marked as WIP, because unit test is needed. Actually the code is complete already.

Post the PR here to:
1) Get initial review. Will add unit test after the PR looks good overall.
2) Demonstrate how to use the metadata store.
3) Demonstrate the code style.

@yujuhong @mikebrow /cc @kubernetes-incubator/maintainers-cri-containerd

PR #22, opened 2017-05-02
Add container implementation

Random-Liu:
Based on https://github.com/kubernetes-incubator/cri-containerd/pull/9 and https://github.com/kubernetes-incubator/cri-containerd/pull/19.

Please only review the last 2 commits. Will add unit test after the code looks good overall.

@kubernetes-incubator/maintainers-cri-containerd

### DIFF
--- a/pkg/metadata/container.go
+++ b/pkg/metadata/container.go
@@ -1,0 +1,183 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	"encoding/json"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
+
+	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+)
+
+// The code is very similar with sandbox.go, but there is no template support
+// in golang, we have to have similar files for different types.
+// TODO(random-liu): Figure out a way to simplify this.
+// TODO(random-liu): Handle versioning with the same mechanism with container.go
+
+// containerMetadataVersion  is current version of container metadata.
+const containerMetadataVersion = "v1" // nolint
+
+// versionedContainerMetadata is the internal versioned container metadata.
+// nolint
+type versionedContainerMetadata struct {
+	// Version indicates the version of the versioned container metadata.
+	Version string
+	ContainerMetadata
+}
+
+// ContainerMetadata is the unversioned container metadata.
+type ContainerMetadata struct {
+	// ID is the container id.
+	ID string
+	// Name is the container name.
+	Name string
+	// SandboxID is the sandbox id the container belongs to.
+	SandboxID string
+	// Config is the CRI container config.
+	Config *runtime.ContainerConfig
+	// ImageRef is the reference of image used by the container.
+	ImageRef string
+	// Pid is the init process id of the container.
+	Pid uint32
+	// CreatedAt is the created timestamp.
+	CreatedAt int64
+	// StartedAt is the started timestamp.
+	StartedAt int64
+	// FinishedAt is the finished timestamp.
+	FinishedAt int64
+	// ExitCode is the container exit code.
+	ExitCode int32
+	// CamelCase string explaining why container is in its current state.
+	Reason string
+	// Human-readable message indicating details about why container is in its
+	// current state.
+	Message string
+	// Removing indicates that the container is in removing state.
+	// In fact, this field doesn't need to be checkpointed.
+	// TODO(random-liu): Skip this during serialization when we put object
+	// into the store directly.
+	// TODO(random-liu): Reset this field to false during state recoverry.
+	Removing bool
+}
+
+// State returns current state of the container based on the metadata.
+func (c *ContainerMetadata) State() runtime.ContainerState {
+	if c.FinishedAt != 0 {
+		return runtime.ContainerState_CONTAINER_EXITED
+	}
+	if c.StartedAt != 0 {
+		return runtime.ContainerState_CONTAINER_RUNNING
+	}
+	if c.CreatedAt != 0 {
+		return runtime.ContainerState_CONTAINER_CREATED
+	}
+	return runtime.ContainerState_CONTAINER_UNKNOWN
+}
+
+// ContainerUpdateFunc is the function used to update ContainerMetadata.
+type ContainerUpdateFunc func(ContainerMetadata) (ContainerMetadata, error)
+
+// ContainerToStoreUpdateFunc generates a metadata store UpdateFunc from ContainerUpdateFunc.
+func ContainerToStoreUpdateFunc(u ContainerUpdateFunc) store.UpdateFunc {
+	return func(data []byte) ([]byte, error) {
+		meta := &ContainerMetadata{}
+		if err := json.Unmarshal(data, meta); err != nil {
+			return nil, err
+		}
+		newMeta, err := u(*meta)
+		if err != nil {
+			return nil, err
+		}
+		return json.Marshal(newMeta)
+	}
+}
+
+// ContainerStore is the store for metadata of all containers.
+type ContainerStore interface {
+	// Create creates a container from ContainerMetadata in the store.
+	Create(ContainerMetadata) error
+	// Get gets a specified container.
+	Get(string) (*ContainerMetadata, error)
+	// Update updates a specified container.
+	Update(string, ContainerUpdateFunc) error
+	// List lists all containers.
+	List() ([]*ContainerMetadata, error)
+	// Delete deletes the container from the store.
+	Delete(string) error
+}
+
+// containerStore is an implmentation of ContainerStore.
+type containerStore struct {
+	store store.MetadataStore
+}
+
+// NewContainerStore creates a ContainerStore from a basic MetadataStore.
+func NewContainerStore(store store.MetadataStore) ContainerStore {
+	return &containerStore{store: store}
+}
+
+// Create creates a container from ContainerMetadata in the store.
+func (c *containerStore) Create(metadata ContainerMetadata) error {
+	data, err := json.Marshal(&metadata)
+	if err != nil {
+		return err
+	}
+	return c.store.Create(metadata.ID, data)
+}
+
+// Get gets a specified container.
+func (c *containerStore) Get(containerID string) (*ContainerMetadata, error) {
+	data, err := c.store.Get(containerID)
+	if err != nil {
+		return nil, err
+	}
+	container := &ContainerMetadata{}
+	if err := json.Unmarshal(data, container); err != nil {
+		return nil, err
+	}
+	return container, nil
+}
+
+// Update updates a specified container. The function is running in a
+// transaction. Update will not be applied when the update function
+// returns error.
+func (c *containerStore) Update(containerID string, u ContainerUpdateFunc) error {
+	return c.store.Update(containerID, ContainerToStoreUpdateFunc(u))
+}
+
+// List lists all containers.
+func (c *containerStore) List() ([]*ContainerMetadata, error) {
+	allData, err := c.store.List()
+	if err != nil {
+		return nil, err
+	}
+	var containers []*ContainerMetadata
+	for _, data := range allData {
+		container := &ContainerMetadata{}
+		if err := json.Unmarshal(data, container); err != nil {
+			return nil, err
+		}
+		containers = append(containers, container)
+	}
+	return containers, nil
+}
+
+// Delete deletes the Container from the store.
+func (c *containerStore) Delete(containerID string) error {
+	return c.store.Delete(containerID)
+}
--- a/pkg/metadata/container_test.go
+++ b/pkg/metadata/container_test.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2017 The Kubernetes Authorc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	"testing"
+	"time"
+
+	assertlib "github.com/stretchr/testify/assert"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
+
+	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+)
+
+func TestContainerState(t *testing.T) {
+	for c, test := range map[string]struct {
+		metadata *ContainerMetadata
+		state    runtime.ContainerState
+	}{
+		"unknown state": {
+			metadata: &ContainerMetadata{
+				ID:   "1",
+				Name: "Container-1",
+			},
+			state: runtime.ContainerState_CONTAINER_UNKNOWN,
+		},
+		"created state": {
+			metadata: &ContainerMetadata{
+				ID:        "2",
+				Name:      "Container-2",
+				CreatedAt: time.Now().UnixNano(),
+			},
+			state: runtime.ContainerState_CONTAINER_CREATED,
+		},
+		"running state": {
+			metadata: &ContainerMetadata{
+				ID:        "3",
+				Name:      "Container-3",
+				CreatedAt: time.Now().UnixNano(),
+				StartedAt: time.Now().UnixNano(),
+			},
+			state: runtime.ContainerState_CONTAINER_RUNNING,
+		},
+		"exited state": {
+			metadata: &ContainerMetadata{
+				ID:         "3",
+				Name:       "Container-3",
+				CreatedAt:  time.Now().UnixNano(),
+				FinishedAt: time.Now().UnixNano(),
+			},
+			state: runtime.ContainerState_CONTAINER_EXITED,
+		},
+	} {
+		t.Logf("TestCase %q", c)
+		assertlib.Equal(t, test.state, test.metadata.State())
+	}
+}
+
+func TestContainerStore(t *testing.T) {
+	containers := map[string]*ContainerMetadata{
+		"1": {
+			ID:        "1",
+			Name:      "Container-1",
+			SandboxID: "Sandbox-1",
+			Config: &runtime.ContainerConfig{
+				Metadata: &runtime.ContainerMetadata{
+					Name:    "TestPod-1",
+					Attempt: 1,
+				},
+			},
+			ImageRef:   "TestImage-1",
+			Pid:        1,
+			CreatedAt:  time.Now().UnixNano(),
+			StartedAt:  time.Now().UnixNano(),
+			FinishedAt: time.Now().UnixNano(),
+			ExitCode:   1,
+			Reason:     "TestReason-1",
+			Message:    "TestMessage-1",
+		},
+		"2": {
+			ID:        "2",
+			Name:      "Container-2",
+			SandboxID: "Sandbox-2",
+			Config: &runtime.ContainerConfig{
+				Metadata: &runtime.ContainerMetadata{
+					Name:    "TestPod-2",
+					Attempt: 2,
+				},
+			},
+			ImageRef:   "TestImage-2",
+			Pid:        2,
+			CreatedAt:  time.Now().UnixNano(),
+			StartedAt:  time.Now().UnixNano(),
+			FinishedAt: time.Now().UnixNano(),
+			ExitCode:   2,
+			Reason:     "TestReason-2",
+			Message:    "TestMessage-2",
+		},
+		"3": {
+			ID:        "3",
+			Name:      "Container-3",
+			SandboxID: "Sandbox-3",
+			Config: &runtime.ContainerConfig{
+				Metadata: &runtime.ContainerMetadata{
+					Name:    "TestPod-3",
+					Attempt: 3,
+				},
+			},
+			ImageRef:   "TestImage-3",
+			Pid:        3,
+			CreatedAt:  time.Now().UnixNano(),
+			StartedAt:  time.Now().UnixNano(),
+			FinishedAt: time.Now().UnixNano(),
+			ExitCode:   3,
+			Reason:     "TestReason-3",
+			Message:    "TestMessage-3",
+			Removing:   true,
+		},
+	}
+	assert := assertlib.New(t)
+
+	c := NewContainerStore(store.NewMetadataStore())
+
+	t.Logf("should be able to create container metadata")
+	for _, meta := range containers {
+		assert.NoError(c.Create(*meta))
+	}
+
+	t.Logf("should be able to get container metadata")
+	for id, expectMeta := range containers {
+		meta, err := c.Get(id)
+		assert.NoError(err)
+		assert.Equal(expectMeta, meta)
+	}
+
+	t.Logf("should be able to list container metadata")
+	cntrs, err := c.List()
+	assert.NoError(err)
+	assert.Len(cntrs, 3)
+
+	t.Logf("should be able to update container metadata")
+	testID := "2"
+	newCreatedAt := time.Now().UnixNano()
+	expectMeta := *containers[testID]
+	expectMeta.CreatedAt = newCreatedAt
+	err = c.Update(testID, func(o ContainerMetadata) (ContainerMetadata, error) {
+		o.CreatedAt = newCreatedAt
+		return o, nil
+	})
+	assert.NoError(err)
+	newMeta, err := c.Get(testID)
+	assert.NoError(err)
+	assert.Equal(&expectMeta, newMeta)
+
+	t.Logf("should be able to delete container metadata")
+	assert.NoError(c.Delete(testID))
+	cntrs, err = c.List()
+	assert.NoError(err)
+	assert.Len(cntrs, 2)
+
+	t.Logf("get should return nil without error after deletion")
+	meta, err := c.Get(testID)
+	assert.Error(store.ErrNotExist, err)
+	assert.True(meta == nil)
+}
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -17,14 +17,84 @@ limitations under the License.
 package server
 
 import (
-	"errors"
+	"fmt"
+	"time"
 
+	"github.com/golang/glog"
 	"golang.org/x/net/context"
 
 	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata"
 )
 
 // CreateContainer creates a new container in the given PodSandbox.
-func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.CreateContainerRequest) (*runtime.CreateContainerResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.CreateContainerRequest) (retRes *runtime.CreateContainerResponse, retErr error) {
+	glog.V(2).Infof("CreateContainer within sandbox %q with container config %+v and sandbox config %+v",
+		r.GetPodSandboxId(), r.GetConfig(), r.GetSandboxConfig())
+	defer func() {
+		if retErr == nil {
+			glog.V(2).Infof("CreateContainer returns container id %q", retRes.GetContainerId())
+		}
+	}()
+
+	config := r.GetConfig()
+	sandboxConfig := r.GetSandboxConfig()
+	sandbox, err := c.getSandbox(r.GetPodSandboxId())
+	if err != nil {
+		return nil, fmt.Errorf("failed to find sandbox id %q: %v", r.GetPodSandboxId(), err)
+	}
+
+	// Generate unique id and name for the container and reserve the name.
+	// Reserve the container name to avoid concurrent `CreateContainer` request creating
+	// the same container.
+	id := generateID()
+	name := makeContainerName(config.GetMetadata(), sandboxConfig.GetMetadata())
+	if err := c.containerNameIndex.Reserve(name, id); err != nil {
+		return nil, fmt.Errorf("failed to reserve container name %q: %v", name, err)
+	}
+	defer func() {
+		// Release the name if the function returns with an error.
+		if retErr != nil {
+			c.containerNameIndex.ReleaseByName(name)
+		}
+	}()
+
+	// Create initial container metadata.
+	meta := metadata.ContainerMetadata{
+		ID:        id,
+		Name:      name,
+		SandboxID: sandbox.ID,
+		Config:    config,
+	}
+
+	// TODO(random-liu): [P0] Prepare container rootfs.
+
+	// TODO(random-liu): [P0] Set ImageRef in ContainerMetadata with image id.
+
+	// Create container root directory.
+	containerRootDir := getContainerRootDir(c.rootDir, id)
+	if err := c.os.MkdirAll(containerRootDir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create container root directory %q: %v",
+			containerRootDir, err)
+	}
+	defer func() {
+		if retErr != nil {
+			// Cleanup the container root directory.
+			if err := c.os.RemoveAll(containerRootDir); err != nil {
+				glog.Errorf("Failed to remove container root directory %q: %v",
+					containerRootDir, err)
+			}
+		}
+	}()
+
+	// Update container CreatedAt.
+	meta.CreatedAt = time.Now().UnixNano()
+	// Add container into container store.
+	if err := c.containerStore.Create(meta); err != nil {
+		return nil, fmt.Errorf("failed to add container metadata %+v into store: %v",
+			meta, err)
+	}
+
+	return &runtime.CreateContainerResponse{ContainerId: id}, nil
 }
--- a/pkg/server/container_list.go
+++ b/pkg/server/container_list.go
@@ -17,14 +17,87 @@ limitations under the License.
 package server
 
 import (
-	"errors"
+	"fmt"
 
+	"github.com/golang/glog"
 	"golang.org/x/net/context"
 
 	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata"
 )
 
 // ListContainers lists all containers matching the filter.
-func (c *criContainerdService) ListContainers(ctx context.Context, r *runtime.ListContainersRequest) (*runtime.ListContainersResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) ListContainers(ctx context.Context, r *runtime.ListContainersRequest) (retRes *runtime.ListContainersResponse, retErr error) {
+	glog.V(4).Infof("ListContainers with filter %+v", r.GetFilter())
+	defer func() {
+		if retErr == nil {
+			glog.V(4).Infof("ListContainers returns containers %+v", retRes.GetContainers())
+		}
+	}()
+
+	// List all container metadata from store.
+	metas, err := c.containerStore.List()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list metadata from container store: %v", err)
+	}
+
+	var containers []*runtime.Container
+	for _, meta := range metas {
+		containers = append(containers, toCRIContainer(meta))
+	}
+
+	containers = c.filterCRIContainers(containers, r.GetFilter())
+	return &runtime.ListContainersResponse{Containers: containers}, nil
+}
+
+// toCRIContainer converts container metadata into CRI container.
+func toCRIContainer(meta *metadata.ContainerMetadata) *runtime.Container {
+	return &runtime.Container{
+		Id:           meta.ID,
+		PodSandboxId: meta.SandboxID,
+		Metadata:     meta.Config.GetMetadata(),
+		Image:        meta.Config.GetImage(),
+		ImageRef:     meta.ImageRef,
+		State:        meta.State(),
+		CreatedAt:    meta.CreatedAt,
+		Labels:       meta.Config.GetLabels(),
+		Annotations:  meta.Config.GetAnnotations(),
+	}
+}
+
+// filterCRIContainers filters CRIContainers.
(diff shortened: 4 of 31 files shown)


pkg/server/container_stop.go
```diff
@@ -17,14 +17,80 @@ limitations under the License.
...
+	_, err = c.containerService.Kill(ctx, &execution.KillRequest{ID: id, Signal: uint32(stopSignal)})
+	if err != nil && !isContainerdContainerNotExistError(err) {
+		return nil, fmt.Errorf("failed to stop container %q: %v", id, err)
+	}
+	if isContainerdContainerNotExistError(err) {
+		return &runtime.StopContainerResponse{}, nil
+	}
+
+	if err := state.waitStop(time.Duration(r.GetTimeout()) * time.Second); err == nil {
+		return &runtime.StopContainerResponse{}, nil
```
Random-Liu: We should not return here, should continue deleting the container.

pkg/metadata/container.go
```diff
@@ -0,0 +1,183 @@
...
+	// Removing indicates that the container is in removing state.
+	// In fact, this field doesn't need to be checkpointed.
+	// TODO(random-liu): Skip this during serialization when we put object
+	// into the store directly.
+	// TODO(random-liu): Reset this field to false during state recoverry.
+	Removing bool
+}
+
+// State returns current state of the container based on the metadata.
+func (c *ContainerMetadata) State() runtime.ContainerState {
```
yujuhong: If a container fails to start (e.g., wrong command), would that be handled correctly? I haven't read the actual implementation, so maybe that'd answer my question.

pkg/metadata/container.go
```diff
@@ -0,0 +1,183 @@
...
+	Reason string
+	// Human-readable message indicating details about why container is in its
+	// current state.
+	Message string
+	// Removing indicates that the container is in removing state.
+	// In fact, this field doesn't need to be checkpointed.
+	// TODO(random-liu): Skip this during serialization when we put object
+	// into the store directly.
+	// TODO(random-liu): Reset this field to false during state recoverry.
+	Removing bool
```
yujuhong: What's the purpose of this? Is this supposed to ensure only one goroutine is removing the container?
Why didn't we need this for the sandbox or any other operations?

pkg/metadata/container.go
```diff
@@ -0,0 +1,183 @@
...
+// containerStore is an implmentation of ContainerStore.
+type containerStore struct {
+	store store.MetadataStore
+}
+
+// NewContainerStore creates a ContainerStore from a basic MetadataStore.
+func NewContainerStore(store store.MetadataStore) ContainerStore {
+	return &containerStore{store: store}
+}
+
```
yujuhong: I am going to skip the rest of the file assuming that they are just the sandbox metadata store with name replaced.

pkg/metadata/container.go
```diff
@@ -0,0 +1,183 @@
...
+	if c.FinishedAt != 0 {
+		return runtime.ContainerState_CONTAINER_EXITED
+	}
+	if c.StartedAt != 0 {
+		return runtime.ContainerState_CONTAINER_RUNNING
+	}
+	if c.CreatedAt != 0 {
+		return runtime.ContainerState_CONTAINER_CREATED
+	}
+	return runtime.ContainerState_CONTAINER_UNKNOWN
```
yujuhong: Would this ever happen?

pkg/metadata/container_test.go
```diff
@@ -0,0 +1,180 @@
...
+			ExitCode:   3,
+			Reason:     "TestReason-3",
+			Message:    "TestMessage-3",
+			Removing:   true,
+		},
+	}
+	assert := assertlib.New(t)
+
+	c := NewContainerStore(store.NewMetadataStore())
+
```
yujuhong: Since we reuse much of the sandbox metadata store code, how much value does the following tests add?

pkg/metadata/container.go
```diff
@@ -0,0 +1,183 @@
+/*
```
yujuhong: Reviewed e0f748c65f71266dcb7324f6a9226bdf9eae327f.

Random-Liu:
@yujuhong PR ready for review. There is something wrong with the `ContainerStop` unit test. Removed in the current PR, will add it back later.

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	go func() {
+		for {
+			c.handleEvent(events)
+		}
+	}()
+	return nil
+}
+
+// handleEvent receives an event from contaienrd and handles the event.
```
mikebrow: /s/en/ne/

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+// container events.
+func (c *criContainerdService) startEventMonitor() error {
+	events, err := c.containerService.Events(context.Background(), &execution.EventsRequest{})
+	if err != nil {
+		return err
+	}
+	go func() {
+		for {
+			c.handleEvent(events)
+		}
```
mikebrow: if we get a fail this loop gets pretty tight..

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+	}()
+	return nil
+}
+
+// handleEvent receives an event from contaienrd and handles the event.
+func (c *criContainerdService) handleEvent(events execution.ContainerService_EventsClient) {
+	e, err := events.Recv()
+	if err != nil {
+		glog.Errorf("Failed to receive event: %v", err)
+		return
```
mikebrow: failed to receive event goes to log returns then comes right back here right?

pkg/server/container_create.go
```diff
@@ -17,14 +17,84 @@ limitations under the License.
...
+		if retErr == nil {
+			glog.V(2).Infof("CreateContainer returns container id %q", retRes.GetContainerId())
+		}
+	}()
+
+	config := r.GetConfig()
+	sandboxConfig := r.GetSandboxConfig()
+	sandbox, err := c.getSandbox(r.GetPodSandboxId())
+	if err != nil {
+		return nil, fmt.Errorf("failed to find sandbox id %q: %v", r.GetPodSandboxId(), err)
```
mikebrow: did you want to set or use retErr here?

pkg/server/container_create.go
```diff
@@ -17,14 +17,84 @@ limitations under the License.
...
+	}
+
+	// Generate unique id and name for the container and reserve the name.
+	// Reserve the container name to avoid concurrent `CreateContainer` request creating
+	// the same container.
+	id := generateID()
+	name := makeContainerName(config.GetMetadata(), sandboxConfig.GetMetadata())
+	if err := c.containerNameIndex.Reserve(name, id); err != nil {
+		return nil, fmt.Errorf("failed to reserve container name %q: %v", name, err)
```
mikebrow: same here... using retErr for defer flag

pkg/server/container_create.go
```diff
@@ -17,14 +17,84 @@ limitations under the License.
...
+	}
+
+	// TODO(random-liu): [P0] Prepare container rootfs.
+
+	// TODO(random-liu): [P0] Set ImageRef in ContainerMetadata with image id.
+
+	// Create container root directory.
+	containerRootDir := getContainerRootDir(c.rootDir, id)
+	if err := c.os.MkdirAll(containerRootDir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create container root directory %q: %v",
```
mikebrow: same here

pkg/server/helpers.go
```diff
@@ -70,6 +93,19 @@ func makeSandboxName(s *runtime.PodSandboxMetadata) string {
...
+// makeContainerName generates container name from sandbox and container metadata.
+// The name generated is unique as long as the sandbox container combination is
+// unique.
+func makeContainerName(c *runtime.ContainerMetadata, s *runtime.PodSandboxMetadata) string {
+	return strings.Join([]string{
+		c.Name,      // 0
+		s.Name,      // 1: sandbox name
+		s.Namespace, // 2: sandbox namespace
+		s.Uid,       // 3: sandbox uid
+		fmt.Sprintf("%d", c.Attempt), // 4
```
mikebrow: strconv.

pkg/server/events.go
Random-Liu: Good catch. We may need a back-off if `Recv` returns error.

pkg/metadata/container.go
Random-Liu: Offline discussed with @yujuhong.

`Removing` is mainly introduced to avoid race with concurrent remove/start operations (mainly start).

We don't need this for sandbox because sandbox only has `run` instead of `create` and `start`. There is no race between `run` and `remove`.

pkg/metadata/container.go
Random-Liu: Yeah, we'll update `FinishedAt` timestamp.

pkg/metadata/container.go
Random-Liu: Shouldn't happen unless checkpoint corrupted.

pkg/metadata/container.go
Random-Liu: ACK.

pkg/metadata/container_test.go
Random-Liu: Not too much, in fact. But it feels weird that we only have test for one metadata store.

After https://github.com/containerd/containerd/pull/859 lands, this part will be refactored, we may only need in memory data structure for `sandbox` and `image`.

The code and test will be different then. :)

pkg/server/events.go
Random-Liu: Will do.

pkg/server/helpers.go
Random-Liu: Will do.

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+	glog.V(2).Infof("Delete container from containerd %q", id)
+	_, err = c.containerService.Delete(ctx, &execution.DeleteRequest{ID: id})
+	if err != nil && !isContainerdContainerNotExistError(err) {
+		return nil, fmt.Errorf("failed to delete container %q: %v", id, err)
+	}
+	if isContainerdContainerNotExistError(err) {
+		return &runtime.StopContainerResponse{}, nil
+	}
+
+	// Wait forever until container stop is observed by event monitor.
```
yujuhong: I don't see the point of waiting forever. The CRI RPC request will probably time out at some point. If for whatever reason the container can't be stopped, you'll most likely be leaking goroutines.

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+	if err := c.waitContainerStop(id, -1*time.Second); err != nil {
+		return nil, fmt.Errorf("error occurs during waiting for container %q to stop: %v",
+			id, err)
+	}
+	return &runtime.StopContainerResponse{}, nil
+}
+
+// waitContainerStop polls container state until timeout excceeds or container is stopped.
```
yujuhong: type: s/excceeds/exceeds

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+	for {
+		// At least check once.
+		meta, err := c.containerStore.Get(id)
+		if err != nil {
+			if !metadata.IsNotExistError(err) {
+				return fmt.Errorf("failed to get container %q metadata: %v", id, err)
+			}
+			// Do not return error here because container is removed means
+			// it is already stopped.
+			glog.Warningf("Container %q is removed during stopping", id)
```
yujuhong: s/is/was

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+			return nil
+		}
+		// TODO(random-liu): Use channel instead of polling.
+		if meta.State() == runtime.ContainerState_CONTAINER_EXITED {
+			return nil
+		}
+		if timeout >= 0 && time.Since(start) >= timeout {
+			break
+		}
+		<-ticker.C
```
yujuhong: How about using a timer (`timer.NewTimer`) and use `select` to choose between the two?

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+}
+
+// waitContainerStop polls container state until timeout excceeds or container is stopped.
+// Negative timeout means waiting forever.
+func (c *criContainerdService) waitContainerStop(id string, timeout time.Duration) error {
+	ticker := time.NewTicker(stopCheckPollInterval)
+	defer ticker.Stop()
+	start := time.Now()
+	for {
+		// At least check once.
```
yujuhong: nit: Poll once before waiting for `stopCheckPollInterval`

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+		if err != nil {
+			if !metadata.IsNotExistError(err) {
+				return fmt.Errorf("failed to get container %q metadata: %v", id, err)
+			}
+			// Do not return error here because container is removed means
+			// it is already stopped.
+			glog.Warningf("Container %q is removed during stopping", id)
+			return nil
+		}
+		// TODO(random-liu): Use channel instead of polling.
```
yujuhong: Do you mean subscribing to the events using a channel?

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+	}
+	id := r.GetContainerId()
+
+	// Return without error if container is not running. This makes sure that
+	// stop only takes real action after the container is started.
+	if meta.State() != runtime.ContainerState_CONTAINER_RUNNING {
+		return &runtime.StopContainerResponse{}, nil
+	}
+
+	// TODO(random-liu): [P1] Get stop signal from image config.
```
yujuhong: Hm....are there any other variables defined in the image config, but not passed through CRI? We should probably audit and document them.

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/golang/glog"
+	"golang.org/x/net/context"
+
+	"github.com/containerd/containerd/api/services/execution"
+	"github.com/containerd/containerd/api/types/container"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata"
+)
+
+// startEventMonitor starts an event monitor which monitors and handles all
```
yujuhong: Is it possible that events would be dropped? Do we handle missing events?

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		glog.Errorf("Failed to receive event: %v", err)
+		return
+	}
+	glog.V(2).Infof("Received container event: %+v", e)
+	switch e.Type {
+	// If containerd-shim exits unexpectedly, there will be no corresponding event.
+	// However, containerd could not retrieve container state in that case, so it's
+	// fine to leave out that case for now.
+	// TODO(random-liu): [P2] Handle container-shim exit.
```
yujuhong: containerd-shim?

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+	case container.Event_EXIT:
+		meta, err := c.containerStore.Get(e.ID)
+		if err != nil {
+			glog.Errorf("Failed to get container %q metadata: %v", e.ID, err)
+			return
+		}
+		if e.Pid != meta.Pid {
+			// Not init process dies, ignore the event.
```
yujuhong: s/Not init/Non-init
s/dies/died

pkg/server/helpers.go
```diff
@@ -90,11 +131,58 @@ func getStreamingPipes(rootDir string) (string, string, string) {
 	return stdin, stdout, stderr
 }
 
+// TODO(Now): working here.
```
yujuhong: TODO(Now)?

pkg/server/events.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/containerd/containerd/api/services/execution"
+	"github.com/containerd/containerd/api/types/container"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata"
+)
+
+// startEventMonitor starts an event monitor which monitors and handles all
+// container events.
+func (c *criContainerdService) startEventMonitor() error {
```
yujuhong: If the grpc connection terminates, we need to be able to restart the event monitor. Need another level of wrapping to retry this.

pkg/server/helpers.go
```diff
@@ -90,11 +131,58 @@ func getStreamingPipes(rootDir string) (string, string, string) {
...
+// prepareStreamingPipes prepares stream named pipe for container. returns nil
+// io.ReadCloser is corresponding stream path is empty.
```
yujuhong: Comment is not coherent. Do you mean `if corresponding`?

pkg/server/helpers.go
```diff
@@ -90,11 +131,58 @@ func getStreamingPipes(rootDir string) (string, string, string) {
...
+	} {
+		if stream.path == "" {
+			continue
+		}
+		s, err := c.os.OpenFifo(ctx, stream.path, stream.flag, 0700)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to open named pipe %q: %v",
+				stream.path, err)
+		}
+		defer func(c io.Closer) {
```
yujuhong: Just a nit: `c` is already used for *criContainerdService. I know the scope is different, but naming the variable something else might help with readability.

pkg/server/events.go
Random-Liu: Done.

pkg/server/events.go
Random-Liu: Add TODO to do backoff here.

pkg/server/events.go
Random-Liu: Yeah, added TODO to do backoff here.

pkg/server/helpers.go
Random-Liu: Actually I've tried, `fmt.Sprintf` is even cleaner than strconv.FormatUint.

pkg/server/container_create.go
```diff
@@ -17,14 +17,84 @@ limitations under the License.
...
 	"k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata"
 )
 
 // CreateContainer creates a new container in the given PodSandbox.
-func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.CreateContainerRequest) (*runtime.CreateContainerResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.CreateContainerRequest) (retRes *runtime.CreateContainerResponse, retErr error) {
+	glog.V(2).Infof("CreateContainer within sandbox %q with container config %+v and sandbox config %+v",
```
Random-Liu: ACK.

pkg/server/container_list.go
```diff
@@ -17,14 +17,89 @@ limitations under the License.
...
+	if filter == nil {
+		return containers
+	}
+
+	filtered := []*runtime.Container{}
+	for _, cntr := range containers {
+		if filter.GetId() != "" && filter.GetId() != cntr.Id {
+			continue
+		}
+
```
Random-Liu: Done.

pkg/server/container_list.go
```diff
@@ -17,14 +17,89 @@ limitations under the License.
...
+		if filter.GetState() != nil && filter.GetState().GetState() != cntr.State {
+			continue
+		}
+
+		if filter.GetLabelSelector() != nil {
+			match := true
+			for k, v := range filter.GetLabelSelector() {
+				if cntr.Labels[k] != v {
```
Random-Liu: Good catch. Done.

Random-Liu: Also updated sandbox_list.

pkg/server/container_remove.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
 )
 
 // RemoveContainer removes the container.
-func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.RemoveContainerRequest) (*runtime.RemoveContainerResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.RemoveContainerRequest) (retRes *runtime.RemoveContainerResponse, retErr error) {
+	glog.V(2).Infof("RemoveContainer for %q", r.GetContainerId())
+	defer func() {
+		if retErr == nil {
+			glog.V(2).Infof("RemoveContainer returns successfully")
```
Random-Liu: Will do for all other operations.

Random-Liu: Done.

pkg/server/container_remove.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+		}
+	}()
+
+	id := r.GetContainerId()
+
+	// Set removing state to prevent following start and removing.
```
Random-Liu: Done.

pkg/server/container_remove.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+	err := c.setContainerRemoving(id)
+	if err != nil {
```
Random-Liu: Done.

pkg/server/container_remove.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+			c.containerNameIndex.ReleaseByKey(id)
+			return
+		}
+		// Reset removing if remove failed.
+		err := c.containerStore.Update(id, func(meta metadata.ContainerMetadata) (metadata.ContainerMetadata, error) {
+			meta.Removing = false
+			return meta, nil
+		})
+		// TODO(random-liu): Deal with update failure. Actually Removing doesn't need to
+		// be checkpointed, we only need it to have the same lifecycle with container metadata.
```
Random-Liu: Yeah. Actually we don't need to checkpoint the `removing` state, it could be purely in memory, we'll improve that in the future. :)

pkg/server/container_remove.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+	// Delete container metadata.
+	if err := c.containerStore.Delete(id); err != nil {
+		return nil, fmt.Errorf("failed to delete container metadata for %q: %v", id, err)
+	}
+
+	return &runtime.RemoveContainerResponse{}, nil
+}
+
+// setContainerRemoving sets the container into removing state. In removing state, the
+// container will not be started or removed again.
```
Random-Liu: There is no race between `Stop` and `Remove`, because `Stop` could only happen when container is in running state; `Remove` could only happen when container is in created/exited state.

For exited container, it will not be running again, so it's fine;
For created container, because `removing` avoid other operation starting the container again, so it will not go into running state.

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
 )
 
 // StartContainer starts the container.
-func (c *criContainerdService) StartContainer(ctx context.Context, r *runtime.StartContainerRequest) (*runtime.StartContainerResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) StartContainer(ctx context.Context, r *runtime.StartContainerRequest) (retRes *runtime.StartContainerResponse, retErr error) {
+	glog.V(2).Infof("StartContainer for %q", r.GetContainerId())
+	defer func() {
+		if retErr == nil {
+			glog.V(2).Infof("StartContainer returns successfully")
```
Random-Liu: Done

pkg/server/container_remove.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+		return &runtime.RemoveContainerResponse{}, nil
+	}
+	defer func() {
+		if retErr == nil {
+			// Cleanup all index after successfully remove the container.
+			c.containerNameIndex.ReleaseByKey(id)
+			return
+		}
+		// Reset removing if remove failed.
+		err := c.containerStore.Update(id, func(meta metadata.ContainerMetadata) (metadata.ContainerMetadata, error) {
```
Random-Liu: Done.

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+	container, err := c.containerStore.Get(r.GetContainerId())
+	if err != nil {
+		return nil, fmt.Errorf("failed to find container %q: %v", r.GetContainerId(), err)
+	}
+	id := container.ID
+
+	var startErr error
+	// start container in one transaction to avoid race with event monitor.
+	if err := c.containerStore.Update(id, func(meta metadata.ContainerMetadata) (metadata.ContainerMetadata, error) {
+		// Always apply metadata change no matter startContainer fails or not.
```
Random-Liu: Done.

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+		startErr = c.startContainer(ctx, id, &meta)
+		return meta, nil
+	}); startErr != nil {
+		return nil, startErr
+	} else if err != nil {
+		return nil, fmt.Errorf("failed to update container %q metadata: %v", id, err)
```
Random-Liu: When error is nil, the response must not be nil, or else grpc will report error.
When error is not nil, the response can be whatever, and nil seems better to me. :)

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+	}
+	return &runtime.StartContainerResponse{}, nil
+}
+
+// startContainer actually starts the container. The function is running in one transaction. Any updates
```
Random-Liu: Done

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+// to the metadata passed in will be applied to container store no matter the function returns error or not.
+func (c *criContainerdService) startContainer(ctx context.Context, id string, meta *metadata.ContainerMetadata) (retErr error) {
+	config := meta.Config
+	// Return error if container is not in created state.
+	if meta.State() != runtime.ContainerState_CONTAINER_CREATED {
+		return fmt.Errorf("container %q is in %s state", id,
+			runtime.ContainerState_name[int32(meta.State())])
```
Random-Liu: Done

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+		return fmt.Errorf("sandbox %q not found: %v", meta.SandboxID, err)
+	}
+	sandboxConfig := sandboxMeta.Config
+	sandboxID := meta.SandboxID
+	// Make sure sandbox is running.
+	sandboxInfo, err := c.containerService.Info(ctx, &execution.InfoRequest{ID: sandboxID})
+	if err != nil {
+		return fmt.Errorf("failed to get sandbox container %q info: %v", sandboxID, err)
+	}
+	// This is only a best effort check, sandbox may still exit after this.
```
Random-Liu: We'll not be able to start this container: https://github.com/opencontainers/runtime-spec/blob/30362737548149b000128dbc059d7e2b13d1032e/config-linux.md#namespaces

Will update the comment.

Random-Liu: Done.

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+		id, meta.Name, createOpts)
+	createResp, err := c.containerService.Create(ctx, createOpts)
+	if err != nil {
+		return fmt.Errorf("failed to create containerd container: %v", err)
+	}
+	defer func() {
+		if retErr != nil {
+			// Cleanup the containerd container if an error is returned.
+			if _, err := c.containerService.Delete(ctx, &execution.DeleteRequest{ID: id}); err != nil {
+				glog.Errorf("Failed to delete containerd container %q: %v",
```
Random-Liu: Done.

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+	meta.StartedAt = time.Now().UnixNano()
+	return nil
+}
+
+func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint32, config *runtime.ContainerConfig, sandboxConfig *runtime.PodSandboxConfig) (*runtimespec.Spec, error) {
+	// Creates a spec Generator with the default spec.
+	// TODO(random-liu): [P2] Move container runtime spec generation into a helper function.
+	g := generate.New()
+
+	// Set relative root path, the relative root path is fixed for containerd.
```
Random-Liu: Done.

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+	g.SetRootPath(relativeRootfsPath)
+
+	if len(config.GetCommand()) != 0 || len(config.GetArgs()) != 0 {
```
Random-Liu: The logic is not complete now.

We should get default entrypoint and args from image config, and apply here. If the command is empty, we should use the default entrypoint.

Will finish this part, after the image PR is finished. :)

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+	g.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
+	g.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
+	g.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
+	g.SetLinuxResourcesMemoryLimit(uint64(resources.GetMemoryLimitInBytes()))
+	g.SetLinuxResourcesOOMScoreAdj(int(resources.GetOomScoreAdj()))
+}
+
+// setOCICapabilities adds/drops process capabilities.
+func setOCICapabilities(g *generate.Generator, capabilities *runtime.Capability) error {
+	if capabilities == nil {
```
Random-Liu: Actually, they may be default. I have to TODO for this in the sandbox code": https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_run.go#L200

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+const stopCheckPollInterval = 100 * time.Millisecond
+
 // StopContainer stops a running container with a grace period (i.e., timeout).
-func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.StopContainerRequest) (*runtime.StopContainerResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.StopContainerRequest) (retRes *runtime.StopContainerResponse, retErr error) {
+	glog.V(2).Infof("StopContainer for %q with timeout %d (s)", r.GetContainerId(), r.GetTimeout())
+	defer func() {
+		if retErr == nil {
+			glog.V(2).Infof("StopContainer returns successfully")
```
Random-Liu: Done

pkg/server/container_stop.go
Random-Liu: There are several image configuration, I'm not sure how we should deal with (See https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/config.go#L24-L51).

E.g. `ExposedPorts` and `Volumes`. I'll take a look at this part when I work on the PR to apply image default.

pkg/server/container_stop.go
```diff
@@ -17,14 +17,106 @@ limitations under the License.
...
+	glog.Errorf("Stop container %q timed out: %v", id, err)
+
+	glog.V(2).Infof("Delete container from containerd %q", id)
+	_, err = c.containerService.Delete(ctx, &execution.DeleteRequest{ID: id})
+	if err != nil && !isContainerdContainerNotExistError(err) {
+		return nil, fmt.Errorf("failed to delete container %q: %v", id, err)
+	}
+	if isContainerdContainerNotExistError(err) {
+		return &runtime.StopContainerResponse{}, nil
+	}
```
Random-Liu: Done. Good idea.

pkg/server/container_stop.go
Random-Liu: I just keep the docker behavior here: https://github.com/moby/moby/blob/master/daemon/stop.go#L83-L84

pkg/server/container_stop.go
Random-Liu: Done

pkg/server/container_stop.go

pkg/server/container_stop.go

pkg/server/container_stop.go
Random-Liu: Nope. I mean we could open a channel between event monitor and stop operation, so that we could wait on the channel for the result.

pkg/server/events.go
Random-Liu: Not yet. I think the assumption is that if containerd is running, no event should be dropped.
Upon cri-containerd startup, it will recover state from containerd actual state.

However, I'm not sure what happens if containerd dies. Will add a TODO here to figure it out.

Random-Liu: Done.

pkg/server/container_stop.go
Random-Liu: This seems similar to me, :)

pkg/server/events.go
Random-Liu: Yeah, that's one thing I'm not sure yet. How to deal with containerd unresponsive. Added a todo, and will also include the retry there.

pkg/server/events.go
Random-Liu: Done

pkg/server/events.go
Random-Liu: Done.

pkg/server/helpers.go
Random-Liu: Sorry, left from developing. Done.

pkg/server/helpers.go
Random-Liu: Yeah, Done.

pkg/server/helpers.go
Random-Liu: Done

pkg/server/container_start.go
```diff
@@ -17,14 +17,327 @@ limitations under the License.
...
+	}
+
+	return nil
+}
+
+// setOCINamespaces sets namespaces.
+func setOCINamespaces(g *generate.Generator, namespaces *runtime.NamespaceOption, sandboxPid uint32) {
+	g.AddOrReplaceLinuxNamespace(string(runtimespec.NetworkNamespace), getNetworkNamespace(sandboxPid)) // nolint: errcheck
+	g.AddOrReplaceLinuxNamespace(string(runtimespec.IPCNamespace), getIPCNamespace(sandboxPid))         // nolint: errcheck
+	g.AddOrReplaceLinuxNamespace(string(runtimespec.UTSNamespace), getUTSNamespace(sandboxPid))         // nolint: errcheck
```
Random-Liu: Offline discussed with @yujuhong, we'll share UTS namespace for now. But file a issue to track this to figure out why docker doesn't support this, whether there is any problem.

Random-Liu:
@yujuhong Addressed comments.

pkg/server/container_remove.go
yujuhong: ACK

pkg/server/container_start.go
yujuhong: Ack

pkg/server/helpers.go
mikebrow: yeah just odd using a "" formatter inside a join... two styles of formatting for the same string

pkg/server/container_start.go
yujuhong: Ack

pkg/server/container_stop.go
yujuhong: If we can't find a reason why we should not time out, I'd add the timeout.

pkg/server/events.go
mikebrow: cool..

pkg/server/container_stop.go
yujuhong: They are not completely the same because you only wake up every stopCheckPollInterval, so you could already be past the timeout when waking up :-)
Up to you.

pkg/server/container_stop.go
yujuhong: That's what I meant: subscribing to the events from the monitor via a channel?

yujuhong:
Looks overall except for the forever timeout. I don't understand why we need it, so I'd rather not have that. At least file an issue if you'd like to address that later.

Also, didn't get to review the unit tests yet. I can review early Monday morning.

Random-Liu:
> Also, didn't get to review the unit tests yet. I can review early Monday morning.

SGTM. Thanks for reviewing! :)

pkg/server/container_stop.go
Random-Liu: I see, makes sense. Will update.

pkg/server/container_stop.go
Random-Liu: Yeah, updated the comment.

pkg/server/container_stop.go
Random-Liu: Done. Set default timeout to 2 minutes.

pkg/server/events_test.go
```diff
@@ -0,0 +1,165 @@
...
+		"should not update state when exited process is not init process": {
+			event: &container.Event{
+				ID:         testID,
+				Type:       container.Event_EXIT,
+				Pid:        9999,
+				ExitStatus: 1,
+				ExitedAt:   testExitedAt,
+			},
+			metadata: &testMetadata,
+			containerdContainer: &container.Container{
```
Random-Liu: Will do.

yujuhong:
LGTM with nits.

pkg/server/container_list_test.go
```diff
@@ -0,0 +1,227 @@
...
+			Id:           "2",
+			PodSandboxId: "s-2",
+			Metadata:     &runtime.ContainerMetadata{Name: "name-2", Attempt: 2},
+			State:        runtime.ContainerState_CONTAINER_EXITED,
+			Labels:       map[string]string{"a": "b"},
+		},
+		{
+			Id:           "3",
+			PodSandboxId: "s-2",
+			Metadata:     &runtime.ContainerMetadata{Name: "name-2", Attempt: 2},
```
Random-Liu: True. Will do.

pkg/server/container_list_test.go
```diff
@@ -0,0 +1,227 @@
...
+	} {
+		filtered := c.filterCRIContainers(testContainers, test.filter)
+		assert.Equal(t, test.expect, filtered, desc)
+	}
+}
+
+func TestListContainers(t *testing.T) {
+	c := newTestCRIContainerdService()
+
+	now := time.Now().UnixNano()
```
Random-Liu: Will do.

pkg/server/container_status_test.go
```diff
@@ -0,0 +1,173 @@
...
+		Reason:      completeExitReason,
+		Labels:      config.GetLabels(),
+		Annotations: config.GetAnnotations(),
+		Mounts:      config.GetMounts(),
+	}
+
+	return metadata, expected
+}
+
+func TestToCRIContainerStatus(t *testing.T) {
```
Random-Liu: ? They are added in getContainerStatusTestData, right?

pkg/server/events_test.go
Random-Liu: Done.

pkg/server/container_list_test.go
Random-Liu: Done.

pkg/server/container_list_test.go
Random-Liu: Done.

pkg/server/container_status_test.go
yujuhong: ack

Random-Liu:
@yujuhong Addressed comments. Will squash after LGTM.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/22#issuecomment-303179768.

Issue #20, opened 2017-04-27
We need fake containerd services.


Random-Liu:
To better unit test functions of cri-containerd, we need a group of fake containerd services (similar with [fake docker client](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/fake_docker_client.go))

Containerd now has 4 kinds of services:
* content
* execution
* images
* rootfs

We can add a fake execution service first, because the execution api is more stable. And add others later when image management code is added and needs test.

/cc @kubernetes-incubator/maintainers-cri-containerd 
@heartlock Are you interested in this? This is a good start to get familiar with containerd and the code base.

heartlock:
yes, I will take this task

Random-Liu:
@heartlock Thanks!

Random-Liu:
I assigned this to @xlgao-zju, because he's in the org now.

/cc @heartlock @Helen-Xie Please take a look.

Helen-Xie:
ok, I will do it.

Random-Liu:
I found that image store service and rootfs service are easier to fake, but content store is a little bit harder. And I'm not sure whether it's useful for us to fake at the content store level, given its interface is just every low level read write.

Let's only work on the image store and rootfs service now. For the content store, we could fake at higher level in the future.

Random-Liu:
Just FYI, containerd is adding a new client library https://github.com/containerd/containerd/pull/904.

We should integrate and test at that level after it lands. It will be much easier.

Random-Liu:
We'll add new fake of the containerd client after the client switch, punt to v0.2.0.

Random-Liu:
This is unit test specific change. May not have time to get to it this release.

Random-Liu:
As is mentioned in https://github.com/containerd/cri-containerd/issues/103. I think we have enough test coverage today.

We can add more integration test instead of adding fake service.

PR #23, opened 2017-05-03
Add fake execution service

heartlock:
Signed-off-by: heartlock <21521209@zju.edu.cn>
ref: #20


Random-Liu:
@heartlock Will take a look today.

pkg/server/fake_execution_client.go
```diff
@@ -0,0 +1,238 @@
...
+	return event, nil
+}
+
+// FakeExecutionClient is a simple fake execution client, so that cri-containerd can be run for testing without requiring a real containerd setup.
+type FakeExecutionClient struct {
+	sync.Mutex
+	called        []calledDetail
+	ContainerList map[string]*container.Container
+	EventsQueue   chan *container.Event
+	EventClients  map[*EventClient]struct{}
```
Random-Liu: An array is enough.

pkg/server/fake_execution_client.go
```diff
@@ -0,0 +1,238 @@
...
+	event := <-cli.Events
+	return event, nil
+}
+
+// FakeExecutionClient is a simple fake execution client, so that cri-containerd can be run for testing without requiring a real containerd setup.
+type FakeExecutionClient struct {
+	sync.Mutex
+	called        []calledDetail
+	ContainerList map[string]*container.Container
+	EventsQueue   chan *container.Event
```
Random-Liu: `EventsQueue` and `EventClients` could be private.

pkg/server/fake_execution_client.go
```diff
@@ -0,0 +1,238 @@
...
+
+// NewFakeExecutionClient new a FakeExecutionClient
+func NewFakeExecutionClient() *FakeExecutionClient {
+	return &FakeExecutionClient{}
+}
+func (f *FakeExecutionClient) withEvents() {
+	f.EventsQueue = make(chan *container.Event, 1024)
+	go func() {
+		for e := range f.EventsQueue {
+			for client := range f.EventClients {
```
Random-Liu: We may need a lock here. Array or map is not thread safe.

pkg/server/fake_execution_client.go
```diff
@@ -0,0 +1,238 @@
...
+func (f *FakeExecutionClient) getCalledNames() []string {
+	names := []string{}
+	for _, detail := range f.called {
+		names = append(names, detail.name)
+	}
+	return names
+}
+
+// Create is test implementation of execution.Create.
+func (f *FakeExecutionClient) Create(ctx context.Context, createOpts *execution.CreateRequest) (*execution.CreateResponse, error) {
```
Random-Liu: We should support injecting error for each call.

heartlock:
@Random-Liu updated. 

1. Fix nits and make vendor package match containerd version
2. support inject error for calls

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,315 @@
...
+		ContainerList: make(map[string]*container.Container),
+	}
+}
+
+// WithEvents setup events publisher for FakeExecutionClient
+func (f *FakeExecutionClient) WithEvents() {
+	f.eventsQueue = make(chan *container.Event, 1024)
+	go func() {
+		for e := range f.eventsQueue {
+			for _, client := range f.eventClients {
```
Random-Liu: I mean you need to lock the whole for clause.

Or else there will be a race between this routine and `Events()`.

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,315 @@
...
+		f.eventsQueue <- event
+	}
+}
+
+func (f *FakeExecutionClient) appendCalled(name string, argument interface{}) {
+	call := calledDetail{name: name, argument: argument}
+	f.called = append(f.called, call)
+}
+
+func (f *FakeExecutionClient) getCalledNames() []string {
```
Random-Liu: Should this be public?

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,315 @@
...
+
+// Info is a test implementation of execution.Info
+func (f *FakeExecutionClient) Info(ctx context.Context, infoOpts *execution.InfoRequest, opts ...grpc.CallOption) (*container.Container, error) {
+	f.Lock()
+	defer f.Unlock()
+	f.appendCalled("info", infoOpts)
+	if err := f.popError("info"); err != nil {
+		return nil, err
+	}
+	c, ok := f.ContainerList[infoOpts.ID]
```
Random-Liu: Same problem here. How about just change `ContainerList` to `map[string]container.Container`? We don't actually need a pointer.

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,315 @@
...
+// NewFakeExecutionClient creates a FakeExecutionClient
+func NewFakeExecutionClient() *FakeExecutionClient {
+	return &FakeExecutionClient{
+		Errors:        make(map[string]error),
+		ContainerList: make(map[string]*container.Container),
+	}
+}
+
+// WithEvents setup events publisher for FakeExecutionClient
+func (f *FakeExecutionClient) WithEvents() {
```
Random-Liu: return the `*FakeExecutionClient`, we usually use `WithXXX` in this way:
```
f := NewFakeExecutionClient().WithAAA().WithBBB().
```

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,315 @@
...
+	event := <-cli.Events
+	return event, nil
+}
+
+// FakeExecutionClient is a simple fake execution client, so that cri-containerd
+// can be run for testing without requiring a real containerd setup.
+type FakeExecutionClient struct {
+	sync.Mutex
+	called        []calledDetail
+	Errors        map[string]error
```
Random-Liu: Make this private because we already have corresponding inject functions.

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,315 @@
...
+func (cli *EventClient) Recv() (*container.Event, error) {
+	event := <-cli.Events
+	return event, nil
+}
+
+// FakeExecutionClient is a simple fake execution client, so that cri-containerd
+// can be run for testing without requiring a real containerd setup.
+type FakeExecutionClient struct {
+	sync.Mutex
+	called        []calledDetail
```
Random-Liu: Actually, we need assert functions to check ContainerList and called list. However, I'm fine with adding them in the future when we actually use them.

pkg/server/testing/fake_execution_client.go
heartlock: I think it should be

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,313 @@
...
+func (f *FakeExecutionClient) List(ctx context.Context, listOpts *execution.ListRequest, opts ...grpc.CallOption) (*execution.ListResponse, error) {
+	f.Lock()
+	defer f.Unlock()
+	f.appendCalled("list", listOpts)
+	if err := f.popError("list"); err != nil {
+		return nil, err
+	}
+	resp := &execution.ListResponse{}
+	for _, c := range f.ContainerList {
+		resp.Containers = append(resp.Containers, &c)
```
Random-Liu: You may not want to get a pointer of the iterator.

Try this https://play.golang.org/p/6Au4lw5Zi6

pkg/server/testing/fake_execution_client.go
```diff
@@ -0,0 +1,317 @@
...
+func (f *FakeExecutionClient) List(ctx context.Context, listOpts *execution.ListRequest, opts ...grpc.CallOption) (*execution.ListResponse, error) {
+	f.Lock()
+	defer f.Unlock()
+	f.appendCalled("list", listOpts)
+	if err := f.popError("list"); err != nil {
+		return nil, err
+	}
+	resp := &execution.ListResponse{}
+	for _, c := range f.ContainerList {
+		resp.Containers = append(resp.Containers, &container.Container{
```
Random-Liu: Actually, you could just change this to:
```
cntr := c
resp.Containers = append(resp.Containers, &cntr)
```
However, I'm fine with this, too.

Random-Liu:
LGTM

Issue #24, opened 2017-05-03
We need a fake OS.


Random-Liu:
We need a fake OS for unit test, similar with https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/container/testing/os.go.

Random-Liu:
I'll pick up this one and send a PR soon.

PR #25, opened 2017-05-03
Add OS interface and fake implementation.

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/24.

Just a very small PR adding the OS interface for unit testing. So far, I mainly used `MkdirAll` and `RemoveAll` in the sandbox and container lifecycle PRs. So only added those 2 functions for now. We could always extend in the future.

I'm not using mockgen because the code generated could not pass lint...

### DIFF
--- a/pkg/os/os.go
+++ b/pkg/os/os.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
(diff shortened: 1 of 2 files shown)


Random-Liu:
@mikebrow Thanks for reviewing!

PR #38, opened 2017-05-15
Return not exist error in metadata store

Random-Liu:
Let metadata store return not exist error instead of `nil, nil`.

We made this change to:
1) Make it more convenient and intuitive to use the metadata store;
2) We need the not exist error for `Update` function in container lifecycle management.

### DIFF
--- a/pkg/metadata/image_metadata_test.go
+++ b/pkg/metadata/image_metadata_test.go
@@ -80,8 +80,8 @@ func TestImageMetadataStore(t *testing.T) {
 	assert.NoError(err)
 	assert.Len(imgs, 2)
 
-	t.Logf("get should return nil without error after deletion")
+	t.Logf("get should return nil not exist error after deletion")
 	meta, err := s.Get(testID)
-	assert.NoError(err)
+	assert.Error(store.ErrNotExist, err)
 	assert.Nil(meta)
 }
--- a/pkg/metadata/sandbox.go
+++ b/pkg/metadata/sandbox.go
@@ -115,11 +115,6 @@ func (s *sandboxStore) Get(sandboxID string) (*SandboxMetadata, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Return nil without error if the corresponding metadata
-	// does not exist.
-	if data == nil {
-		return nil, nil
-	}
 	sandbox := &SandboxMetadata{}
 	if err := json.Unmarshal(data, sandbox); err != nil {
 		return nil, err
--- a/pkg/metadata/sandbox_test.go
+++ b/pkg/metadata/sandbox_test.go
@@ -113,8 +113,8 @@ func TestSandboxStore(t *testing.T) {
 	assert.NoError(err)
 	assert.Len(sbs, 2)
 
-	t.Logf("get should return nil without error after deletion")
+	t.Logf("get should return nil with not exist error after deletion")
 	meta, err := s.Get(testID)
-	assert.NoError(err)
+	assert.Error(store.ErrNotExist, err)
 	assert.Nil(meta)
 }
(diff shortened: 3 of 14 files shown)


pkg/server/sandbox_remove.go
```diff
@@ -39,9 +41,9 @@ func (c *criContainerdService) RemovePodSandbox(ctx context.Context, r *runtime.
 
 	sandbox, err := c.getSandbox(r.GetPodSandboxId())
 	if err != nil {
-		return nil, fmt.Errorf("failed to find sandbox %q: %v", r.GetPodSandboxId(), err)
-	}
-	if sandbox == nil {
+		if !metadata.IsNotExistError(err) {
```
yujuhong: When will the metadata store return an error that is not "NotFound" for a Get request? The "failed to find sandbox" error message below seem to suggest it's the NotFound error anyway...

mikebrow:
Cool.. I prefer it this way anyway.

pkg/metadata/image_metadata_test.go
```diff
@@ -82,6 +82,6 @@ func TestImageMetadataStore(t *testing.T) {
 
 	t.Logf("get should return nil without error after deletion")
```
mikebrow: /s/without/with/

pkg/metadata/sandbox_test.go
```diff
@@ -115,6 +115,6 @@ func TestSandboxStore(t *testing.T) {
 
 	t.Logf("get should return nil without error after deletion")
```
mikebrow: /s/without/with/

pkg/metadata/sandbox.go
```diff
@@ -115,11 +115,6 @@ func (s *sandboxStore) Get(sandboxID string) (*SandboxMetadata, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Return nil without error if the corresponding metadata
-	// does not exist.
-	if data == nil {
-		return nil, nil
```
mikebrow: just making sure there's no chance for data to be a nil here? otherwise you'll get the json unmarshal error

pkg/server/sandbox_remove.go
Random-Liu: Currently, `Get` returns error only when the metadata does not exist.
I could change it to `an error occurred when try to find sandbox %q: %v`. Is that better?

Random-Liu: Currently, Get returns error only when the metadata does not exist.
I could change it to an error occurred when try to find sandbox %q: %v. Is that better?

yujuhong: Sounds good

pkg/metadata/image_metadata_test.go
Random-Liu: Will do. Good catch

pkg/metadata/sandbox_test.go
Random-Liu: Will do. Good catch

pkg/server/sandbox_remove.go
Random-Liu: Done.

pkg/metadata/image_metadata_test.go
Random-Liu: Done

pkg/metadata/sandbox_test.go
Random-Liu: Done

pkg/metadata/sandbox.go
Random-Liu: Get json unmarshal error is expected behavior if the data is invalid. :)

Random-Liu:
@yujuhong @mikebrow Addressed comments.

yujuhong:
lgtm

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/38#issuecomment-301899133

pkg/metadata/sandbox.go
mikebrow: yes.. it's nil.. just checking to make sure that's the error you wanted vs ErrNotExist

PR #40, opened 2017-05-17
Add network plugin

### LINKED ISSUE #27: Integrate with kubernetes network plugin.
CRI expects the runtime to handle container networking.

As is expected, containerd itself doesn't provide container networking. We should integrate cri-containerd with kubernetes network plugin to initialize the network namespace.

Things need to do:
1) Add flags for network plugin settings and initialize network plugin in cri-containerd. (See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_service.go#L197-L204)
2) Implement interface needed by network plugin, include `namespaceGetter` and `portMappingGetter`. (See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_service.go#L119-L137)
3) Setup and teardown network when run/stop pod sandbox. (See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_sandbox.go#L128)

/cc @xlgao-zju

xlgao-zju:
- [x] Add flags for network plugin settings and initialize network plugin in cri-containerd. 
- [x] Setup and teardown network when run/stop pod sandbox. 
- [x] And `GetContainerNetworkStatus` in PodSandboxStatus in order to get IP.
- [x] And some fake function to make travis happy.

close #27 
Signed-off-by: Xianglin Gao <xlgao@zju.edu.cn>

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -147,6 +147,26 @@
 			"Rev": "bbeaab5ee38568daa84b8c233a8067b524588355"
 		},
 		{
+			"ImportPath": "github.com/containernetworking/cni/libcni",
+			"Comment": "v0.4.0",
+			"Rev": "530dd71e193012895ff7beee42cd64a02e1993da"
+		},
+		{
+			"ImportPath": "github.com/containernetworking/cni/pkg/invoke",
+			"Comment": "v0.4.0",
+			"Rev": "530dd71e193012895ff7beee42cd64a02e1993da"
+		},
+		{
+			"ImportPath": "github.com/containernetworking/cni/pkg/types",
+			"Comment": "v0.4.0",
+			"Rev": "530dd71e193012895ff7beee42cd64a02e1993da"
+		},
+		{
+			"ImportPath": "github.com/containernetworking/cni/pkg/version",
+			"Comment": "v0.4.0",
+			"Rev": "530dd71e193012895ff7beee42cd64a02e1993da"
+		},
+		{
 			"ImportPath": "github.com/davecgh/go-spew/spew",
 			"Comment": "v1.1.0",
 			"Rev": "346938d642f2ec3594ed81d874461961cd0faa76"
@@ -175,6 +195,11 @@
 			"ImportPath": "github.com/docker/docker/pkg/truncindex",
 			"Comment": "v1.13.1",
 			"Rev": "092cba3727bb9b4a2f0e922cd6c0f93ea270e363"
+		},
+		{
+			"ImportPath": "github.com/fsnotify/fsnotify",
+			"Comment": "v1.4.2-4-g7d7316e",
+			"Rev": "7d7316ed6e1ed2de075aab8dfc76de5d158d66e1"
 		},
 		{
 			"ImportPath": "github.com/gogo/protobuf/gogoproto",
@@ -212,6 +237,10 @@
 		{
 			"ImportPath": "github.com/golang/protobuf/ptypes/empty",
 			"Rev": "8ee79997227bf9b34611aee7946ae64735e6fd93"
+		},
+		{
+			"ImportPath": "github.com/kubernetes-incubator/cri-o/pkg/ocicni",
+			"Rev": "f648cd6e60948e4da391040e5c75d8175fea4fb7"
 		},
 		{
 			"ImportPath": "github.com/nightlyone/lockfile",
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -36,6 +36,10 @@ type CRIContainerdOptions struct {
 	ContainerdEndpoint string
 	// ContainerdConnectionTimeout is the connection timeout for containerd client.
 	ContainerdConnectionTimeout time.Duration
+	// NetworkPluginBinDir is the directory in which the binaries for the plugin is kept.
+	NetworkPluginBinDir string
+	// NetworkPluginConfDir is the directory in which the admin places a CNI conf.
+	NetworkPluginConfDir string
 }
 
 // NewCRIContainerdOptions returns a reference to CRIContainerdOptions
@@ -55,6 +59,10 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 		2*time.Minute, "Connection timeout for containerd client.")
 	fs.BoolVar(&c.PrintVersion, "version",
 		false, "Print cri-containerd version information and quit.")
+	fs.StringVar(&c.NetworkPluginBinDir, "network-bin-dir",
+		"/etc/cni/net.d", "The directory for putting network binaries.")
+	fs.StringVar(&c.NetworkPluginConfDir, "network-conf-dir",
+		"/opt/cni/bin", "The directory for putting network plugin configuration files.")
 }
 
 // InitFlags must be called after adding all cli options flags are defined and
--- a/pkg/os/testing/fake_os.go
+++ b/pkg/os/testing/fake_os.go
@@ -19,6 +19,7 @@ package testing
 import (
 	"io"
 	"os"
+	"sync"
 
 	"golang.org/x/net/context"
 
@@ -29,20 +30,64 @@ import (
 // If a member of the form `*Fn` is set, that function will be called in place
 // of the real call.
 type FakeOS struct {
+	sync.Mutex
 	MkdirAllFn  func(string, os.FileMode) error
 	RemoveAllFn func(string) error
 	OpenFifoFn  func(context.Context, string, int, os.FileMode) (io.ReadWriteCloser, error)
+	StatFn      func(name string) (os.FileInfo, error)
+	errors      map[string]error
 }
 
 var _ osInterface.OS = &FakeOS{}
 
+// getError get error for call
+func (f *FakeOS) getError(op string) error {
+	f.Lock()
+	defer f.Unlock()
+	err, ok := f.errors[op]
+	if ok {
+		delete(f.errors, op)
+		return err
+	}
+	return nil
+}
+
+// InjectError inject error for call
+func (f *FakeOS) InjectError(fn string, err error) {
+	f.Lock()
+	defer f.Unlock()
+	f.errors[fn] = err
+}
+
+// InjectErrors inject errors for calls
+func (f *FakeOS) InjectErrors(errs map[string]error) {
+	f.Lock()
+	defer f.Unlock()
+	for fn, err := range errs {
+		f.errors[fn] = err
+	}
+}
+
+// ClearErrors clear errors for call
+func (f *FakeOS) ClearErrors() {
+	f.Lock()
+	defer f.Unlock()
+	f.errors = make(map[string]error)
+}
+
 // NewFakeOS creates a FakeOS.
 func NewFakeOS() *FakeOS {
-	return &FakeOS{}
+	return &FakeOS{
+		errors: make(map[string]error),
+	}
 }
 
 // MkdirAll is a fake call that invokes MkdirAllFn or just returns nil.
 func (f *FakeOS) MkdirAll(path string, perm os.FileMode) error {
+	if err := f.getError("MkdirAll"); err != nil {
+		return err
+	}
+
 	if f.MkdirAllFn != nil {
 		return f.MkdirAllFn(path, perm)
 	}
@@ -51,6 +96,10 @@ func (f *FakeOS) MkdirAll(path string, perm os.FileMode) error {
 
 // RemoveAll is a fake call that invokes RemoveAllFn or just returns nil.
 func (f *FakeOS) RemoveAll(path string) error {
+	if err := f.getError("RemoveAll"); err != nil {
+		return err
+	}
+
 	if f.RemoveAllFn != nil {
 		return f.RemoveAllFn(path)
 	}
@@ -59,8 +108,24 @@ func (f *FakeOS) RemoveAll(path string) error {
 
 // OpenFifo is a fake call that invokes OpenFifoFn or just returns nil.
 func (f *FakeOS) OpenFifo(ctx context.Context, fn string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
+	if err := f.getError("OpenFifo"); err != nil {
+		return nil, err
+	}
+
 	if f.OpenFifoFn != nil {
 		return f.OpenFifoFn(ctx, fn, flag, perm)
+	}
+	return nil, nil
+}
+
+// Stat is a fake call that invokes Stat or just return nil.
+func (f *FakeOS) Stat(name string) (os.FileInfo, error) {
+	if err := f.getError("Stat"); err != nil {
+		return nil, err
+	}
+
+	if f.StatFn != nil {
+		return f.StatFn(name)
 	}
 	return nil, nil
 }
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -86,8 +86,6 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	// Use fixed rootfs path and sleep command.
 	const rootPath = "/"
 
-	// TODO(random-liu): [P0] Set up sandbox network with network plugin.
-
 	// Create sandbox container root directory.
 	// Prepare streaming named pipe.
 	sandboxRootDir := getSandboxRootDir(c.rootDir, id)
@@ -173,6 +171,24 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 		}
 	}()
 
+	meta.NetNS = getNetworkNamespace(createResp.Pid)
+	if !config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
+		// Setup network for sandbox.
+		// TODO(random-liu): [P2] Replace with permanent network namespace.
+		podName := config.GetMetadata().GetName()
+		if err = c.netPlugin.SetUpPod(meta.NetNS, config.GetMetadata().GetNamespace(), podName, id); err != nil {
+			return nil, fmt.Errorf("failed to setup network for sandbox %q: %v", id, err)
+		}
+		defer func() {
+			if retErr != nil {
+				// Teardown network if an error is returned.
+				if err := c.netPlugin.TearDownPod(meta.NetNS, config.GetMetadata().GetNamespace(), podName, id); err != nil {
+					glog.Errorf("failed to destroy network for sandbox %q: %v", id, err)
+				}
+			}
+		}()
+	}
+
 	// Start sandbox container in containerd.
 	if _, err := c.containerService.Start(ctx, &execution.StartRequest{ID: id}); err != nil {
 		return nil, fmt.Errorf("failed to start sandbox container %q: %v",
@@ -181,8 +197,6 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 
 	// Add sandbox into sandbox store.
 	meta.CreatedAt = time.Now().UnixNano()
-	// TODO(random-liu): [P2] Replace with permanent network namespace.
-	meta.NetNS = getNetworkNamespace(createResp.Pid)
 	if err := c.sandboxStore.Create(meta); err != nil {
 		return nil, fmt.Errorf("failed to add sandbox metadata %+v into store: %v",
 			meta, err)
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -164,7 +164,7 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	defer func() {
 		if retErr != nil {
 			// Cleanup the sandbox container if an error is returned.
-			if _, err := c.containerService.Delete(ctx, &execution.DeleteRequest{ID: id}); err != nil {
+			if _, err = c.containerService.Delete(ctx, &execution.DeleteRequest{ID: id}); err != nil {
 				glog.Errorf("Failed to delete sandbox container %q: %v",
 					id, err)
 			}
--- a/pkg/server/sandbox_status.go
+++ b/pkg/server/sandbox_status.go
@@ -59,11 +59,18 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
 		state = runtime.PodSandboxState_SANDBOX_READY
 	}
 
-	return &runtime.PodSandboxStatusResponse{Status: toCRISandboxStatus(sandbox, state)}, nil
+	ip, err := c.netPlugin.GetContainerNetworkStatus(sandbox.NetNS, sandbox.Config.GetMetadata().GetNamespace(), sandbox.Config.GetMetadata().GetName(), id)
+	if err != nil {
+		// Ignore the error on network status
+		ip = ""
+		glog.V(4).Infof("GetContainerNetworkStatus returns error: %v", err)
+	}
+
+	return &runtime.PodSandboxStatusResponse{Status: toCRISandboxStatus(sandbox, state, ip)}, nil
 }
 
 // toCRISandboxStatus converts sandbox metadata into CRI pod sandbox status.
-func toCRISandboxStatus(meta *metadata.SandboxMetadata, state runtime.PodSandboxState) *runtime.PodSandboxStatus {
+func toCRISandboxStatus(meta *metadata.SandboxMetadata, state runtime.PodSandboxState, ip string) *runtime.PodSandboxStatus {
 	nsOpts := meta.Config.GetLinux().GetSecurityContext().GetNamespaceOptions()
 	netNS := meta.NetNS
 	if state == runtime.PodSandboxState_SANDBOX_NOTREADY {
@@ -79,8 +86,7 @@ func toCRISandboxStatus(meta *metadata.SandboxMetadata, state runtime.PodSandbox
 		Metadata:  meta.Config.GetMetadata(),
 		State:     state,
 		CreatedAt: meta.CreatedAt,
-		// TODO(random-liu): [P0] Get sandbox ip from network plugin.
-		Network: &runtime.PodSandboxNetworkStatus{},
+		Network:   &runtime.PodSandboxNetworkStatus{Ip: ip},
 		Linux: &runtime.LinuxPodSandboxStatus{
 			Namespaces: &runtime.Namespace{
 				// TODO(random-liu): Revendor new CRI version and get
--- a/pkg/server/sandbox_status_test.go
+++ b/pkg/server/sandbox_status_test.go
@@ -35,7 +35,11 @@ import (
 
 // Variables used in the following test.
 
-const sandboxStatusTestID = "test-id"
+const (
+	sandboxStatusTestID    = "test-id"
+	sandboxStatusTestIP    = "10.10.10.10"
+	sandboxStatusTestNetNS = "test-netns"
+)
 
 func getSandboxStatusTestData() (*metadata.SandboxMetadata, *runtime.PodSandboxStatus) {
 	config := &runtime.PodSandboxConfig{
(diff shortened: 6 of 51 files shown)


xlgao-zju:
I find there is something conflicts in `github.com/docker/distribution/digest`, when I update the vendor. I will update vendor when we fix it.

/cc @feiskyer @Random-Liu

Random-Liu:
@xlgao-zju Talked with @freehan and @yujuhong, the kubernetes network plugin introduces too many unnecessary dependencies, which makes the code more complicated and may cause version conflict.

Could you take a look at https://github.com/kubernetes-incubator/cri-o/tree/master/pkg/ocicni? Let's use it for our alpha integration, which is simpler and cleaner.

xlgao-zju:
@Random-Liu I reference the logic in `cri-o` to call the package `kubernetes-incubator/cri-o/pkg/ocicni`. And this PR is ready for review now. :)

pkg/server/sandbox_stop.go
```diff
@@ -45,14 +46,24 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
 	// Use the full sandbox id.
 	id := sandbox.ID
 
+	// Teardown network.
+	if _, err = os.Stat(sandbox.NetNS); err == nil {
```
Random-Liu: Add unit test for this, just:
1) When Stat returns err, `StopPodSandbox` should return error, and `TearDownPod` should not be called;
2) When Stat returns not exist err, `StopPodSandbox` should not return error, and `TearDownPod` should not be called;
3) When Stat doesn't return error, `TearDownPod` should be called, if `TearDownPod` fails, `StopPodSandbox` should return error;
4) When Stat doesn't return error, `TearDownPod` should be called, if `TearDownPod` succeeds, `StopPodSandbox` should succeed.

You can add the test case into current `StopPodSandbox` unit test. This is not overkill, because the cleanup logic is critical.

pkg/server/service.go
```diff
@@ -88,15 +91,17 @@ type criContainerdService struct {
 	// imageStoreService is the containerd service to store and track
 	// image metadata.
 	imageStoreService images.Store
+	// netPlugin is used to setup and teardown network when run/stop pod sandbox.
+	netPlugin ocicni.CNIPlugin
 }
 
 // NewCRIContainerdService returns a new instance of CRIContainerdService
-func NewCRIContainerdService(conn *grpc.ClientConn, rootDir string) CRIContainerdService {
+func NewCRIContainerdService(conn *grpc.ClientConn, opt *options.CRIContainerdOptions) (CRIContainerdService, error) {
```
Random-Liu: I prefer passes in parameters one by one, or else in the future it will be hard to figure out which option is used in containerd service, which is not.

pkg/server/service.go
```diff
@@ -108,4 +113,12 @@ func NewCRIContainerdService(conn *grpc.ClientConn, rootDir string) CRIContainer
 		contentProvider:   contentservice.NewProviderFromClient(contentapi.NewContentClient(conn)),
 		rootfsUnpacker:    rootfsservice.NewUnpackerFromClient(rootfsapi.NewRootFSClient(conn)),
 	}
+
+	netPlugin, err := ocicni.InitCNI(opt.NetworkPluginBinDir, opt.NetworkPluginConfDir)
+	if err != nil {
+		return nil, err
```
Random-Liu: fmt.Errorf("failed to initialize cni plugin: %v", err)

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+package testing
+
+import (
+	"sync"
+
+	"github.com/kubernetes-incubator/cri-o/pkg/ocicni"
+)
+
+// FakeCNIPlugin is a fake plugin used for test.
+type FakeCNIPlugin struct {
```
Random-Liu: Take a look at https://github.com/xlgao-zju/cri-containerd/blob/d9a3854d2595e98604a3ef6a46de11b9a68d992c/pkg/server/testing/fake_execution_client.go, we need to do similar thing.

All functions should track down call details.
All functions should be able to inject error.
We can use some test mock generator, but it's simple enough to write ourselves.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+// Name return the name of fake CNI plugin.
+func (f *FakeCNIPlugin) Name() string {
+	return "fake-CNI-plugin"
+}
+
+// SetUpPod setup the network of PodSandbox.
+func (f *FakeCNIPlugin) SetUpPod(netnsPath string, namespace string, name string, containerID string) error {
+	f.Lock()
+	defer f.Unlock()
+	return nil
```
Random-Liu: Generate a random ip for the pod.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// TearDownPod teardown the network of PodSandbox.
+func (f *FakeCNIPlugin) TearDownPod(netnsPath string, namespace string, name string, containerID string) error {
+	f.Lock()
+	defer f.Unlock()
```
Random-Liu: Release the ip.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+// GetContainerNetworkStatus get the status of network.
+func (f *FakeCNIPlugin) GetContainerNetworkStatus(netnsPath string, namespace string, name string, containerID string) (string, error) {
+	f.Lock()
+	defer f.Unlock()
+	return "", nil
```
Random-Liu: Return the ip in the internal ip map.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Status get the status of the plugin.
+func (f *FakeCNIPlugin) Status() error {
+	return nil
+}
+
+// AddIP add an IP in the IPMap.
+func (f *FakeCNIPlugin) AddIP(podID, IP string) {
```
Random-Liu: `SetFakePodNetwork` this could used to inject fake pod ip.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+	f.Lock()
+	defer f.Unlock()
+	f.IPMap[podID] = IP
+}
+
+// DeleteIP delete an IP in the IPMap.
+func (f *FakeCNIPlugin) DeleteIP(podID string) {
```
Random-Liu: May not need this for now.

pkg/server/sandbox_run.go
```diff
@@ -180,10 +180,24 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 			id, err)
 	}
 
-	// Add sandbox into sandbox store.
-	meta.CreatedAt = time.Now().UnixNano()
+	// Setup network.
 	// TODO(random-liu): [P2] Replace with permanent network namespace.
 	meta.NetNS = getNetworkNamespace(createResp.Pid)
+	if err = c.netPlugin.SetUpPod(meta.NetNS, config.GetMetadata().GetNamespace(), name, id); err != nil {
```
Random-Liu: Add unit test for this. The unit test of `RunPodSandbox` is simple now, just assert `SetUpPod` is called with correct namespace, name and id.

We'll need to test cleanup in future PRs.

Godeps/Godeps.json
```diff
@@ -147,18 +147,38 @@
 			"Rev": "bbeaab5ee38568daa84b8c233a8067b524588355"
 		},
 		{
+			"ImportPath": "github.com/containernetworking/cni/libcni",
+			"Comment": "v0.3.0-61-g9d5e6e6",
```
xlgao-zju: This is imported by the `cri-o`, so we I think we should let the version be even with the version which `cri-o` is using.

Godeps/Godeps.json
```diff
@@ -214,6 +239,10 @@
 			"Rev": "8ee79997227bf9b34611aee7946ae64735e6fd93"
 		},
 		{
+			"ImportPath": "github.com/kubernetes-incubator/cri-o/pkg/ocicni",
+			"Rev": "0e2af44d72ec7b79cd14f2489a4fcaf54701b53f"
```
xlgao-zju: will do.

cmd/cri-containerd/options/options.go
```diff
@@ -36,6 +36,14 @@ type CRIContainerdOptions struct {
 	ContainerdEndpoint string
 	// ContainerdConnectionTimeout is the connection timeout for containerd client.
 	ContainerdConnectionTimeout time.Duration
+	// NetworkPluginBinDir is the directory in which the binaries for the plugin with
+	// PluginName is kept. The admin is responsible for provisioning these
+	// binaries before-hand.
```
xlgao-zju: Will fix.

cmd/cri-containerd/options/options.go
```diff
@@ -36,6 +36,14 @@ type CRIContainerdOptions struct {
...
+	NetworkPluginBinDir string
+	// NetworkPluginConfDir is the directory in which the admin places a CNI conf.
+	// Depending on the plugin, this may be an optional field, eg: kubenet
```
xlgao-zju: will fix.

pkg/server/sandbox_run.go
xlgao-zju: So, should use `config.GetMetadata().GetName()`?

xlgao-zju: So, should use `config.GetMetadata().GetName()`?

xlgao-zju: Why?

pkg/server/sandbox_run.go
```diff
@@ -180,10 +180,24 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+		return nil, fmt.Errorf("failed to setup network %v", err)
+	}
+	defer func() {
+		if retErr != nil {
+			// Teardown network if an error is returned.
+			if err := c.netPlugin.TearDownPod(meta.NetNS, config.GetMetadata().GetNamespace(), name, id); err != nil {
+				glog.Errorf("failed to destroy network for container %s in sandbox %s: %v",
```
xlgao-zju: Wonder what's the difference between `%s` and `%q`...

pkg/server/sandbox_status_test.go
```diff
@@ -71,7 +71,7 @@ func getSandboxStatusTestData() (*metadata.SandboxMetadata, *runtime.PodSandboxS
 		Id:        sandboxStatusTestID,
 		Metadata:  config.GetMetadata(),
 		CreatedAt: createdAt,
-		Network:   &runtime.PodSandboxNetworkStatus{},
+		Network:   &runtime.PodSandboxNetworkStatus{Ip: ""},
```
xlgao-zju: Will do.

pkg/server/sandbox_stop.go
xlgao-zju: Yep, that is cleaner.

xlgao-zju: Will do.

xlgao-zju: Cool, that is comprehensive. And this will make the cleanup reliable.

pkg/server/service.go
xlgao-zju: But we will have too may parameters when we call `NewCRIContainerdService ` in the future.

pkg/server/service.go
xlgao-zju: Will do.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"sync"
+
+	"github.com/kubernetes-incubator/cri-o/pkg/ocicni"
+)
+
+// FakeCNIPlugin is a fake plugin used for test.
+type FakeCNIPlugin struct {
+	sync.Mutex
+	IPMap map[string]string
```
xlgao-zju: Which type should the key of `IPMap` be? `netnsPath` or a struct like this?
```
type IPMapKey struct {
netnsPath string
namespace string
name string
containerID string
}
```

pkg/server/testing/fake_cni_plugin.go
xlgao-zju: will do

pkg/server/testing/fake_cni_plugin.go
xlgao-zju: Use a random number as a random IP like`123454321`? Or should be the IP form like `123.456.789.123`? I think a random number maybe enough...

pkg/server/sandbox_stop.go
xlgao-zju: When should `Stat` return err?

pkg/server/sandbox_run.go
xlgao-zju: We start the container first in the dockershim...

xlgao-zju:
@Random-Liu Addressed all comments.

Godeps/Godeps.json
Random-Liu: They are using v0.4.0: https://github.com/kubernetes-incubator/cri-o/blob/v0.3/lock.json#L82.

Let's make sure we are using the same version.

Godeps/Godeps.json
```diff
@@ -177,6 +197,11 @@
 			"Rev": "092cba3727bb9b4a2f0e922cd6c0f93ea270e363"
 		},
 		{
+			"ImportPath": "github.com/fsnotify/fsnotify",
+			"Comment": "v1.3.1-1-gf12c623",
+			"Rev": "f12c6236fe7b5cf6bcf30e5935d08cb079d78334"
```
Random-Liu: Let's make sure we are using the same version with cri-o to avoid unnecessary conflict in the future.
7d7316ed6e1ed2de075aab8dfc76de5d158d66e1

pkg/os/testing/fake_os.go
```diff
@@ -29,20 +30,62 @@ import (
...
 	RemoveAllFn func(string) error
 	OpenFifoFn  func(context.Context, string, int, os.FileMode) (io.ReadWriteCloser, error)
+	StatFn      func(name string) (os.FileInfo, error)
+	errors      map[string]error
 }
 
 var _ osInterface.OS = &FakeOS{}
 
+// getError get error for call
+func (f *FakeOS) getError(op string) error {
```
Random-Liu: Lock here.

pkg/server/sandbox_run.go
Random-Liu: Yeah

Random-Liu: Let's do it in this PR, it's only one line change. :)

Random-Liu: That is because `docker create` doesn't really start the process. Actually docker start creates and starts the containerd container.

`create` in containerd/runc already starts the process, so we could initialize its network namespace.

pkg/server/sandbox_run.go
Random-Liu: `%q	a single-quoted character literal safely escaped with Go syntax.`.

For empty string, `%s` will print nothing ``, `%q` will print `""`.

pkg/server/sandbox_run.go
```diff
@@ -167,13 +165,32 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
 				glog.Errorf("Failed to delete sandbox container %q: %v",
 					id, err)
 			}
 		}
 	}()
 
+	if !config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
+		// Setup network for sandbox.
+		// TODO(random-liu): [P2] Replace with permanent network namespace.
+		meta.NetNS = getNetworkNamespace(createResp.Pid)
```
Random-Liu: Move this out. We need to set network namespace even when using hostnetwork, so that application container could share the sandbox's namespace and use hostnetwork as well.

pkg/server/sandbox_run.go
```diff
@@ -167,13 +165,32 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+		podName := config.GetMetadata().GetName()
+		if err = c.netPlugin.SetUpPod(meta.NetNS, config.GetMetadata().GetNamespace(), podName, id); err != nil {
+			return nil, fmt.Errorf("failed to setup network %v", err)
```
Random-Liu: `failed to setup network for sandbox %q: %v`

pkg/server/sandbox_run.go
```diff
@@ -167,13 +165,32 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+		}
+		defer func() {
+			if retErr != nil {
+				// Teardown network if an error is returned.
+				if err := c.netPlugin.TearDownPod(meta.NetNS, config.GetMetadata().GetNamespace(), podName, id); err != nil {
```
Random-Liu: `for sandbox %q: %v`

pkg/server/sandbox_status.go
```diff
@@ -59,11 +59,18 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
 		state = runtime.PodSandboxState_SANDBOX_READY
 	}
 
-	return &runtime.PodSandboxStatusResponse{Status: toCRISandboxStatus(sandbox, state)}, nil
+	ip, err := c.netPlugin.GetContainerNetworkStatus(sandbox.NetNS, sandbox.Config.GetMetadata().GetNamespace(), sandbox.Config.GetMetadata().GetName(), id)
+	if err != nil {
+		// Ignore the error on network status
+		ip = ""
+		glog.V(4).Infof("GetContainerNetworkStatus returns error %v", err)
```
Random-Liu: nit: `returns error: %q`, add a colon.

pkg/server/sandbox_status_test.go
```diff
@@ -35,7 +35,10 @@ import (
 
 // Variables used in the following test.
 
-const sandboxStatusTestID = "test-id"
+const (
+	sandboxStatusTestID = "test-id"
+	testIP              = "10.10.10.10"
```
Random-Liu: `s/testIP/sandboxStatusTestIP`

Use a longer name to avoid conflict with other global variables in unit test in the same package.

pkg/server/sandbox_status_test.go
```diff
@@ -119,54 +124,91 @@ func TestPodSandboxStatus(t *testing.T) {
 		expectState       runtime.PodSandboxState
 		expectErr         bool
 		expectCalls       []string
+		expectIP          bool
+		expectedCNICalls  []string
+		injectCNIErr      error
```
Random-Liu: nit: Move "inject" fields up to near other "inject" fields.

Godeps/Godeps.json
xlgao-zju: Will do.

pkg/os/testing/fake_os.go
xlgao-zju: Will do.

pkg/server/sandbox_status_test.go
```diff
@@ -119,54 +124,91 @@ func TestPodSandboxStatus(t *testing.T) {
...
+		expectCNIErr      bool
```
Random-Liu: This field is not used, may not be needed.

pkg/server/sandbox_status_test.go
```diff
@@ -119,54 +124,91 @@ func TestPodSandboxStatus(t *testing.T) {
 		expectState       runtime.PodSandboxState
 		expectErr         bool
 		expectCalls       []string
+		expectIP          bool
```
Random-Liu: feels more like `injectIP`.

pkg/server/sandbox_status_test.go
```diff
@@ -65,15 +68,17 @@ func getSandboxStatusTestData() (*metadata.SandboxMetadata, *runtime.PodSandboxS
 		Name:      "test-name",
 		Config:    config,
 		CreatedAt: createdAt,
+		NetNS:     "test-netns",
```
Random-Liu: nit: Since we use this outside of the function, make it a global var.

pkg/server/sandbox_stop.go
```diff
@@ -45,14 +46,26 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
 	// Use the full sandbox id.
 	id := sandbox.ID
 
+	// Teardown network for sandbox.
+	_, err = c.os.Stat(sandbox.NetNS)
+	if err == nil {
+		if teardownErr := c.netPlugin.TearDownPod(sandbox.NetNS, sandbox.Config.GetMetadata().GetNamespace(), sandbox.Config.GetMetadata().GetName(), id); teardownErr != nil {
```
Random-Liu: nit: The line is too long, make a new line here.

pkg/server/sandbox_stop.go
```diff
@@ -45,14 +46,26 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
+			return nil, fmt.Errorf("failed to destroy network for sandbox %q: %v",
+				id, teardownErr)
+		}
+	} else if !os.IsNotExist(err) { // It's ok for sandbox.NetNS to *not* exist
+		return nil, fmt.Errorf("failed to stat netns path for container %q in sandbox %q before tearing down the network: %v",
```
Random-Liu: `for sandbox %q`.

pkg/server/sandbox_stop_test.go
```diff
@@ -51,36 +61,78 @@ func TestStopPodSandbox(t *testing.T) {
 		injectErr         error
 		expectErr         bool
 		expectCalls       []string
+		injectCNIErr      error
+		expectCNIErr      bool
+		expectedCNICalls  []string
+		injectOSErr       error
+		expectOSErr       bool
```
Random-Liu: Not used.

pkg/server/sandbox_stop_test.go
```diff
@@ -51,36 +61,78 @@ func TestStopPodSandbox(t *testing.T) {
 		injectErr         error
 		expectErr         bool
 		expectCalls       []string
+		injectCNIErr      error
+		expectCNIErr      bool
```
Random-Liu: not used.

pkg/server/sandbox_stop_test.go
```diff
@@ -51,36 +61,78 @@ func TestStopPodSandbox(t *testing.T) {
...
+		expectedCNICalls  []string
+		injectOSErr       error
```
Random-Liu: nit: Move `inject` up.

pkg/server/sandbox_stop.go
Random-Liu: Just inject an error into the `Stat` function directly, a random error.

pkg/server/service.go
Random-Liu: Let's do it now. In the future when necessary we could create a containerd service config object.

pkg/server/testing/fake_cni_plugin.go
Random-Liu: Let's use IP form if it's easy. Just `rand.Int(256)` for 4 times.

pkg/server/service.go
```diff
@@ -31,6 +33,8 @@ import (
 	imagesservice "github.com/containerd/containerd/services/images"
 	rootfsservice "github.com/containerd/containerd/services/rootfs"
 
+	"github.com/kubernetes-incubator/cri-o/pkg/ocicni"
```
Random-Liu: My bad, we should not have so many different groups of imports. Just add this to the second group. :)

pkg/server/service_test.go
```diff
@@ -110,6 +113,10 @@ func TestSandboxOperations(t *testing.T) {
 	require.NotNil(t, statusRes)
 	status := statusRes.GetStatus()
 	expectSandboxStatus.CreatedAt = status.GetCreatedAt()
+
```
Random-Liu: unnecessary empty line.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// Status get the status of the plugin.
+func (f *FakeCNIPlugin) Status() error {
+	return nil
+}
+
+func generateIP() string {
+	rand.Seed(time.Now().Unix())
+	return strconv.Itoa(rand.Intn(10000))
```
Random-Liu: call `rand.Intn(266)` for 4 times.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,187 @@
...
+// GetCalledDetails get detail of each call.
+func (f *FakeCNIPlugin) GetCalledDetails() []CalledDetail {
+	f.Lock()
+	defer f.Unlock()
+	// Copy the list and return.
+	return append([]CalledDetail{}, f.called...)
+}
+
+// SetFakePodNetwork sets the given IP for given arguments.
+func (f *FakeCNIPlugin) SetFakePodNetwork(netnsPath string, namespace string, name string, containerID string, ip string) {
```
Random-Liu: lock

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,187 @@
...
+	arg := CNIPluginArgument{netnsPath, namespace, name, containerID}
+	f.appendCalled("TearDownPod", arg)
+
+	if err := f.getError("TearDownPod"); err != nil {
+		return err
+	}
+
+	_, ok := f.IPMap[arg]
+	if !ok {
+		return fmt.Errorf("failed to find the IP")
```
Random-Liu: I'm not sure whether we should return error here. CNI doesn't define this, some plugin returns error in this case, some not. Let's treat it as it will return an error then.

pkg/server/testing/fake_cni_plugin.go
```diff
@@ -0,0 +1,187 @@
...
+// Name return the name of fake CNI plugin.
+func (f *FakeCNIPlugin) Name() string {
+	return "fake-CNI-plugin"
+}
+
+// SetUpPod setup the network of PodSandbox.
+func (f *FakeCNIPlugin) SetUpPod(netnsPath string, namespace string, name string, containerID string) error {
+	f.Lock()
+	defer f.Unlock()
+
```
Random-Liu: We don't need so many spaces. In google's code style, we want to avoid unnecessary space as much as possible to make the code shorter, and easier to read.

pkg/server/testing/fake_cni_plugin.go
xlgao-zju: > If Teardown is not expected to return error when there is no existing pod network, we don't need to inject ip.

As you said before, I think we should not return error...

Random-Liu: @xlgao-zju Discussed with our network guy @freehan. Currently cni binary returns an error in this case. Then let's return error here.

Random-Liu:
@xlgao-zju We may need a follow up PR for this PR.

The teardown network logic in this PR assumes that once the network namespace is gone, we don't need to teardown the network again.

However, it doesn't work well with current implementation of cri-containerd. Currently, cri-container creates a sandbox container for each sandbox and use the network namespace of this container. But the problem is that, once the container dies, we'll not be able to find the network namespace by `/proc/${PID}/ns/net`.

So based on current logic, once the sandbox container dies itself, we'll not be able to find the network namespace, thus not teardown network for it anymore. This will cause resource leakage.

We need to maintain a permanent network namespace for each sandbox, which is similar with cri-o. https://github.com/kubernetes-incubator/cri-o/blob/master/server/sandbox_run.go#L355
We should remove the permanent network namespace as soon as we successfully tear down the network, so as to avoid multiple teardown to the same network in most cases.

Random-Liu:
@xlgao-zju Filed an issue for this https://github.com/kubernetes-incubator/cri-containerd/issues/43.

pkg/server/testing/fake_cni_plugin.go
xlgao-zju: @Random-Liu So, If Teardown is not expected to return error, we should inject an IP, right?

Random-Liu: @xlgao-zju Yeah, but we could always inject an ip in that test. It doesn't seem like we need a field to switch between inject or not, unless you add a test case to test when ip is not injected, `StopPodSandbox` should return error.

xlgao-zju:
Updated, I think we can add permanent network namespace in another PR.

Random-Liu:
> Updated, I think we can add permanent network namespace in another PR.
@xlgao-zju SGTM.

The PR looks good to me. Thanks for helping on this! Do you have time to work on the permanent network namespace? :)

Issue #28, opened 2017-05-09
Maintain resolv.conf for pod.


Random-Liu:
Previously, Docker maintains a resolv.conf for containers sharing the same network namespace.

Now, when using containerd, we need to do these ourselves. We need to:
1) Maintain a resolv.conf when creating/removing sandbox if not using `HostNetwork`. And initialize resolv.conf with `DNSOptions` in `PodSandboxConfig`.
2) Bind mount the resolv.conf into each container.

This logic is similar with cri-o. (See https://github.com/kubernetes-incubator/cri-o/blob/master/server/sandbox_run.go#L144-L159)

/cc @xlgao-zju

Crazykev:
Since #22 has been merged, I'll get this done ASAP.

Crazykev:
@Random-Liu To implement this, we need to maintain `resolv.conf` file for each pod(expect for hostNetwork pod) and `resolvPath` for this file. This `resolvPath` is related with on-disk layout(not implemented yet), do you have any suggestion or preference on this? Thanks.

Random-Liu:
@Crazykev Just put the resolv file into sandbox root. https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_run.go#L91

Random-Liu:
Close this one since the PR is merged, thanks! @Crazykev

PR #50, opened 2017-05-25
Generate and maintain resolv.conf for sandbox

Crazykev:
This is implement part of #28, I'll continue to write some related unit test for this.
> And we may need to maintain it for hostNetwork pod, let me double check it and rely later.

@Random-Liu I copy a host resolv.conf for each pod with hostNetwork, since user may change this file in container. Let me know if you change your mind.

Signed-off-by: Crazykev <crazykev@zju.edu.cn>

### DIFF
--- a/pkg/server/container_start.go
+++ b/pkg/server/container_start.go
@@ -292,10 +292,6 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 
 	// TODO(random-liu): [P2] Add apparmor and seccomp.
 
-	// TODO(random-liu): [P1] Bind mount sandbox /dev/shm.
-
-	// TODO(random-liu): [P0] Bind mount sandbox resolv.conf.
-
 	return g.Spec(), nil
 }
 
@@ -307,9 +303,17 @@ func (c *criContainerdService) generateContainerMounts(sandboxRootDir string, co
 	mounts = append(mounts, &runtime.Mount{
 		ContainerPath: etcHosts,
 		HostPath:      getSandboxHosts(sandboxRootDir),
-		Readonly:      securityContext.ReadonlyRootfs,
+		Readonly:      securityContext.GetReadonlyRootfs(),
 	})
-	// TODO(random-liu): [P0] Mount sandbox resolv.config.
+
+	// Mount sandbox resolv.config.
+	// TODO: Need to figure out whether we should always mount it as read-only
+	mounts = append(mounts, &runtime.Mount{
+		ContainerPath: resolvConfPath,
+		HostPath:      getResolvPath(sandboxRootDir),
+		Readonly:      securityContext.GetReadonlyRootfs(),
+	})
+
 	// TODO(random-liu): [P0] Mount sandbox /dev/shm.
 	return mounts
 }
--- a/pkg/server/container_start_test.go
+++ b/pkg/server/container_start_test.go
@@ -173,24 +173,22 @@ func getStartContainerTestData() (*runtime.ContainerConfig, *runtime.PodSandboxC
 
 func TestGeneralContainerSpec(t *testing.T) {
 	testID := "test-id"
-	testPodID := "test-pod-id"
 	testPid := uint32(1234)
 	config, sandboxConfig, imageConfig, specCheck := getStartContainerTestData()
 	c := newTestCRIContainerdService()
-	spec, err := c.generateContainerSpec(testID, testPodID, testPid, config, sandboxConfig, imageConfig, nil)
+	spec, err := c.generateContainerSpec(testID, testPid, config, sandboxConfig, imageConfig, nil)
 	assert.NoError(t, err)
 	specCheck(t, testID, testPid, spec)
 }
 
 func TestContainerSpecTty(t *testing.T) {
 	testID := "test-id"
-	testPodID := "test-pod-id"
 	testPid := uint32(1234)
 	config, sandboxConfig, imageConfig, specCheck := getStartContainerTestData()
 	c := newTestCRIContainerdService()
 	for _, tty := range []bool{true, false} {
 		config.Tty = tty
-		spec, err := c.generateContainerSpec(testID, testPodID, testPid, config, sandboxConfig, imageConfig, nil)
+		spec, err := c.generateContainerSpec(testID, testPid, config, sandboxConfig, imageConfig, nil)
 		assert.NoError(t, err)
 		specCheck(t, testID, testPid, spec)
 		assert.Equal(t, tty, spec.Process.Terminal)
@@ -199,13 +197,12 @@ func TestContainerSpecTty(t *testing.T) {
 
 func TestContainerSpecReadonlyRootfs(t *testing.T) {
 	testID := "test-id"
-	testPodID := "test-pod-id"
 	testPid := uint32(1234)
 	config, sandboxConfig, imageConfig, specCheck := getStartContainerTestData()
 	c := newTestCRIContainerdService()
 	for _, readonly := range []bool{true, false} {
 		config.Linux.SecurityContext.ReadonlyRootfs = readonly
-		spec, err := c.generateContainerSpec(testID, testPodID, testPid, config, sandboxConfig, imageConfig, nil)
+		spec, err := c.generateContainerSpec(testID, testPid, config, sandboxConfig, imageConfig, nil)
 		assert.NoError(t, err)
 		specCheck(t, testID, testPid, spec)
 		assert.Equal(t, readonly, spec.Root.Readonly)
@@ -214,7 +211,6 @@ func TestContainerSpecReadonlyRootfs(t *testing.T) {
 
 func TestContainerSpecWithExtraMounts(t *testing.T) {
 	testID := "test-id"
-	testPodID := "test-pod-id"
 	testPid := uint32(1234)
 	config, sandboxConfig, imageConfig, specCheck := getStartContainerTestData()
 	c := newTestCRIContainerdService()
@@ -229,7 +225,7 @@ func TestContainerSpecWithExtraMounts(t *testing.T) {
 		HostPath:      "test-host-path-extra",
 		Readonly:      true,
 	}
-	spec, err := c.generateContainerSpec(testID, testPodID, testPid, config, sandboxConfig, imageConfig, []*runtime.Mount{extraMount})
+	spec, err := c.generateContainerSpec(testID, testPid, config, sandboxConfig, imageConfig, []*runtime.Mount{extraMount})
 	assert.NoError(t, err)
 	specCheck(t, testID, testPid, spec)
 	var mounts []runtimespec.Mount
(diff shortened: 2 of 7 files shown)


Random-Liu:
Will take a look tomorrow, and let's try to get it merged this week.

pkg/server/container_start.go
```diff
@@ -283,7 +283,10 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 
 	// TODO(random-liu): [P1] Bind mount sandbox /dev/shm.
 
-	// TODO(random-liu): [P0] Bind mount sandbox resolv.conf.
+	// Bind mount sandbox resolv.conf.
+	if resolvPath := getResolvPath(getSandboxRootDir(c.rootDir, sandboxID), sandboxConfig); resolvPath != "" {
+		g.AddBindMount(resolvPath, resolvConfPath, []string{"ro"})
```
Random-Liu: Let's mount it as "rw" for now, and add TODO to figure out whether we should mount it as read-only.

Docker mount it as `ro`, and let's avoid regression for now, there may be user relying on this behavior. For example, there may be user doing hacky things with it.

pkg/server/sandbox_run_test.go
```diff
@@ -113,7 +113,8 @@ func TestGenerateSandboxContainerSpec(t *testing.T) {
 		if test.configChange != nil {
 			test.configChange(config)
 		}
-		spec := c.generateSandboxContainerSpec(testID, config)
+		spec, err := c.generateSandboxContainerSpec(testID, config)
```
Random-Liu: Test resolve.conf bind mount.

Random-Liu:
@Crazykev Please rebase your PR and address comments

pkg/server/sandbox_run.go
```diff
@@ -225,7 +228,20 @@ func (c *criContainerdService) generateSandboxContainerSpec(id string, config *r
 	// Set hostname.
 	g.SetHostname(config.GetHostname())
 
-	// TODO(random-liu): [P0] Set DNS options. Maintain a resolv.conf for the sandbox.
```
Random-Liu: NVM, the whole sandbox root directory will be removed.

pkg/server/sandbox_run.go
```diff
@@ -225,7 +228,20 @@ func (c *criContainerdService) generateSandboxContainerSpec(id string, config *r
...
+	// Set DNS options.
+	if dnsConfig := config.GetDnsConfig(); dnsConfig != nil {
+		// resolvPath is not "" guaranteed by 'dnsConfig != nil'
+		resolvPath := getResolvPath(getSandboxRootDir(c.rootDir, id), config)
+		err := parseDNSOptions(dnsConfig.Servers, dnsConfig.Searches, dnsConfig.Options, resolvPath)
+		if err != nil {
+			err1 := removeFile(resolvPath)
```
Random-Liu: I found that we've already had code cleaning up the sandbox root. https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_run.go#L112

We can rely on that to cleanup.

pkg/server/container_start.go
Random-Liu: We should decide mount option based on `securityContext.GetReadonlyRootfs()`.

pkg/server/sandbox_run.go
```diff
@@ -225,7 +228,20 @@ func (c *criContainerdService) generateSandboxContainerSpec(id string, config *r
 	// Set hostname.
 	g.SetHostname(config.GetHostname())
 
-	// TODO(random-liu): [P0] Set DNS options. Maintain a resolv.conf for the sandbox.
+	// Set DNS options.
+	if dnsConfig := config.GetDnsConfig(); dnsConfig != nil {
```
Crazykev: Yep, this is my original thought, not sure why cri-o guys do this assumption...

Crazykev:
@Random-Liu  Addressed comments and added related unit test. PTAL.

pkg/server/sandbox_run.go
```diff
@@ -133,6 +133,26 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+		resolvContent, err = parseDNSOptions(dnsConfig.Servers, dnsConfig.Searches, dnsConfig.Options)
+		if err != nil {
+			return nil, err
+		}
+	}
+	resolvPath := getResolvPath(sandboxRootDir)
+	if len(resolvContent) == 0 && hostNetwork {
+		// copy host's resolv.conf to resolvPath
+		// err = c.os.CopyFile(resolvConfPath, resolvPath)
```
Crazykev: will uncomment this after #60 merged.

Crazykev:
@Random-Liu Rebased.

pkg/server/sandbox_run.go
```diff
@@ -303,8 +308,61 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
 	if err := c.os.CopyFile(etcHosts, sandboxEtcHosts, 0666); err != nil {
 		return fmt.Errorf("failed to generate sandbox hosts file %q: %v", sandboxEtcHosts, err)
 	}
-	// TODO(random-liu): [P0] Set DNS options. Maintain a resolv.conf for the sandbox.
+	// Set DNS options. Maintain a resolv.conf for the sandbox.
+	resolvContent := []byte{}
+	var err error
+	if dnsConfig := config.GetDnsConfig(); dnsConfig != nil {
+		resolvContent, err = parseDNSOptions(dnsConfig.Servers, dnsConfig.Searches, dnsConfig.Options)
```
Random-Liu: nit: Pass in `DNSConfigs`? Seems more clean.

pkg/server/sandbox_run.go
```diff
@@ -303,8 +308,61 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
...
+			return fmt.Errorf("failed to parse sandbox dnsConfig: %v", err)
+		}
+	}
+	resolvPath := getResolvPath(rootDir)
+	hostNetwork := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
+	if len(resolvContent) == 0 && hostNetwork {
+		// copy host's resolv.conf to resolvPath
+		err = c.os.CopyFile(resolvConfPath, resolvPath, 0755)
+	} else {
+		err = c.os.WriteFile(resolvPath, resolvContent, 0755)
```
Random-Liu: Same, should be `0644`.

pkg/server/sandbox_run.go
```diff
@@ -303,8 +308,61 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
...
+	}
+	if err != nil {
+		return fmt.Errorf("failed to copy/write file to %q: %v", resolvPath, err)
```
Random-Liu: Let's return err separately.

pkg/server/sandbox_run.go
```diff
@@ -303,8 +308,61 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
...
 		return fmt.Errorf("failed to generate sandbox hosts file %q: %v", sandboxEtcHosts, err)
 	}
-	// TODO(random-liu): [P0] Set DNS options. Maintain a resolv.conf for the sandbox.
+	// Set DNS options. Maintain a resolv.conf for the sandbox.
+	resolvContent := []byte{}
+	var err error
+	if dnsConfig := config.GetDnsConfig(); dnsConfig != nil {
+		resolvContent, err = parseDNSOptions(dnsConfig.Servers, dnsConfig.Searches, dnsConfig.Options)
+		if err != nil {
+			return fmt.Errorf("failed to parse sandbox dnsConfig: %v", err)
```
Random-Liu: `fmt.Errorf("failed to parse sandbox DNSConfig %+v: %v", DNSConfig, err)`

pkg/server/sandbox_run_test.go
```diff
@@ -269,5 +286,36 @@ func TestRunPodSandbox(t *testing.T) {
...
+		servers  []string
+		searches []string
+		options  []string
+	}{
+		"empty dns options": {},
+		"non-empty dns options": {
+			servers:  []string{"8.8.8.8", "server.google.com"},
+			searches: []string{"114.114.114.114"},
+			options:  []string{"timeout:1"},
+		},
```
Random-Liu: 1) Let's add test case to check whether the function returns error when searches is more than 6
2) Let's add `expectedOutput` into the test case, and directly check `assert.Equal(t, expectedOutput, resolvContent)`, e.g.:
```golang
map[string]struct{
dnsConfig *runtime.DNSConfig
expectedContent string
expectErr bool
} {
 "empty dns options should return empty content": {
    dnsConfig: &runtime.DNSConfig{},
    expectedOutput: "",
  },
  "non-empty dns options should return correct content": {
   dnsConfig: &runtime.DNSConfig{/*...*/},
   expectedOutput: `search 114.114.114.114
nameserver 8.8.8.8
nameserver server.google.com
...
`
  },
  "should return error if dns search exceeds limit": {
   /*...*/
  },
}
```

pkg/server/sandbox_run.go
```diff
@@ -303,8 +308,61 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
...
+	}
+
 	// TODO(random-liu): [P0] Deal with /dev/shm. Use host for HostIpc, and create and mount for
 	// non-HostIpc. What about mqueue?
 	return nil
 }
+
+// parseDNSOptions parse DNS options into resolv.conf format content,
+// if none option is specified, will return empty with no error.
+func parseDNSOptions(servers, searches, options []string) ([]byte, error) {
```
Random-Liu: nit: Can we use `string` in this function, and use `resolvContent += fmt.Sprintf("search %s\n", strings.Join(searches, " "))`, which is much more cleaner.

Random-Liu:
@Crazykev Reviewed. Nits are not that important, but are be nice to address. Others please take a look.

Crazykev:
@Random-Liu Updated, PTAL

pkg/server/container_start.go
```diff
@@ -307,9 +303,18 @@ func (c *criContainerdService) generateContainerMounts(sandboxRootDir string, co
...
-		Readonly:      securityContext.ReadonlyRootfs,
+		Readonly:      securityContext.GetReadonlyRootfs(),
+	})
+
+	// Mount sandbox resolv.config.
+	resolvPath := getResolvPath(sandboxRootDir)
+	// TODO: Need to figure out whether we should always mount it as read-only
+	mounts = append(mounts, &runtime.Mount{
+		ContainerPath: resolvConfPath,
+		HostPath:      resolvPath,
```
Random-Liu: nit: `getResolvPath(sandboxRootDir)` to save one line.

pkg/server/container_start_test.go
```diff
@@ -309,23 +309,37 @@ func TestGenerateContainerMounts(t *testing.T) {
...
-			}},
+			expectedMounts: []*runtime.Mount{
+				{
+					ContainerPath: "/etc/hosts",
+					HostPath:      testSandboxRootDir + "/hosts",
+					Readonly:      true,
+				},
+				{
+					ContainerPath: resolvConfPath,
+					HostPath:      getResolvPath(testSandboxRootDir),
```
Random-Liu: nit: Use `testSandboxRootDir + "/resolv.conf"`, because we don't know whether `getResolvPath` works or not, which is also what we want to test.

pkg/server/sandbox_run.go
```diff
@@ -303,8 +304,62 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
...
+	resolvContent := ""
+	if dnsConfig := config.GetDnsConfig(); dnsConfig != nil {
+		resolvContent, err = parseDNSOptions(dnsConfig.Servers, dnsConfig.Searches, dnsConfig.Options)
+		if err != nil {
+			return fmt.Errorf("failed to parse sandbox DNSConfig %+v: %v", dnsConfig, err)
+		}
+	}
+	resolvPath := getResolvPath(rootDir)
+	hostNetwork := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
+	if resolvContent == "" && hostNetwork {
```
Random-Liu: Based on https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_sandbox.go#L616-L651, we should overwrite `resolv.conf` no matter it's using hostNetwork or not.

Random-Liu: Actually we also need unit test for this function. However, I'm fine with leave it for now. I'll add it in https://github.com/kubernetes-incubator/cri-containerd/pull/67.

pkg/server/sandbox_run.go
```diff
@@ -303,8 +304,62 @@ func (c *criContainerdService) setupSandboxFiles(rootDir string, config *runtime
...
 }
+
+// parseDNSOptions parse DNS options into resolv.conf format content,
+// if none option is specified, will return empty with no error.
+func parseDNSOptions(servers, searches, options []string) (string, error) {
+	nServers := len(servers)
+	nSearches := len(searches)
+	nOptions := len(options)
+	resolvContent := ""
+	if nServers == 0 && nSearches == 0 && nOptions == 0 {
```
Random-Liu: nit: This seems unnecessary, the function will return empty string anyway if all are empty.

And if we get rid of this block, we could directly use `len` in the following if clause, because we only need the result once.

pkg/server/sandbox_run.go
Crazykev: @Random-Liu 
> we should overwrite resolv.conf no matter it's using hostNetwork or not as long as resolvContent != "".

Yep, we should. While if it's not hostNetwork and resolvContent is empty(DnsConfig is nil or content is empty), should we just generate a empty resolv.conf(maybe one with default setting?) for this pod? In your logic it like we should pass host's resolv.conf to pod in this scene.

Random-Liu: @Crazykev Based on current dockershim implementation, we only rewrite when `resolvContent` is not empty. Let's keep the same behavior now.

Crazykev:
@Random-Liu Addressed comments except adding a new test case, left it to you ;-)

pkg/server/container_start_test.go
```diff
@@ -309,23 +309,37 @@ func TestGenerateContainerMounts(t *testing.T) {
...
-			}},
+			expectedMounts: []*runtime.Mount{
+				{
+					ContainerPath: "/etc/hosts",
+					HostPath:      testSandboxRootDir + "/hosts",
+					Readonly:      false,
+				},
+				{
+					ContainerPath: resolvConfPath,
+					HostPath:      getResolvPath(testSandboxRootDir),
```
Random-Liu: nit: This also needs to be updated, but I could do it in my PR.

PR #52, opened 2017-05-25
Apply container image config

Random-Liu:
Based on #41, #46 and #48.

This PR pulls container image and applies image config.

Only the last commit is new. Will rebase the PR and previous PRs soon.

PR #48, opened 2017-05-24
Apply sandbox image config

Random-Liu:
Based on https://github.com/kubernetes-incubator/cri-containerd/pull/41 and https://github.com/kubernetes-incubator/cri-containerd/pull/46.

This PR let sandbox container use pause image and applies image config.

Only the last commit is new. Will rebase the PR and previous PRs soon.

PR #46, opened 2017-05-24
Wait and check image pulling progress.

Random-Liu:
Based on https://github.com/kubernetes-incubator/cri-containerd/pull/41.

This PR:
1) Update contained to https://github.com/containerd/containerd/pull/821/commits/193abed96e06a45fead2ed86cfbb11c2a0a001a4 which contains some content api update and implements `Status` function in content api.
2) When pulling image, get references of all resources associated to the image, wait for those pulling to be done by checking content status. This makes sure that concurrent pull which pulls the same resource won't return before the resource is fully pulled.

/cc @yujuhong @mikebrow

PR #41, opened 2017-05-18
Finish image management.

Random-Liu:
This PR finishes the containerd image management. There are still many TODOs to be addressed, I'll do them in some following PRs.

TODOs for this PR:
- [x] Add unit test as much as I can. Since containerd API is changing fast, I'll not add unit test for `rootfs` service related unit test now.
- [x] ~~Add image list filters.~~

TODOs for future PRs:
- [x] 1) Create rootfs and apply image config when creating sandbox/container. (#48 and #52)

2) Some work for operation synchronization:
- [x] a) When pulling image, get digests of all resources associated to the image, wait for those pulling to be done by checking content status. This makes sure that concurrent pull which pulls the same resource won't return before the resource is fully pulled. (content store returns immediately for ongoing pulling). (#46)
  b) We'll not need checkpoint for image management. Write a cleaner and easier to use image metadata which could lock more operations together because all data is in memory.
  c) Add some synchronization between concurrent pulling/removing against the same image.

3) Add authentication support

4) Update containerd version to use the [new snapshot service](https://github.com/containerd/containerd/blob/master/api/services/snapshot/snapshots.proto)

This PR also depends on https://github.com/containerd/containerd/pull/875.

PR #53, opened 2017-05-26
Add other small functions

Random-Liu:
Add `Version`, `UpdateRuntimeConfig` and `Status`.

Based on #41, #46, #48 and #52. Only last 2 commits are new.

We could basically achieve our alpha goal after this PR is merged, and more tests are needed after code freeze. @yujuhong @dchen1107 @mikebrow 
/cc @abgworrall @matchstick

With this PR, we could pass half CRI validation test now:
```
Ran 36 of 36 Specs in 93.276 seconds
FAIL! -- 18 Passed | 18 Failed | 0 Pending | 0 Skipped --- FAIL: TestE2ECRI (93.28s)
FAIL

Ginkgo ran 1 suite in 1m33.350842999s
Test Suite Failed
exit status 1
```

Failed tests are caused by:
1) Container logging
2) Exec
3) Other unsupported features: HostPort, Schema 1 image support etc.


pkg/server/status.go
```diff
@@ -17,14 +17,22 @@ limitations under the License.
...
 // Status returns the status of the runtime.
 func (c *criContainerdService) Status(ctx context.Context, r *runtime.StatusRequest) (*runtime.StatusResponse, error) {
-	return nil, errors.New("not implemented")
+	runtimeReady := &runtime.RuntimeCondition{
+		Type:   runtime.RuntimeReady,
+		Status: true,
+	}
+	networkReady := &runtime.RuntimeCondition{
+		Type:   runtime.NetworkReady,
+		Status: true,
```
yujuhong: Add a TODO to check whether the plugins are ready or they exist? I didn't review the networking code, so this may not make sense.

pkg/server/status.go
```diff
@@ -17,14 +17,22 @@ limitations under the License.
...
-
 	"golang.org/x/net/context"
 
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1"
 )
 
 // Status returns the status of the runtime.
 func (c *criContainerdService) Status(ctx context.Context, r *runtime.StatusRequest) (*runtime.StatusResponse, error) {
-	return nil, errors.New("not implemented")
+	runtimeReady := &runtime.RuntimeCondition{
```
yujuhong: Do we want to check whether containerd is alive? E.g., call containerd version().

Random-Liu: Will do in this PR.

pkg/server/status.go
Random-Liu: cri-containerd doesn't support cider allocation now (we are directly using cri-o cni plugin package now).

The networking team is also planing to deprecate kubenet. In the future, the cider should be handled by a daemonset outside of the container.

/cc @freehan

yujuhong: It doesn't have to be CIDR. Network could be not ready if cri-containerd cannot find any network plugins.

Random-Liu: True, will update.

Random-Liu: Done

pkg/server/status.go
Random-Liu: Done

Random-Liu:
@yujuhong Addressed comments.

pkg/server/status.go
```diff
@@ -17,14 +17,52 @@ limitations under the License.
...
+	// runtimeNotReadyReason is the reason reported when runtime is not ready.
+	runtimeNotReadyReason = "ContainerdNotReady"
+	// networkNotReadyReason is the reason reported when network is not ready.
+	networkNotReadyReason = "NetworkPluginNotReady"
+)
+
 // Status returns the status of the runtime.
 func (c *criContainerdService) Status(ctx context.Context, r *runtime.StatusRequest) (*runtime.StatusResponse, error) {
-	return nil, errors.New("not implemented")
+	runtimeCondition := &runtime.RuntimeCondition{
```
mikebrow: maybe check the health of the meta stores?

Random-Liu: @mikebrow What do you mean?

pkg/server/update_runtime_config.go
```diff
@@ -17,14 +17,12 @@ limitations under the License.
...
-
 	"golang.org/x/net/context"
 
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1"
 )
 
 // UpdateRuntimeConfig updates the runtime config. Currently only handles podCIDR updates.
 func (c *criContainerdService) UpdateRuntimeConfig(ctx context.Context, r *runtime.UpdateRuntimeConfigRequest) (*runtime.UpdateRuntimeConfigResponse, error) {
-	return nil, errors.New("not implemented")
+	return &runtime.UpdateRuntimeConfigResponse{}, nil
```
mikebrow: ?

Random-Liu: We don't need this function now. In current dockershim, this is used to update pod CIDR, which we don't support now and may not support in the future.
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_service.go#L304-L315

We may have a separate daemonset to update the CIDR in the future.

pkg/server/version.go
```diff
@@ -17,14 +17,35 @@ limitations under the License.
...
 
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1"
 )
 
+const (
+	containerName        = "containerd"
+	containerdAPIVersion = "0.0.0"
+	containerdVersion    = "0.0.0"
+	// kubeAPIVersion is the api version of kubernetes.
+	kubeAPIVersion = "0.1.0"
```
mikebrow: Couple questions.. 1) Should / could this be set at build vs being hard coded? 2) Why kubeAPIVersion and not kubelet and / or kubelet/cri?

pkg/server/version.go
```diff
@@ -17,14 +17,35 @@ limitations under the License.
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get containerd version: %v", err)
+	}
+	return &runtime.VersionResponse{
+		Version:     kubeAPIVersion,
+		RuntimeName: containerName,
+		// Containerd doesn't return semver because of a bug.
+		// TODO(random-liu): Replace this with version from containerd.
+		RuntimeVersion: containerdVersion,
+		// Containerd doesn't have an api version now.
```
mikebrow: so containerd isn't going to follow the new docker/moby model for using the date to set the version and let the api version be the more traditional major change tag?

pkg/server/status.go
mikebrow: that which implements a MetadataStore..

SandboxStore, ImageMetadataStore, ...

pkg/server/update_runtime_config.go
mikebrow: maybe a comment or todo to investigate if / when we need to update..  Returning no error here with no update if they passed in something to be updated could lead the user astray?

pkg/server/status.go
Random-Liu: Here, it's checking whether containerd is working or not.

pkg/server/version.go
Random-Liu: Copied from dockershim.

This field doesn't make too much sense to me now.

pkg/server/version.go
Random-Liu: containerd won't use date to set version I think.

They should expose api version, but I think it's to early to them to talk about api version now, given that they are changing api everyday before GA.

pkg/server/update_runtime_config.go
Random-Liu: Will do.

Random-Liu: Done

pkg/server/status.go
mikebrow: kk didn't realize runtimeCondition=="containerdRuntimeCondition" is the networkCondition the containerdNetworkCondition?

Random-Liu:
@mikebrow Thanks for reiviewing!
@yujuhong Are you ok with this one?

yujuhong:
LGTM

Random-Liu:
@yujuhong Thanks for reviewing!

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/53#pullrequestreview-41350180 and https://github.com/kubernetes-incubator/cri-containerd/pull/53#issuecomment-305326241.

Issue #43, opened 2017-05-22
Create permanent network namespace for sandbox.


Random-Liu:
Currently, the network teardown logic assumes that once the network namespace is gone, we don't need to teardown the network again.

However, it doesn't work well with current implementation of cri-containerd. Currently, cri-containerd creates a sandbox container for each sandbox and use the network namespace of this container. But the problem is that, once the container dies, we'll not be able to find the network namespace by /proc/${PID}/ns/net.

So based on current logic, once the sandbox container dies itself, we'll not be able to find the network namespace, thus not teardown network for it. This will cause resource leakage.

We need to maintain a permanent network namespace for each sandbox, which is [similar with cri-o](https://github.com/kubernetes-incubator/cri-o/blob/master/server/sandbox_run.go#L355).
We should remove the permanent network namespace as soon as we successfully tear down the network, so as to avoid multiple teardown to the same network in most cases.

/cc @xlgao-zju

xlgao-zju:
I would like to make this issue two parts:

- [ ] Create permanent network namespace for sandbox.
- [ ] Add unit test.

xlgao-zju:
Since I am not maintainer of `cri-containerd`, I can not self-assigned. So ping @Random-Liu.

Random-Liu:
Moved this to v0.2.0.

We still need this enhancement, and it's very useful. However, I don't think we could get all depdencencies merged this release, so punt this one to next release.

Random-Liu:
/cc @abhinandanpb 
This is the issue for the network namespace.

And there is a pending PR for it https://github.com/kubernetes-incubator/cri-containerd/pull/54, which was originally blocked by something else.

Random-Liu:
#138 is merged. @abhinandanpb Thanks!

PR #54, opened 2017-05-26
Add permanent netns

xlgao-zju:
part of #43

### DIFF
--- a/pkg/metadata/sandbox.go
+++ b/pkg/metadata/sandbox.go
@@ -18,6 +18,11 @@ package metadata
 
 import (
 	"encoding/json"
+	"fmt"
+	"sync"
+
+	"github.com/containernetworking/cni/pkg/ns"
+	"github.com/golang/glog"
 
 	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
 
@@ -33,7 +38,10 @@ import (
 // metadata into unversioned metadata.
 
 // sandboxMetadataVersion is current version of sandbox metadata.
-const sandboxMetadataVersion = "v1" // nolint
+const (
+	sandboxMetadataVersion = "v1" // nolint
+	nsRunDir               = "/var/run/netns"
+)
 
 // versionedSandboxMetadata is the internal struct representing the versioned
 // sandbox metadata
@@ -55,7 +63,9 @@ type SandboxMetadata struct {
 	// CreatedAt is the created timestamp.
 	CreatedAt int64
 	// NetNS is the network namespace used by the sandbox.
-	NetNS string
+	NetNS *SandboxNetNS
+	// SandboxPID is the pid of sandbox container.
+	SandboxPID uint32
 }
 
 // SandboxUpdateFunc is the function used to update SandboxMetadata.
@@ -149,4 +159,52 @@ func (s *sandboxStore) List() ([]*SandboxMetadata, error) {
 // Delete deletes the sandbox from the store.
 func (s *sandboxStore) Delete(sandboxID string) error {
 	return s.store.Delete(sandboxID)
+}
+
+type SandboxNetNS struct {
+	sync.Mutex
+	NS     ns.NetNS
+	Closed bool
+}
+
+// NetNsPath returns the path to the network namespace of the container.
+func (s *SandboxMetadata) NetNSPath() string {
+	return s.NetNS.NS.Path()
+}
+
+// NetNsCreate creates netns and symlink it with a file.
+func (s *SandboxMetadata) NetNSCreate() error {
+	if s.NetNS != nil {
+		return fmt.Errorf("net NS already created")
+	}
+	netNS, err := ns.NewNS()
+	if err != nil {
+		return fmt.Errorf("failed to create new net NS: %v", err)
+	}
+	s.NetNS = &SandboxNetNS{
+		NS:     netNS,
+		Closed: false,
+	}
+	return nil
+}
+
+// NetNsRemove removes the netns store in sandbox meta data.
+func (s *SandboxMetadata) NetNSRemove() error {
+	if s.NetNS == nil {
+		glog.Error("no networking namespace to be removed.")
+		return nil
+	}
+	s.NetNS.Lock()
+	defer s.NetNS.Unlock()
+	if s.NetNS.Closed {
+		// netNsRemove() can be called multiple times without returning an error.
+		return nil
+	}
+	if err := s.NetNS.NS.Close(); err != nil {
+		return fmt.Errorf("failed to close net NS: %v", err)
+	}
+	// TODO: Handle restore logic with netNsGet.
+	// TODO: Get rid of the forcibly cleanup on restore after containernetworking/cni#342 is fixed.
+	s.NetNS.Closed = true
+	return nil
 }
--- a/pkg/server/container_start.go
+++ b/pkg/server/container_start.go
@@ -119,7 +119,7 @@ func (c *criContainerdService) startContainer(ctx context.Context, id string, me
 	if err != nil {
 		return fmt.Errorf("failed to get container image %q: %v", meta.ImageRef, err)
 	}
-	spec, err := c.generateContainerSpec(id, sandboxPid, config, sandboxConfig, imageMeta.Config)
+	spec, err := c.generateContainerSpec(id, sandboxMeta.NetNSPath(), sandboxPid, config, sandboxConfig, imageMeta.Config)
 	if err != nil {
 		return fmt.Errorf("failed to generate container %q spec: %v", id, err)
 	}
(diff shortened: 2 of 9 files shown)


xlgao-zju:
@Random-Liu I don't know how to unmarshal `ns.NetNS`. Any ideas?

When I get status of sandbox, will get the following error:

```sh
root@Joker:/home/xlgao/gocode/bin# ./crictl -r=/var/run/cri-containerd.sock sb status f5f5967e918bb548504feef152d3ee5a818e94b77a0d145815d37f623ffd3031

getting the pod sandbox status failed: rpc error: code = 2 desc = an error occurred when try to find sandbox "f5f5967e918bb548504feef152d3ee5a818e94b77a0d145815d37f623ffd3031": sandbox metadata not found: json: cannot unmarshal object into Go struct field sandboxNetNS.NS of type ns.NetNS
```

Random-Liu:
Will review this tonight.

pkg/metadata/sandbox.go
```diff
@@ -150,3 +164,115 @@ func (s *sandboxStore) List() ([]*SandboxMetadata, error) {
...
+		// netNsRemove() can be called multiple times without returning an error.
+		return nil
+	}
+	if err := s.NetNS.symlinkRemove(); err != nil {
+		return fmt.Errorf("failed to remove symlink: %v", err)
+	}
+	if err := s.NetNS.NS.Close(); err != nil {
+		return fmt.Errorf("failed to close net NS: %v", err)
+	}
+	if s.NetNS.Restored {
```
Random-Liu: This code is not useful because you didn't add the `netNsGet` function.

We don't need the restore logic for now, because we don't support restore after `cri-containerd` restart.
We can remove this code, and add a TODOs here:
1) We should handle restore logic with `netNsGet`.
2) We should get read of the forcibly cleanup on restore after https://github.com/containernetworking/cni/pull/342 is fixed and link the original issue https://github.com/kubernetes-incubator/cri-o/pull/276.

pkg/metadata/sandbox.go
```diff
@@ -150,3 +164,115 @@ func (s *sandboxStore) List() ([]*SandboxMetadata, error) {
...
+			return fmt.Errorf("net NS can not be close: %v", closeErr)
+		}
+		return fmt.Errorf("failed to create symlink for sandbox %q: %v", s.Name, err)
+	}
+	return nil
+}
+
+// NetNsRemove removes the netns store in sandbox meta data.
+func (s *SandboxMetadata) NetNSRemove() error {
+	glog.Error("remove bei diaoyongle")
```
Random-Liu: Remove the log.

pkg/metadata/sandbox.go
```diff
@@ -55,7 +67,9 @@ type SandboxMetadata struct {
 	// CreatedAt is the created timestamp.
 	CreatedAt int64
 	// NetNS is the network namespace used by the sandbox.
-	NetNS string
+	NetNS *SandboxNetNS
```
Random-Liu: This is my fault.

Currently, `SandboxMetadata` is designed to be "pure data", it's not designed to hold this kind of data structure, and all update should go through `Update()`.

Let me think about this a little more. We need a better design here.

pkg/metadata/util.go
```diff
@@ -16,10 +16,22 @@ limitations under the License.
...
+const netNSFormat = "/proc/%v/ns/net"
 
 // IsNotExistError is a helper function to check whether the error returned
 // by metadata store is not exist error.
 func IsNotExistError(err error) bool {
 	return err.Error() == store.ErrNotExist.Error()
 }
+
+// getNetworkNamespace returns the network namespace of a process.
+func getNetworkNamespace(pid uint32) string {
```
Random-Liu: We should use the permanent network namespace path.

One of the reason we need permanent path is that, once the sandbox process is gone, you will not be able to access `/proc/pid/ns/net`, which is not what we want.

We want to cleanup the network namespace even though the sandbox exits itself. That's why we want to have a known permanent network namespace.

pkg/server/container_start.go
```diff
@@ -266,7 +265,7 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 	// TODO(random-liu): [P0] Handle privileged.
 
 	// Set namespaces, share namespace with sandbox container.
-	setOCINamespaces(&g, securityContext.GetNamespaceOptions(), sandboxPid)
+	setOCINamespaces(&g, securityContext.GetNamespaceOptions(), sb)
```
Random-Liu: Do not pass in the SandboxMetadata. I want to make these functions `cri-containerd` independent, so that in the future we may be able to consolidate the logic with cri-o.

pkg/server/sandbox_run.go
```diff
@@ -124,7 +124,28 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+	hostNetwork := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
+	var netNSPath string
+	if hostNetwork {
+		netNSPath, err = hostNetNsPath()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get host netns path: %v", err)
+		}
+	} else {
+		// Create the sandbox network namespace
+		if err = meta.NetNSCreate(); err != nil {
```
Random-Liu: We should be aware of that, if `cri-containerd` crashes after we create the namespace, but before we checkpoint the namespace path, we'll not be able to cleanup the namespace.

Add a TODO here to handle that case in the future.

pkg/server/sandbox_stop.go
```diff
@@ -47,15 +47,19 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
 	if err == nil {
-		if teardownErr := c.netPlugin.TearDownPod(sandbox.NetNS, sandbox.Config.GetMetadata().GetNamespace(),
+		if teardownErr := c.netPlugin.TearDownPod(netNS, sandbox.Config.GetMetadata().GetNamespace(),
 			sandbox.Config.GetMetadata().GetName(), id); teardownErr != nil {
 			return nil, fmt.Errorf("failed to destroy network for sandbox %q: %v", id, teardownErr)
 		}
 	} else if !os.IsNotExist(err) { // It's ok for sandbox.NetNS to *not* exist
 		return nil, fmt.Errorf("failed to stat netns path for sandbox %q before tearing down the network: %v", id, err)
 	}
+	if err = sandbox.NetNSRemove(); err != nil {
```
Random-Liu: Add TODO that, it's still possible that `cri-containerd` crashes after we teardown the network, but before we remove the network namespace.

In that case, we'll not be able to remove the sandbox anymore. The chance is slim, but we should be aware of that.

In the future, once `TearDownPod` is idempotent, this will be fixed.

pkg/metadata/sandbox.go
Random-Liu: Will send out a PR to refactor the metadata store soon. After that this will be much easier.

pkg/server/helpers.go
```diff
@@ -216,3 +210,13 @@ func (c *criContainerdService) getSandbox(id string) (*metadata.SandboxMetadata,
 func criContainerStateToString(state runtime.ContainerState) string {
 	return runtime.ContainerState_name[int32(state)]
 }
+
+// hostNetNsPath gets the host netns path.
+func hostNetNsPath() (string, error) {
```
xlgao-zju: will do

pkg/server/sandbox_run.go
xlgao-zju: will do

pkg/metadata/sandbox.go
xlgao-zju: Will update this PR after your PR is merged.

pkg/metadata/sandbox.go
```diff
@@ -150,3 +164,115 @@ func (s *sandboxStore) List() ([]*SandboxMetadata, error) {
...
+
+type SandboxNetNS struct {
+	sync.Mutex
+	NS       ns.NetNS
+	Symlink  *os.File
+	Closed   bool
+	Restored bool
+}
+
+func (ns *SandboxNetNS) symlinkCreate(name string) error {
```
xlgao-zju: Since I referenced the cri-o v0.3, so I may write something outdated.
So shell I reference the newest version?

pkg/metadata/sandbox.go
Random-Liu: The PR is out. https://github.com/kubernetes-incubator/cri-containerd/pull/66
`NetNS` should be in https://github.com/kubernetes-incubator/cri-containerd/pull/66/files#diff-e723511e95258721117711bb80b67136R27.

I'll try to get it merged as soon as possible.

pkg/metadata/sandbox.go
Random-Liu: Nope. They still have this code, but they plan to remove it. We don't need it on our side.

pkg/metadata/sandbox.go
xlgao-zju: Cool

Random-Liu:
@xlgao-zju I'll close this in favor of #138. :)

PR #59, opened 2017-05-31
Retry and backoff when lost connection with containerd.

Random-Liu:
Retry to reconnect containerd event stream when containerd is down.

### DIFF
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -54,7 +54,7 @@ import (
 
 // CRIContainerdService is the interface implement CRI remote service server.
 type CRIContainerdService interface {
-	Start() error
+	Start()
 	runtime.RuntimeServiceServer
 	runtime.ImageServiceServer
 }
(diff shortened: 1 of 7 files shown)


Random-Liu:
Currently, if containerd is down when `cri-containerd` is running, `cri-containerd` will start printing error log in an infinity loop, which is quite spammy.

This PR also fixed that.

pkg/server/service.go
```diff
@@ -129,5 +129,6 @@ func NewCRIContainerdService(conn *grpc.ClientConn, rootDir, networkPluginBinDir
 }
 
 func (c *criContainerdService) Start() error {
-	return c.startEventMonitor()
+	c.startEventMonitor()
+	return nil
```
mikebrow: nit.. don't need an error now...

Random-Liu: May need in the future. But I'm fine with removing it for now.

Random-Liu: Done

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/59#pullrequestreview-41367282.

PR #60, opened 2017-06-01
Add sandbox /etc/hosts.

Random-Liu:
Add sandbox `/etc/hosts` and mount it inside each container of the sandbox.

Currently, in Kubelet,
* For `/etc/hosts`, we mount it when the sandbox is not using host network, and rely on the runtime to maintain it if the sandbox is using host network.
* For `/etc/resolv.conf`, we rely on the runtime to maintain it (dockershim overwrites it after creates sandbox container).

From the perspective of a runtime, it's weird that we only maintain the file in some cases, and not in other cases.

So in this PR, I'll let `cri-containerd` always maintain `/etc/hosts` no matter the sandbox is using host network or not.

In the future, we should make it consistent:
1) Either let kubelet manage all those files;
2) Let runtime manage all those files.

/cc @yujuhong

### DIFF
--- a/pkg/os/os.go
+++ b/pkg/os/os.go
@@ -32,6 +32,7 @@ type OS interface {
 	RemoveAll(path string) error
 	OpenFifo(ctx context.Context, fn string, flag int, perm os.FileMode) (io.ReadWriteCloser, error)
 	Stat(name string) (os.FileInfo, error)
+	CopyFile(src, dest string, perm os.FileMode) error
 }
 
 // RealOS is used to dispatch the real system level operations.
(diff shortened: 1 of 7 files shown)


Random-Liu:
@Crazykev This PR is very similar with yours. Please take a look at this PR and address the comments in your PR correspondingly.

Crazykev:
@Random-Liu THX, good to see `CopyFile` in Os interface, should I wait this PR get merged first?

Random-Liu:
@Crazykev You could fetch this commit to your PR, and send out.

Will merge your PR after this one.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/60#pullrequestreview-42404873.

PR #64, opened 2017-06-02
Cleanup code.

Random-Liu:
Remove some out-dated TODOs and code.


Random-Liu:
@mikebrow Thanks for reviewing

PR #65, opened 2017-06-02
adds support for v prefix in release tags

mikebrow:
Fixes the issue where we could not use the `v` prefix for release tags.  semver 2.0 removed support for the v prefix.  Still the v prefix is commonly used and expected on release tags in git repos. 

The solution here is to allow for the v prefix on the release tag, but to remove the prefix before setting the version variable in the version package. This way the reported version is sans the v.. but the displayed version includes the v in git hub. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,9 @@ BUILD_DIR ?= _output
 # VERSION is derived from the current tag for HEAD plus amends. Version is used
 # to set/overide the criContainerdVersion variable in the verison package for
 # cri-containerd.
-VERSION:=$(shell git describe --tags --dirty)
+VERSION := $(shell git describe --tags --dirty)
+# strip the first char of the tag if it's a `v`
+VERSION := $(VERSION:v%=%)
 BUILD_TAGS:= -ldflags '-X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)'
 
 all: binaries


Makefile
```diff
@@ -20,7 +20,9 @@ BUILD_DIR ?= _output
 # VERSION is derived from the current tag for HEAD plus amends. Version is used
 # to set/overide the criContainerdVersion variable in the verison package for
 # cri-containerd.
-VERSION:=$(shell git describe --tags --dirty)
+VERSION := $(shell git describe --tags --dirty)
+# strip the first char of the tag if it's a `v`
+VERSION := $(firstword $(subst v, ,$(VERSION)))
```
Random-Liu: Why `firstword`?

mikebrow: I did it in case someone puts a second v in there my thought was to strip everything to the right of the second v .. but yeah $(subst v, ,$(VERSION)) would work just as well

Random-Liu: IIUC, `subst` replaces all occurrence of `v`, which seems to be problematic.
Does `firstword` solve that problem? Don't quite get it.

Random-Liu: `VERSION:=$(VERSION:v%=%)` seems to be better.
I think we could assume that in the future, all versions should start with `v`.

mikebrow: sure that works too

Random-Liu:
LGTM

PR #66, opened 2017-06-06
Rewrite metadata store

Random-Liu:
Previously, containerd only managed running container, so we have to do checkpoint to maintain complete lifecycle of a container and sandbox. That's why we introduced the metadata store, which was targeted to be a checkpoint store for container, sandbox and image if needed.

However, it seems that we should not rely on the metadata store now, because:
1) Containerd added container metadata store, and also supports container label, which could help us checkpoint constant data. We only need to maintain checkpoint of container status now.
2) The metadata store was originally designed for pure data checkpoint, it only supports update via `Update` function, which has some limitations:
  a) `Update` function is not flexible enough, e.g. for image, we need `CreateOrUpdate` operation to create a metadata if it does not exsit, or update it if does.
  b) `Update` and checkpoint doesn't make sense to some in-memory data we need to associate with the metadata lifecycle, e.g. containerd container/image client, cni netns object, stop container channel etc.
  c) The whole metadata store logic is unnecessarily complex given that we don't need a general checkpoint solution now.

Thus I made this change. Relatively large, but help us avoid future pain.

Use container store as an example, the store is organized like this:
```
Store (a simple in-memory map)
   |--> Container (a collection of all resources associated with the container)
             | --> Metadata (constant metadata of the container, will be checkpointed in containerd container label)
             | --> Status (container status maintained in our own checkpoint, the only checkpoint we need now)
             | --> Containerd Container Client (container client of containerd, added soon)
             | --> Container StopCh (a channel indicates container is stopped or not, added soon)
             | --> More... (If needed in the future)
```

For sandbox, we don't need status.
For image, we don't event need a separate `Metadata` field, because containerd provided everything we need.


Random-Liu:
CRI Validation test result is not changed with this PR:
```
Ran 36 of 36 Specs in 94.648 seconds
FAIL! -- 23 Passed | 13 Failed | 0 Pending | 0 Skipped --- FAIL: TestE2ECRI (94.65s)
FAIL
```

hack/verify-lint.sh
```diff
@@ -24,6 +24,7 @@ for d in $(find . -type d -a \( -iwholename './pkg*' -o -iwholename './cmd*' \))
 		 --exclude='.*_test\.go:.*error return value not checked.*\(errcheck\)$' \
 		 --exclude='duplicate of.*_test.go.*\(dupl\)$' \
 		 --exclude='.*/mock_.*\.go:.*\(golint\)$' \
+		 --exclude='declaration of "err" shadows declaration at.*\(vetshadow\)' \
```
mikebrow: Removing will introduce the possibility of programming errors with `err` vars.

Random-Liu: The problem is that the check is known as error prone. It sometimes report error and sometimes not, and I don't think something like:
```
if err := func1(); err != nil {
}
if err := func2(); err != nil {
}
```
should be an error.

mikebrow: only with err? Or are you arguing for a different recommended pattern of use of variables in go?

Random-Liu: Only with err. I mean.

mikebrow: Obviously `err := foo` is just a shortcut for `var err error =` In your above use case you are repetitivly stating var err error, when all you are really doing is assigning with an =. In your use case the cleaner way is to declare first then just use err = since you don't like using := sometimes and = other times, yes?

mikebrow: I guess my question is then, why is err a special use case of variable declaration and assignment? I understand the thought for keeping each test independent but the variable redeclaration doesn't make the test more independent than using assignment. Symmetry?

mikebrow:
When you refactor you really refactor! :-) Will review from the perspective of it being a complete replacement.

hack/verify-lint.sh
Random-Liu: Offline discussed. Keep the check and update the code.

yujuhong:
Will review today.

Random-Liu:
@yujuhong Thanks~

pkg/store/container/status.go
```diff
@@ -0,0 +1,146 @@
...
+		return err
+	}
+	// TODO(random-liu) *Update* existing status on disk atomically,
+	// return error if checkpoint failed.
+	m.status = newStatus
+	return nil
+}
+
+// Delete deletes the container status from disk atomically.
+func (m *statusStorage) Delete() error {
```
yujuhong: Why can't we delete `m.status` in this function? It looks like `Delete()` is only meaningful if we back the data on the disk.

pkg/store/image/image.go
```diff
@@ -0,0 +1,120 @@
...
+	// TODO(random-liu): Add trunc index.
+}
+
+// LoadStore loads images from runtime.
+// TODO(random-liu): Implement LoadStore.
+func LoadStore() *Store { return nil }
+
+// NewStore creates an image store.
+func NewStore() *Store {
+	return &Store{images: make(map[string]Image)}
```
yujuhong: Same question about the rationale of storing `Image` vs `*Image`

pkg/store/image/image.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"sync"
+
+	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/store"
+)
+
+// Image contains all resources associated with the image. All fields
+// are **read-only**.
```
yujuhong: Is that true? In `Add()`, you merge RepoTags and RepoDigest and write it back to the store.

pkg/store/container/container.go
```diff
@@ -0,0 +1,108 @@
...
+package container
+
+import (
+	"sync"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/store"
+)
+
+// Container contains all resources associated with the container. All fields of
+// Container are thread safe.
```
yujuhong: I'm not sure how fields can be thread-safe. I think you mean the methods to mutate the status are thread-safe?

pkg/store/sandbox/sandbox.go
```diff
@@ -0,0 +1,88 @@
...
+	// Metadata is the metadata of the sandbox, it is **read-only**.
+	Metadata
+	// TODO(random-liu): Add containerd container client.
+	// TODO(random-liu): Add cni network namespace client.
+}
+
+// Store stores all sandboxes.
+type Store struct {
+	lock      sync.RWMutex
+	sandboxes map[string]Sandbox
```
yujuhong: Same question about storing pointers vs. structs.

Random-Liu:
Will rebase this and address comments.

Random-Liu:
Moved to v0.2.0, because this PR mainly:
1) Eases the work to switching to new containerd client, and add permanent network namespace.
2) Make it possible to checkpoint most metadata in container label.
3) Make it easier to make future change.

There is no new feature introduced.

Since we don't have time to finish the containerd client switch and permanent network namespace work. Let's also punt this one to v0.2.0.

pkg/store/container/container.go
```diff
@@ -0,0 +1,108 @@
...
+	// TODO(random-liu): Add trunc index.
+}
+
+// LoadStore loads containers from runtime.
+// TODO(random-liu): Implement LoadStore.
+func LoadStore() *Store { return nil }
+
+// NewStore creates a container store.
+func NewStore() *Store {
+	return &Store{containers: make(map[string]Container)}
```
Random-Liu: I use `Container` rather than `*Container` because:
1) The value copy could prevent callers from mutating the struct fields in some way, because most fields are primitive types.
2) Most importantly, when caller using the api, usually they won't expect to be able to mutate fields of a returned value, but may expect to mutate a returned pointer.

As for the overhead, metadata is small, so there shouldn't be too much overhead. :)

pkg/store/container/status.go
Random-Liu: It doesn't matter whether we delete `m.status` or not, because once the container is removed from container list, no one should be able to retrieve and access it.

Delete `m.status` will only cause race condition between operations which already get the reference.

pkg/store/image/image.go
Random-Liu: Will update the comment to "All fields **MUST** not be mutated directly".

pkg/store/image/image.go
Random-Liu: Same with above.

pkg/store/sandbox/sandbox.go
Random-Liu: Same.

Random-Liu:
@yujuhong Rebased the PR. Could you help me take another look? Maybe focus on the first commit.

@mikebrow Could you help me review this PR?

pkg/server/container_create.go
```diff
@@ -158,12 +158,23 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
+			}
+		}
+	}()
+
 	// Add container into container store.
-	if err := c.containerStore.Create(meta); err != nil {
-		return nil, fmt.Errorf("failed to add container metadata %+v into store: %v",
-			meta, err)
+	if err := c.containerStore.Add(container); err != nil {
+		return nil, fmt.Errorf("failed to add container %q into store: %v", id, err)
```
mikebrow: I see the storing of the status in the NewContainer() call but don't understand why the status is stored before the container. Is that to make sure the container always has status? maybe the container store.add could be changed to add container with status to make it more atomic?

pkg/store/container/container.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// Container contains all resources associated with the container. All fields of
+// Container are thread safe.
+type Container struct {
+	// Metadata is the metadata of the container, it is **read-only**.
+	Metadata
+	// Status stores the status of the container.
+	Status StatusStorage
+	// TODO(random-liu): Add containerd container client.
```
mikebrow: ?

pkg/store/container/metadata.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+// NOTE(random-liu):
+// 1) Metadata is read-only.
```
mikebrow: ? you mean can't be updated after create?

pkg/store/container/metadata.go
```diff
@@ -0,0 +1,76 @@
...
+// 2) Metadata is checkpointed as containerd container label.
```
mikebrow: just confirming we'll use labels not annotations for all metadata so we can find the container by the metadata labels or... maybe some combination of labels and annotations...?

pkg/store/image/image.go
mikebrow: Can we do this change universally on the other read only comments?

pkg/metadata/container.go
```diff
@@ -1,183 +0,0 @@
-/*
```
Random-Liu: You won't see the diff even after I merge those 2. Will change to "rewrite".

pkg/server/container_create.go
Random-Liu: `Add` need to lock the container list. I don't think we want to lock the list and do some disk operation.

pkg/server/container_execsync.go
```diff
@@ -45,15 +45,16 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
 		}
 	}()
 
-	// Get container metadata from our container store.
-	meta, err := c.containerStore.Get(r.GetContainerId())
+	// Get container from our container store.
```
Random-Liu: Yeah, that's a problem... :(

pkg/store/container/container.go
Random-Liu: I mean in the future, we could [this](https://github.com/containerd/containerd/blob/master/container.go#L22) here, after we start to use containerd client.

pkg/store/container/metadata.go
Random-Liu: Yeah, will update.

pkg/store/container/metadata.go
Random-Liu: containerd only has label, there is no notion of annotation in containerd.

pkg/store/container/container.go
mikebrow: kk

pkg/server/sandbox_run.go
```diff
@@ -241,10 +232,9 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	}
 
 	// Add sandbox into sandbox store.
-	meta.CreatedAt = time.Now().UnixNano()
-	if err := c.sandboxStore.Create(meta); err != nil {
-		return nil, fmt.Errorf("failed to add sandbox metadata %+v into store: %v",
-			meta, err)
+	sandbox.CreatedAt = time.Now().UnixNano()
```
Random-Liu: We don't maintain state of sandbox. We infer that directly from sandbox container state.

We could not do that for container because we also need exit code, finished timestamp etc. which we have to checkpoint ourselves. But for sandbox, we only care about whether it is running or not.

pkg/server/container_list.go
```diff
@@ -36,33 +34,31 @@ func (c *criContainerdService) ListContainers(ctx context.Context, r *runtime.Li
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to list metadata from container store: %v", err)
-	}
+	// List all containers from store.
+	containersInStore := c.containerStore.List()
 
 	var containers []*runtime.Container
-	for _, meta := range metas {
-		containers = append(containers, toCRIContainer(meta))
+	for _, container := range containersInStore {
```
Random-Liu: What do you mean here? :)

pkg/server/sandbox_run.go
mikebrow: ok yeah been having some debug issues on the update PR for sandboxes that have a messed up state (can't be stopped) then later on can't be created cause /pause isn't there.. the rootfs is messed up.

pkg/server/container_list.go
mikebrow: you renamed range metas to range containersInStore presumably cause you didn't like the term meta... and didn't have a better option than just calling the data container again... :-)

pkg/store/container/metadata.go
mikebrow: oci will have the annotations.. and I suspect containerd will let them pass through without providing first class support

pkg/server/container_create.go
Random-Liu: Done. Added `Container.Delete` function.

pkg/store/container/container.go
Random-Liu: Done.

pkg/store/container/container.go
```diff
@@ -0,0 +1,108 @@
...
+	// TODO(random-liu): Add stop channel to get rid of stop poll waiting.
+}
+
+// NewContainer creates a internal used container type.
```
Random-Liu: Done.

pkg/store/container/container.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return Container{}, err
+	}
+	return Container{
+		Metadata: metadata,
+		Status:   s,
+	}, nil
+}
+
+// LoadContainer loads the internal used container type.
```
Random-Liu: Done.

pkg/store/image/image.go
Random-Liu: Done.

pkg/store/container/metadata.go
Random-Liu: Done.

Random-Liu:
@mikebrow Addressed comments.

The code could be cleaned up more, especially when we start to use containerd client. We could do that in the future.

pkg/server/sandbox_run.go
Random-Liu: I'll help debug your PR over the weekend. :)

PR #72, opened 2017-06-11
Add ExecSync.

Random-Liu:
Add simple `ExecSync` based on current containerd api.

Some TODOs in the future which will be addressed in future PRs:
1) Switch to containerd client and add unit test.
2) Add `ExecSync` timeout support, including timeout waiting and graceful termination after timeout.

### DIFF
--- a/pkg/metadata/container.go
+++ b/pkg/metadata/container.go
@@ -19,9 +19,10 @@ package metadata
 import (
 	"encoding/json"
 
-	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
-
+	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/metadata/store"
 )
 
 // The code is very similar with sandbox.go, but there is no template support
@@ -71,8 +72,13 @@ type ContainerMetadata struct {
 	// In fact, this field doesn't need to be checkpointed.
 	// TODO(random-liu): Skip this during serialization when we put object
 	// into the store directly.
-	// TODO(random-liu): Reset this field to false during state recoverry.
+	// TODO(random-liu): Reset this field to false during state recovery.
 	Removing bool
+	// TODO(random-liu): Remove following field after switching to new containerd
+	// client.
+	// Not including them in unit test now because they will be removed soon.
+	// Spec is the oci runtime spec used to run the container.
+	Spec *runtimespec.Spec
 }
 
 // State returns current state of the container based on the metadata.
--- a/pkg/server/container_execsync.go
+++ b/pkg/server/container_execsync.go
@@ -17,15 +17,136 @@ limitations under the License.
 package server
 
 import (
-	"errors"
+	"bytes"
+	"encoding/json"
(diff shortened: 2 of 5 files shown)


pkg/server/container_execsync.go
```diff
@@ -17,15 +17,134 @@ limitations under the License.
...
+			TypeUrl: runtimespec.Version,
+			Value:   rawSpec,
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to exec in container %q: %v", id, err)
+	}
+	exitCode, err := waitContainerExec(cancel, events, id, resp.Pid, r.GetTimeout())
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait for exec in container %q to finish: %v", id, err)
```
mikebrow: this return eats the unknownExitCode.. What's that for if we are going to nil it here?

pkg/metadata/container.go
```diff
@@ -73,6 +74,11 @@ type ContainerMetadata struct {
 	// into the store directly.
 	// TODO(random-liu): Reset this field to false during state recoverry.
```
Random-Liu: Will do.

Random-Liu: Done.

pkg/metadata/container.go
```diff
@@ -73,6 +74,11 @@ type ContainerMetadata struct {
...
 	Removing bool
+	// TODO(random-liu): Remove following fields after switching to new containerd
+	// client.
+	// Not including them in unit test now because they will be removed soon.
+	// Spec is the oci runtime spec used to run the container.
```
Random-Liu: Initially there are more. Will change to field.

Random-Liu: Done.

pkg/server/container_execsync.go
```diff
@@ -17,15 +17,134 @@ limitations under the License.
...
+				retRes.GetStdout(), retRes.GetStderr())
+		}
+	}()
+
+	// Get container config from container store.
+	meta, err := c.containerStore.Get(r.GetContainerId())
+	if err != nil {
+		return nil, fmt.Errorf("an error occurred when try to find container %q: %v", r.GetContainerId(), err)
+	}
+	id := r.GetContainerId()
```
Random-Liu: Good catch, will do.

Random-Liu: Changed to `id := meta.ID`.

The metadata store may support truncated index soon, so get the id from metadata.

pkg/server/container_execsync.go
Random-Liu: Good catch! We do have some special logic here in dockershim. Let me check.

pkg/server/container_execsync.go
Random-Liu: Done.

pkg/server/container_execsync.go
Random-Liu: If containerd dies during waiting, the exec process may still be running, so I think we should return error instead of 255 exit code without error.

I still keep the unknownExitCode, because:
1) We may still need it when adding the recovery logic after daemon restart.
2) Just in case we use the 0 exit code by mistake.

mikebrow: ok.. will watch for the recovery logic code to see unknownExitCode in action then :-)

PR #73, opened 2017-06-13
Fix Delete race.

Random-Liu:
1. Fix a `Delete` race between `container_stop.go` and `event.go`.

Initially, I thought `Delete` is thread safe. However, I found that if there are 2 on-going `Delete`s, one may stop the shim already, but the other one may have get the container index from containerd and try to access shim and get a `"transport is closing"` error. This seems to be expected behavior. /cc @crosbymichael

Actually I think we should not let `container_stop.go` `Delete` the container before event handler properly handles the exit event, so we should use `Kill` in `container_stop.go`.

2. Ignore `Kill` errors.
Another change made by this PR is to ignore `Kill` error, because when process is already stopped, `Kill` will return an `os: process already finished` error, [which is expected](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#kill).

So in this PR, we only log an error for `Kill` error, and rely on the following `WaitStop` to figure out whether the container is successfully stopped or not. Which is also similar with [what docker is doing today](https://github.com/moby/moby/blob/master/daemon/stop.go#L85).

This PR also shortens the kill container timeout so that if containerd is down or `Kill` returns error, we could get an error sooner.

### DIFF
--- a/pkg/server/container_stop.go
+++ b/pkg/server/container_stop.go
@@ -71,10 +71,10 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
 		glog.V(2).Infof("Stop container %q with signal %v", id, stopSignal)
 		_, err = c.containerService.Kill(ctx, &execution.KillRequest{ID: id, Signal: uint32(stopSignal)})
 		if err != nil {
-			if isContainerdContainerNotExistError(err) {
-				return &runtime.StopContainerResponse{}, nil
+			if !isContainerdContainerNotExistError(err) && !isRuncProcessAlreadyFinishedError(err) {
+				return nil, fmt.Errorf("failed to stop container %q: %v", id, err)
 			}
-			return nil, fmt.Errorf("failed to stop container %q: %v", id, err)
+			// Move on to make sure container status is updated.
 		}
 
 		err = c.waitContainerStop(ctx, id, time.Duration(r.GetTimeout())*time.Second)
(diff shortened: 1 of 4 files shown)


pkg/server/container_stop.go
```diff
@@ -38,7 +38,7 @@ const (
 
 	// killContainerTimeout is the timeout that we wait for the container to
 	// be SIGKILLed.
-	killContainerTimeout = 2 * time.Minute
+	killContainerTimeout = 10 * time.Second
```
yujuhong: Why changing the timeout?

pkg/server/container_stop.go
```diff
@@ -73,7 +73,9 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
 		if isContainerdContainerNotExistError(err) {
 			return &runtime.StopContainerResponse{}, nil
 		}
-		return nil, fmt.Errorf("failed to stop container %q: %v", id, err)
+		// Do not return error here because the container process may
+		// not be running.
+		glog.Errorf("Failed to stop container %q: %v", id, err)
```
yujuhong: Use `Warning` instead based on the comment.

pkg/server/container_stop.go
yujuhong: Oh...because now we always wait to see if the container has been stopped.

pkg/server/container_stop.go
```diff
@@ -82,13 +84,15 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
...
 		if isContainerdContainerNotExistError(err) {
 			return &runtime.StopContainerResponse{}, nil
 		}
-		return nil, fmt.Errorf("failed to delete container %q: %v", id, err)
+		// Do not return error here because the container process may
+		// not be running.
+		glog.Errorf("Failed to kill container %q: %v", id, err)
 	}
 
 	// Wait forever until container stop is observed by event monitor.
```
yujuhong: nit: not *forever* any more.

pkg/server/container_stop.go
```diff
@@ -82,13 +84,15 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
 	}
 	glog.Errorf("Stop container %q timed out: %v", id, err)
 
-	glog.V(2).Infof("Delete container from containerd %q", id)
-	_, err = c.containerService.Delete(ctx, &execution.DeleteRequest{ID: id})
+	glog.V(2).Infof("Kill container %q", id)
+	_, err = c.containerService.Kill(ctx, &execution.KillRequest{ID: id, Signal: uint32(unix.SIGKILL)})
```
yujuhong: I think I've missed something. Where do you `Delete` the container after this change?

pkg/server/container_stop.go
```diff
@@ -82,13 +84,15 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
...
 	if err != nil {
 		if isContainerdContainerNotExistError(err) {
 			return &runtime.StopContainerResponse{}, nil
 		}
-		return nil, fmt.Errorf("failed to delete container %q: %v", id, err)
+		// Do not return error here because the container process may
+		// not be running.
+		glog.Errorf("Failed to kill container %q: %v", id, err)
```
yujuhong: s/Error/Warning

Random-Liu:
@yujuhong Sorry, as is discussed. I updated the PR to parse the returned error, and filed issue to containerd https://github.com/containerd/containerd/issues/999.

pkg/server/container_stop.go
Random-Liu: Changed back.

pkg/server/container_stop.go
Random-Liu: Go with parsing the error, will not log the error now.

pkg/server/container_stop.go
Random-Liu: In `events.go`, will add a comment to explain here.

pkg/server/container_stop.go
Random-Liu: Ditto.

pkg/server/container_stop.go

pkg/server/container_stop.go

yujuhong:
LGTM

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/73#issuecomment-308543685.

PR #75, opened 2017-06-13
Kill container directly if timeout is 0.

Random-Liu:
Based on [CRI Defination](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/cri/v1alpha1/runtime/api.proto#L652-L654), timeout 0 means `forcibly terminate the container immediately`.

So skip graceful deletion when timeout is 0.

### DIFF
--- a/pkg/server/container_stop.go
+++ b/pkg/server/container_stop.go
@@ -65,24 +65,28 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
 		return &runtime.StopContainerResponse{}, nil
 	}
 
-	// TODO(random-liu): [P1] Get stop signal from image config.
-	stopSignal := unix.SIGTERM
-	glog.V(2).Infof("Stop container %q with signal %v", id, stopSignal)
-	_, err = c.containerService.Kill(ctx, &execution.KillRequest{ID: id, Signal: uint32(stopSignal)})
-	if err != nil {
-		if isContainerdContainerNotExistError(err) {
+	if r.GetTimeout() > 0 {
+		// TODO(random-liu): [P1] Get stop signal from image config.
+		stopSignal := unix.SIGTERM
+		glog.V(2).Infof("Stop container %q with signal %v", id, stopSignal)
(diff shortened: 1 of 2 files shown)


pkg/server/container_stop.go
```diff
@@ -65,22 +65,24 @@ func (c *criContainerdService) StopContainer(ctx context.Context, r *runtime.Sto
...
-	// TODO(random-liu): [P1] Get stop signal from image config.
-	stopSignal := unix.SIGTERM
-	glog.V(2).Infof("Stop container %q with signal %v", id, stopSignal)
-	_, err = c.containerService.Kill(ctx, &execution.KillRequest{ID: id, Signal: uint32(stopSignal)})
-	if err != nil {
-		if isContainerdContainerNotExistError(err) {
-			return &runtime.StopContainerResponse{}, nil
+	if r.GetTimeout() > 0 {
+		// TODO(random-liu): [P1] Get stop signal from image config.
+		stopSignal := unix.SIGTERM
```
mikebrow: did we want to send a SIGKILL if the SIGTERM times out?

Random-Liu: This means that if a timeout is set, we use `SIGTERM` to gracefully stop the container;
If not, we `SIGKILL` the container directly.

mikebrow: if zero no kill?

Random-Liu: `Delete` includes a `SIGKILL`.

mikebrow: Please comment below that calling c.containerService.Delete results in a SIGKILL being sent

Random-Liu: Will do.

mikebrow: calling delete on a running container will not kill.. so probably need to send a kill if timeout is zero...

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/75#pullrequestreview-43831790.

Random-Liu:
Offline discussed with @yujuhong, she's also ok with the change. Merge it.

PR #76, opened 2017-06-14
Fix a race that fake execution client sends event to closed channel.

Random-Liu:
We should clear the client list after close all event client channels. Or else [this line](https://github.com/Random-Liu/cri-containerd/blob/2ae22b33b7de6d327369aa35da66f6d4acd9c194/pkg/server/testing/fake_execution_client.go#L102) may send event to the closed channel which causes a panic.

### DIFF
--- a/pkg/server/testing/fake_execution_client.go
+++ b/pkg/server/testing/fake_execution_client.go
@@ -89,6 +89,7 @@ func (f *FakeExecutionClient) Stop() {
 	for _, client := range f.eventClients {
 		close(client.Events)
 	}
+	f.eventClients = nil
 }
 
 // WithEvents setup events publisher for FakeExecutionClient


hack/verify-lint.sh
```diff
@@ -30,5 +30,5 @@ for d in $(find . -type d -a \( -iwholename './pkg*' -o -iwholename './cmd*' \))
 		 --cyclo-over=60 \
 		 --dupl-threshold=100 \
 		 --tests \
-		 --deadline=300s "${d}"
+		 --deadline=600s "${d}"
```
mikebrow: I had to go all the way to 900.. but really it's random issue based on memory use.. fix is in progress on gometalinter.. PR #289

pkg/server/testing/fake_execution_client.go
```diff
@@ -89,6 +89,7 @@ func (f *FakeExecutionClient) Stop() {
 	for _, client := range f.eventClients {
 		close(client.Events)
 	}
+	f.eventClients = nil
```
mikebrow: isn't the underlying root issue here that f.eventsQueue isn't thread safe? 

Did you want to kill eventClients or empty it?  `f.eventClients = f.eventClients[:0]`

Random-Liu: > isn't the underlying root issue here that f.eventsQueue isn't thread safe?

That's why we have the lock above.

>  Did you want to kill eventClients or empty it? f.eventClients = f.eventClients[:0]

For golang, either is ok. :)

hack/verify-lint.sh
Random-Liu: I updated https://github.com/kubernetes-incubator/cri-containerd/pull/77. Let's see whether it's a travis error, maybe it's fixed today.

pkg/server/testing/fake_execution_client.go
mikebrow: kk yeah saw in a blog nil is acceptable learned something new!

mikebrow: don't follow the lock is below the access to f.eventsQueue

PR #80, opened 2017-06-19
Register all possible repo tags and repo digests.

Random-Liu:
Register both repo tag and repo digest when pulling image.

Currently, Kubelet assumes that once image is pulled with repo tag, the repo digest also needs to be stored as existing reference.

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -31,7 +31,6 @@ import (
 	"github.com/containerd/containerd/remotes/docker/schema1"
 	containerdrootfs "github.com/containerd/containerd/rootfs"
 	"github.com/golang/glog"
-	imagedigest "github.com/opencontainers/go-digest"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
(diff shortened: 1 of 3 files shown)


pkg/server/image_pull.go
```diff
@@ -265,34 +264,43 @@ func (c *criContainerdService) pullImage(ctx context.Context, ref string) (
...
-			ref, desc, err)
+	repoDigest, repoTag := getRepoDigestAndTag(namedRef, desc.Digest)
+	if ref != repoTag && ref != repoDigest {
+		return "", "", "", fmt.Errorf("unexpected repo tag %q and repo digest %q for %q", repoTag, repoDigest, ref)
+	}
+	for _, r := range []string{repoTag, repoDigest} {
+		if r == "" {
+			continue
+		}
+		if err = c.imageStoreService.Put(ctx, r, desc); err != nil {
```
yujuhong: nit: `err :=` for the scope.

pkg/server/image_pull.go
```diff
@@ -265,34 +264,43 @@ func (c *criContainerdService) pullImage(ctx context.Context, ref string) (
 	// 2) We need desc returned by schema1 converter.
 	// So just put the image metadata after downloading now.
 	// TODO(random-liu): Fix the potential garbage collection race.
-	if err = c.imageStoreService.Put(ctx, ref, desc); err != nil {
-		return "", "", fmt.Errorf("failed to put image %q desc %v into containerd image store: %v",
-			ref, desc, err)
+	repoDigest, repoTag := getRepoDigestAndTag(namedRef, desc.Digest)
+	if ref != repoTag && ref != repoDigest {
+		return "", "", "", fmt.Errorf("unexpected repo tag %q and repo digest %q for %q", repoTag, repoDigest, ref)
```
yujuhong: Is the sanity check necessary? You pass `namedRef` to the function, and `ref` is just `namedRef.String()`...
If you must have this check, I think it should be in the `getRepoDigestAndTag` function.

Random-Liu: Because in code below, we use `ref` to get image metadata from the containerd image store, which is making the assumption that `ref` is either `repoTag` or `repoDigest`.

Initially, I thought I should add a comment here, but then I remember one time I read in the toilet that sanity check is better than comment. So...

pkg/server/image_pull.go
Random-Liu: Will do.

mikebrow: https://github.com/kubernetes-incubator/cri-containerd/pull/82/commits/97063a0e3483e61427340f1fac93909da4e687b9#diff-74b3ed4ae41ffcda29899d0eacf5c279R27

Random-Liu: Done.

pkg/server/image_pull.go
Random-Liu: Changed to comment.

Random-Liu: @yujuhong I added the sanity check back, because it did help me catch a bug for schema 1 image.

Please take a look at the latest commit.

yujuhong:
lgtm

Random-Liu:
Will apply label and merge after squash. @yujuhong Thanks for reviewing!.

PR #98, opened 2017-06-29
Do not teardown network namespace when using host network.

Random-Liu:
We should not teardown pod network when sandbox is using host network.

@xlgao-zju Could you help me review this?


xlgao-zju:
LGTM
Will merge this if the travis is happy.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/98#issuecomment-311842705

Thanks for reviewing! @xlgao-zju

PR #99, opened 2017-06-29
Change containerd log level to debug for cri validation test.

Random-Liu:
CHange containerd log level to debug for better debugging.

### DIFF
--- a/hack/test-cri.sh
+++ b/hack/test-cri.sh
@@ -60,7 +60,7 @@ if [ ! -x "$(command -v containerd)" ]; then
   exit 1
 fi
 sudo pkill containerd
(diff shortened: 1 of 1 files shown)


hack/test-cri.sh
```diff
@@ -60,7 +60,7 @@ if [ ! -x "$(command -v containerd)" ]; then
   exit 1
 fi
 sudo pkill containerd
-sudo containerd &> ${REPORT_DIR}/containerd.log &
+sudo containerd -l debug &> ${REPORT_DIR}/containerd.log &
```
mikebrow: irony :-)

Random-Liu: We can enable it in our test, but should not let user add the flag by default when following our dobument.

PR #102, opened 2017-07-28
Get spec from metadata

Random-Liu:
Remove spec from our metadata and get spec from containerd container metadata.

This PR also updated cri validation test version to include https://github.com/kubernetes-incubator/cri-tools/pull/115.


hack/versions
```diff
@@ -1,4 +1,4 @@
 RUNC_VERSION=639454475cb9c8b861cc599f8bcd5c8c790ae402
 CNI_VERSION=v0.4.0
 CONTAINERD_VERSION=8ed1e24ae925b5c6d8195858ee89dddb0507d65f
-CRITEST_VERSION=v0.1
+CRITEST_VERSION=dbfd7f14a457fc95cf736c5361b244c087cc9688
```
mikebrow: ? Don't see that commit

mikebrow: Ah  the merged commit is .. 74bbd4e142f752f13c648d9dde23defed3e472a2

Random-Liu: Sorry, will update.

pkg/server/container_execsync.go
```diff
@@ -55,6 +56,22 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
 		return nil, fmt.Errorf("container %q is in %s state", id, criContainerStateToString(meta.State()))
 	}
 
+	// Get exec process spec.
```
mikebrow: above comment // Get container config from container store .. should be changed to // Get metadata for container from our container store .. or something like that

pkg/server/container_execsync.go
```diff
@@ -55,6 +56,22 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
...
+	cntrResp, err := c.containerService.Get(ctx, &containers.GetContainerRequest{ID: id})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get container %q from containerd: %v", id, err)
```
mikebrow: failed to get container struct.. or container information..

hack/versions
Random-Liu: Done.

pkg/server/container_execsync.go
Random-Liu: I don't quite get it.

pkg/server/container_execsync.go
Random-Liu: The metadata is just the container in containerd. :p

pkg/server/container_execsync.go
mikebrow: see the comment above for call to get metadata for the container from the container metadata store

Random-Liu: I see. Will do.

pkg/server/container_execsync.go
mikebrow: right, just trying to be more specific in the error message.. get container doesn't say much. The response includes the container struct that has the spec in it. Was thinking more detail might help down the road.

mikebrow: iow what's the difference between failing to get the container when pulling an image vs failing to get the container here?

mikebrow: ah well the line number is enough.. either way is ok by me 👍

PR #108, opened 2017-08-02
loop waiting for containerd server to respond to client connection 

mikebrow:
Adds a loop that increases in delay five times while attempting to wait for the containerd server to come up enough that it can respond to a version request from the client.

### DIFF
--- a/hack/test-cri.sh
+++ b/hack/test-cri.sh
@@ -26,9 +26,9 @@ SKIP=${SKIP:-"Streaming|RunAsUser|host port"}
 REPORT_DIR=${REPORT_DIR:-"/tmp"}
 
 if [[ -z "${GOPATH}" ]]; then
-  echo "GOPATH is not set"	
+  echo "GOPATH is not set"
   exit 1
-fi       
+fi
 
 if [[ ! "${PATH}" =~ (^|:)${GOPATH}/bin(|/)(:|$) ]]; then
   echo "GOPATH/bin is not in path"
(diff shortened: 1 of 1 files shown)


hack/test-cri.sh
```diff
@@ -26,9 +26,9 @@ SKIP=${SKIP:-"Streaming|RunAsUser|host port"}
 REPORT_DIR=${REPORT_DIR:-"/tmp"}
 
 if [[ -z "${GOPATH}" ]]; then
-  echo "GOPATH is not set"	
+  echo "GOPATH is not set"
```
Random-Liu: Thanks for the cleanup!

hack/test-cri.sh
```diff
@@ -62,9 +62,17 @@ fi
 sudo pkill containerd
 sudo containerd -l debug &> ${REPORT_DIR}/containerd.log &
 
+MAX_ATTEMPTS=5
+attempt_num=1
+until sudo ctr version | grep -m 1 "Server:" || (( attempt_num == MAX_ATTEMPTS ))
```
Random-Liu: Good idea to use `ctr`. No need to grep, just:
```shell
until sudo ctr version &> /dev/null || (( attemp_num == MAX_ATTEMPTS ))
do
  ...
done
```

mikebrow: kk.. will check that out in the morning :-)

mikebrow:
rebased.. removed sleep 1 from master, updated to address above comment from @Random-Liu...

Random-Liu:
LGTM

PR #110, opened 2017-08-03
Use vndr instead of godep.

Random-Liu:
To support container streaming, I need to import the streaming server from kubernetes. However, it's impossible to do that correctly with Godep.
You'll keep getting following error although no packages we are depending on are using it:
```
$ godep save ./...
godep: Package (github.com/docker/distribution/digest) not found
```

So as to avoid future trouble and make cri-containerd consistent with moby, containerd and cri-o, let's switch to `vndr` instead.


Random-Liu:
Also updated the `README.md` to include information about how to update dependencies.

abhinandanpb:
LGTM

Random-Liu:
> Over in CONTRIBUTING.md you'll need to modify or delete the part about godep...
https://github.com/kubernetes-incubator/cri-containerd/blob/master/CONTRIBUTING.md#adding-dependencies
Or just move this readme edit to contributing.md

Sorry. I thought contributing is linking to outside the repo, and just found that it's inside the repo.

Random-Liu:
@mikebrow Done.

PR #111, opened 2017-08-04
Use containerd client for container execsync

Random-Liu:
I'll be using new containerd client when implementing containerd streaming, so update container execsync first.

/cc @abhinandanpb 

Note that I'm not using `process.Wait`. Will use it after https://github.com/containerd/containerd/pull/1287 and https://github.com/containerd/containerd/pull/1279 are merged.

### DIFF
--- a/pkg/server/container_execsync.go
+++ b/pkg/server/container_execsync.go
@@ -18,17 +18,12 @@ package server
 
 import (
 	"bytes"
-	"encoding/json"
 	"fmt"
-	"io"
-	"io/ioutil"
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/api/services/events/v1"
-	"github.com/containerd/containerd/api/services/tasks/v1"
 	"github.com/containerd/containerd/typeurl"
-	prototypes "github.com/gogo/protobuf/types"
 	"github.com/golang/glog"
-	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 )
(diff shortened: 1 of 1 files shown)


pkg/server/container_execsync.go
```diff
@@ -48,92 +43,68 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
...
-		ExecID:      execID,
-	}); err != nil && !isContainerdGRPCNotFoundError(err) {
-		return nil, fmt.Errorf("failed to delete exec %q in container %q: %v", execID, id, err)
-	}
 	// TODO(random-liu): [P1] Deal with timeout, kill and wait again on timeout.
 
-	// TODO(random-liu): Make sure stdout/stderr are drained.
+	// Wait for the io to be drained.
+	task.IO().Wait()
+
```
mikebrow: task.IO().Close() ?

Random-Liu: That's included in `task.Delete`.

pkg/server/container_execsync.go
```diff
@@ -48,92 +43,68 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
...
+	process, err := task.Exec(ctx, execID, pspec, io)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create exec streaming directory: %v", err)
+		return nil, fmt.Errorf("failed to create exec %q: %v", execID, err)
 	}
 	defer func() {
-		if err = c.os.RemoveAll(execDir); err != nil {
-			glog.Errorf("Failed to remove exec streaming directory %q: %v", execDir, err)
+		if _, err := process.Delete(ctx); err != nil {
+			glog.Errorf("Failed to delete exec process %q for container %q: %v", execID, id, err)
```
mikebrow: for sanity can we check the exit code from this Delete call against the exit code from c.waitContainerExec? Or am I missing something?

pkg/server/container_execsync.go
mikebrow: i see then probably just need to close if there's an error on task.exec above..

pkg/server/container_execsync.go
```diff
@@ -48,92 +43,68 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
...
-		},
-		ExecID: execID,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to exec in container %q: %v", id, err)
+	if err := process.Start(ctx); err != nil {
+		return nil, fmt.Errorf("failed to start exec %q: %v", execID, err)
 	}
+
 	exitCode, err := c.waitContainerExec(eventstream, id, execID)
```
mikebrow: should this be process.Wait() now?

pkg/server/container_execsync.go
Random-Liu: `Exec` has handled cleanup on error inside.

pkg/server/container_execsync.go
Random-Liu: Seems not necessary. :)

pkg/server/container_execsync.go
Random-Liu: Let's use `Wait` after we bump up containerd version including the fixes for `Wait`.

Random-Liu:
@mikebrow The cleanup issue will be addressed in https://github.com/containerd/containerd/pull/1290.

pkg/server/container_execsync.go
mikebrow: kk a // TODO then.

Random-Liu: There is one above.

pkg/server/container_execsync.go
mikebrow: still a sanity check won't hurt... you can add this should not happen if you like to the error message that is un-necessary :-)

mikebrow: just makes me uneasy getting the exit code earlier than the Delete()

Random-Liu: Check ctr/run.go, or ctr/exec. They are doing similar thing.

Random-Liu: In the future , this will be Wait. Wait is guaranteed to return exit status if there is no error.

mikebrow: right and the delete invokes a request that also retrieves and exit status..

Sorry just my intuition here that there may come a day when one of the two ways to get the exit status will have an error or a different value and if so we'd want to investigate cause.

Random-Liu:
@mikebrow Sorry for the delay. Went to take a nap just now. :p 
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/111#pullrequestreview-54470601.

Issue #106, opened 2017-08-02
Change all Sirupsen to sirupsen


Random-Liu:
Ref
- [x] https://github.com/containerd/containerd/issues/1266;
- [x] https://github.com/moby/moby/pull/34272;
- [x] https://github.com/kubernetes-incubator/cri-o/issues/712.

We should change vendored `Sirupsen` to `sirupsen` after dependencies are resolved.

mikebrow:
Yeah problem is some folks have moved and some have not leaving us in a golang black hole.

Random-Liu:
`docker/docker/pkg/signal` has removed the logrus usage, we could upgrade docker version to get rid of that.

PR #112, opened 2017-08-05
Add script to rewrite `Sirupsen` to `sirupsen` in vendor.

Random-Liu:
`sirupsen` was changed to upper case in https://github.com/kubernetes-incubator/cri-containerd/pull/110, which causes problem.

Travis could still build, because there is `Sirupsen` cached in `$GOPATH`, and upper case `Sirupsen` in `$GOPATH` still works but doesn't work in `vendor`, which seems to be a golang problem.

Anyway, we need to rewrite until https://github.com/kubernetes-incubator/cri-containerd/issues/106 is fixed.

### DIFF
--- a/hack/update-vendor.sh
+++ b/hack/update-vendor.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


ijc:
FYI https://github.com/moby/moby/commit/cc4da8112814cdbb00dbf23370f9ed764383de1f recently removed the use of `logrus` from `github.com/docker/docker/pkg/signal` so for that one at least you also have the option of bumping `vendor.conf`.

Random-Liu:
@ijc Cool! Good to know! Thanks! Will do that in another PR.

Issue #86, opened 2017-06-21
Upgrade containerd api.


Random-Liu:
Containerd is still making changes to API, e.g. https://github.com/containerd/containerd/pull/1047, https://github.com/containerd/containerd/pull/1062.

We should upgrade once more after they finalize the api.

mikebrow:
A quick review of 1040 and 1047 they don't look like major changes just some renames.

Random-Liu:
> A quick review of 1040 and 1047 they don't look like major changes just some renames.

Yeah, but without rename, cri-containerd won't work with new containerd because grpc requires both sides using the same package name. :)

crosbymichael:
Ya, we are doing our final API review so we have something consistent and good for the long term where its versioned as well.  Sorry for the churn right now but its best to do this now, before a release, then later.

Random-Liu:
@crosbymichael Yeah, no worry. :)

Just track the api change here, so that we do rebase we have better ideas what need to be changed.

crosbymichael:
When we cut the 1.0 alpha it would be a good time to rebase on it

mikebrow:
main changes:
* rename execution->tasks
* execution/task events are now implemented by a new events service
* all services now have the /v1 suffix

Random-Liu:
@mikebrow Thanks! We could keep updating here, so as to make our future rebase easier.

crosbymichael:
We still have 1-2 API level changes to make before cutting the Alpha.  I'll let you know when we do.

mikebrow:
Work in progress synching up to the containerd/containerd current master commit 4888d40ac7ca94b914c5f37c1f8129eab2df3790

mikebrow:
Moving to the latest commit to pick up some more client apis 
4118a256e9f17199ed2ab647fad6946cc59fcd81

Issue #49, opened 2017-05-25
Use containerd client.


Random-Liu:
Containerd is adding a new client, which will make integration and our unit test much easier.

We should use it after it lands.

https://github.com/containerd/containerd/pull/904

mikebrow:
Looks nice.. will have to see if it does all we need.

Random-Liu:
Just a note here, the so-called `AgentFactory` could be removed after we switch to new containerd client.

Random-Liu:
Given that:
1) The timeframe is pretty tight;
2) Containerd client may miss some features we need, e.g. some image functions.

I propose the following plan:
- [ ] [P0] Upgrade to new containerd **API**, and make it work.
- [ ] [P1] Replace all service clients with "client.service" (containerd expose all services directly).
- [ ] [P0] Cut the alpha.1 release, which I think highly possibly will be our final alpha release version.
- [ ] [P2] Switching each function to use new containerd client one by one.

@kubernetes-incubator/maintainers-cri-containerd 
/cc @crosbymichael @stevvooe

Random-Liu:
Change this to v0.2.0. I don't think we have time to switch to the new client. Making sure we are using new containerd api is good enough for this release.

@mikebrow Let's finish the second step in this release and leave the client switch to next one.

PR #113, opened 2017-08-05
Replacing containerd GRPC API with client

abhinandanpb:
This commit 
1) Replaces the usage of containerd GRPC APIs with the containerd client for all operations related to containerd.
2) Updated containerd to v1.0alpha4+
3) Updated runc to v1.0.0


Random-Liu:
@abhinandanpb Thanks! Ping me when you are ready, and feel free to tell me if you have any questions. :)

abhinandanpb:
I have signed the linux foundation cla

pkg/store/sandbox/sandbox_test.go
```diff
@@ -78,7 +78,7 @@ func TestSandboxStore(t *testing.T) {
 	assert := assertlib.New(t)
 	sandboxes := map[string]Sandbox{}
 	for _, id := range ids {
-		sandboxes[id] = Sandbox{metadatas[id]}
```
Random-Liu:
```
Sandbox{Metadata: metadatas[id]}
```

pkg/store/sandbox/sandbox.go
```diff
@@ -28,6 +28,8 @@ type Sandbox struct {
 	// Metadata is the metadata of the sandbox, it is immutable after created.
 	Metadata
 	// TODO(random-liu): Add containerd container client.
```
Random-Liu: Remove the TODO.

pkg/store/sandbox/sandbox.go
```diff
@@ -28,6 +28,8 @@ type Sandbox struct {
...
+	// Containerd sandbox
+	Container containerd.Container
```
Random-Liu: Seems better to call this `sandbox container`.
```
// Containerd sandbox container
```

pkg/store/sandbox/sandbox.go
```diff
@@ -17,9 +17,9 @@ limitations under the License.
 package sandbox
 
 import (
-	"sync"
-
+	"github.com/containerd/containerd"
 	"github.com/kubernetes-incubator/cri-containerd/pkg/store"
+	"sync"
```
Random-Liu: nit: Move "sync" up. Our style is usually:
```go
import (
"sync" // golang libraries

"github.com/containerd/containerd" // Other libraries

"github.com/kubernetes-incubator/cri-containerd/pkg/store" // cri-containerd libraries
)
```

pkg/store/image/image.go
```diff
@@ -39,6 +39,8 @@ type Image struct {
 	Size int64
 	// Config is the oci image config of the image.
 	Config *imagespec.ImageConfig
+	// container image reference
+	Image containerd.Image
 	// TODO(random-liu): Add containerd image client.
```
Random-Liu: Remove the TODO.

And `// Containerd image reference` seems better.

pkg/store/image/image.go
```diff
@@ -19,9 +19,9 @@ package image
 import (
 	"sync"
 
-	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
-
+	"github.com/containerd/containerd"
 	"github.com/kubernetes-incubator/cri-containerd/pkg/store"
+	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
```
Random-Liu: Same rule for import.

pkg/store/container/container_test.go
```diff
@@ -99,7 +99,7 @@ func TestContainerStore(t *testing.T) {
 	assert := assertlib.New(t)
 	containers := map[string]Container{}
 	for _, id := range ids {
-		container, err := NewContainer(metadatas[id], statuses[id])
+		container, err := NewContainer(metadatas[id], statuses[id],nil)
```
Random-Liu: Won't gofmt complain here? you need a space before `nil`.

pkg/store/container/container.go
```diff
@@ -30,18 +31,22 @@ type Container struct {
 	// Status stores the status of the container.
 	Status StatusStorage
 	// TODO(random-liu): Add containerd container client.
```
Random-Liu: Remove the `TODO`.

pkg/store/container/container.go
```diff
@@ -19,6 +19,7 @@ package container
 import (
 	"sync"
 
+	"github.com/containerd/containerd"
```
Random-Liu: Follow the import convention.

pkg/server/status.go
```diff
@@ -38,18 +37,24 @@ func (c *criContainerdService) Status(ctx context.Context, r *runtime.StatusRequ
 		Type:   runtime.RuntimeReady,
 		Status: true,
 	}
-	// Use containerd grpc server healthcheck service to check its readiness.
-	resp, err := c.healthService.Check(ctx, &healthapi.HealthCheckRequest{})
-	if err != nil || resp.Status != healthapi.HealthCheckResponse_SERVING {
+
+	if c.client == nil {
```
Random-Liu: Don't think we'll start `cri-containerd` if client is nil, so maybe we should ignore this case.

pkg/server/sandbox_run.go
```diff
@@ -77,55 +73,34 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
-		Runtime: containers.RuntimeInfo{Name: defaultRuntime},
-		Spec: &prototypes.Any{
-			TypeUrl: runtimespec.Version,
-			Value:   rawSpec,
-		},
-		RootFS: id,
-	}); err != nil {
+
+	// TODO(random-liu): Checkpoint metadata into container labels.
+	var opts []containerd.NewContainerOpts
```
Random-Liu: Why not:
```go
opts := []containerd.NewContainerOpts{
containerd.WithSpec(spec),
containerd.WithImage(image.Image),
containerd.WithRuntime(defaultRuntime),
containerd.WithNewSnapshotView(id, image.Image),
}
```

pkg/server/sandbox_run.go
```diff
@@ -77,55 +73,34 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+	opts = append(opts, []containerd.NewContainerOpts{
+		containerd.WithSpec(spec),
+		containerd.WithImage(image.Image),
+		containerd.WithRuntime(defaultRuntime),
+		containerd.WithNewSnapshotView(id, image.Image)}...)
+
+	var container containerd.Container
+	if container, err = c.client.NewContainer(ctx, id, opts...); err != nil {
```
Random-Liu: nit:
```go
container, err := c.client.NewContainer(ctx, id, opts...)
if err != nil {
}
```

pkg/server/sandbox_run.go
```diff
@@ -77,55 +73,34 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
-		RootFS: id,
-	}); err != nil {
+
+	// TODO(random-liu): Checkpoint metadata into container labels.
+	var opts []containerd.NewContainerOpts
+	opts = append(opts, []containerd.NewContainerOpts{
+		containerd.WithSpec(spec),
+		containerd.WithImage(image.Image),
+		containerd.WithRuntime(defaultRuntime),
+		containerd.WithNewSnapshotView(id, image.Image)}...)
```
Random-Liu: There is a problem with the containerd client. If `NewContainer` fails, no one will cleanup the snapshot.

In our original implementation, we'll remove the snapshot. But in their client, it's not removed.

This is not your problem, will file an issue to them. But could you add a TODO here to track that?

pkg/server/sandbox_run.go
```diff
@@ -222,14 +194,23 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 		}()
 	}
 
-	// Start sandbox container in containerd.
-	if _, err := c.taskService.Start(ctx, &tasks.StartTaskRequest{ContainerID: id}); err != nil {
-		return nil, fmt.Errorf("failed to start sandbox container %q: %v",
+	err = task.Start(ctx)
```
Random-Liu: nit to save one line:
```go
if err := task.Start(ctx); err != nil {
}
```

pkg/server/sandbox_run.go
```diff
@@ -179,32 +154,29 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to create sandbox container %q: %v",
-			id, err)
+		return nil, fmt.Errorf("failed to create task for sandbox:%q: %v", container.ID(), err)
 	}
 	defer func() {
 		if retErr != nil {
-			// Cleanup the sandbox container if an error is returned.
-			if err := c.stopSandboxContainer(ctx, id); err != nil {
+			if _, err := task.Delete(ctx); err != nil {
```
Random-Liu: You could not `Delete` a task once it's created, unless you use the new `WithProcessKill` option https://github.com/containerd/containerd/pull/1301.

We need the orignal code here, and replace it with `Delete` `WithProcessKill` after we bump up containerd version.

pkg/server/sandbox_run.go
```diff
@@ -179,32 +154,29 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
 	}
-	// Create sandbox task in containerd.
-	glog.V(5).Infof("Create sandbox container (id=%q, name=%q) with options %+v.",
-		id, name, createOpts)
-	createResp, err := c.taskService.Create(ctx, createOpts)
+	task, err := container.NewTask(ctx, containerd.NewIOWithFifoSet(os.Stdin, os.Stdout, os.Stderr, fifo, false))
 	if err != nil {
-		return nil, fmt.Errorf("failed to create sandbox container %q: %v",
-			id, err)
+		return nil, fmt.Errorf("failed to create task for sandbox:%q: %v", container.ID(), err)
```
Random-Liu: nit: s/sandbox:%q/sandbox %q.

pkg/server/sandbox_run.go
abhinandanpb: sure I can fix it in containerd

pkg/server/status.go
abhinandanpb: I did that only for this if some test cases we expect the error in that format. Happy to remove it . Dont have a strong opinion on this one.

pkg/server/sandbox_run.go
abhinandanpb: For this PR to work we anyways need to bump containerd version. I can incorporate the change. WDYT ?

pkg/store/sandbox/sandbox.go
```diff
@@ -19,6 +19,7 @@ package sandbox
 import (
 	"sync"
 
+	"github.com/containerd/containerd"
```
Random-Liu: nit: Maybe too picky, but may help with future PRs, :p 
```
import(
"sync" // golang libraries
// empty line
"github.com/containerd/containerd" // other libraries
// empty line
"github.com/kubernetes-incubator/cri-containerd/pkg/store" // CRI containerd libraries
)
```

And also the same with the other imports.

pkg/server/status.go
Random-Liu: Seems not required. Because we've been using `c.client.XXX` all around, :)

pkg/server/sandbox_run.go
Random-Liu: Cool. Then let's add TODO on our side, and try to fix it in containerd. :)

pkg/server/sandbox_run.go
Random-Liu: Sure. Let's update containerd version then.

There may be some change in their api and client, you may need to update your PR correspondingly.

mikebrow: Delete with kill looks useful 
https://github.com/containerd/containerd/commit/d8c075aadc3f2c42684030f0b7cf17e9575a94c7

pkg/server/container_create.go
```diff
@@ -79,12 +77,11 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	// Prepare container image snapshot. For container, the image should have
 	// been pulled before creating the container, so do not ensure the image.
 	imageRef := config.GetImage().GetImage()
-	image, err := c.localResolve(ctx, imageRef)
+
+	// Ensure sandbox container image snapshot.
+	image, err := c.ensureImageExists(ctx, imageRef)
```
Random-Liu: In CRI, the assumption is that when create container, the image should have already been pulled, if not create should return error. So we could not use `ensureImageExists` in container create.

We use `ensureImageExists` in `RunPodSandbox`, because sandbox container image is runtime detail, CRI doesn't care about it.

pkg/server/container_create.go
```diff
@@ -79,12 +77,11 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to resolve image %q: %v", imageRef, err)
-	}
-	if image == nil {
```
Random-Liu: For `localResolve`, we still need this.

pkg/server/container_create.go
```diff
@@ -134,33 +119,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
-		// TODO(random-liu): Checkpoint metadata into container labels.
-		Image:   image.ID,
-		Runtime: containers.RuntimeInfo{Name: defaultRuntime},
-		Spec: &prototypes.Any{
-			TypeUrl: runtimespec.Version,
-			Value:   rawSpec,
-		},
-		RootFS: id,
-	}); err != nil {
+	opts = append(opts, []containerd.NewContainerOpts{containerd.WithSpec(spec), containerd.WithRuntime(defaultRuntime)}...)
```
Random-Liu: Just:
```go
opts = append(opts,
  containerd.WithSpec(spec),
  containerd.WithRuntime(defaultRuntime))
```

pkg/server/container_create.go
```diff
@@ -134,33 +119,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
 		}
 	}()
 
-	container, err := containerstore.NewContainer(meta, containerstore.Status{CreatedAt: time.Now().UnixNano()})
+	container, err := containerstore.NewContainer(meta, containerstore.Status{CreatedAt: time.Now().UnixNano()}, newContainer)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create internal container object for %q: %v",
 			id, err)
 	}
+
```
Random-Liu: Why add a new line before defer? I feel like we should group defer with the resource we want it to release?
Or is there any other convention?

pkg/server/container_create.go
```diff
@@ -134,33 +119,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
 		if retErr != nil {
-			if err := c.containerService.Delete(ctx, id); err != nil {
+			if err := newContainer.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
 				glog.Errorf("Failed to delete containerd container %q: %v", id, err)
 			}
 		}
 	}()
 
-	container, err := containerstore.NewContainer(meta, containerstore.Status{CreatedAt: time.Now().UnixNano()})
+	container, err := containerstore.NewContainer(meta, containerstore.Status{CreatedAt: time.Now().UnixNano()}, newContainer)
```
Random-Liu: nit: The line is too long, add a new line?

pkg/server/container_create.go
```diff
@@ -134,33 +119,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
-		Spec: &prototypes.Any{
-			TypeUrl: runtimespec.Version,
-			Value:   rawSpec,
-		},
-		RootFS: id,
-	}); err != nil {
+	opts = append(opts, []containerd.NewContainerOpts{containerd.WithSpec(spec), containerd.WithRuntime(defaultRuntime)}...)
+
+	var newContainer containerd.Container
+	if newContainer, err = c.client.NewContainer(ctx, id, opts...); err != nil {
```
Random-Liu: nit:
```go
newContainer, err := c.client.XXXXXXXX
if err != nil {
  return nil, fmt.Errorf("failed to create containerd container: %v", err)
}
```
`newContainer`/`container` seems a bit confusing...Let's use `cntr` or `cContainer` or something else here?

pkg/server/container_start.go
```diff
@@ -66,16 +64,16 @@ func (c *criContainerdService) StartContainer(ctx context.Context, r *runtime.St
 
 // startContainer actually starts the container. The function needs to be run in one transaction. Any updates
 // to the status passed in will be applied no matter the function returns error or not.
-func (c *criContainerdService) startContainer(ctx context.Context, id string, meta containerstore.Metadata, status *containerstore.Status) (retErr error) {
+func (c *criContainerdService) startContainer(ctx context.Context, container containerd.Container, meta containerstore.Metadata, status *containerstore.Status) (retErr error) {
```
Random-Liu: nit: The line is too long, add a new line.

pkg/server/container_start.go
```diff
@@ -147,52 +147,33 @@ func (c *criContainerdService) startContainer(ctx context.Context, id string, me
...
 		return fmt.Errorf("failed to create containerd task: %v", err)
 	}
+
 	defer func() {
 		if retErr != nil {
-			// Cleanup the containerd task if an error is returned.
-			if _, err := c.taskService.Delete(ctx, &tasks.DeleteTaskRequest{ContainerID: id}); err != nil {
-				glog.Errorf("Failed to delete containerd task %q: %v", id, err)
+			if _, err := task.Delete(ctx); err != nil {
+				glog.Errorf("Failed to delete container %q: %v", container.ID(), err)
```
Random-Liu: original error message seems better.

pkg/server/container_start.go
```diff
@@ -147,52 +147,33 @@ func (c *criContainerdService) startContainer(ctx context.Context, id string, me
...
 	if err != nil {
 		return fmt.Errorf("failed to create containerd task: %v", err)
 	}
+
 	defer func() {
 		if retErr != nil {
-			// Cleanup the containerd task if an error is returned.
-			if _, err := c.taskService.Delete(ctx, &tasks.DeleteTaskRequest{ContainerID: id}); err != nil {
-				glog.Errorf("Failed to delete containerd task %q: %v", id, err)
+			if _, err := task.Delete(ctx); err != nil {
```
Random-Liu: Should add `WithProcessKill`.
My original code was not correct.

pkg/server/container_stop.go
```diff
@@ -94,11 +93,16 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
...
+
+		task, err := container.Container.Task(ctx, nil)
+		if err != nil {
+			if !isContainerdGRPCNotFoundError(err) {
+				return fmt.Errorf("failed to stop container, task not found for container %q: %v", id, err)
+			}
+		}
+
+		err = task.Kill(ctx, stopSignal)
+
```
Random-Liu: ? unnecessary empty line? There are several empty lines, seem like something wrong with your editor?

```go
if err := task.Kill(ctx, stopSignal); err != nil {
}
```

pkg/server/container_stop.go
```diff
@@ -113,13 +117,16 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
 		glog.Errorf("Stop container %q timed out: %v", id, err)
 	}
 
+	task, err := container.Container.Task(ctx, nil)
```
Random-Liu: Why does this again?

pkg/server/container_stop.go
```diff
@@ -113,13 +117,16 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
...
+		}
+	}
 	// Event handler will Delete the container from containerd after it handles the Exited event.
 	glog.V(2).Infof("Kill container %q", id)
-	_, err := c.taskService.Kill(ctx, &tasks.KillRequest{
-		ContainerID: id,
-		Signal:      uint32(unix.SIGKILL),
-		All:         true,
-	})
+	err = task.Kill(ctx, unix.SIGKILL)
```
Random-Liu:
```go
if err := task.Kill(ctx, unix.SIGKILL); err != nil {
}
```

pkg/store/sandbox/sandbox.go
abhinandanpb: I use goimports to format. May I will skip using that.

pkg/server/sandbox_run.go
```diff
@@ -179,32 +153,29 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+	fifo := &containerd.FIFOSet{
+		Dir: sandboxRootDir,
+		Out: stdout,
+		Err: stderr,
 	}
-	// Create sandbox task in containerd.
-	glog.V(5).Infof("Create sandbox container (id=%q, name=%q) with options %+v.",
-		id, name, createOpts)
-	createResp, err := c.taskService.Create(ctx, createOpts)
+	task, err := container.NewTask(ctx, containerd.NewIOWithFifoSet(os.Stdin, os.Stdout, os.Stderr, fifo, false))
```
abhinandanpb: This is still possible. The only reason I avoided using NewIO as is -> we need to pass the customer path. The tests were failing because CRI validation checks for logs in /tmp/Podxxxx where containerd by default (NewIO) logs into /tmp/containerd/XXXX.

pkg/server/sandbox_remove.go
```diff
@@ -96,7 +85,7 @@ func (c *criContainerdService) RemovePodSandbox(ctx context.Context, r *runtime.
 	}
 
 	// Delete sandbox container.
-	if err := c.containerService.Delete(ctx, id); err != nil {
+	if err := sandbox.Container.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
 		if !isContainerdGRPCNotFoundError(err) {
```
abhinandanpb: Yes I realised that based on your comments in containerd PR

pkg/server/sandbox_remove.go
```diff
@@ -18,14 +18,11 @@ package server
 
 import (
```
abhinandanpb: nope , goimports

pkg/server/sandbox_run.go
Random-Liu: @abhinandanpb That's not how it works.

Containerd create FIFOs for each container. Currently, we create those fifos and give the path to containerd. In fact, we could use containerd client directly, which creates fifos for us in `/tmp/containerd/XXX`. We don't really care where the FIFOs are.

We create a **Logger** for each container, which read log from the FIFO, decorate the log, and write to the CRI log file, which is `/var/log/pods/XXX`. The test is also checking this path, the test doesn't care where the FIFOs are, either.

pkg/server/container_create.go
abhinandanpb: Got it. Thanks for explaining

pkg/server/container_stop.go
```diff
@@ -94,11 +93,16 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
...
-			ContainerID: id,
-			Signal:      uint32(stopSignal),
-			All:         true,
-		})
+
+		task, err := container.Container.Task(ctx, nil)
+		if err != nil {
+			if !isContainerdGRPCNotFoundError(err) {
+				return fmt.Errorf("failed to stop container, task not found for container %q: %v", id, err)
+			}
```
Random-Liu: `task` will be nil if there is an error, we may want to skip the `task.Kill` when there is an error, even it's not found error.

Random-Liu: And we could get task outside of the if clause, so that we don't need to do it twice.

pkg/server/container_stop.go
```diff
@@ -113,13 +117,16 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
 		glog.Errorf("Stop container %q timed out: %v", id, err)
 	}
 
+	task, err := container.Container.Task(ctx, nil)
+	if err != nil {
+		if !isContainerdGRPCNotFoundError(err) {
```
Random-Liu: Skip `task.Kill` if not found.

pkg/server/container_stop_test.go
```diff
@@ -70,7 +70,7 @@ func TestWaitContainerStop(t *testing.T) {
 		if test.status != nil {
 			container, err := containerstore.NewContainer(
 				containerstore.Metadata{ID: id},
-				*test.status,
+				*test.status, nil,
```
Random-Liu: New line.

pkg/server/helpers.go
```diff
@@ -272,18 +273,9 @@ func normalizeImageRef(ref string) (reference.Named, error) {
...
-	if err != nil {
-		return "", 0, nil, fmt.Errorf("failed to normalize image reference %q: %v", ref, err)
-	}
-	normalizedRef := normalized.String()
-	image, err := c.imageStoreService.Get(ctx, normalizedRef)
-	if err != nil {
-		return "", 0, nil, fmt.Errorf("failed to get image %q from containerd image store: %v",
-			normalizedRef, err)
-	}
+
```
Random-Liu: We should add following functions into containerd client.
`Image.Config`, `Image.RootFS`, `Image.Size` etc.

pkg/server/image_pull.go
```diff
@@ -182,216 +166,3 @@ func ParseAuth(auth *runtime.AuthConfig) (string, string, error) {
...
-
-	// In the future, containerd will rely on the information in the image store to perform image
-	// garbage collection.
-	// For now, we simply use it to store and retrieve information required for pulling an image.
-	// @stevvooe said we should `Put` before downloading content, However:
-	// 1) Containerd client put image metadata after downloading;
-	// 2) We need desc returned by schema1 converter.
-	// So just put the image metadata after downloading now.
-	// TODO(random-liu): Fix the potential garbage collection race.
-	repoDigest, repoTag := getRepoDigestAndTag(namedRef, desc.Digest, schema1Converter != nil)
```
Random-Liu: We still need to write `repoDigest`, `repoTag` and `imageID` into containerd image metadata store.

pkg/server/container_create.go
abhinandanpb: i usally let gofmt/goimports take care of formatting and indentation

Random-Liu: @abhinandanpb Weird. In my case, `gofmt` never add unnecessary new lines for me.

pkg/server/image_pull.go
abhinandanpb: Can you elaborate on this .? This has to be written on the containerd side ?

Random-Liu: Currently, our internal image store is maintaining the Image ID -> RepoDigests/RepoTags mapping, which is for the requirement of CRI.

However, when user create a container or pulling a image, they usually give us RepoDigest/RepoTag. So we have to do the RepoDigest/RepoTag -> Image ID (and other Information) conversion, which we are currently leveraging containerd image metadata store to do that. That's why we should put all `RepoDigest/RepoTag` into the metadata store.

As for why we also need to put image ID - I was told that containerd will do image gc based on reference, an entry in image metadata store is a reference to the corresponding snapshot and content. Other references are not unique enough to identify an image, e.g. repo tag could be reused by another image; schema 1 image doesn't have a repo digest etc.; image id seems to be the only thing we can use the reference the image and keep it from being garbage collected.
I think it's because other references are not unique to identify

pkg/server/image_pull.go
```diff
@@ -88,70 +81,66 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
-
 	// Get image information.
 	chainID, size, config, err := c.getImageInfo(ctx, imageRef)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get image %q information: %v", imageRef, err)
 	}
-	image := imagestore.Image{
-		ID:      imageID,
+	img := imagestore.Image{
+		ID:      image.Target().Digest.String(),
```
Random-Liu: Same here, this is not `ImageID`.
See https://github.com/opencontainers/image-spec/blob/d9ffa2f05d4dbe917edfdffcafaf427287a26d1c/config.md#imageid.

pkg/server/sandbox_list.go
```diff
@@ -19,15 +19,12 @@ package server
 import (
 	"fmt"
 
+	tasks "github.com/containerd/containerd/api/services/tasks/v1"
+	"github.com/containerd/containerd/api/types/task"
 	"github.com/golang/glog"
+	sandboxstore "github.com/kubernetes-incubator/cri-containerd/pkg/store/sandbox"
```
Random-Liu: nit: Move cri-containerd dependency into a separate group.

pkg/server/sandbox_list.go
```diff
@@ -42,16 +39,15 @@ func (c *criContainerdService) ListPodSandbox(ctx context.Context, r *runtime.Li
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to list sandbox containers: %v", err)
 	}
-	sandboxesInContainerd := resp.Tasks
 
 	var sandboxes []*runtime.PodSandbox
 	for _, sandboxInStore := range sandboxesInStore {
-		var sandboxInContainerd *task.Task
-		for _, s := range sandboxesInContainerd {
+		var sandboxInContainerd *task.Process
```
Random-Liu: Why change to `task.Process`?

Random-Liu: `Status()` underlying?

pkg/server/events.go
```diff
@@ -104,13 +105,21 @@ func (c *criContainerdService) handleEvent(evt *events.Envelope) {
...
-			return
+		task, err := cntr.Container.Task(context.Background(), nil)
+		if err != nil {
+			if !errdefs.IsNotFound(err) {
+				glog.Errorf("failed to stop container, task not found for container %q: %v", e.ContainerID, err)
+				return
+			}
+		}
+		if task != nil {
+			if _, err = task.Delete(context.Background(), containerd.WithProcessKill); err != nil {
```
Random-Liu: Don't think we want to `KillProcess` here.
If the process is still running, it's a bug.

pkg/server/sandbox_run.go
```diff
@@ -78,61 +74,35 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
-			TypeUrl: runtimespec.Version,
-			Value:   rawSpec,
-		},
-		RootFS: id,
-	}); err != nil {
+	// TODO(random-liu): Checkpoint metadata into container labels.
+
+	opts := []containerd.NewContainerOpts{
+		containerd.WithSpec(spec),
+		containerd.WithImage(image.Image),
```
Random-Liu: Not needed. Image is specified in `WithNewSnapshotXXX` already.

pkg/server/sandbox_run.go
```diff
@@ -180,32 +147,24 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
-		Stdout: stdout,
-		Stderr: stderr,
-	}
 	// Create sandbox task in containerd.
-	glog.V(5).Infof("Create sandbox container (id=%q, name=%q) with options %+v.",
-		id, name, createOpts)
-	createResp, err := c.taskService.Create(ctx, createOpts)
+	glog.V(5).Infof("Create sandbox container (id=%q, name=%q).",
+		id, name)
+	task, err := container.NewTask(ctx, containerd.NewIO(os.Stdin, wStdoutPipe, wStderrPipe))
```
Random-Liu: Add TODO to disable stdin.
s/os.Stdin/new(bytes.Buffer)

pkg/server/sandbox_stop.go
```diff
@@ -83,69 +80,33 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
-			continue
-		}
-		any, err := typeurl.UnmarshalAny(evt.Event)
-		if err != nil {
-			return err
-		}
-		e := any.(*events.TaskExit)
-		if e.ContainerID == id && e.Pid == pid {
-			return nil
+	if taskStatus.Status != containerd.Stopped {
```
Random-Liu: Since we have `WithProcessKill` now, just `Delete`, we don't need check the status anymore.

In your current code, we'll never delete the task once the task is stopped.

pkg/server/sandbox_stop.go
```diff
@@ -83,69 +80,33 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
-			return fmt.Errorf("failed to kill sandbox container: %v", err)
-		}
 
-		if err := c.waitSandboxContainer(eventstream, id, resp.Task.Pid); err != nil {
-			return fmt.Errorf("failed to wait for pod sandbox to stop: %v", err)
-		}
-	}
+	taskStatus, err := task.Status(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get task status for sandbox %q : %v", container.ID(), err)
```
Random-Liu: Remove space before `:`.

Actually, now I prefer not to include id inside error message, unless the caller doesn't know that. Currently, error message is a bit messy. :p We need to gradually clean them up.

pkg/server/sandbox_run.go
```diff
@@ -180,32 +147,24 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to create sandbox container %q: %v",
-			id, err)
+		return nil, fmt.Errorf("failed to create task for sandbox %q: %v", id, err)
 	}
 	defer func() {
 		if retErr != nil {
 			// Cleanup the sandbox container if an error is returned.
-			if err := c.stopSandboxContainer(ctx, id); err != nil {
+			if err := c.stopSandboxContainer(ctx, container); err != nil {
```
Random-Liu: Since you've got task here, just call `task.Delete(ctx, WithProcessKill)`.

pkg/server/sandbox_stop.go
```diff
@@ -83,69 +80,33 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
 		return nil, fmt.Errorf("failed to stop sandbox container %q: %v", id, err)
 	}
 	return &runtime.StopPodSandboxResponse{}, nil
 }
 
 // stopSandboxContainer kills and deletes sandbox container.
-func (c *criContainerdService) stopSandboxContainer(ctx context.Context, id string) error {
-	cancellable, cancel := context.WithCancel(ctx)
-	eventstream, err := c.eventService.Subscribe(cancellable, &events.SubscribeRequest{})
+func (c *criContainerdService) stopSandboxContainer(ctx context.Context, container containerd.Container) error {
```
Random-Liu: This function is not that useful now, maybe just embed into `StopContainer`.

pkg/server/sandbox_status.go
```diff
@@ -47,18 +46,24 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
...
 	// Set sandbox state to NOTREADY by default.
 	state := runtime.PodSandboxState_SANDBOX_NOTREADY
 	// If the sandbox container is running, treat it as READY.
-	if info != nil && info.Task.Status == task.StatusRunning {
-		state = runtime.PodSandboxState_SANDBOX_READY
-	}
+	if task != nil {
+		taskStatus, err := task.Status(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get task status for sandbox container %q : %v", id, err)
```
Random-Liu: nit: space before `:`.

pkg/server/image_pull.go
```diff
@@ -88,70 +81,66 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
 	// TODO(mikebrow): add truncIndex for image id
-	imageID, repoTag, repoDigest, err := c.pullImage(ctx, imageRef, r.GetAuth())
+	image, err := c.client.Pull(ctx, ref, containerd.WithPullUnpack, containerd.WithSchema1Conversion, containerd.WithResolver(resolver))
 	if err != nil {
-		return nil, fmt.Errorf("failed to pull image %q: %v", imageRef, err)
+		return nil, fmt.Errorf("failed to pull image %q: %v", ref, err)
 	}
-	glog.V(4).Infof("Pulled image %q with image id %q, repo tag %q, repo digest %q", imageRef, imageID,
+	repoDigest, repoTag := getRepoDigestAndTag(namedRef, image.Target().Digest, image.Target().MediaType == containerdimages.MediaTypeDockerSchema1Manifest)
+	for _, r := range []string{repoTag, repoDigest} {
```
Random-Liu: We also need to put image ID. :)

pkg/server/image_pull.go
```diff
@@ -88,70 +81,66 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
+		}
+		if err := c.createImageReference(ctx, r, image.Target()); err != nil {
+			return nil, fmt.Errorf("failed to update image reference %q: %v", r, err)
+		}
+	}
+	glog.V(4).Infof("Pulled image %q with image id %q, repo tag %q, repo digest %q", imageRef, image.Name(),
 		repoTag, repoDigest)
-
 	// Get image information.
 	chainID, size, config, err := c.getImageInfo(ctx, imageRef)
```
Random-Liu: `getImageInfo` may not be needed, if `Image.Size()`, `Image.RootFS()` could provide enough information.

pkg/server/image_pull.go
abhinandanpb: Thats already taken care of by the client ? https://github.com/containerd/containerd/blob/master/client.go#L252

Random-Liu: @abhinandanpb The client only puts the image reference in.

There are several different image references:
* Image ID: Image config digest.
* Image Pullable Digest: Image manifest digest.
* Image ChainID: Calculated from image snapshot.
* Image Reference: Readable image name, e.g docker.io/library/busybox:latest, docker.io/library/nginx:latest etc.

Random-Liu: This issue https://github.com/kubernetes/kubernetes/issues/46255, and the OCI image spec document may be useful here. :)

pkg/server/sandbox_list.go
abhinandanpb: yep thats changed recently.

pkg/server/events.go
```diff
@@ -104,13 +105,21 @@ func (c *criContainerdService) handleEvent(evt *events.Envelope) {
...
-			// TODO(random-liu): [P0] Enqueue the event and retry.
-			glog.Errorf("Failed to delete container %q: %v", e.ContainerID, err)
-			return
+		task, err := cntr.Container.Task(context.Background(), nil)
+		if err != nil {
+			if !errdefs.IsNotFound(err) {
+				glog.Errorf("failed to stop container, task not found for container %q: %v", e.ContainerID, err)
+				return
+			}
+		}
```
abhinandanpb: else? --> we just update the status ?

Random-Liu: I mean:
```go
if err != nil {
  if !errdefs.IsNotFound(err) {
    glog.Errorf("failed to stop container, task not found for container %q: %v", e.ContainerID, err)
    return
  }
} else {
  if _, err = task.Delete(context.Background(), containerd.WithProcessKill); err != nil {
  // ....
  }
}
```
Just a nit.

pkg/server/sandbox_list.go
Random-Liu: Hm, you are write, that's changed to `task.Process`.

And you are correct, I mistook this with client process. In the api, `Status` is actually just a field. My bad. :p

pkg/server/image_pull.go
abhinandanpb: Got it . Thanks for explaining. Is there anyway we could avoid storing these many-to-one reference copies in the store ? I mean any change we can do on the container to resolve the image ? Not in the scope of this PR. But moving forward.

pkg/server/container_start.go
```diff
@@ -97,102 +99,65 @@ func (c *criContainerdService) startContainer(ctx context.Context, id string, me
...
 	if err != nil {
 		return fmt.Errorf("failed to get sandbox container %q info: %v", sandboxID, err)
 	}
 	// This is only a best effort check, sandbox may still exit after this. If sandbox fails
 	// before starting the container, the start will fail.
-	if sandboxInfo.Task.Status != task.StatusRunning {
-		return fmt.Errorf("sandbox container %q is not running", sandboxID)
+	taskStatus, err := s.Status(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get task status for sandbox container %q : %v", id, err)
```
Random-Liu: nit: remove space before `:`.

pkg/server/image_pull.go
```diff
@@ -88,70 +81,70 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
+	repoDigest, repoTag := getRepoDigestAndTag(namedRef, image.Target().Digest, image.Target().MediaType == containerdimages.MediaTypeDockerSchema1Manifest)
+	for _, r := range []string{repoTag, repoDigest} {
+		if r == "" {
+			continue
+		}
+		if err := c.createImageReference(ctx, r, image.Target()); err != nil {
+			return nil, fmt.Errorf("failed to update image reference %q: %v", r, err)
+		}
+	}
+	glog.V(4).Infof("Pulled image %q with image id %q, repo tag %q, repo digest %q", imageRef, image.Name(),
```
Random-Liu: Log this after you get the image id, so that you could log the image id out.

pkg/server/helpers.go
```diff
@@ -272,43 +244,45 @@ func normalizeImageRef(ref string) (reference.Named, error) {
 // getImageInfo returns image chainID, compressed size and oci config. Note that getImageInfo
```
Random-Liu: and image id

pkg/server/helpers.go
```diff
@@ -272,43 +244,45 @@ func normalizeImageRef(ref string) (reference.Named, error) {
...
 // assumes that the image has been pulled or it will return an error.
 func (c *criContainerdService) getImageInfo(ctx context.Context, ref string) (
```
Random-Liu: nit: Return image id as `imagedigest.Digest`.

pkg/server/image_pull.go
```diff
@@ -88,70 +81,70 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
 	// Get image information.
-	chainID, size, config, err := c.getImageInfo(ctx, imageRef)
+	chainID, size, config, desc, err := c.getImageInfo(ctx, imageRef)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get image %q information: %v", imageRef, err)
 	}
-	image := imagestore.Image{
+	imageID := desc.Digest.String()
+	if err := c.createImageReference(ctx, imageID, image.Target()); err != nil {
+		return nil, fmt.Errorf("failed to update image reference %q: %v", r, err)
```
Random-Liu: Why `r` here?
Actually, we could get image id before updating repoDigest, repoTag, so that we could update them all together in a single loop.

abhinandanpb:
@Random-Liu updated PR. Thanks for reviewing the changes multiple times :)

Random-Liu:
@abhinandanpb LGTM, let's wait for the test result.

pkg/server/helpers.go
```diff
@@ -247,47 +241,48 @@ func normalizeImageRef(ref string) (reference.Named, error) {
 	return reference.TagNameOnly(named), nil
 }
 
-// getImageInfo returns image chainID, compressed size and oci config. Note that getImageInfo
+// getImageInfo returns image chainID, compressed size, imageID and oci config. Note that getImageInfo
```
Random-Liu: oci config and imageID.

To make it the same with the return order.

pkg/server/image_pull.go
```diff
@@ -112,15 +112,19 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
-	glog.V(4).Infof("Pulled image %q with image id %q, repo tag %q, repo digest %q", imageRef, image.Name(),
-		repoTag, repoDigest)
 	// Get image information.
-	chainID, size, config, err := c.getImageInfo(ctx, imageRef)
+	chainID, size, config, id, err := c.getImageInfo(ctx, imageRef)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get image %q information: %v", imageRef, err)
 	}
+	imageID := id.String()
+	glog.V(4).Infof("Pulled image %q with image id %q, repo tag %q, repo digest %q", imageRef, imageID,
```
Random-Liu: Move this down after the following if clause.

abhinandanpb:
Done.

Random-Liu:
LGTM
Will merge after test passes.
@abhinandanpb Thanks a lot! :)

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/86.
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/49.

PR #114, opened 2017-08-07
Allow setting of compiler options

ijc:
To allow e.g. building a static binary with:

    make binaries GO_GCFLAGS="--ldflags '-extldflags \"-fno-PIC -static\"'"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,10 +70,11 @@ boiler:
 cri-containerd: check-gopath
 	$(GO) build -o $(BUILD_DIR)/$@ \
 	   $(BUILD_TAGS) \
+	   $(GO_LDFLAGS) $(GO_GCFLAGS) \
 	   $(PROJECT)/cmd/cri-containerd
 
 test:
-	go test -timeout=10m -race ./pkg/... $(BUILD_TAGS)
+	go test -timeout=10m -race ./pkg/... $(BUILD_TAGS) $(GO_LDFLAGS) $(GO_GCFLAGS)
 
 test-cri:
 	@./hack/test-cri.sh
--- a/Makefile
+++ b/Makefile
@@ -32,16 +32,17 @@ default: help
 help:
 	@echo "Usage: make <target>"
 	@echo
-	@echo " * 'install'       - Install binaries to system locations"
-	@echo " * 'binaries'      - Build cri-containerd"
-	@echo " * 'test'          - Test cri-containerd"
-	@echo " * 'test-cri'      - Test cri-containerd with cri validation test"
-	@echo " * 'clean'         - Clean artifacts"
-	@echo " * 'verify'        - Execute the source code verification tools"
-	@echo " * 'install.tools' - Install tools used by verify"
-	@echo " * 'install.deps'  - Install dependencies of cri-containerd (containerd, runc, cni)"
-	@echo " * 'uninstall'     - Remove installed binaries from system locations"
-	@echo " * 'version'       - Print current cri-containerd release version"
+	@echo " * 'install'        - Install binaries to system locations"
+	@echo " * 'binaries'       - Build cri-containerd"
+	@echo " * 'static-binaries - Build static cri-containerd"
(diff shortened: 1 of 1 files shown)


ijc:
I believe I have now signed the CLA. This comment should retrigger the check.

ijc:
@mikebrow Happy to do so. Getting the quoting right for the echo to print exactly the right thing is pretty tricky. How would you feel about just adding a `static-binaries` target which just recurses to `binaries` with the right flags?

Self-nit: The example should really be using `GO_LDFLAGS` not `GO_CFLAGS` I think.

mikebrow:
Dunno.. I sort of prefer GO_GCFLAGS over GO_CFLAGS... and containerd/containerd uses GO_GCFLAGS. Presume the goal here is a static of all 3, runc, containerd, and cri-containerd?

Yes, a static or static-binaries target would be nice. In that case may only need the LD flags to be used as and environment variable.  Just thinking out loud.

ijc:
> Presume the goal here is a static of all 3, `runc`, `containerd`, and `cri-containerd`?

Correct. The context is [LinuxKit](https://github.com/linuxkit/linuxkit) and in particular the [projects/kubernetes](https://github.com/linuxkit/linuxkit/tree/master/projects/kubernetes) in there which I'd like be using `cri-containerd` (at least as an option).

FWIW LinuxKit's containerd build is using:
```
make binaries EXTRA_FLAGS="-buildmode pie" EXTRA_LDFLAGS="-extldflags \\\"-fno-PIC -static\\\""
```
Where `EXTRA_FLAGS` is passed to `go build` directly and `EXTRA_LDFLAGS` is appended to `GO_LDFLAGS`. It's quite the proliferation of options!

I'll go ahead and add a `static-binaries` target, but if you prefer I could do a slightly larger change modelled on containerd's use of all those flag names.

Makefile
```diff
@@ -83,6 +85,9 @@ clean:
 
 binaries: cri-containerd
 
+static-binaries: GO_LDFLAGS=--ldflags '-extldflags "-fno-PIC -static"'
+static-binaries: cri-containerd
```
Random-Liu: Add to `.PHONY`?

Random-Liu:
> Correct. The context is LinuxKit and in particular the projects/kubernetes in there which I'd like be using cri-containerd (at least as an option).

@ijc I'm interested in this part. LInuxKit generated OS is using containerd to manage services, right? Will kubernetes inside talk to this containerd? Just curious. :)

Makefile
```diff
@@ -83,6 +85,9 @@ clean:
 
 binaries: cri-containerd
 
+static-binaries: GO_LDFLAGS=--ldflags '-extldflags "-fno-PIC -static"'
```
mikebrow: didn't realize where you were heading with this.. can we make these LD flags more specific and leave the environment variables to other uses?  say make your flags GO_STATIC_LDFLAGS?

Random-Liu: We probably want to use `+=` here.

Makefile
ijc: Ack, will do.

Makefile
ijc: Shall I copy the `EXTRA_LDFLAGS` and `EXTRA_FLAGS` pattern from containerd perhaps?

ijc:
> @ijc I'm interested in this part. LInuxKit generated OS is using containerd to manage services, right? Will kubernetes inside talk to this containerd? Just curious. :)

Yes, that's the plan, cri-containerd is running in a system container but with access to the system containerd control socket.

I have it working locally just need to clean up the patch and submit it to linuxkit. Currently I'm preempted with something else (completely unrelated) but I hope to get back to it in the next couple of days and post something. This PR isn't a blocker for my sending an initial patch (but it would be a nice to have of course!).

Random-Liu:
> Yes, that's the plan, cri-containerd is running in a system container but with access to the system containerd control socket.

@ijc Thanks! Good to know.

Random-Liu:
LGTM

PR #115, opened 2017-08-07
Add container Exec support.

Random-Liu:
This PR:
1) Vendored Kubernetes streaming server inside cri-containerd, which will be used for all container streaming operations: Exec, Portforward, Attach.
2) Added container `Exec` support.
3) Enabled `Exec` CRI validation test.


pkg/server/container_execsync.go
```diff
@@ -100,16 +149,11 @@ func (c *criContainerdService) ExecSync(ctx context.Context, r *runtime.ExecSync
 	if err != nil {
 		return nil, fmt.Errorf("failed to wait for exec in container %q to finish: %v", id, err)
```
mikebrow: always throw out the exit code if there is an error? unknownExitCode?

pkg/server/service.go
```diff
@@ -84,6 +86,8 @@ type criContainerdService struct {
 	// imageStoreService is the containerd service to store and track
 	// image metadata.
 	imageStoreService images.Store
+	// eventsService is the containerd task service client
```
mikebrow: just curious why this one is better up here?

pkg/server/service.go
```diff
@@ -136,9 +142,20 @@ func NewCRIContainerdService(containerdEndpoint, rootDir, networkPluginBinDir, n
...
+	}
+
 	return c, nil
 }
 
 func (c *criContainerdService) Start() {
 	c.startEventMonitor()
+	go func() {
+		if err := c.streamServer.Start(true); err != nil {
+			glog.Errorf("Failed to start streaing server: %v", err)
```
mikebrow: /s/streaing/streaming/

pkg/server/container_execsync.go
Random-Liu: Maybe we should let `waitContainerExec` also return pointer, so that we don't need the `unknownExitStatus` for now.

We don't use the exit code when there is an error now.

pkg/server/service.go
Random-Liu: Just group containerd services together. No particular reason. :)

pkg/server/service.go
Random-Liu: Will do.

pkg/server/container_execsync.go
mikebrow: right.. just trying to figure out if that's on purpose. Seems at least possible to have an exit code and an error. Not a fan of throwing it out unless we need to.

Random-Liu: Done.

pkg/server/service.go
Random-Liu: Done.

Random-Liu:
@mikebrow Addressed comments.

PR #116, opened 2017-08-07
Handle device symlink.

Random-Liu:
I found that cri-containerd prints out a lot of following logs when running privileged container test:
```
WARNING: The same type, major and minor should not be used for multiple devices.
```

It turns out that device list returned by `HostDevices` also contains all the device symlinks with 0 major & minor.

This PR:
1) Skip these 0 major minor devices for privileged container;
2) Follow symlink for non-privileged container, which is [the same with docker](https://github.com/moby/moby/blob/7238cca42c3d024adfa030306ad3e3ec4232baed/oci/devices_linux.go#L45).

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -341,6 +341,10 @@ func addOCIDevices(g *generate.Generator, devs []*runtime.Device, privileged boo
 				UID:   &hostDevice.Uid,
 				GID:   &hostDevice.Gid,
 			}
+			if hostDevice.Major == 0 && hostDevice.Minor == 0 {
+				// Invalid device, most likely a symbolic link, skip it.
+				continue
+			}
 			g.AddDevice(rd)
 		}
 		spec.Linux.Resources.Devices = []runtimespec.LinuxDeviceCgroup{
@@ -352,7 +356,11 @@ func addOCIDevices(g *generate.Generator, devs []*runtime.Device, privileged boo
 		return nil
 	}
 	for _, device := range devs {
-		dev, err := devices.DeviceFromPath(device.HostPath, device.Permissions)
+		path, err := resolveSymbolicLink(device.HostPath)
+		if err != nil {
+			return err
+		}
+		dev, err := devices.DeviceFromPath(path, device.Permissions)
 		if err != nil {
 			return err
 		}
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


pkg/server/helpers.go
```diff
@@ -404,3 +405,20 @@ func (c *criContainerdService) ensureImageExists(ctx context.Context, ref string
...
+// path and true; if the path is not a symlink, returns original path and false.
+func resolveSymbolicLink(path string) (string, bool, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		return "", false, err
+	}
+	if info.Mode()&os.ModeSymlink != os.ModeSymlink {
+		return path, false, nil
+	}
+	resolved, err := filepath.EvalSymlinks(path)
```
mikebrow: might result in an absolute path.. do these have to be/stay relative?

pkg/server/helpers.go
```diff
@@ -404,3 +405,20 @@ func (c *criContainerdService) ensureImageExists(ctx context.Context, ref string
 	}
 	return &newImage, nil
 }
+
+// resolveSymbolicLink resolves a possbile symlink path. If the path is a symlink, returns resolved
+// path and true; if the path is not a symlink, returns original path and false.
```
mikebrow: you may have had a thought or purpose for the bool, but not used it?

pkg/server/container_create.go
```diff
@@ -341,6 +341,10 @@ func addOCIDevices(g *generate.Generator, devs []*runtime.Device, privileged boo
 				UID:   &hostDevice.Uid,
 				GID:   &hostDevice.Gid,
 			}
+			if hostDevice.Major == 0 && hostDevice.Minor == 0 {
```
Random-Liu: Good idea.

pkg/server/helpers.go
Random-Liu: Nope. I will mention in the comment that the returned path is absolute.

Random-Liu:
```
If path is relative the result will be relative to the current directory, unless one of the components is an absolute symbolic link. EvalSymlinks calls Clean on the result.
```
https://golang.org/pkg/path/filepath/#EvalSymlinks

It seems that the function keeps relative/absolute.

pkg/server/helpers.go
Random-Liu: Thought it will be useful in the future. But yeah, remove it for now.

pkg/server/container_create.go
Random-Liu: On second thought, given that:
1) Major 0 is never used`http://www.fsl.cs.sunysb.edu/kernel-api/re941.html`
2) I'd like to make the spec generation reusable library.

I'd prefer not to add log in this function. :)

pkg/server/helpers.go
Random-Liu: Done.

Random-Liu:
@mikebrow Addressed comments.

pkg/server/helpers.go
mikebrow: I read it as keeps relative unless the symlink resolves to an absolute.. thus my question? I wasn't sure if it needed to stay relative. One would hope it doesn't matter.

PR #126, opened 2017-08-10
Replace the original default spec with containerd default spec

miaoyq:
The original default spec contain `seccomp` configuration, but some OS do not support this feature, such as ubuntu14.04, and `make test-cri` always fail. The containerd default spec dosen't contain `seccomp`, so I think we could replace the default spec with containerd default spec.

The difference between original default spec and containerd default spec:

| original default spec | containerd default spec |
|---------------------------|-------------------------------|
| - | mounts: /run |
| seccomp | - |
| process.capabilities.ambient | - |
| env: TERM=xterm | - |

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/containers"
 	prototypes "github.com/gogo/protobuf/types"
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


Random-Liu:
@miaoyq Thanks a lot for the comparison! Nice table!

I'm fine with changing to containerd default, and I don't think the default should contain seccomp config, either.
Only one question I have is that why does containerd create `/run`, that doesn't seem to be the behavior of docker. Is there any particular reason?

Random-Liu:
I asked on slack channel, let's wait for the response. :)

Random-Liu:
I'm fine with this PR, and we'll be able to get rid of the `/run` mount after https://github.com/containerd/containerd/pull/1333 is merged.

Random-Liu:
LGTM

PR #127, opened 2017-08-10
Use containerd spec generator to set tty.

Random-Liu:
In https://github.com/kubernetes-incubator/cri-containerd/pull/126, we changed to use containerd default spec.

However, because containerd doesn't add `TERM=xterm` by default, it needs to be added separately when tty is enabled.

This is not done in runtime-tools, but is handled in containerd `WithTTY`, so we'll use containerd spec generator to set TTY.

/cc @miaoyq

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -202,6 +202,11 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 		g.SetProcessCwd(imageConfig.WorkingDir)
 	}
 
+	g.SetProcessTerminal(config.GetTty())
+	if config.GetTty() {
+		g.AddProcessEnv("TERM", "xterm")
+	}
+
 	// Apply envs from image config first, so that envs from container config
 	// can override them.
 	if err := addImageEnvs(&g, imageConfig.Env); err != nil {
@@ -229,8 +234,6 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 		cgroupsPath := getCgroupsPath(sandboxConfig.GetLinux().GetCgroupParent(), id)
 		g.SetLinuxCgroupsPath(cgroupsPath)
 	}
-
-	g.SetProcessTerminal(config.GetTty())
 
 	if err := setOCICapabilities(&g, securityContext.GetCapabilities(), securityContext.GetPrivileged()); err != nil {
 		return nil, fmt.Errorf("failed to set capabilities %+v: %v",
--- a/pkg/server/container_create_test.go
+++ b/pkg/server/container_create_test.go
@@ -191,6 +191,11 @@ func TestContainerSpecTty(t *testing.T) {
 		assert.NoError(t, err)
 		specCheck(t, testID, testPid, spec)
 		assert.Equal(t, tty, spec.Process.Terminal)
+		if tty {
+			assert.Contains(t, spec.Process.Env, "TERM=xterm")
+		} else {
+			assert.NotContains(t, spec.Process.Env, "TERM=xterm")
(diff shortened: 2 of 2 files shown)


miaoyq:
@Random-Liu  It's fine! 
I think the rest of the configuration items could also be set by this way, similar in containerd. In addition, we can match the different operating systems through the conditional compilation.

abhinandanpb:
LGTM. Agree with @mikebrow on using corresponding options.

pkg/server/container_create.go
```diff
@@ -181,8 +181,15 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
 func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint32, config *runtime.ContainerConfig,
 	sandboxConfig *runtime.PodSandboxConfig, imageConfig *imagespec.ImageConfig, extraMounts []*runtime.Mount) (*runtimespec.Spec, error) {
+	// We use containerd spec generator for tty because we need the TERM env.
+	// TODO(random-liu): Consider whether we should completely move to containerd spec generator.
+	var opts []containerd.SpecOpts
+	if config.GetTty() {
+		opts = append(opts, containerd.WithTTY)
+	}
+
 	// Creates a spec Generator with the default spec.
```
mikebrow: /s/with the default spec./with the default spec and above specified options./

pkg/server/container_create.go
```diff
@@ -181,8 +181,15 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 
 func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint32, config *runtime.ContainerConfig,
 	sandboxConfig *runtime.PodSandboxConfig, imageConfig *imagespec.ImageConfig, extraMounts []*runtime.Mount) (*runtimespec.Spec, error) {
+	// We use containerd spec generator for tty because we need the TERM env.
+	// TODO(random-liu): Consider whether we should completely move to containerd spec generator.
```
mikebrow: I think we should pick one spec api here and make it work. I mention the runtime tools option to set the TERM env entry below. There are pros and cons to both apis. I don't have a preference for which one we go with, but yeah doing it piece meal and switching back and forth and we're probably going to miss something :-)

pkg/server/container_create.go
```diff
@@ -230,8 +237,6 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 		g.SetLinuxCgroupsPath(cgroupsPath)
 	}
 
-	g.SetProcessTerminal(config.GetTty())
```
mikebrow: or ```g.AddProcessEnv("TERM", "xterm")```

pkg/server/container_create.go
Random-Liu: Make sense to me. Will add env with runtime-tools.

pkg/server/container_create.go
Random-Liu: Will do.

Random-Liu: Done.

pkg/server/container_create.go
Random-Liu: Done.

pkg/server/container_create.go
Random-Liu: Not needed now.

Random-Liu:
@miaoyq @abhinandanpb @mikebrow Thanks you all for reviewing! I think what @mikebrow said makes sense, we should be using one tool instead at a time to avoid confusion. Changed the PR to use runtime-tools to set `TERM=xterm`.

PR #130, opened 2017-08-14
Add portforward support.

Random-Liu:
Add port forward support. The implementation is the same with [the dockershim](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_streaming.go#L162-L217).

### DIFF
--- a/pkg/server/sandbox_portforward.go
+++ b/pkg/server/sandbox_portforward.go
@@ -17,14 +17,109 @@ limitations under the License.
 package server
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
+	"io"
+	"os/exec"
+	"strings"
 
+	"github.com/containerd/containerd"
+	"github.com/golang/glog"
 	"golang.org/x/net/context"
-
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 )
 
 // PortForward prepares a streaming endpoint to forward ports from a PodSandbox, and returns the address.
-func (c *criContainerdService) PortForward(ctx context.Context, r *runtime.PortForwardRequest) (*runtime.PortForwardResponse, error) {
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) PortForward(ctx context.Context, r *runtime.PortForwardRequest) (retRes *runtime.PortForwardResponse, retErr error) {
+	// TODO(random-liu): Run a socat container inside the sandbox to do portforward.
+	glog.V(2).Infof("Portforward for sandbox %q port %v", r.GetPodSandboxId(), r.GetPort())
+	defer func() {
+		if retErr == nil {
+			glog.V(2).Infof("Portforward for %q returns URL %q", r.GetPodSandboxId(), retRes.GetUrl())
+		}
+	}()
+
+	sandbox, err := c.sandboxStore.Get(r.GetPodSandboxId())
+	if err != nil {
+		return nil, fmt.Errorf("failed to find sandbox: %v", err)
+	}
+
+	t, err := sandbox.Container.Task(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get sandbox container: %v", err)
+	}
+	status, err := t.Status(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get sandbox container status: %v", err)
+	}
+	if status.Status != containerd.Running {
+		return nil, errors.New("sandbox container is not running")
+	}
+	// TODO(random-liu): Verify that ports are exposed.
+	return c.streamServer.GetPortForward(r)
+}
+
+// portForward requires `nsenter` and `socat` on the node, it uses `nsenter` to enter the
+// sandbox namespace, and run `socat` inside the namespace to forward stream for a specific
+// port. The `socat` command keeps running until it exits or client disconnect.
+func (c *criContainerdService) portForward(id string, port int32, stream io.ReadWriteCloser) error {
+	s, err := c.sandboxStore.Get(id)
+	if err != nil {
+		return fmt.Errorf("failed to find sandbox in store: %v", err)
+	}
+	pid := s.Pid
+
+	socat, err := exec.LookPath("socat")
+	if err != nil {
+		return fmt.Errorf("failed to find socat: %v", err)
+	}
+
+	// Check following links for meaning of the options:
+	// * socat: https://linux.die.net/man/1/socat
+	// * nsenter: http://man7.org/linux/man-pages/man1/nsenter.1.html
+	args := []string{"-t", fmt.Sprintf("%d", pid), "-n", socat,
+		"-", fmt.Sprintf("TCP4:localhost:%d", port)}
+
+	nsenter, err := exec.LookPath("nsenter")
+	if err != nil {
+		return fmt.Errorf("failed to find nsenter: %v", err)
+	}
+
+	glog.V(2).Infof("Executing port forwarding command: %s %s", nsenter, strings.Join(args, " "))
+
+	cmd := exec.Command(nsenter, args...)
+	cmd.Stdout = stream
+
+	stderr := new(bytes.Buffer)
+	cmd.Stderr = stderr
+
+	// If we use Stdin, command.Run() won't return until the goroutine that's copying
+	// from stream finishes. Unfortunately, if you have a client like telnet connected
+	// via port forwarding, as long as the user's telnet client is connected to the user's
+	// local listener that port forwarding sets up, the telnet session never exits. This
+	// means that even if socat has finished running, command.Run() won't ever return
+	// (because the client still has the connection and stream open).
+	//
+	// The work around is to use StdinPipe(), as Wait() (called by Run()) closes the pipe
+	// when the command (socat) exits.
+	in, err := cmd.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("failed to create stdin pipe: %v", err)
+	}
+	go func() {
+		if _, err := io.Copy(in, stream); err != nil {
+			glog.Errorf("Failed to copy port forward input for %q port %d: %v", id, port, err)
+		}
+		in.Close()
+		glog.V(4).Infof("Finish copy port forward input for %q port %d: %v", id, port)
+	}()
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("nsenter command returns error: %v, stderr: %q", err, stderr.String())
+	}
+
+	glog.V(2).Infof("Finish port forwarding for %q port %d", id, port)
+
+	return nil
 }
--- a/pkg/server/streaming.go
+++ b/pkg/server/streaming.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net"
 
 	"golang.org/x/net/context"
@@ -83,7 +84,10 @@ func (s *streamRuntime) Attach(containerID string, in io.Reader, out, err io.Wri
 }
 
 func (s *streamRuntime) PortForward(podSandboxID string, port int32, stream io.ReadWriteCloser) error {
-	return errors.New("not implemented")
+	if port <= 0 || port > math.MaxUint16 {
+		return fmt.Errorf("invalid port %d", port)
+	}
+	return s.c.portForward(podSandboxID, port, stream)
 }
 
 // handleResizing spawns a goroutine that processes the resize channel, calling resizeFunc for each
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
     - sudo apt-get install btrfs-tools
     - sudo apt-get install libseccomp2/trusty-backports
     - sudo apt-get install libseccomp-dev/trusty-backports
+    - sudo apt-get install socat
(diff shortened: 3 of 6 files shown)


pkg/server/streaming.go
```diff
@@ -83,7 +84,10 @@ func (s *streamRuntime) Attach(containerID string, in io.Reader, out, err io.Wri
 }
 
 func (s *streamRuntime) PortForward(podSandboxID string, port int32, stream io.ReadWriteCloser) error {
-	return errors.New("not implemented")
+	if port < 0 || port > math.MaxUint16 {
```
AkihiroSuda: port <= 0 ?

Random-Liu: The orignal code in kubelet is using `port < 0`. However, you are right, we should check `==0` as well. Will update.

Random-Liu: Done.

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,99 @@ limitations under the License.
...
 
+	"github.com/containerd/containerd/api/services/tasks/v1"
+	"github.com/containerd/containerd/api/types/task"
+	"github.com/golang/glog"
 	"golang.org/x/net/context"
-
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 )
 
 // PortForward prepares a streaming endpoint to forward ports from a PodSandbox, and returns the address.
```
mikebrow: Comment does not follow what is happening in the api.

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,99 @@ limitations under the License.
...
+
+	if info.Task.Status != task.StatusRunning {
+		return nil, errors.New("sandbox container is not running")
+	}
+	// TODO(random-liu): Verify that ports are exposed.
+	return c.streamServer.GetPortForward(r)
+}
+
+// portForward requires `nsenter` and `socat` on the node.
+func (c *criContainerdService) portForward(id string, port int32, stream io.ReadWriteCloser) error {
```
mikebrow: should this api be here or over in streaming.go

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,99 @@ limitations under the License.
...
+	info, err := c.taskService.Get(ctx, &tasks.GetTaskRequest{ContainerID: id})
+	if err != nil && !isContainerdGRPCNotFoundError(err) {
+		return nil, fmt.Errorf("unable to get sandbox container info: %v", err)
+	}
+
+	if info.Task.Status != task.StatusRunning {
+		return nil, errors.New("sandbox container is not running")
+	}
+	// TODO(random-liu): Verify that ports are exposed.
+	return c.streamServer.GetPortForward(r)
```
mikebrow: Curious.. why all the above validation on the running status? Are there expected errors if the sandbox isn't running?

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,99 @@ limitations under the License.
...
+}
+
+// portForward requires `nsenter` and `socat` on the node.
```
mikebrow: maybe some explanation about what the api does.

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,99 @@ limitations under the License.
...
+		return fmt.Errorf("unable to find sandbox in store: %v", err)
+	}
+	pid := s.Pid
+
+	socat, err := exec.LookPath("socat")
+	if err != nil {
+		return fmt.Errorf("unable to find socat: %v", err)
+	}
+
+	args := []string{"-t", fmt.Sprintf("%d", pid), "-n", socat,
```
mikebrow: maybe a comment about the options being used.

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,99 @@ limitations under the License.
...
+		if _, err := io.Copy(in, stream); err != nil {
+			glog.Errorf("Failed to copy port forward stdin for %q: %v", id, err)
+		}
+		in.Close()
+	}()
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("nsenter command returns error: %v, stderr: %q", err, stderr.String())
+	}
+
```
mikebrow: wait for success on in.Close() before returning nil?

Random-Liu: I think we want the goroutine to keep serving even after the function returns, until the client disconnects.

pkg/server/sandbox_portforward.go
Random-Liu: Good idea. Will do.

pkg/server/sandbox_portforward.go
Random-Liu: Will do.

pkg/server/sandbox_portforward.go
Random-Liu: Portforward nsenters a the pause container namespace, and running a `socat` server to do portforwarding.

I believe this requires the pause container is running, because we use the pid to find it's namespace.

We could probably use permanent network namespace to prevent this in the future.

Random-Liu: Portforward nsenters a the pause container namespace, and running a `socat` server to do portforwarding.

I believe this requires the pause container is running, because we use the pid to find it's namespace.

We could probably use permanent network namespace to prevent this in the future.

pkg/server/sandbox_portforward.go
mikebrow: kk didn't know that was a long running task. Maybe a comment about that..

Random-Liu: I'll add some comment to describe what port forward is doing.

pkg/server/sandbox_portforward.go
mikebrow: Does the GetPortForward api require we make sure this is the case, or will it return an appropriate error if we have not. IOW are we double checking what the API is going to check anyway?

pkg/server/sandbox_portforward.go
Random-Liu: The comment is the same with dockershim https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_streaming.go#L115.
And I think it explains what the function is doing. :)

pkg/server/sandbox_portforward.go
Random-Liu: I put `attach`, `portForward` and `exec` in corresponding source file.

pkg/server/sandbox_portforward.go
Random-Liu: streamServer doesn't even know CRI, how does it check container status?

pkg/server/sandbox_portforward.go
Random-Liu: Done.

pkg/server/sandbox_portforward.go
Random-Liu: Done.

pkg/server/sandbox_portforward.go
Random-Liu: Actually, you are correct, we should wait on the go func. Will update.

pkg/server/sandbox_portforward.go
mikebrow: not what I meant but ok..

I presumed you wanted to ensure there is a pid and associated namespace before calling.. and also presumed that the streamServer would also check and return an error message if not vs using a static table of some sort. But I'm not looking at it's code so..

pkg/server/sandbox_portforward.go
Random-Liu: The implementation is the same with https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_streaming.go#L162. And based on the original comment and code analysis, the code should be safe.

I added more logs so that if in the future there are any problem, we could figure it out.

Random-Liu:
@mikebrow Addressed comments.

pkg/server/sandbox_portforward.go
mikebrow: ok I'm confused .. where does the api "prepare a streaming endpoint to forward ports from a PodSandbox" all I see is ensuring the running state for the sandbox and calling GetPortForward which I presume just does a get?

Random-Liu: `GetPortForward` prepares the streaming endpoint and returns the URL.

pkg/server/sandbox_portforward.go
Random-Liu: @mikebrow We want to avoid returning useless streaming endpoint as much as possible. If we know the sandbox is not running, we should just return error instead of returning a streaming endpoint which doesn't work.

If the sandbox exits after we return the endpoint, the actual port forward will fail, but that's fine. We just need to do best effort.

As for `GetPortForward`, actually it just save the entry inside a map, and wait for apiserver to invoke the actual port forward. It will success even if sandbox is not running.

pkg/server/sandbox_portforward.go
mikebrow: Logging works if we have the right test buckets...  hate leaving what might be a dangling pipe thread but at least we'll have an unbalanced lvl 4 log to identify it if it happens.

pkg/server/sandbox_portforward.go
mikebrow: kk thx!

pkg/server/sandbox_portforward.go
mikebrow: I see.. one of "those" apis :-) should probably be called Reserve instead of Get.. now it makes sense.

pkg/server/sandbox_portforward.go
Random-Liu: @mikebrow `Run` already contains a `Wait`, and the wait makes sure that the stream will be closed.

If `Run` doesn't do that, I'd prefer add a channel and do select. :)

pkg/server/sandbox_portforward.go
mikebrow: Don't understand the answer. There is a PortForward api in streaming.go that calls portForward.  The PortForward api in sandbox_portforward.go does not call portForward. Thus my question.

pkg/server/sandbox_portforward.go
mikebrow: Makes sense.

pkg/server/sandbox_portforward.go
```diff
@@ -17,14 +17,107 @@ limitations under the License.
...
+		}
+	}()
+
+	sandbox, err := c.sandboxStore.Get(r.GetPodSandboxId())
+	if err != nil {
+		return nil, fmt.Errorf("unable to find sandbox: %v", err)
+	}
+	id := sandbox.ID
+
+	info, err := c.taskService.Get(ctx, &tasks.GetTaskRequest{ContainerID: id})
```
abhinandanpb: Can you please add a TODO to replace this with client API

abhinandanpb: or TODO to use permanent network namespace ( discussed below)

pkg/server/sandbox_portforward.go
abhinandanpb: Just for my understanding : 
Even if the task is paused/stopped the namespace is still available to nsenter right? 
How is this supposed to work with windows in the future.

abhinandanpb:
Minor comments, LGTM

pkg/server/sandbox_portforward.go
Random-Liu: > Even if the task is paused/stopped the namespace is still available to nsenter right?

Offline talked with @yujuhong. Yeah, we only need network namespace here. I had a concern that we run a pod specific process but only in its network namespace, it means that the process could access other host namespaces. However, socat is on the node, the cluster admin should make sure it's friendly. :) Actually, we have other issues not resolved, e.g. the socat is not in pod cgroup, so the resource usage will be charged to the system.

> How is this supposed to work with windows in the future.

@abhinandanpb Good question. Currently, Kubelet only has limited windows support now. Many critical kubelet features doesn't have corresponding windows support, e.g. pod level cgroup, cadvisor monitoring etc.

Actually, we should be able to run a port forward container inside the pod sandbox to do port forward, which makes more sense because:
1) The resource overhead will be charged to the pod;
2) If containerd container exec support different platforms, this solution could be platform agnostic.

I'll add a TODO here, we could do this in near future. :)

pkg/server/sandbox_portforward.go
Random-Liu: Will do.

pkg/server/sandbox_portforward.go
Random-Liu: Done.

pkg/server/sandbox_portforward.go
Random-Liu: Done.

Random-Liu:
@abhinandanpb Addressed comments. Thanks for reviewing

pkg/server/sandbox_portforward.go
abhinandanpb: got it. Thanks for explaining.

Random-Liu:
Just updated this PR to use the containerd client.

PR #132, opened 2017-08-14
Split the travis Build in different Build and Test Stage

asifdxtreme:
Split the travis Build to different logical stages so that it's easy to understand what went wrong in CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 go:
-    - 1.8.x
-    - tip
+    - 1.8
 
 sudo: required
 
@@ -24,12 +23,20 @@ install:
 before_script:
     - export PATH=$HOME/gopath/bin:$PATH
 
-script:
-    - make .gitvalidation
-    - make verify
-    - make binaries
-    - make test
-    - make test-cri
+jobs:
+  include:
+    - stage: Static Check
+      script: 
+        - make .gitvalidation
+        - make verify
+    - stage: Build
+      script: make binaries
+    - stage: Unit Test
+      script: make test
+    - stage: Integration Test
+      script:
+        - make binaries
+        - make test-cri
 after_script:
     # Abuse travis to preserve the log.
     # TODO(random-liu): Use prow for integration test.
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,26 +17,25 @@ install:
     - sudo apt-get install libseccomp2/trusty-backports
     - sudo apt-get install libseccomp-dev/trusty-backports
     - docker run --rm -v /usr/local/bin:/target jpetazzo/nsenter 
-    - make install.tools
-    - make install.deps
 
 before_script:
     - export PATH=$HOME/gopath/bin:$PATH
 
 jobs:
   include:
-    - stage: Static Check
+    - stage: Build
       script: 
+        - make install.tools
         - make .gitvalidation
         - make verify
-    - stage: Build
-      script: make binaries
-    - stage: Unit Test
-      script: make test
-    - stage: Integration Test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,12 +24,16 @@ install:
...
-    - make verify
-    - make binaries
-    - make test
-    - make test-cri
+jobs:
+  include:
+    - stage: Static Check and Build
+      script: 
+        - make .gitvalidation
+        - make verify
```
mikebrow: make binaries is missing

mikebrow: really odd output

I see a build output with make being called in the log. But I don't see the call to make in the travis.yml

I can't see any output from the stages either?

.travis.yml
```diff
@@ -24,12 +23,17 @@ install:
...
+  include:
+    - stage: Static Check
+      script: 
+        - make .gitvalidation
+        - make verify
+    - stage: Build and Test
+      script:
+        - make binaries
+        - make test
+        - make test-cri
```
mikebrow: yes make test-cri has a dependency on the binaries. make test should not .. if that matters.

mikebrow:
each commit has to be signed .. :-)

asifdxtreme:
@mikebrow Thanks, signed off all my commits

Random-Liu:
This feature is really cool!
Is there any limit to the number of stages? Could we split the test into 4 stages?
```yaml
  include:
     - stage: Static Check
       script: 
         - make .gitvalidation
         - make verify
     - stage: Build
       script:
         - make binaries
     - state: Unit Test
       script:
         - make test
     - state: Integration Test
       script:
         - make test-cri
```

asifdxtreme:
@Random-Liu : As Integration test depends on the output of 'make binaries' so we cannot split integration test and build, maximum we can split UT in different stage

Random-Liu:
@asifdxtreme Hm, so these stage can't access result generated by last stage?

asifdxtreme:
@Random-Liu No, the stages are the way to seperate out the each step, anyhow I have made changes for 4 stages, please have a look at it

Random-Liu:
@asifdxtreme Thanks for explaining. Yeah, you are right:
```
It is important to note that jobs do not share storage, as each job runs in a fresh VM or container. If your jobs need to share files (e.g., using build artifacts from the “Test” stage for deployment in the subsequent “Deploy” stage), you need to use an external storage mechanism such as S3 and a remote scp server.
```

Random-Liu:
@asifdxtreme If so, this seems less useful then. :(

It's time consuming to install dependencies for each VM.

asifdxtreme:
@Random-Liu but this will make your CI more stringent because this will ensure each and every step is carried out in isolation and no other steps interfere or effect the output of others, moreover its easy for new users to interpret what is happening in CI

asifdxtreme:
@Random-Liu yes this makes sense to reduce into 2 steps, I will make the changes now

asifdxtreme:
@Random-Liu the changes has been done as per suggestion and you can see the build time has reduced considerably to 11mins from 18mins

Random-Liu:
LGTM

mikebrow:
@Random-Liu missed the part about changing which go versions we are testing against.

asifdxtreme:
@Random-Liu yes currently we are testing it against 1.8, Currently stages are difficult to run with multiple versions of any language, for example if run test against go 1.8 and tip then it will use around 4 containers and in return double the time of build.

mikebrow:
@asifdxtreme  We used two scripts in each stage to run the two golang versions in parallel.

PR #134, opened 2017-08-15
Putting in the go versions we want to test against

mikebrow:
Fixing the go versions being tested against after the last travis PR introducing stages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,29 +21,25 @@ before_script:
 
 jobs:
   include:
-    - stage: Build
+    - stage: Verify
       script:
         - make install.tools
         - make .gitvalidation
         - make verify
-        - make binaries
       go: 1.8.x
-    - stage: Build
-      script:
+    - script:
         - make install.tools
         - make .gitvalidation
         - make verify
-        - make binaries
       go: tip
-    - stage: Test
+    - stage: Build and Test
       script:
         - make install.deps
         - make test
         - make binaries
         - make test-cri
       go: 1.8.x
-    - stage: Test
-      script:
+    - script:
         - make install.deps
         - make test
         - make binaries
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: go
-go:
-    - 1.8
 
 sudo: required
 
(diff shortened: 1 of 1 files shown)


mikebrow:
And there's a bug with stages where you can't matrix in multiple go versions.. great. I'll push a different fix.

mikebrow:
So the trick with this beta stages tool and the matrix feature seems to be to create your own matrix of stages by hand for the various combinations you want.  Since we had two go versions to test against both stages that doubles the number of stages, but they have the same name so it looks correct in the travis log. If / when they fix this, we can go back to the old go matrix style. Apparently half the conversation around this beta feature is about this particular issue... https://github.com/travis-ci/beta-features/issues/11#issuecomment-301875447

Random-Liu:
@mikebrow Actually, I'm fine with only build with 1.8, which is also what containerd doing.

mikebrow:
They are testing 1.8.x which is currently v1.8.3. We are testing v1.8 now. Pretty much all of the incubator projects are testing against 1.8.3 and tip. The value of tip is to make sure we don't enter any changes that will become incompatible with future releases of go.. and to aid in forcing golang to generate future releases that are compatible with prior ones..

mikebrow:
That worked.. going to try to add stage scripts so there's less copying

mikebrow:
I like it... @Random-Liu what do you think?

Random-Liu:
@mikebrow Different stages will run in serial, previously the test only took 11min, but with this change it takes 26min. 

According to the document https://docs.travis-ci.com/user/build-stages/, you should change it to:
```
    - stage: Build
      script: ./hack/travis-build.sh
      go: 1.8.x
    - script: ./hack/travis-build.sh
      go: tip
    - stage: Test
      script: ./hack/travis-test.sh
      go: 1.8.x
    - script: ./hack/travis-test.sh
      go: tip
```

And after changing to `./hack/travis-test.sh`, it's hard to figure out what commands are running, I kind of prefer the previous solution.

mikebrow:
Agreed. Updated per comments and offline discussion.

mikebrow:
Per discussion streamlined the two stages into Verify; and Build and Test

Random-Liu:
LGTM

PR #137, opened 2017-08-17
Some cleanup after switching to new client.

Random-Liu:
Some cleanup after switching to new containerd client.

### DIFF
--- a/pkg/server/container_execsync.go
+++ b/pkg/server/container_execsync.go
@@ -23,10 +23,10 @@ import (
 	"time"
 
 	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/api/services/events/v1"
-	"github.com/containerd/containerd/typeurl"
+	"github.com/containerd/containerd/errdefs"
 	"github.com/golang/glog"
 	"golang.org/x/net/context"
+	"golang.org/x/sys/unix"
 	"k8s.io/client-go/tools/remotecommand"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 )
@@ -72,10 +72,21 @@ type execOptions struct {
 	timeout time.Duration
 }
 
+// execResult is the result returned by exec.
+type execResult struct {
+	exitCode uint32
+	err      error
+}
+
 // execInContainer executes a command inside the container synchronously, and
 // redirects stdio stream properly.
-// TODO(random-liu): Support timeout.
 func (c *criContainerdService) execInContainer(ctx context.Context, id string, opts execOptions) (*uint32, error) {
+	// Cancel the context before returning to ensure goroutines are stopped.
+	// This is important, because if `Start` returns error, `Wait` will hang
+	// forever unless we cancel the context.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	// Get container from our container store.
 	cntr, err := c.containerStore.Get(id)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/server/container_execsync.go
```diff
@@ -131,50 +131,22 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
-	// TODO(random-liu): Use `Wait` after is fixed. (containerd#1279, containerd#1287)
-	cancellable, cancel := context.WithCancel(ctx)
-	eventstream, err := c.eventService.Subscribe(cancellable, &events.SubscribeRequest{})
-	if err != nil {
-		return nil, fmt.Errorf("failed to subscribe event stream: %v", err)
-	}
-	defer cancel()
-
+	resCh := make(chan execResult, 1)
+	go func() {
```
abhinandanpb: Will this go routine be stuck if the ```process.Start(ctx)``` errors out ? Because there will be no listener on the channel.

abhinandanpb: actually it won't since its buffered channel

abhinandanpb:
LGTM

ijc:
LGTM (FWIW)

pkg/server/container_execsync.go
Random-Liu: If `Start` fails, we'll go to defer to delete the container, and then `Wait` will return. :)

However, this is not that obvious... I think we should make it more clear, let me see.

Random-Liu: You are correct, I even found a containerd bug when trying to address this https://github.com/containerd/containerd/issues/1376.

Random-Liu:
@abhinandanpb Addressed comments. PTAL.

Random-Liu:
The new commit also added the `ExecSync` timeout.

pkg/server/container_execsync.go
```diff
@@ -120,6 +127,8 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 		return nil, fmt.Errorf("failed to create exec %q: %v", execID, err)
 	}
 	defer func() {
+		// TODO(random-liu): There is a containerd bug here containerd#1376, revisit this
+		// after that is fixed.
 		if _, err := process.Delete(ctx); err != nil {
```
abhinandanpb: wondering if we should just defer on ```process.Delete(ctx, containerd.WithProcessKill)``` ?

pkg/server/container_execsync.go
```diff
@@ -131,50 +140,40 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
+		glog.V(2).Infof("Exec process %q exits with exit code %d and error %v", execID, exitCode, err)
+	}()
 	if err := process.Start(ctx); err != nil {
 		return nil, fmt.Errorf("failed to start exec %q: %v", execID, err)
 	}
 
-	exitCode, err := c.waitContainerExec(eventstream, id, execID)
-	if err != nil {
-		return nil, fmt.Errorf("failed to wait for exec in container %q to finish: %v", id, err)
+	var timeoutCh <-chan time.Time
```
abhinandanpb: And then use context.WithDeadline if the timeout is set and just return  
```
select{
     <-ctx.Done() : return err
                               ....
     <-resCh      :  ....
}
```

abhinandanpb: Just a thought. This works great too. WDYT ?

pkg/server/container_execsync.go
Random-Liu: As is mentioned in the containerd issue https://github.com/containerd/containerd/issues/1376, it doesn't work.

pkg/server/container_execsync.go
Random-Liu: I've tried that. The problem is that once context is cancelled, we could not use it to call `Delete` or `Kill`. Of course, we could use `context.Background()` for the those operations, but I'm not sure what the code style should be.

That is also why I filed https://github.com/kubernetes-incubator/cri-containerd/issues/142.

abhinandanpb: Got it . makes sense

pkg/server/container_execsync.go
abhinandanpb: Yep yep. forgot about it.

abhinandanpb:
LGTM

Random-Liu:
@abhinandanpb Thanks for reviewing! :)

Random-Liu:
Squashed commits. Will merge after test passes.

PR #138, opened 2017-08-17
Creating sandbox namespace

abhinandanpb:
This commit contains changes to create/delete permanent namespace
for a sandbox container.

### DIFF
--- a/pkg/server/sandbox_list_test.go
+++ b/pkg/server/sandbox_list_test.go
@@ -43,7 +43,7 @@ func TestToCRISandbox(t *testing.T) {
 		Name:      "test-name",
 		Config:    config,
 		CreatedAt: createdAt,
-		NetNS:     "test-netns",
+		NetNSPath: "test-netns",
 	}
 	state := runtime.PodSandboxState_SANDBOX_READY
 	expect := &runtime.PodSandbox{
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -77,9 +77,49 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	if err != nil {
 		return nil, fmt.Errorf("failed to get sandbox image %q: %v", c.sandboxImage, err)
 	}
+	//Create Network Namespace if it is not in host network
+	hostNet := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
+	if !hostNet {
+		// If it is not in host network namespace then create a namespace and set the sandbox
+		// handle. NetNSPath in sandbox metadata and NetNS is non empty only for non host network
+		// namespaces. If the pod is in host network namespace then both are empty and should not
+		// be used.
+		sandbox.NetNS, err = sandboxstore.NewNetNS()
+		if err != nil {
+			return nil, fmt.Errorf("failed to create network namespace for sandbox %q: %v", id, err)
+		}
+		sandbox.NetNSPath = sandbox.NetNS.GetPath()
+		defer func() {
+			if retErr != nil {
+				if err := sandbox.NetNS.Remove(); err != nil {
+					glog.Errorf("Failed to remove network namespace %s for sandbox %q: %v", sandbox.NetNSPath, id, err)
+				}
+				sandbox.NetNSPath = ""
+			}
+		}()
+		// Setup network for sandbox.
+		podNetwork := ocicni.PodNetwork{
+			Name:         config.GetMetadata().GetName(),
+			Namespace:    config.GetMetadata().GetNamespace(),
+			ID:           id,
+			NetNS:        sandbox.NetNSPath,
+			PortMappings: toCNIPortMappings(config.GetPortMappings()),
+		}
+		if err = c.netPlugin.SetUpPod(podNetwork); err != nil {
+			return nil, fmt.Errorf("failed to setup network for sandbox %q: %v", id, err)
+		}
+		defer func() {
+			if retErr != nil {
+				// Teardown network if an error is returned.
+				if err := c.netPlugin.TearDownPod(podNetwork); err != nil {
+					glog.Errorf("Failed to destroy network for sandbox %q: %v", id, err)
+				}
+			}
+		}()
+	}
 
 	// Create sandbox container.
-	spec, err := c.generateSandboxContainerSpec(id, config, image.Config)
+	spec, err := c.generateSandboxContainerSpec(id, config, image.Config, sandbox.NetNSPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate sandbox container spec: %v", err)
 	}
@@ -165,30 +205,6 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	}()
 
 	sandbox.Pid = task.Pid()
-	sandbox.NetNS = getNetworkNamespace(task.Pid())
-	if !config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
-		// Setup network for sandbox.
-		// TODO(random-liu): [P2] Replace with permanent network namespace.
-		podNetwork := ocicni.PodNetwork{
-			Name:         config.GetMetadata().GetName(),
-			Namespace:    config.GetMetadata().GetNamespace(),
-			ID:           id,
-			NetNS:        sandbox.NetNS,
-			PortMappings: toCNIPortMappings(config.GetPortMappings()),
-		}
-		if err = c.netPlugin.SetUpPod(podNetwork); err != nil {
-			return nil, fmt.Errorf("failed to setup network for sandbox %q: %v", id, err)
-		}
-		defer func() {
-			if retErr != nil {
-				// Teardown network if an error is returned.
-				if err := c.netPlugin.TearDownPod(podNetwork); err != nil {
-					glog.Errorf("failed to destroy network for sandbox %q: %v", id, err)
-				}
-			}
-		}()
-	}
-
 	if err = task.Start(ctx); err != nil {
 		return nil, fmt.Errorf("failed to start sandbox container task %q: %v",
 			id, err)
@@ -205,7 +221,7 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 }
 
 func (c *criContainerdService) generateSandboxContainerSpec(id string, config *runtime.PodSandboxConfig,
-	imageConfig *imagespec.ImageConfig) (*runtimespec.Spec, error) {
+	imageConfig *imagespec.ImageConfig, nsPath string) (*runtimespec.Spec, error) {
 	// Creates a spec Generator with the default spec.
 	// TODO(random-liu): [P1] Compare the default settings with docker and containerd default.
 	spec, err := containerd.GenerateSpec()
@@ -252,15 +268,12 @@ func (c *criContainerdService) generateSandboxContainerSpec(id string, config *r
 
 	// Set namespace options.
 	nsOptions := config.GetLinux().GetSecurityContext().GetNamespaceOptions()
-	// TODO(random-liu): [P1] Create permanent network namespace, so that we could still cleanup
-	// network namespace after sandbox container dies unexpectedly.
-	// By default, all namespaces are enabled for the container, runc will create a new namespace
-	// for it. By removing the namespace, the container will inherit the namespace of the runtime.
 	if nsOptions.GetHostNetwork() {
 		g.RemoveLinuxNamespace(string(runtimespec.NetworkNamespace)) // nolint: errcheck
-		// TODO(random-liu): [P1] Figure out how to handle UTS namespace.
+	} else {
+		//TODO(Abhi): May be move this to containerd spec opts (WithLinuxSpaceOption)
+		g.AddOrReplaceLinuxNamespace(string(runtimespec.NetworkNamespace), nsPath) // nolint: errcheck
 	}
-
 	if nsOptions.GetHostPid() {
 		g.RemoveLinuxNamespace(string(runtimespec.PIDNamespace)) // nolint: errcheck
 	}
--- a/pkg/server/sandbox_run_test.go
+++ b/pkg/server/sandbox_run_test.go
@@ -68,6 +68,7 @@ func getRunPodSandboxTestData() (*runtime.PodSandboxConfig, *imagespec.ImageConf
 
 func TestGenerateSandboxContainerSpec(t *testing.T) {
 	testID := "test-id"
+	nsPath := "test-cni"
 	for desc, test := range map[string]struct {
 		configChange      func(*runtime.PodSandboxConfig)
 		imageConfigChange func(*imagespec.ImageConfig)
@@ -80,6 +81,7 @@ func TestGenerateSandboxContainerSpec(t *testing.T) {
 				require.NotNil(t, spec.Linux)
 				assert.Contains(t, spec.Linux.Namespaces, runtimespec.LinuxNamespace{
 					Type: runtimespec.NetworkNamespace,
+					Path: nsPath,
 				})
 				assert.Contains(t, spec.Linux.Namespaces, runtimespec.LinuxNamespace{
 					Type: runtimespec.PIDNamespace,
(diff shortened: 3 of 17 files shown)


Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/43.

pkg/server/container_create.go
```diff
@@ -440,9 +441,9 @@ func setOCICapabilities(g *generate.Generator, capabilities *runtime.Capability,
 }
 
 // setOCINamespaces sets namespaces.
-func setOCINamespaces(g *generate.Generator, namespaces *runtime.NamespaceOption, sandboxPid uint32) {
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.NetworkNamespace), getNetworkNamespace(sandboxPid)) // nolint: errcheck
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.IPCNamespace), getIPCNamespace(sandboxPid))         // nolint: errcheck
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.UTSNamespace), getUTSNamespace(sandboxPid))         // nolint: errcheck
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.PIDNamespace), getPIDNamespace(sandboxPid))         // nolint: errcheck
+func setOCINamespaces(g *generate.Generator, namespaces *runtime.NamespaceOption, sandbox *sandboxstore.Sandbox) {
+	g.AddOrReplaceLinuxNamespace(string(runtimespec.NetworkNamespace), sandbox.NetNS)            // nolint: errcheck
```
Random-Liu: Whether we could continue using `getNetworkNamespace(sandboxPid)` here? Need to think about it.

pkg/store/sandbox/sandbox.go
```diff
@@ -20,7 +20,7 @@ import (
 	"sync"
 
 	"github.com/containerd/containerd"
-
+	cnins "github.com/containernetworking/plugins/pkg/ns"
```
abhinandanpb: Damn !

pkg/store/sandbox/sandbox.go
```diff
@@ -31,7 +31,9 @@ type Sandbox struct {
 	Metadata
 	// Containerd sandbox container
 	Container containerd.Container
-	// TODO(random-liu): Add cni network namespace client.
+	// CNI network namespace client
+	//TODO(Abhi): Should we handle multi network scenario
```
abhinandanpb: It was a note to myself to see if there is a case where pod can be attached to multiple networks. Then how does sandbox hold the network namespace.

pkg/server/sandbox_run.go
```diff
@@ -77,9 +82,30 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	if err != nil {
 		return nil, fmt.Errorf("failed to get sandbox image %q: %v", c.sandboxImage, err)
 	}
+	//Create Network Namespace if it is not in host network
+	hostNet := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
+	nsPath := HostNamespacePath
```
abhinandanpb: or we could do ns.GetCurrentNS() to get the current namespace and the path.

pkg/server/sandbox_run.go
```diff
@@ -165,10 +191,9 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	}()
 
 	sandbox.Pid = task.Pid()
-	sandbox.NetNS = getNetworkNamespace(task.Pid())
-	if !config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
+	sandbox.NetNS = nsPath
+	if !hostNet {
```
abhinandanpb: sure will move the logic

pkg/store/sandbox/sandbox.go
Random-Liu: @abhinandanpb For now, I believe the answer is no. :) When you do `PodSandboxStatus`, we only expect 1 IP to be returned.

pkg/server/sandbox_run.go
Random-Liu: @abhinandanpb SGTM.

abhinandanpb:
@Random-Liu updated. PTAL

pkg/server/container_create.go
```diff
@@ -155,7 +156,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	return &runtime.CreateContainerResponse{ContainerId: id}, nil
 }
 
-func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint32, config *runtime.ContainerConfig,
+func (c *criContainerdService) generateContainerSpec(id string, sandbox sandboxstore.Sandbox, config *runtime.ContainerConfig,
```
abhinandanpb: For now passing the whole struct because I dont see this to be heavy. I feel its a tradeoff between ever increasing arguements/object. WDYT?

Random-Liu:
@abhinandanpb Will look at this this morning.

pkg/store/sandbox/metadata.go
```diff
@@ -51,8 +51,8 @@ type Metadata struct {
 	CreatedAt int64
 	// Pid is the process id of the sandbox.
 	Pid uint32
-	// NetNS is the network namespace used by the sandbox.
-	NetNS string
+	// NetNSPath is the network namespace used by the sandbox.
+	NetNSPath string
```
Random-Liu: We still need to use and initialize this field. The reason why we have this here is that cri-containerd may die.

After cri-containerd comes up, you won't know what network namespace a sandbox is using, thus you won't be able to load and recover it.

The sandbox metadata is checkpointed, so we could read network namespace path from the checkpoint, and reload the network namespace.

abhinandanpb: Got it. will do it

pkg/server/sandbox_stop.go
```diff
@@ -62,17 +61,19 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
 	}
 
 	// Teardown network for sandbox.
-	_, err = c.os.Stat(sandbox.NetNS)
-	if err == nil {
-		if !sandbox.Config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
-			if teardownErr := c.netPlugin.TearDownPod(sandbox.NetNS, sandbox.Config.GetMetadata().GetNamespace(),
+	if sandbox.GetNetNsPath() != "" {
+		if err == nil {
```
abhinandanpb: bad rebase

abhinandanpb:
@Random-Liu Done

pkg/server/container_create.go
```diff
@@ -453,9 +454,13 @@ func setOCICapabilities(g *generate.Generator, capabilities *runtime.Capability)
 }
 
 // setOCINamespaces sets namespaces.
-func setOCINamespaces(g *generate.Generator, namespaces *runtime.NamespaceOption, sandboxPid uint32) {
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.NetworkNamespace), getNetworkNamespace(sandboxPid)) // nolint: errcheck
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.IPCNamespace), getIPCNamespace(sandboxPid))         // nolint: errcheck
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.UTSNamespace), getUTSNamespace(sandboxPid))         // nolint: errcheck
-	g.AddOrReplaceLinuxNamespace(string(runtimespec.PIDNamespace), getPIDNamespace(sandboxPid))         // nolint: errcheck
+func setOCINamespaces(g *generate.Generator, namespaces *runtime.NamespaceOption, sandbox sandboxstore.Sandbox) {
+	nsPath := sandbox.GetNetNsPath()
```
Random-Liu: Why can't we use keep using sandboxPid here? Seems cleaner and less change.

pkg/server/sandbox_list_test.go
```diff
@@ -43,7 +43,6 @@ func TestToCRISandbox(t *testing.T) {
 		Name:      "test-name",
 		Config:    config,
 		CreatedAt: createdAt,
-		NetNS:     "test-netns",
```
Random-Liu: Add this back.

pkg/server/sandbox_run.go
```diff
@@ -78,9 +78,45 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
 		return nil, fmt.Errorf("failed to get sandbox image %q: %v", c.sandboxImage, err)
 	}
+	//Create Network Namespace if it is not in host network
+	hostNet := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
+	if !hostNet {
+		// If it is not in host network namespace then create a namespace and set the sandbox
+		// handle
+		err := sandbox.CreateNetNS()
+		if err != nil {
+			return nil, err
```
Random-Liu: error description.

pkg/server/sandbox_run.go
```diff
@@ -78,9 +78,45 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+		}
+		defer func() {
+			if retErr != nil {
+				if err := sandbox.RemoveNetNS(); err != nil {
+					glog.Errorf("failed to remove network namespace for sandbox %q: %v", id, err)
```
Random-Liu: Include network namespace path here.

Random-Liu: s/failed/Failed.

Not your fault, but let's fix it here.

Our convention in Kubernetes is that, error must start with lower case; log must start with upper case.

pkg/server/sandbox_run.go
```diff
@@ -78,9 +78,45 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+			PortMappings: toCNIPortMappings(config.GetPortMappings()),
+		}
+		if err = c.netPlugin.SetUpPod(podNetwork); err != nil {
+			return nil, fmt.Errorf("failed to setup network for sandbox %q: %v", id, err)
+		}
+		defer func() {
+			if retErr != nil {
+				// Teardown network if an error is returned.
+				if err := c.netPlugin.TearDownPod(podNetwork); err != nil {
+					glog.Errorf("failed to destroy network for sandbox %q: %v", id, err)
```
Random-Liu: ditto.

pkg/server/sandbox_run_test.go
```diff
@@ -136,7 +136,7 @@ func TestGenerateSandboxContainerSpec(t *testing.T) {
 		if test.imageConfigChange != nil {
 			test.imageConfigChange(imageConfig)
 		}
-		spec, err := c.generateSandboxContainerSpec(testID, config, imageConfig)
+		spec, err := c.generateSandboxContainerSpec(testID, config, imageConfig, "")
```
Random-Liu: Please add a test case in this test.

pkg/server/sandbox_stop.go
```diff
@@ -63,22 +62,22 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
-			if teardownErr := c.netPlugin.TearDownPod(ocicni.PodNetwork{
-				Name:         sandbox.Config.GetMetadata().GetName(),
-				Namespace:    sandbox.Config.GetMetadata().GetNamespace(),
-				ID:           id,
-				NetNS:        sandbox.NetNS,
-				PortMappings: toCNIPortMappings(sandbox.Config.GetPortMappings()),
-			}); teardownErr != nil {
-				return nil, fmt.Errorf("failed to destroy network for sandbox %q: %v", id, teardownErr)
-			}
+	if netns := sandbox.GetNetNsPath(); netns != "" {
```
Random-Liu: We still need to `Stat`. `Stat` is to make sure that we won't try to teardown the network when the namespace is not there, or removed by someone else.

pkg/store/sandbox/sandbox.go
```diff
@@ -41,6 +45,49 @@ type Store struct {
 	// TODO(random-liu): Add trunc index.
 }
 
+// NetNS holds network namespace for sandbox
```
Random-Liu: 1) Can we move this to a separate file? netns.go or something?
2) Why following functions are member function of `Sandbox` instead of `NetNS`?

Basically, I want Sandbox/Container to be a simple collection of resources, each resource is thread safe, and you should be able to operate them separately without worrying about any concurrent issue. However, `CreateNetNS` doesn't seem so. It seems better to:
  a) Have a standalone NewNetNS() function;
  b) Add `Remove` and `GetPath` function to `NetNS`. Actually, we don't even need the `GetPath` function, because we could always get it from sandbox metadata.
  c) In the future we'll add `LoadNetNS()` to load existing `NetNS`, which will be useful for restart recovery.

Random-Liu:
@abhinandanpb Review done. :)

Random-Liu:
@abhinandanpb Please run `make verify`.

pkg/server/sandbox_run_test.go
```diff
@@ -113,6 +115,31 @@ func TestGenerateSandboxContainerSpec(t *testing.T) {
 				})
 			},
 		},
+		"non host namespace": {
```
Random-Liu: This seems been covered by the first case already. :)

pkg/server/sandbox_run.go
```diff
@@ -78,9 +78,47 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	if err != nil {
 		return nil, fmt.Errorf("failed to get sandbox image %q: %v", c.sandboxImage, err)
 	}
+	//Create Network Namespace if it is not in host network
+	hostNet := config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork()
```
Random-Liu: Please add a comment to explain that if not using host network, sandbox.NetNSPath and sandbox.NetNS will be left empty, we don't and shouldn't use it.

pkg/server/sandbox_stop.go
```diff
@@ -63,22 +63,27 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
+			NetNS:        sandbox.NetNSPath,
+			PortMappings: toCNIPortMappings(sandbox.Config.GetPortMappings()),
+		}); teardownErr != nil {
+			return nil, fmt.Errorf("failed to destroy network for sandbox %q: %v", id, teardownErr)
+		}
+		//Close the sandbox network namespace if it was created
+		if err = sandbox.NetNS.RemoveNetNS(); err != nil {
+			return nil, fmt.Errorf("failed to remove network namespace for sandbox %q:  %v", id, err)
+		}
+		sandbox.NetNSPath = ""
```
Random-Liu: Let's not rely on `NetNSPath` is empty or not. Sandbox metadata is designed to be immutable after creation, if you update it here, there will be race between `StopSandbox` and `SandboxStatus`.

Just keep `NetNSPath` as it is, I believe it's ok. And in that way we could double check whether network namespace is removed when remove sandbox (See https://github.com/kubernetes-incubator/cri-containerd/issues/164).

pkg/server/sandbox_stop.go
```diff
@@ -63,22 +63,27 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
...
-		return nil, fmt.Errorf("failed to stat netns path for sandbox %q before tearing down the network: %v", id, err)
+		if teardownErr := c.netPlugin.TearDownPod(ocicni.PodNetwork{
+			Name:         sandbox.Config.GetMetadata().GetName(),
+			Namespace:    sandbox.Config.GetMetadata().GetNamespace(),
+			ID:           id,
+			NetNS:        sandbox.NetNSPath,
+			PortMappings: toCNIPortMappings(sandbox.Config.GetPortMappings()),
+		}); teardownErr != nil {
+			return nil, fmt.Errorf("failed to destroy network for sandbox %q: %v", id, teardownErr)
+		}
```
Random-Liu: Mentioned above https://github.com/kubernetes-incubator/cri-containerd/pull/138#discussion_r134025419.
Please add the TODO.

`TearDownPod` is not idempotent. If we successfully `TearDownPod`, but fail to remove network namespace, we'll never be able to stop the sandbox, because `TearDownPod` will always return error.

This is a bug which did happen and caused trouble to us, we should fix it soon. Let's add the TODO here.

pkg/server/sandbox_run_test.go
abhinandanpb: Yes thats why I asked you earlier :)

Random-Liu: Sorry, I misunderstood get your question.

Random-Liu:
LGTM

Random-Liu:
Please squash your commits. Ready to merge then! :)

Random-Liu:
Will merge after test passes.

abhinandanpb:
Done! Thanks for reviewing @Random-Liu 👍

PR #139, opened 2017-08-17
[WIP] Replacing nsenter with netns

abhi:
This commit has changes to replace the usage of nsenter with
netns. This can relax the requirement of requiring nsenter
installation.

### DIFF
--- a/pkg/server/sandbox_portforward.go
+++ b/pkg/server/sandbox_portforward.go
@@ -20,14 +20,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"io"
-	"os/exec"
-	"strings"
-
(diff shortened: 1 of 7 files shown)


abhinandanpb:
Will update doc if we want to take this path

Random-Liu:
Need to figure out how to solve https://github.com/vishvananda/netns/issues/17 and https://github.com/golang/go/issues/20676.

Random-Liu:
The golang issue will be solved in golang 1.10. At that time, we could use `NetNS.Do` to run code inside the network namespace.

abhinandanpb:
Woot Woot ! Thanks for the update @Random-Liu

PR #128, opened 2017-08-12
Add container attach

Random-Liu:
This PR added container attach support in cri-containerd, and also enabled the test.

The attach code is really error prone, so please pay extra attention to see whether there is goroutine leakage or deadlock.

@kubernetes-incubator/maintainers-cri-containerd

PR #151, opened 2017-08-22
Add instrumented service.

Random-Liu:
Move all grpc handler logs into a instrumented service, so that we don't need to care about it in the real service.

### DIFF
--- a/pkg/server/instrumented_service.go
+++ b/pkg/server/instrumented_service.go
@@ -1,0 +1,282 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"github.com/golang/glog"
+	"golang.org/x/net/context"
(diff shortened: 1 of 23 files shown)


abhinandanpb:
Should we convert those logs to something like debugf in logrus. I see that glog has severity level may be print those logs only if the logging level is set appropriately ? Instead of an instrumented code.

Random-Liu:
@abhinandanpb I did this not for the log level. :) Just extract the log logic into an outside wrapper, so that we don't need to see this everywhere. :)

pkg/server/instrumented_service.go
```diff
@@ -0,0 +1,270 @@
...
+
+func newInstrumentedService(c *criContainerdService) CRIContainerdService {
+	return &instrumentedService{criContainerdService: c}
+}
+
+func (in *instrumentedService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandboxRequest) (res *runtime.RunPodSandboxResponse, err error) {
+	glog.V(2).Infof("RunPodSandbox with config %+v", r.GetConfig())
+	defer func() {
+		if err != nil {
+			glog.Errorf("RunPodSandbox for %+v returns error %v", r.GetConfig().GetMetadata(), err)
```
abhinandanpb: ```glog.Errorf("RunPodSandbox for %+v failed, err: %v", r.GetConfig().GetMetadata(), err)```  
would be better ?

Random-Liu: Will update all.

Random-Liu: Done.

abhinandanpb:
LGTM

Random-Liu:
This touches many files, and relatively lower priority. I'll hold this until https://github.com/kubernetes-incubator/cri-containerd/pull/138 and https://github.com/kubernetes-incubator/cri-containerd/pull/128 get merged.

/cc @abhinandanpb

PR #152, opened 2017-08-22
Fix node e2e test.

Random-Liu:
Our first cron build fails, because the test command returns exit code 1, which travis thought the test was failing.

@kubernetes-incubator/maintainers-cri-containerd


Random-Liu:
Merged. @mikebrow Thanks for reviewing!

PR #155, opened 2017-08-23
Support NoNewPrivileges

### LINKED ISSUE #117: Support NoNewPrivileges
Kubernetes 1.8 and runc both support it, we should update CRI version and support it.

miaoyq:
fixes #117

/cc @Random-Liu 


pkg/server/container_create.go
```diff
@@ -214,6 +214,8 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 		// TODO(random-liu): [P1] Set selinux options.
 
 		// TODO(random-liu): [P2] Add apparmor and seccomp.
+
+		g.SetProcessNoNewPrivileges(securityContext.GetNoNewPrivs())
```
Random-Liu: Please add unit test for this. You could take the existing unit test as a reference. :)

Random-Liu:
The PR LGTM overall, please add unit test.

One question I have: Should we set no new privilege for sandbox container by default?

miaoyq:
@Random-Liu 
I have no idea :disappointed: 
But I think that if sandbox container need be set by default, k8s should give instructions clearly for this.
Also, sandbox container is simple enough, and it should be security.
Just my own opinion :-)

pkg/server/container_create.go
miaoyq: Thanks for review,  will do! : )

miaoyq: @Random-Liu Done!

pkg/server/container_create.go
```diff
@@ -214,6 +214,9 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 		// TODO(random-liu): [P1] Set selinux options.
 
 		// TODO(random-liu): [P2] Add apparmor and seccomp.
+
+		// TODO: Figure out weather we should set no new privilege for sandbox container by default
```
Random-Liu: s/weather/whether.

pkg/server/container_create_test.go
```diff
@@ -146,6 +147,9 @@ func getCreateContainerTestData() (*runtime.ContainerConfig, *runtime.PodSandbox
 		assert.Contains(t, spec.Process.User.AdditionalGids, uint32(1111))
 		assert.Contains(t, spec.Process.User.AdditionalGids, uint32(2222))
 
+		t.Logf("Check NoNewPrivs")
```
Random-Liu: nit: Check no_new_privs.

This is linux term, and also lower case which is consistent with other logs.

Just a small nit. :)

Random-Liu:
@miaoyq LGTM other than a typo. Please fix and I'll apply the label and merge.

pkg/server/container_create.go
miaoyq: Done!

pkg/server/container_create_test.go
miaoyq: Done

miaoyq:
@Random-Liu  Thanks for your review!  Updated

Random-Liu:
LGTM.
Will merge after test passes.

PR #156, opened 2017-08-23
Checkpoint and restart recovery

### LINKED ISSUE #120: Checkpoint and restart recovery
There are several restart recovery problems with current cri-containerd:
1) cri-containerd restart. Because cri-containerd maintains all internal state in-memory, including sandbox list, container list and image list, once restarted all state will be lost.
2) containerd restart. When containerd restart and reconnect, there may be state mismatch between containerd and cri-containerd, e.g. a container dies during containerd is down.

To fix this, we should recover/reconcile state during cri-containerd start or after containerd restart and reconnect.

There are 3 kinds of internal state:
1) Image list. Containerd has all the information we need, we just need to list images from containerd and recover the image list.
2) Sandbox/container metadata: Most of the metadata is not provided by containerd, we need to checkpoint them for restart recovery. However, because metadata is constant, we could save it into containerd container label so as to leverage containerd metadata store to save it for us.
3) Container status: Container status is not persisted by containerd, we need to persist it ourselves. And because it's constantly changing, we may not want to abuse containerd container label to save it. So we need to maintain its checkpoint ourselves.

yanxuean:
fix #120 

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -118,7 +118,18 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 		}
 	}()
 
-	opts = append(opts, containerd.WithSpec(spec), containerd.WithRuntime(defaultRuntime))
+	metaBytes, err := meta.Encode()
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert sandbox metadata: %+v, %v", meta, err)
+	}
+	labels := map[string]string{
+		containerMetadataLabel: string(metaBytes),
+	}
+
+	opts = append(opts,
+		containerd.WithSpec(spec),
+		containerd.WithRuntime(defaultRuntime),
+		containerd.WithContainerLabels(labels))
 	var cntr containerd.Container
 	if cntr, err = c.client.NewContainer(ctx, id, opts...); err != nil {
 		return nil, fmt.Errorf("failed to create containerd container: %v", err)
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -95,6 +95,13 @@ const (
 	resolvConfPath = "/etc/resolv.conf"
 )
 
+const (
+	// sandboxMetadataLabel is label name that identify metadata of sandbox in CreateContainerRequest
+	sandboxMetadataLabel = "io.cri-containerd.sandbox.metadata"
+	// sandboxMetadataLabel is label name that identify metadata of container in CreateContainerRequest
+	containerMetadataLabel = "io.cri-containerd.container.metadata"
+)
+
 // generateID generates a random unique id.
 func generateID() string {
 	return stringid.GenerateNonCryptoID()
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -84,10 +84,26 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 		return nil, fmt.Errorf("failed to generate sandbox container spec: %v", err)
 	}
 	glog.V(4).Infof("Sandbox container spec: %+v", spec)
-	// TODO(random-liu): Checkpoint metadata into container labels.
+
+	// Checkpoint metadata into container
+	// TODO(random-liu): Switch to extensions(after merge containerd #1378).
+	// Actually the Pid, NetNS and CreatedAt underlying are not included here.
+	// I'm fine with this for now. After this PR is merged, we could:
+	// 1.Get Pid from containerd, so that we don't need to checkpoint it ourselves;
+	// 2.Use permanent NetNS after #138 is merged, so that we'll have network namespace here;
+	// 3.Get CreatedAt from containerd, which have been checkpointed by containerd
+	// https://github.com/containerd/containerd/blob/master/containers/containers.go#L14.
+	metaBytes, err := sandbox.Metadata.Encode()
+	if err != nil {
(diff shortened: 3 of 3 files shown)


yanxuean:
@Random-Liu   PTAL

pkg/server/container_create.go
```diff
@@ -118,7 +119,15 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 		}
 	}()
 
-	opts = append(opts, containerd.WithSpec(spec), containerd.WithRuntime(defaultRuntime))
+	b,err := meta.Encode()
```
Random-Liu: add space before `err`.

Random-Liu: Let's use a little bit longer var name, e.g. `metaBytes`.

pkg/server/container_create.go
```diff
@@ -118,7 +119,15 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert sandbox metadata: %+v, %v", meta, err)
+	}
+	labels[containerMetadataLabel] = string(b)
```
Random-Liu: Why make labels so far away? :) Just:
```go
labels := map[string]string {
  containerMetadataLabel: string(metaBytes),
}
```

pkg/server/container_create.go
```diff
@@ -118,7 +119,15 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
+
+
+	opts = append(opts, containerd.WithSpec(spec), containerd.WithRuntime(defaultRuntime),
```
Random-Liu: new line for each `WithXXX` if you've already added new line.

pkg/server/helpers.go
```diff
@@ -52,6 +52,9 @@ const (
 	errorExitReason = "Error"
 	// oomExitReason is the exit reason when process in container is oom killed.
 	oomExitReason = "OOMKilled"
+
+	sandboxMetadataLabel ="io.cri-containerd.sandbox.metadata"
```
Random-Liu: Add comment to describe what the constant is used for.

pkg/server/helpers.go
```diff
@@ -52,6 +52,9 @@ const (
...
+	containerMetadataLabel ="io.cri-containerd.container.metadata"
```
Random-Liu: ditto.

pkg/server/sandbox_run.go
```diff
@@ -86,8 +87,15 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	glog.V(4).Infof("Sandbox container spec: %+v", spec)
 	// TODO(random-liu): Checkpoint metadata into container labels.
 
+	b,err := sandbox.Metadata.Encode()
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert sandbox metadata: %+v, %v", sandbox.Metadata, err)
+	}
+	labels[sandboxMetadataLabel] = string(b)
```
Random-Liu: Same comments with container_create.go

Random-Liu: Actually the `Pid`, `NetNS` and `CreatedAt` underlying are not included here.

I'm fine with this for now. After this PR is merged, we could:
1) Get `Pid` from containerd, so that we don't need to checkpoint it ourselves;
2) Use permanent `NetNS` after https://github.com/kubernetes-incubator/cri-containerd/pull/138 is merged, so that we'll have network namespace here;
3) Get `CreatedAt` from containerd, which have been checkpointed by containerd https://github.com/containerd/containerd/blob/master/containers/containers.go#L14.

Please add a TODO for this.

pkg/server/sandbox_run.go
```diff
@@ -86,8 +87,15 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	glog.V(4).Infof("Sandbox container spec: %+v", spec)
 	// TODO(random-liu): Checkpoint metadata into container labels.
 
+	b,err := sandbox.Metadata.Encode()
```
Random-Liu: space.

Random-Liu:
Please run `make verify`, your code doesn't pass gofmt.

Random-Liu:
@yanxuean And this PR doesn't completely fix #120. We should not close #120 after this is merged.

pkg/server/sandbox_run.go
Random-Liu: And please add TODO to switch to extension.

pkg/server/container_create.go
yanxuean: If  other will  add new label in future, It has not affect

pkg/server/helpers.go
yanxuean: Are the place and the name ok?  I'm not sure if it should be here

Random-Liu: Probably add another const section? Or put them into the second const section. :)

pkg/server/container_create.go
Random-Liu: Even we add new label, I prefer group the logic together, so that we know what labels are added at a glance.

pkg/server/sandbox_run.go
yanxuean: Your memory is perfect.

pkg/server/container_create.go
yanxuean: ok

yanxuean:
I see.  Thanks for your  review

Random-Liu:
LGTM
@yanxuean Thanks a lot for this! :)

PR #157, opened 2017-08-23
Support selinux options/label

miaoyq:
Support selinux optios/label

/cc @Random-Liu

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -27,7 +27,9 @@ import (
 	"github.com/opencontainers/runc/libcontainer/devices"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/selinux/go-selinux/label"
 	"golang.org/x/net/context"
+	"golang.org/x/sys/unix"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
 	cio "github.com/kubernetes-incubator/cri-containerd/pkg/server/io"
@@ -234,9 +236,17 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 	}
 
 	securityContext := config.GetLinux().GetSecurityContext()
+	selinuxOpt := securityContext.GetSelinuxOptions()
+	processLabel, mountLabel, err := initSelinuxOpts(selinuxOpt)
+	if err != nil {
+		return nil, fmt.Errorf("failed to init selinux options %+v: %v", securityContext.GetSelinuxOptions(), err)
+	}
 
 	// Add extra mounts first so that CRI specified mounts can override.
-	addOCIBindMounts(&g, append(extraMounts, config.GetMounts()...))
+	mounts := append(extraMounts, config.GetMounts()...)
+	if err := addOCIBindMounts(&g, mounts, mountLabel); err != nil {
+		return nil, fmt.Errorf("failed to set OCI bind mounts %+v: %v", mounts, err)
+	}
 
 	if securityContext.GetPrivileged() {
 		if !sandboxConfig.GetLinux().GetSecurityContext().GetPrivileged() {
@@ -258,6 +268,9 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 		// TODO(random-liu): [P2] Add apparmor and seccomp.
 
 	}
+
+	g.SetProcessSelinuxLabel(processLabel)
+	g.SetLinuxMountLabel(mountLabel)
 
 	// TODO: Figure out whether we should set no new privilege for sandbox container by default
 	g.SetProcessNoNewPrivileges(securityContext.GetNoNewPrivs())
@@ -435,7 +448,7 @@ func setOCIDevicesPrivileged(g *generate.Generator) error {
 // addOCIBindMounts adds bind mounts.
 // TODO(random-liu): Figure out whether we need to change all CRI mounts to readonly when
 // rootfs is readonly. (https://github.com/moby/moby/blob/master/daemon/oci_linux.go)
-func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount) {
+func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount, mountLabel string) error {
 	// Mount cgroup into the container as readonly, which inherits docker's behavior.
 	g.AddCgroupsMount("ro") // nolint: errcheck
 	for _, mount := range mounts {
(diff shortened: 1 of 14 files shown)


Random-Liu:
@miaoyq Thanks! We do need this, but haven't got time to work on it. Will review it.

Random-Liu:
Also assign @mikebrow because he's recently working on security related stuff. :)

Random-Liu:
@miaoyq Do you have a selinux environment? Have you tested this? :)

miaoyq:
@Random-Liu Not yet:pensive:, I will add this part

mikebrow:
I'm rooting for ya!

crosbymichael:
I think the approach is good.  

Labels should be generated on the CRI side because you will have to make sure the mount labels match up for various containers in a pod.

I'll add a couple of helper Opts in containerd for setting the labels but I think this is the right approach by generating them here.

pkg/server/container_create.go
```diff
@@ -27,10 +27,12 @@ import (
 	"github.com/opencontainers/runc/libcontainer/devices"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/opencontainers/selinux/go-selinux/label"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
 	containerstore "github.com/kubernetes-incubator/cri-containerd/pkg/store/container"
+	"syscall"
```
Random-Liu: Use `golang.org/x/sys/unix` and move this to the first group, `syscall` is deprecated.

pkg/server/container_create.go
```diff
@@ -399,9 +410,16 @@ func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount) {
 		if mount.GetReadonly() {
 			options = []string{"ro"}
 		}
-		// TODO(random-liu): [P1] Apply selinux label
+
+		if mount.GetSelinuxRelabel() {
+			if err := label.Relabel(src, mountLabel, true); err != nil && err != syscall.ENOTSUP {
+				return fmt.Errorf("relabel failed %s: %v", src, err)
```
Random-Liu: also include `mountlabel` in the error message.

pkg/server/container_create.go
```diff
@@ -194,8 +196,16 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 
 	securityContext := config.GetLinux().GetSecurityContext()
 
+	processLabel, mountLabel, err := initSelinuxOpts(securityContext.GetSelinuxOptions())
```
Random-Liu: nit: Create a local variable for `securityContext.GetSelinuxOptions()`, so that we could make the next line shorter. :p

pkg/server/container_create.go
```diff
@@ -211,7 +221,8 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 				securityContext.GetCapabilities(), err)
 		}
 
-		// TODO(random-liu): [P1] Set selinux options.
```
Random-Liu: I think we should also set this for privileged container, let's keep it consistent with docker. https://github.com/moby/moby/blob/master/daemon/oci_linux.go#L831-L834

pkg/server/helpers.go
```diff
@@ -391,3 +392,18 @@ func resolveSymbolicLink(path string) (string, error) {
 	}
 	return filepath.EvalSymlinks(path)
 }
+
+func initSelinuxOpts(selinuxOpt *runtime.SELinuxOption) (string, string, error) {
+	var labelOpts []string
```
Random-Liu: nit: Move this down.

pkg/server/sandbox_run.go
```diff
@@ -263,7 +263,15 @@ func (c *criContainerdService) generateSandboxContainerSpec(id string, config *r
 		g.RemoveLinuxNamespace(string(runtimespec.IPCNamespace)) // nolint: errcheck
 	}
 
-	// TODO(random-liu): [P1] Apply SeLinux options.
+	securityContext := config.GetLinux().GetSecurityContext()
+
```
Random-Liu: nit: unnecessary empty line.

pkg/server/sandbox_run.go
```diff
@@ -263,7 +263,15 @@ func (c *criContainerdService) generateSandboxContainerSpec(id string, config *r
...
+	processLabel, mountLabel, err := initSelinuxOpts(securityContext.GetSelinuxOptions())
+	if err != nil {
+		return nil, fmt.Errorf("failed to init selinux options %+v: %v", securityContext.GetSelinuxOptions(), err)
+	}
+
```
Random-Liu: nit: ditto.

pkg/server/helpers.go
Random-Liu: And this seems cleaner https://github.com/opencontainers/selinux/blob/63bd19516561f1973f86c9f777b263ee1e0db140/go-selinux/selinux.go#L572

label.InitLabels(label.DupSecOpt("user:role:type:level")).

pkg/server/container_create.go
Random-Liu: Please add unit test for this. Basically, just set selinux options, and check whether `ProcessSelinuxLabel` and `LinuxMountLabel` are set correctly.

We don't need to add `SelinuxRelabel` for any mount, so that `Relabel` won't be actually run in the unit test.

Random-Liu:
@miaoyq Please rebase the PR and address comments.

We don't have selinux test and test environment for CRI validation test.
We do have cluster e2e test for it, but we are not running in any of our own environment https://github.com/kubernetes/kubernetes/blob/master/test/e2e/node/security_context.go#L98-L108.

If possible, could you try to verify this in a selinux environment? If it's hard for you, please file an issue, and we could get back to verify this later.

miaoyq:
@Random-Liu  Thanks for review!
Sorry for my late reply, I was busy with other things.
Now I'm free. I'll rebase the PR and address comments.
It's hard for you to verify this actually, but I have built a selinux environment, if it is not urgent, I would like to complete this verification with your guidance. :-)

Random-Liu:
@miaoyq Thanks! Feel free to comment, file issue or ping me if you encountered any problem on that distro.

miaoyq:
@Random-Liu  OK, I will, thanks!

pkg/server/container_create.go
miaoyq: @Random-Liu  I have tried to add unit test for this, but the result of test depends on whether 
selinux have been enabled, see https://github.com/opencontainers/selinux/blob/4a2974bf1ee960774ffd517717f1f45325af0206/go-selinux/label/label_selinux.go#L28.

miaoyq: @Random-Liu I have add the unit test with build tag `selinux`.

miaoyq:
@Random-Liu  Rebased and addressed comments, PTAL

Random-Liu:
LGTM. @miaoyq Please help validate this after merged. Thanks! :)

PR #163, opened 2017-08-24
Updating to container1.0-alpha

abhinandanpb:
Updating to containerd 1.0 alpha6 with minor changes.


pkg/server/container_execsync.go
```diff
@@ -140,40 +132,15 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
-		// Wait will return if context is cancelled.
-		exitCode, err := process.Wait(ctx)
-		resCh <- execResult{
-			exitCode: exitCode,
-			err:      err,
-		}
-		glog.V(2).Infof("Exec process %q exits with exit code %d and error %v", execID, exitCode, err)
-	}()
+	// Wait will return if context is cancelled.
+	exitCh, err := process.Wait(ctx)
```
Random-Liu: Why remove the timeout?

Random-Liu: Why remove the timeout? We still need the timeout.

pkg/server/container_execsync.go
```diff
@@ -140,40 +132,15 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait for process")
```
Random-Liu: Include error.

pkg/server/container_execsync.go
```diff
@@ -127,9 +119,7 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 		return nil, fmt.Errorf("failed to create exec %q: %v", execID, err)
 	}
 	defer func() {
-		// TODO(random-liu): There is a containerd bug here containerd#1376, revisit this
-		// after that is fixed.
-		if _, err := process.Delete(ctx); err != nil {
+		if _, err := process.Delete(ctx, containerd.WithProcessKill); err != nil {
```
Random-Liu: Should not `WithProcessKill` here, it will fail and skip delete if `Exec` is not started yet.

pkg/server/container_execsync.go
```diff
@@ -127,8 +120,6 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 		return nil, fmt.Errorf("failed to create exec %q: %v", execID, err)
 	}
 	defer func() {
-		// TODO(random-liu): There is a containerd bug here containerd#1376, revisit this
```
Random-Liu: Thanks for removing this! :) Just about to do that.

pkg/server/container_execsync.go
```diff
@@ -140,40 +131,33 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
+	case <-ctx.Done():
+		// If parent context cancel, the ctx done chanel would be closed as well.
 		// Ignore the not found error because the process may exit itself before killing.
-		if err := process.Kill(ctx, unix.SIGKILL); err != nil && !errdefs.IsNotFound(err) {
+		if _, err := process.Delete(ctx, containerd.WithProcessKill); err != nil && !errdefs.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to kill exec %q: %v", execID, err)
 		}
 		// Wait for the process to be killed.
-		<-resCh
+		<-exitCh
```
Random-Liu: Can we add a log here explain there is a timeout, and the exitCode, error we get here.
I think V(2) should be good enough.

I should have done that before. Thanks!

pkg/server/container_execsync.go
```diff
@@ -140,40 +131,33 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
 		return nil, fmt.Errorf("timeout %v exceeded", opts.timeout)
-	case res := <-resCh:
-		if res.err != nil {
-			return nil, fmt.Errorf("failed to wait for exec %q: %v", execID, res.err)
+	case exitRes := <-exitCh:
+		code, _, err := exitRes.Result()
```
Random-Liu: Can we add the original log back?
```
glog.V(2).Infof("Exec process %q exits with exit code %d and error %v", execID, exitCode, err)
```

pkg/server/container_execsync.go
```diff
@@ -140,40 +131,33 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
 	}
 
-	var timeoutCh <-chan time.Time
-	if opts.timeout == 0 {
-		// Do not set timeout if it's 0.
-		timeoutCh = make(chan time.Time)
-	} else {
-		timeoutCh = time.After(opts.timeout)
+	if opts.timeout != 0 {
+		ctx, cancel = context.WithTimeout(ctx, opts.timeout)
```
Random-Liu: We should not use context timeout here. Once the context is cancelled, you'll not even be able to call `Delete`. It means that once there is a timeout, there will be resource leakage, because we won't be able to delete the process.

Although we could pass `context.Background()` to `Delete`, I prefer keeping the original timeout mechanism. :)

pkg/server/container_execsync.go
```diff
@@ -140,40 +131,33 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
-		exitCode, err := process.Wait(ctx)
-		resCh <- execResult{
-			exitCode: exitCode,
-			err:      err,
-		}
-		glog.V(2).Infof("Exec process %q exits with exit code %d and error %v", execID, exitCode, err)
-	}()
+	exitCh, err := process.Wait(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait for process %v", err)
```
Random-Liu: Let's include the execID here (reference the other error). We should include it in the error here because caller doesn't know it.

We don't need to include container id or sandbox id, because caller already knows that.

Random-Liu: And let's make the error different from the one below, so that once there is an error, we know which lines feels when debugging.

How about:
* Here: `failed to start waiting for exec %q: %v`
* Below: `failed when waiting for exec %q: %v`

Something like this, your English is better than mine. :)

pkg/server/container_execsync.go
```diff
@@ -140,40 +131,33 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
...
+		ctx, cancel = context.WithTimeout(ctx, opts.timeout)
+		defer cancel()
 	}
 	select {
-	case <-timeoutCh:
+	case <-ctx.Done():
+		// If parent context cancel, the ctx done chanel would be closed as well.
 		// Ignore the not found error because the process may exit itself before killing.
-		if err := process.Kill(ctx, unix.SIGKILL); err != nil && !errdefs.IsNotFound(err) {
+		if _, err := process.Delete(ctx, containerd.WithProcessKill); err != nil && !errdefs.IsNotFound(err) {
```
Random-Liu: Let's only kill the process here. If we `Delete` it here, the `Delete` in defer will always log error, which is annoying and confusing. :)

pkg/server/container_execsync.go
abhinandanpb: If the context is cancelled then the ctx.Done() channel will be closed which will enable the Delete as well. As explained here the ctx.Done() is closed when the parent context cancels or the deadline expires. https://golang.org/pkg/context/#WithDeadline.
If you are talking about Delete() in defer that should be invoked too.

Random-Liu: Delete will return error immediately if context is cancelled, that's how grpc works. You could try it. :)

abhinandanpb: ok will add a TODO :) I want to check the issue.

Random-Liu:
LGTM

PR #171, opened 2017-08-25
Do not checkpoint sandbox pid.

Random-Liu:
Address TODOs added in https://github.com/kubernetes-incubator/cri-containerd/pull/156.

* On one hand, we don't need to checkpoint sandbox pid, because we use sandbox when it's running, and we'll always be able to its pid from containerd.
* On the other hand, if we want to use containerd label to checkpoint sandbox pid, we have to update the label. Because we usually create label when create container, but we'll only be able to get pid after start task.

/cc @yanxuean 


yanxuean:
The Metadata.NetNSPath  is associated with pid.  Shouldn't  we remove it?
@Random-Liu

PR #174, opened 2017-08-28
Do not teardown when network namespace is removed already.

Random-Liu:
@abhinandanpb 


abhinandanpb:
LGTM. Thanks for taking care if it. It will avoid unwanted errors when ```StopPodSandbox``` is called multiple times :)

Random-Liu:
@abhinandanpb Yeah. Thanks for reviewing! ;0

Issue #173, opened 2017-08-26
Handle the Events about sandbox container from containerd 


miaoyq:
Now we only handle the events about app container, I think we should handle the events about the sandbox container as well.

Random-Liu:
@miaoyq We don't handle events for sandbox container, because we don't need to maintain its state inside cri-containerd.

We purely rely on the existence of the corresponding containerd task: if it's running, sandbox container is ready; if not, then not ready. So every time `PodSandboxStatus` is called, we get real time information from containerd, and generate the sandbox status.

That's why we don't maintain the state of the sandbox container, and that's also why we don't care about its event.

miaoyq:
@Random-Liu  I understand, Thanks! : )
We should exclude the event sandbox containers from event stream in order to avoid outputting unexpected error print.

Random-Liu:
Close this one since #177 is merged.

PR #177, opened 2017-08-28
Exclude the event of sandbox containers from event stream

miaoyq:
We should exclude the event of sandbox containers from event stream in order to avoid outputting unexpected error print.

related #173

/cc @Random-Liu 


Random-Liu:
@miaoyq Thanks! This gets rid of the confusing log output.

LGTM

PR #179, opened 2017-08-29
Checkpoint container status onto disk.

Random-Liu:
Checkpoint container status onto disk. With this we could recover container state from container metadata store and status checkpoint during restart.


abhinandanpb:
LGTM

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/179#issuecomment-327595811.

PR #183, opened 2017-08-29
Cri containerd exits with containerd

Random-Liu:
Based on https://github.com/kubernetes/kubernetes/pull/51377.

This PR combines `pkg/server/service.go` and `pkg/server/server.go` and most importantly lets cri-containerd exit if containerd is not running (event channel disconnects). We do this because:
1) cri-containerd could do nothing if containerd is not running, most requests will failure, `ListContainers` may still work, but it's returning staled data;
2) It's tricky to handle reconnection with containerd, because grpc server is still running. There could still be new requests during state recovery, so we have to lock the whole system to do state recovery. Given so, it's simpler to just let cri-containerd exit with containerd, so that we'll only need one state recovery logic before starting serving traffic.

We could handle reconnection in the future, but for now, this seems to be an easier temporary solution.

@kubernetes-incubator/maintainers-cri-containerd @crosbymichael Any thoughts?


crosbymichael:
As long as kube handles reconnecting to the CRI when it goes away I think this would work.

Random-Liu:
> As long as kube handles reconnecting to the CRI when it goes away I think this would work.

@crosbymichael Kubelet does so. :) Kubelet will only report that the node is not ready, but will come back normal if cri-containerd and containerd come back in time.

crosbymichael:
@Random-Liu cool, let them do the hard work ;)

pkg/server/service.go
```diff
@@ -123,14 +135,79 @@ func NewCRIContainerdService(containerdEndpoint, rootDir, networkPluginBinDir, n
...
+	// Stop the whole cri-containerd service if any of the critical service exits.
+	select {
+	case <-eventMonitorClose:
+		c.Stop()
+	case <-streamServerClose:
+		c.Stop()
+	case <-grpcServerClose:
+		c.Stop()
+	}
+	<-eventMonitorClose
```
abhinandanpb: For my understanding - why are we waiting in select and here as well ? Wont select block on one of the channels to close ?

pkg/server/service.go
```diff
@@ -123,14 +135,79 @@ func NewCRIContainerdService(containerdEndpoint, rootDir, networkPluginBinDir, n
...
+	l, err := net.Listen(unixProtocol, c.serverAddress)
+	if err != nil {
+		return fmt.Errorf("failed to listen on %q: %v", c.serverAddress, err)
+	}
+	grpcServerClose := make(chan struct{})
+	go func() {
+		if err := c.server.Serve(l); err != nil {
+			glog.Errorf("Failed to serve grpc grpc request: %v", err)
+		}
+		close(grpcServerClose)
```
abhinandanpb: if we close - won't the criContainerdService stop below?

Random-Liu: Yeah, that's what we want.

We want criContainerdService to exit when a critical service is down.

pkg/server/service.go
Random-Liu: The select means that: "If any of the critical service down, we should stop and exit"
The following 3 selects make sure that all critical services are properly stopped before exit.

pkg/server/events.go
```diff
@@ -17,72 +17,69 @@ limitations under the License.
...
-func (c *criContainerdService) startEventMonitor() {
-	b := backoff.Backoff{
-		Min:    minRetryInterval,
-		Max:    maxRetryInterval,
-		Factor: exponentialFactor,
+type eventMonitor struct {
+	c           *criContainerdService
+	eventstream events.Events_SubscribeClient
+	cancel      context.CancelFunc
+	close       chan struct{}
```
abhinandanpb: nit: can we rename to close to closeChan or something on those lines.

pkg/server/service.go
abhinandanpb: @Random-Liu one question. Is there any code path that would have closed the channel already ? Also is there a code path where we would end up being blocked on c.server.Serve ? Just wondering if this go routine could get stuck and also ensuring we dont double close the channel.

pkg/server/events.go
Random-Liu: Will do.

pkg/server/service.go
Random-Liu: We only close the channel here, we don't even pass it to outside of the function. :)

Random-Liu:
@abhinandanpb Addressed comments. Please take a look.

abhinandanpb:
LGTM

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/183#issuecomment-326703249.

Will merge after test passes.

Random-Liu:
Ha, I forgot there is a commit update vendor directly. I'll update kubernetes version now to get rid of that.

PR #184, opened 2017-08-29
Adding option to configure cgroup to start cri-containerd

abhinandanpb:
Manual test output:

```
ps -aux | grep cri-containerd
root     26047  7.0  0.7 224224 28872 pts/1    Sl+  15:11   0:00 ./cri-containerd --cgroup-path /containerd

$ cat /proc/26047/cgroup
11:net_cls,net_prio:/containerd
10:perf_event:/containerd
9:pids:/containerd
8:hugetlb:/containerd
7:memory:/containerd
6:freezer:/containerd
5:blkio:/containerd
4:devices:/containerd
3:cpuset:/containerd
2:cpu,cpuacct:/containerd
1:name=systemd:/containerd
```

### DIFF
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -24,12 +24,14 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/containerd/cgroups"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/docker/distribution/reference"
 	imagedigest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/identity"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
@@ -351,4 +353,24 @@ func resolveSymbolicLink(path string) (string, error) {
 		return path, nil
 	}
 	return filepath.EvalSymlinks(path)
+}
+
+// loadCgroup loads the cgroup associated with path if it exists and moves the current process into the cgroup. If the cgroup
+// is not created it is created and returned.
+func loadCgroup(cgroupPath string) (cgroups.Cgroup, error) {
+	cg, err := cgroups.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
+	if err != nil {
+		if err != cgroups.ErrCgroupDeleted {
+			return nil, err
+		}
+		if cg, err = cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupPath), &specs.LinuxResources{}); err != nil {
+			return nil, err
+		}
+	}
+	if err := cg.Add(cgroups.Process{
+		Pid: os.Getpid(),
+	}); err != nil {
+		return nil, err
+	}
+	return cg, nil
 }
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -84,6 +84,8 @@ type criContainerdService struct {
 	client *containerd.Client
 	// streamServer is the streaming server serves container streaming request.
 	streamServer streaming.Server
+	// cgroupPath in which the cri-containerd is placed in
+	cgroupPath string
 }
 
 // NewCRIContainerdService returns a new instance of CRIContainerdService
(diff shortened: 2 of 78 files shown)


pkg/server/service.go
```diff
@@ -82,18 +85,37 @@ type criContainerdService struct {
 	client *containerd.Client
 	// streamServer is the streaming server serves container streaming request.
 	streamServer streaming.Server
+	// cgroup in which the cri-containerd is placed in
+	cgroup cgroups.Cgroup
```
Random-Liu: Seems like that we only need the path (as the backup for parent cgroup). Can we only save the path?

pkg/server/service.go
```diff
@@ -82,18 +85,37 @@ type criContainerdService struct {
...
 		return nil, fmt.Errorf("failed to initialize containerd client with endpoint %q: %v", containerdEndpoint, err)
 	}
+	if cgroupPath != "" {
+		cg, err = cgroups.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
+		if err != nil {
+			if err != cgroups.ErrCgroupDeleted {
+				return nil, err
+			}
+			if cg, err = cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupPath), &specs.LinuxResources{}); err != nil {
+				return nil, err
```
Random-Liu: ditto.

pkg/server/service.go
```diff
@@ -82,18 +85,37 @@ type criContainerdService struct {
...
+			}
+		}
+		if err := cg.Add(cgroups.Process{
+			Pid: os.Getpid(),
```
Random-Liu: ditto.

pkg/server/service.go
```diff
@@ -82,18 +85,37 @@ type criContainerdService struct {
...
+	streamAddress, streamPort, cgroupPath string) (CRIContainerdService, error) {
 	// TODO(random-liu): [P2] Recover from runtime state and checkpoint.
 
+	var cg cgroups.Cgroup
 	client, err := containerd.New(containerdEndpoint, containerd.WithDefaultNamespace(k8sContainerdNamespace))
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize containerd client with endpoint %q: %v", containerdEndpoint, err)
 	}
+	if cgroupPath != "" {
+		cg, err = cgroups.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
```
Random-Liu: nit: Can we move this logic into a separate function? Just keep the code cleaner.

vendor.conf
```diff
@@ -3,6 +3,8 @@ github.com/boltdb/bolt v1.3.0-58-ge9cf4fa
 github.com/containerd/containerd cf09e32618398fc59fcb45bcfe9b4c0335972733
 github.com/containerd/continuity cf279e6ac893682272b4479d4c67fd3abf878b4e
 github.com/containerd/fifo fbfb6a11ec671efbe94ad1c12c2e98773f19e1e6
+github.com/containerd/cgroups 7a5fdd8330119dc70d850260db8f3594d89d6943
```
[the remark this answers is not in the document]
abhinandanpb: cgroups doesn't vendor dependency. In such cases we will need to manually vendor the external dependencies. 
I think it should be ok. But I can  check it to match with containerd.

pkg/server/service.go
abhinandanpb: will do so

pkg/server/service.go
```diff
@@ -82,18 +85,37 @@ type criContainerdService struct {
...
+		if err != nil {
+			if err != cgroups.ErrCgroupDeleted {
+				return nil, err
```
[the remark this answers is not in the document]
abhinandanpb: will update

Random-Liu:
LGTM overall with nits.

abhinandanpb:
@Random-Liu addressed comments. PTAL.

pkg/server/helpers.go
```diff
@@ -352,3 +354,23 @@ func resolveSymbolicLink(path string) (string, error) {
 	}
 	return filepath.EvalSymlinks(path)
 }
+
+// LoadCgroup loads the cgroup associated with path if it exits. If the cgroup
+// is not created it is created and returned.
+func LoadCgroup(cgroupPath string) (cgroups.Cgroup, error) {
```
Random-Liu: s/ LoadCgroup loads the cgroup associated with path if it exits/ loadCgroup loads the cgroup associated with path if it exits, and move the current process into the cgroup.

nit: Make it private if we don't want to expose it.

Random-Liu:
LGTM. Will merge after test passes.

Random-Liu:
LGTM. Will merge after test passes.

abhinandanpb:
@Random-Liu Cool! Thanks for reviewing

Issue #185, opened 2017-08-30
Support mount propagation.


Random-Liu:
We should support mount propagation after CRI supports it https://github.com/kubernetes/kubernetes/pull/46444.

Random-Liu:
kubernetes/kubernetes#46444 is going to be merged soon. We should update kubernetes version and support it after it is merged.

Random-Liu:
The Kubernetes PR has been merged. We should add the mount propagation support.

Feel free to grab this one if you are interested @yanxuean @miaoyq. :)

yanxuean:
I don't understand this PR, Can you give me more detail.

miaoyq:
@Random-Liu @yanxuean  I will work for this.

miaoyq:
@Random-Liu  First of all, we should sync `k8s.io/kubernetes/staging/src/k8s.io/api` to `k8s.io/api` project.
Cloud you help me sync this?  I don't have permission.

Random-Liu:
@miaoyq Hm, their bot is broken, we may have to wait for them to sync.

Random-Liu:
@miaoyq @yanxuean I've sent Collaborators invitation to you, so that I could assign issue to you. :)

miaoyq:
@Random-Liu  Looks like we don't need to update `api`, the change is used in `k8s`.

I have created a pr for this.

PR #187, opened 2017-08-30
Use rbind and rprivate in bind mount.

Random-Liu:
Use `rbind` and `rpivate` option for bind mount.

The QoS test mount host cgroup inside the container, and check whether QoS cgroups are created as expected. Without `rbind`, this won't work.

Now with this PR we could run QoS test.

Also ref https://github.com/kubernetes-incubator/cri-containerd/issues/185.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -438,9 +438,11 @@ func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount) {
 	for _, mount := range mounts {
 		dst := mount.GetContainerPath()
 		src := mount.GetHostPath()
-		options := []string{"rw"}
+		options := []string{"rbind", "rprivate"}
 		if mount.GetReadonly() {
-			options = []string{"ro"}
+			options = append(options, "ro")
+		} else {
+			options = append(options, "rw")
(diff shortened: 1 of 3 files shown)


pkg/server/container_create.go
```diff
@@ -438,9 +438,11 @@ func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount) {
 	for _, mount := range mounts {
 		dst := mount.GetContainerPath()
 		src := mount.GetHostPath()
-		options := []string{"rw"}
+		options := []string{"rbind", "rprivate"}
```
mikebrow: I feel like these options, rbind (recursive) and rprivate (isolated), should be be derived through the CRI create request, possibly also the pod. Is this the current default for dockershim? If so it makes sense to make it the default if there is no override from the kublet through CRI. Thoughts?

crosbymichael:
LGTM

`rbind` and `rprivate` are both default for docker volumes

abhinandanpb:
LGTM

Random-Liu:
Yeah, these are docker default. And we'll support mount propagation after CRI supports it #185.

PR #189, opened 2017-08-30
Update cri validation test to add supplemental groups test.

Random-Liu:
Update cri validation test to include supplemental groups test https://github.com/kubernetes-incubator/cri-tools/pull/127.


Random-Liu:
Thanks for reviewing!

Issue #136, opened 2017-08-16
Cleanup the image management code to use containerd client


Random-Liu:
After https://github.com/kubernetes-incubator/cri-containerd/pull/113, cri-containerd will be using containerd client.

However, there are still some image specific code needs to cleanup, e.g. `getImageInfo`, `localResolve` etc.

We should add required functions into containerd client, and clean those functions up.

/cc @kubernetes-incubator/maintainers-cri-containerd 
@abhinandanpb

PR #190, opened 2017-08-30
Cleanup image operations

Random-Liu:
Based on https://github.com/containerd/containerd/pull/1444.
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/136.

I prioritize this because this will also simplify checkpoint recovery code, it will be much cleaner to recover image information.

/cc @abhinandanpb


abhinandanpb:
I almost missed this. Need to keep better track of my TODOs. I would suggest we wait till the containerd PR gets merged. I have feeling the interface might be asked to be changed to Info() or something else.

Random-Liu:
@abhinandanpb No worry. Just found this useful when work on the checkpoint and recovery, thus do it btw.

PR #198, opened 2017-08-31
[DO NOT MERGE] Abuse travis to debug why exec fails on it.

Random-Liu:
exec tests are failing on travis. However, it works well in my environment.

So I pushed this PR to let travis run the test and collect log for me. Will close this PR after I figure out why the test doesn't pass on travis.


Random-Liu:
Close this one because #201 is merged.

Issue #196, opened 2017-08-31
gcr.io private registry doesn't work


Random-Liu:
Ref https://github.com/containerd/containerd/issues/1318.

We need to resolve this.

mikebrow:
yeah breaking all my e2e tests

Random-Liu:
The containerd side change has been merged https://github.com/containerd/containerd/pull/1465.

We should try our test again to see whether it's fixed.

PR #202, opened 2017-09-01
Fix repo digest for schema 1 image.

Random-Liu:
For schema1 image, we should not use it's manifest digest as pullable repo digest, because it is not pullable.

However, the current logic is wrong, because `image.Target().MediaType` has already been converted https://github.com/containerd/containerd/blob/master/remotes/docker/schema1/converter.go#L156.

We could only figure out whether the image is schema1 ourselves.

With this PR and https://github.com/kubernetes/kubernetes/pull/51775, we could pass the image digest node e2e test.

### DIFF
--- a/pkg/server/container_status.go
+++ b/pkg/server/container_status.go
@@ -32,13 +32,23 @@ func (c *criContainerdService) ContainerStatus(ctx context.Context, r *runtime.C
 		return nil, fmt.Errorf("an error occurred when try to find container %q: %v", r.GetContainerId(), err)
 	}
 
+	imageRef := container.ImageRef
+	image, err := c.imageStore.Get(imageRef)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get image %q: %v", container.ImageRef, err)
+	}
+	if len(image.RepoDigests) > 0 {
+		// Based on the CRI definition, this field will be consumed by user.
+		imageRef = image.RepoDigests[0]
+	}
+
 	return &runtime.ContainerStatusResponse{
-		Status: toCRIContainerStatus(container),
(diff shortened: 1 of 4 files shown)


Random-Liu:
After all important pending PRs merged, we'll only have 2 node e2e test failures:
* `querying\s\/stats\/summary`: Missing feature. Container stats and image filesystem stats required.
* `pull\sfrom\sprivate\sregistry\swith\ssecret`: Pull image from gcr private registry, tracked by issue https://github.com/kubernetes-incubator/cri-containerd/issues/196. GCR will roll out the fix this week, and we may still need to add work around in containerd for other registries (not sure yet).

@kubernetes-incubator/maintainers-cri-containerd /cc @crosbymichael @stevvooe @dmcgowan @mlaventure

Random-Liu:
@mikebrow @abhinandanpb Per discussion at https://github.com/kubernetes/kubernetes/pull/51775#issuecomment-326646967. I changed cri-containerd to return first repo digest in container status. Please take another look.

pkg/server/container_status.go
```diff
@@ -32,13 +32,23 @@ func (c *criContainerdService) ContainerStatus(ctx context.Context, r *runtime.C
...
 	}
 
+	imageRef := container.ImageRef
+	image, err := c.imageStore.Get(imageRef)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get image %q: %v", container.ImageRef, err)
+	}
+	if len(image.RepoDigests) > 0 {
+		// Based on the CRI definition, this field will be consumed by user.
+		imageRef = image.RepoDigests[0]
```
mikebrow: weird logic but it's the current CRI expectation so...

PR #204, opened 2017-09-02
Revert "Setting containerd shim to Pod cgroup"

Random-Liu:
Reverts kubernetes-incubator/cri-containerd#193

@abhinandanpb I have to revert this. This is breaking our node e2e test. We need to debug it.
https://travis-ci.org/kubernetes-incubator/cri-containerd/builds/270726614

I verified this myself, and kept getting [`failed to load cgroup`](https://github.com/containerd/containerd/blob/4291fb480320faf025c6e0200f55914e377427ea/linux/shim/client_linux.go#L21) error, seems like that pod cgroup is not ready when creating pod sandbox.

PR #205, opened 2017-09-02
Revert "Setting containerd shim cgroup same as pod cgroup"

Random-Liu:
This reverts commit 59008c608e90fdc0b57c468489a792bc56fdaa5b.


Random-Liu:
This is the same with #204.

#204 is not signed, because it's github generated.

abhinandanpb:
I will debug this. Please go ahead and merge it

PR #206, opened 2017-09-03
Use EnsureRemoveAll

Random-Liu:
Use `EnsureRemoveAll` to recursively umount and remove container and sandbox directory.

This function will also be useful during checkpoint recovery.


Random-Liu:
Will do.

Random-Liu:
@mikebrow Done.

PR #209, opened 2017-09-04
Checkpoint recovery

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/120.

Add recovery logic during restart. I've manually tested this, it's working even better than I thought. :)

Based on https://github.com/kubernetes-incubator/cri-containerd/pull/206 and https://github.com/kubernetes-incubator/cri-containerd/pull/179.

@kubernetes-incubator/maintainers-cri-containerd Please review this carefully, especially `pkg/server/restart.go`. This is really important to the production quality.

I'll send another PR to add our own integration test framework, and will add restart test.

### DIFF
--- a/pkg/server/restart.go
+++ b/pkg/server/restart.go
@@ -1,0 +1,421 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/typeurl"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/pkg/system"
+	"github.com/golang/glog"
+	"golang.org/x/net/context"
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+
+	cio "github.com/kubernetes-incubator/cri-containerd/pkg/server/io"
+	containerstore "github.com/kubernetes-incubator/cri-containerd/pkg/store/container"
+	imagestore "github.com/kubernetes-incubator/cri-containerd/pkg/store/image"
+	sandboxstore "github.com/kubernetes-incubator/cri-containerd/pkg/store/sandbox"
+)
+
+// NOTE: The recovery logic has following assumption: when cri-containerd is down:
+// 1) Files (e.g. root directory, netns) and checkpoint maintained by cri-containerd MUST NOT be
+// touched. Or else, recovery logic for those containers/sandboxes may return error.
+// 2) Containerd containers may be deleted, but SHOULD NOT be added. Or else, recovery logic
+// for the newly added container/sandbox will return error, because there is no corresponding root
+// directory created.
+// 3) Containerd container tasks may exit or be stoppped, deleted. Even though current logic could
+// tolerant tasks being created or started, we prefer that not to happen.
+
+// recover recovers system state from containerd and status checkpoint.
+func (c *criContainerdService) recover(ctx context.Context) error {
+	// Recover all sandboxes.
+	sandboxes, err := c.client.Containers(ctx, filterLabel(containerKindLabel, containerKindSandbox))
+	if err != nil {
+		return fmt.Errorf("failed to list sandbox containers: %v", err)
+	}
+	for _, sandbox := range sandboxes {
+		sb, err := loadSandbox(ctx, sandbox)
+		if err != nil {
+			glog.Errorf("Failed to load sandbox %q: %v", sandbox.ID(), err)
+			continue
+		}
+		glog.V(4).Infof("Loaded sandbox %+v", sb)
+		if err := c.sandboxStore.Add(sb); err != nil {
+			return fmt.Errorf("failed to add sandbox %q to store: %v", sandbox.ID(), err)
+		}
+	}
+
+	// Recover all containers.
+	containers, err := c.client.Containers(ctx, filterLabel(containerKindLabel, containerKindContainer))
+	if err != nil {
+		return fmt.Errorf("failed to list containers: %v", err)
+	}
+	for _, container := range containers {
+		containerDir := getContainerRootDir(c.config.RootDir, container.ID())
+		cntr, err := loadContainer(ctx, container, containerDir)
+		if err != nil {
+			glog.Errorf("Failed to load container %q: %v", container.ID(), err)
+			continue
+		}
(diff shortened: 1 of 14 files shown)


pkg/server/restart.go
```diff
@@ -0,0 +1,429 @@
...
+	metaBytes, ok := labels[sandboxMetadataLabel]
+	if !ok {
+		return sandbox, fmt.Errorf("metadata label %q not found", sandboxMetadataLabel)
+	}
+	if err := meta.Decode([]byte(metaBytes)); err != nil {
+		return sandbox, fmt.Errorf("failed to parse metadata %q: %v", metaBytes, err)
+	}
+
+	// Load network namespace.
+	netNS, err := sandboxstore.LoadNetNS(meta.NetNSPath)
```
abhinandanpb: NetNSPath is not set if its in host network namespace. Should we be checking for the network namespace before loading it ?

pkg/server/restart.go
```diff
@@ -0,0 +1,429 @@
...
+		return fmt.Errorf("failed to read sandboxes directory %q: %v", sandboxesRoot, err)
+	}
+	for _, d := range dirs {
+		if !d.IsDir() {
+			glog.Warningf("Invalid file %q found in sandboxes directory", d.Name())
+			continue
+		}
+		id := d.Name()
+		found := false
+		for _, c := range cntrs {
```
abhinandanpb: can we use a map to track the loaded containers/sandboxes from the caller. Just thinking,  looping over containers for every directory may not be a faster method.

pkg/server/restart.go
```diff
@@ -0,0 +1,429 @@
...
+		return fmt.Errorf("failed to read containers directory %q: %v", containersRoot, err)
+	}
+	for _, d := range dirs {
+		if !d.IsDir() {
+			glog.Warningf("Invalid file %q found in containers directory", d.Name())
+			continue
+		}
+		id := d.Name()
+		found := false
+		for _, c := range cntrs {
```
abhinandanpb: same as mentioned for sandboxes

pkg/server/restart.go
abhinandanpb: or may be keep a track of containerDirs in a map once loadContainer succeeds. So we can just loop around the dirs here and check if its in the containerDirs map , if not then clean it up.

pkg/server/restart.go
Random-Liu: Good catch. We should.

Random-Liu: Done.

pkg/server/restart.go
Random-Liu: Will do.

Random-Liu:
@abhinandanpb Addressed comments.

pkg/server/restart.go
Random-Liu: Done.

pkg/server/restart.go
Random-Liu: Done.

pkg/server/restart.go
```diff
@@ -0,0 +1,426 @@
...
+		if err := containerIO.Pipe(); err != nil {
+			return nil, err
+		}
+		return containerIO, nil
+	})
+	if err != nil && !errdefs.IsNotFound(err) {
+		return container, fmt.Errorf("failed to load task: %v", err)
+	}
+	var s containerd.Status
+	notFound := false
```
abhinandanpb: var notFound bool

pkg/server/restart.go
```diff
@@ -0,0 +1,426 @@
...
+// tolerant tasks being created or started, we prefer that not to happen.
+
+// recover recovers system state from containerd and status checkpoint.
+func (c *criContainerdService) recover(ctx context.Context) error {
+	// Recover all sandboxes.
+	sandboxes, err := c.client.Containers(ctx, filterLabel(sandboxMetadataLabel))
+	if err != nil {
+		return fmt.Errorf("failed to list sandbox containers: %v", err)
+	}
+	for _, sandbox := range sandboxes {
```
abhinandanpb: I think you can create a sandboxmap[string]bool here and pass it to cleanuporphansandboxdir instead of creating one there.

pkg/server/restart.go
```diff
@@ -0,0 +1,426 @@
...
+			return fmt.Errorf("failed to add sandbox %q to store: %v", sandbox.ID(), err)
+		}
+	}
+
+	// Recover all containers.
+	containers, err := c.client.Containers(ctx, filterLabel(containerMetadataLabel))
+	if err != nil {
+		return fmt.Errorf("failed to list containers: %v", err)
+	}
+	for _, container := range containers {
```
abhinandanpb: same as sandbox

pkg/store/sandbox/netns.go
```diff
@@ -41,6 +50,19 @@ func NewNetNS() (*NetNS, error) {
 	return n, nil
 }
 
+// LoadNetNS loads existing network namespace. It returns ErrClosedNetNS
+// is the network namespace has already been closed.
```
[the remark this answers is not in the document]
Random-Liu: Done

pkg/store/sandbox/netns.go
```diff
@@ -49,10 +71,39 @@ func (n *NetNS) Remove() error {
...
+		// TODO(random-liu): Add util function for unmount.
+		// Check netns existence.
+		if _, err := os.Stat(path); err != nil {
+			if os.IsNotExist(err) {
+				return nil
+			}
+			return fmt.Errorf("failed to stat netns: %v", err)
+		}
+		// Follow possible /var/run -> /run symlink.
+		path, err := symlink.FollowSymlinkInScope(path, "/")
```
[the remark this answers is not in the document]
Random-Liu: Copied from cri-o. Yeah, it seems that we don't need this even `/var/run -> /run` is a symlink.

pkg/server/restart.go
Random-Liu: Done.

pkg/server/restart.go
Random-Liu: Doesn't make too much difference to me, because we won't use the map here. Given so I prefer move the code into cleanuporphansandboxdir which is the place where the map will be used. :)

pkg/store/sandbox/netns.go
```diff
@@ -41,6 +50,19 @@ func NewNetNS() (*NetNS, error) {
 	return n, nil
 }
 
+// LoadNetNS loads existing network namespace. It returns ErrClosedNetNS
+// is the network namespace has already been closed.
+func LoadNetNS(path string) (*NetNS, error) {
+	if err := cnins.IsNSorErr(path); err != nil {
```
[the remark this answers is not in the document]
abhinandanpb: GetNS would return ```cnins.NSPathNotExistErr``` so you can always check for that error if the nspath doesnt exist. Unless I am missing something. This is fine. If we can avoid an if condition even better.

pkg/server/restart.go
abhinandanpb: I meant instead of passing sandbox list and container list , pass a map for comparison. Not a blocker though

Random-Liu:
> GetNS would return cnins.NSPathNotExistErr so you can always check for that error if the nspath doesnt exist. Unless I am missing something. This is fine. If we can avoid an if condition even better.

@abhinandanpb I read their code. You are right. We should use `cnins.NSPathNotExistErr` instead. We may not want to ignore the error if it's some other transient stat error.

Random-Liu:
@abhinandanpb Addressed comments. I still keeps the `ErrClosedNetNS`, so that caller of `LoadNetNS` doesn't need to make assumption on the underlying implementation. However, I does change `LoadNetNS` to look at `cnins.NSPathNotExistErr`.

Random-Liu:
@abhinandanpb Thanks for reviewing! Will merge after test passes.

PR #210, opened 2017-09-04
Add release tarball.

Random-Liu:
Add `make release` to:
1) Install all dependencies (containerd, runc and cni) into `_output/release-stage`;
2) Create tarball from `_output/release-stage`.

```console
$ tree _output/
_output/
├── cri-containerd
├── cri-containerd-0.1.0-158-g5ea9de0.tar.gz
└── release-stage
    ├── etc
    │   └── cni
    │       └── net.d
    │           └── 10-containerd-net.conflist
    ├── opt
    │   └── cni
    │       └── bin
    │           ├── bridge
    │           ├── dhcp
    │           ├── flannel
    │           ├── host-local
    │           ├── ipvlan
    │           ├── loopback
    │           ├── macvlan
    │           ├── portmap
    │           ├── ptp
    │           ├── sample
    │           ├── tuning
    │           └── vlan
    └── usr
        └── local
            ├── bin
            │   ├── containerd
            │   ├── containerd-shim
            │   ├── containerd-stress
            │   ├── cri-containerd
            │   └── ctr
            └── sbin
                └── runc

11 directories, 21 files
```

With the release tar ball, user could install cri-containerd and all the dependencies by simply **untar the tarball to `/`**, which is similar with golang installation.
```console
$ sudo tar -C / -xzf cri-containerd-0.1.0-158-g5ea9de0.tar.gz
```

However, we may want to consider whether we should put the CNI config inside the tarball, after @abhinandanpb figures out the cluster creation.

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -30,10 +30,27 @@ set -o pipefail
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 . ${ROOT}/hack/versions
 
+# DESTDIR is the dest path to install dependencies.
+DESTDIR=${DESTDIR:-"/"}
+# Convert to absolute path if it's relative.
+if [[ ${DESTDIR} != /* ]]; then
+  DESTDIR=${ROOT}/${DESTDIR}
+fi
+
+# NOSUDO indicates not to use sudo during installation.
+NOSUDO=${NOSUDO:-false}
+sudo="sudo"
+if ${NOSUDO}; then
+  sudo=""
+fi
+
+CONTAINERD_DIR=${DESTDIR}/usr/local
+RUNC_DIR=${DESTDIR}
+CNI_DIR=${DESTDIR}/opt/cni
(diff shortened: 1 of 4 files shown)


hack/install-deps.sh
```diff
@@ -52,19 +69,18 @@ git fetch --all
 git checkout ${RUNC_VERSION}
 BUILDTAGS=${BUILDTAGS:-seccomp apparmor}
 make BUILDTAGS="$BUILDTAGS"
-sudo make install
-which runc
```
mikebrow: why remove the runc and containerd which sanity checks?

Random-Liu: Because now we also use `make install.deps` to install all the binaries to `release-stage`, which are not in path.

mikebrow: Ah.. see it below, makes sense now.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/210#pullrequestreview-60595498.

PR #211, opened 2017-09-05
Upload node e2e test log to gcs

Random-Liu:
This PR:
1) Install gcloud on travis node, and also handles some gcloud authentication stuff. ***The service account is encrypted by travis https://docs.travis-ci.com/user/encrypting-files/, which only merged code has the access. Please keep in mind that we should never merge code trying to print that out, or else I'll have to delete the service account.***
2) Change `make test-e2e-node` to upload test logs to gcs is `UPLOAD_LOG=true` is specified.

At the end of the test output, you will see:
```console
Test logs are uploaed to:
    https://pantheon.corp.google.com/storage/browser/test-e2e-node/0.1.0-158-ga6b0e41-dirty-20170905-053204
```

And the logs are publicly accessible.

For this PR, I've created a Google Cloud Platform project `k8s-cri-containerd`, **which is paid by Google.** In the future, we'll use this project to:
1) Presistent test result;
2) Continuously deploy binaries and node images;
3) Probably publish our official releases.


Random-Liu:
The whole thing works properly on my node, but I'm not sure whether it works on travis.

So I created a temporary service account with only GCS access, and send a temporary commit to verify whether the log update works.

Will remove the last commit and the service account, after we prove it works.

Random-Liu:
The test commit works good, you could access the test log http://gcsweb.k8s.io/gcs/cri-containerd_test-e2e-node/0.1.0-160-gfafee8c-20170905-193606/.

I'll remove the temporary commit, and this PR is ready for review.

.travis.yml
```diff
@@ -5,10 +5,20 @@ sudo: required
 services:
     - docker
 
+cache:
+    directories:
+        - "${HOME}/google-cloud-sdk/"
+
 before_install:
+    # Workaround to make gsutil work (see travis-ci#7940).
+    - sudo rm -f /etc/boto.cfg
```
mikebrow: "export BOTO_CONFIG=/dev/null" is another workaround also referenced in travis-ci#7940

Random-Liu: I've verified the current fix, and successfully use gsutil to upload log. Let's use the verified one now. :)

Random-Liu: Added TODO to remove it.

PR #212, opened 2017-09-05
Add build tags and Improve the test case of selinux

miaoyq:
- Add build tags in Makefile
- Fixes a bug because of my negligence
- Improve the test case of `InitSelinuxOpts`

### DIFF
--- a/pkg/server/helpers_selinux_test.go
+++ b/pkg/server/helpers_selinux_test.go
@@ -36,22 +36,22 @@ func TestInitSelinuxOpts(t *testing.T) {
 		processLabel string
 		mountLabels  []string
 	}{
-		"testNullValue": {
+		"Should return empty strings for processLabel and mountLabel when selinuxOpt is nil": {
 			selinuxOpt:   nil,
 			processLabel: "",
 			mountLabels:  []string{"", ""},
 		},
-		"testEmptyString": {
+		"Should return empty strings for processLabel and mountLabel when selinuxOpt has been initialized partially": {
 			selinuxOpt: &runtime.SELinuxOption{
 				User:  "",
-				Role:  "",
+				Role:  "user_r",
 				Type:  "",
-				Level: "",
+				Level: "s0:c1,c2",
 			},
-			processLabel: ":::",
-			mountLabels:  []string{":object_r:container_file_t:", ":object_r:svirt_sandbox_file_t:"},
+			processLabel: "",
+			mountLabels:  []string{"", ""},
 		},
-		"testUser": {
+		"Should be resolved correctly when selinuxOpt has been initialized completely": {
 			selinuxOpt: &runtime.SELinuxOption{
 				User:  "user_u",
 				Role:  "user_r",
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,10 @@ VERSION := $(shell git describe --tags --dirty)
 # strip the first char of the tag if it's a `v`
 VERSION := $(VERSION:v%=%)
 TARBALL ?= cri-containerd-$(VERSION).tar.gz
-BUILD_TAGS:= -ldflags '-X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)'
+ifdef BUILD_TAGS
+BUILD_TAGS := -tags $(BUILD_TAGS)
+endif
+GO_LDFLAGS := -ldflags '-X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)'
 SOURCES := $(shell find . -name '*.go')
 
 all: binaries
@@ -68,8 +71,7 @@ boiler:
 
 $(BUILD_DIR)/cri-containerd: $(SOURCES)
 	$(GO) build -o $@ \
-	   $(BUILD_TAGS) \
-	   $(GO_LDFLAGS) $(GO_GCFLAGS) \
+	   $(BUILD_TAGS) $(GO_LDFLAGS) $(GO_GCFLAGS) \
 	   $(PROJECT)/cmd/cri-containerd
 
 test:
@@ -86,7 +88,7 @@ clean:
 
 binaries: $(BUILD_DIR)/cri-containerd
 
-static-binaries: GO_LDFLAGS=--ldflags '-extldflags "-fno-PIC -static"'
+static-binaries: GO_LDFLAGS += --ldflags '-extldflags "-fno-PIC -static"'
 static-binaries: $(BUILD_DIR)/cri-containerd
 
 install: binaries
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -368,11 +368,19 @@ func initSelinuxOpts(selinuxOpt *runtime.SELinuxOption) (string, string, error) 
 		return "", "", nil
 	}
 
+	// Should ignored selinuxOpts if they are incomplete.
+	if selinuxOpt.GetUser() == "" ||
+		selinuxOpt.GetRole() == "" ||
+		selinuxOpt.GetType() == "" ||
+		selinuxOpt.GetLevel() == "" {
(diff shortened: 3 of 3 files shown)


pkg/server/helpers_selinux_test.go
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 package server
 
 import (
+	"fmt"
```
Random-Liu: Sorry. Actually I don't quite understand the test.
Could you change the desciption to "should xxx when xxx"? So that I could understand what you are willing to test.

miaoyq: Will do, Thanks :)

miaoyq: Done.

pkg/server/helpers_selinux_test.go
```diff
@@ -65,7 +85,12 @@ func TestInitSelinuxOpts(t *testing.T) {
 		t.Logf("TestCase %q", desc)
 		processLabel, mountLabel, err := initSelinuxOpts(test.selinuxOpt)
 		assert.NoError(t, err)
-		assert.Equal(t, test.processLabel, processLabel)
-		assert.Contains(t, test.mountLabels, mountLabel)
+		if desc == "testUser" {
```
Random-Liu: What about here?

I don't think we want to change test behavior based on description. I'd rather add a filed in the struct to indicate whether we should use equal to do regexp match.

pkg/server/helpers_selinux_test.go
```diff
@@ -65,7 +85,12 @@ func TestInitSelinuxOpts(t *testing.T) {
...
+			assert.Equal(t, test.processLabel, processLabel)
+			assert.Contains(t, test.mountLabels, mountLabel)
+		} else {
+			assert.Regexp(t, "^"+test.processLabel, processLabel)
+			assert.Regexp(t, "^"+test.mountLabels[0], mountLabel)
```
Random-Liu: What about `test.mountLabels[1]`?

pkg/server/helpers_selinux_test.go
```diff
@@ -31,27 +32,46 @@ func TestInitSelinuxOpts(t *testing.T) {
...
 				Role:  "",
 				Type:  "",
 				Level: "",
 			},
-			processLabel: ":::",
-			mountLabels:  []string{":object_r:container_file_t:", ":object_r:svirt_sandbox_file_t:"},
+			processLabel: pLabelNolevel,
+			mountLabels:  []string{mLabelNolevel, ""},
+		},
+		"Should be resolved correctly when selinuxOpt has been initialized partially": {
```
Random-Liu: So these should be ignored if it's incomplete?

pkg/server/helpers_selinux_test.go
miaoyq: Sorry, should change `testUser` to `Should be resolved correctly when selinuxOpt has been initialized completely`.

pkg/server/helpers_selinux_test.go
miaoyq: Only `test.mountLabels[0]` is valid values except last one.

pkg/server/helpers_selinux_test.go
Random-Liu: Let's add field in the struct instead.

pkg/server/helpers_selinux_test.go
Random-Liu: Then why do we have it in the test case? Seems that we don't use it at all.

pkg/server/helpers_selinux_test.go
miaoyq: I think this should depend on the implementation of k8s. 

These should be ignored if K8s don't allow to set incomplete selinux options.

pkg/server/helpers_selinux_test.go
miaoyq: >I'd rather add a filed in the struct to indicate whether we should use equal to do regexp match.
That's better. :)

Random-Liu:
@miaoyq Anyway, we don't have test environment for selinux, and I'm not really familiar with it. :) I'll completely trust you.

Please address the comments, and make sure the test could pass on your selinux environment.

pkg/server/helpers_selinux_test.go
Random-Liu: So `InitLabels` will ignore them if they are incomplete?

pkg/server/helpers_selinux_test.go
miaoyq: It will be used in `assert.Contains(t, test.mountLabels, mountLabel)`

miaoyq:
Thank you for your trust, the test had pass on my environment, but I have not make test when addressed [this comment](https://github.com/kubernetes-incubator/cri-containerd/pull/212#discussion_r139272231).
I' sorry for this, I will address the last comments, and provided the test result on my selinux environment. :)

pkg/server/helpers_selinux_test.go
miaoyq: No, `InitLabels` have no error returned when the selinux option is incomplete.

miaoyq: But must be in this form: `user: , role: , type: , level: ,`
For example, `user: , role:user_r, type: , level:s0:c1,c2,`

miaoyq: Refer to the [test case](
 https://github.com/opencontainers/selinux/blob/4a2974bf1ee960774ffd517717f1f45325af0206/go-selinux/label/label_selinux_test.go#L49-L54) for `InitLabels`, maybe we should ignore these?

But the options is compliant here.

Random-Liu:
@miaoyq Are you still working on this? Does the test pass?

miaoyq:
@Random-Liu Yes,  sorry for delay , I filed an issue in [opencontainers/selinux](https://github.com/opencontainers/selinux/issues/17)
Have get a reply from @rhatdan,  Options should be ignored if it's incomplete.

I am working on this. : )

miaoyq:
@Random-Liu  Done. 
If the `sellinuxOpts` is incomplete, it will be ignored.
I have changed both `initSelinuxOpts` and unit test related, PTAL. Thanks!
```
[root@localhost cri-containerd]# go test --tags selinux -c -i -o /tmp/selinux-test github.com/kubernetes-incubator/cri-containerd/pkg/server
[root@localhost cri-containerd]# /tmp/selinux-test -test.v -test.run ^TestInitSelinuxOpts
=== RUN   TestInitSelinuxOpts
--- PASS: TestInitSelinuxOpts (0.00s)
	helpers_selinux_test.go:65: TestCase "Should return empty strings for processLabel and mountLabel when selinuxOpt is nil"
	helpers_selinux_test.go:65: TestCase "Should return empty strings for processLabel and mountLabel when selinuxOpt has been initialized partially"
	helpers_selinux_test.go:65: TestCase "Should be resolved correctly when selinuxOpt has been initialized completely"
PASS
```

Random-Liu:
@miaoyq Seems like you changed the test back? So previous version didn't pass?

miaoyq:
@Random-Liu  The previous version have passed, 
Because I added the following code in `InitSelinuxOpts`, so get the different test result.
```go
	// Should ignored selinuxOpts if they are incomplete.
	if selinuxOpt.GetUser() == "" ||
		selinuxOpt.GetRole() == "" ||
		selinuxOpt.GetType() == "" ||
		selinuxOpt.GetLevel() == "" {
		return "", "", nil
	}
```

miaoyq:
@Random-Liu I will continue to follow this feature, and improve it.  Try to manually verify it.

Random-Liu:
@miaoyq Thanks! :)

PR #214, opened 2017-09-05
Handle environment variables which containe spaces

ijc:
This avoids errors such as:

    spec: invalid environment variable "JAVA_OPTS=-Djava.security.egd=file:/dev/urandom"

use `SplitN(2)` to get the envvar name and value while allowing the value to
contain `=`.

Add some variables to the test data which have one or more `=` in the value.
Since this makes the resulting list of variables to check rather long split the
check in two and check the container config and image config derived values
independently.


Random-Liu:
LGTM. Thanks for the fix!

PR #221, opened 2017-09-06
Always use a writeable snapshot as the rootfs.

### LINKED ISSUE #220: using ro snapshot for rootfs fails if all required mount points are not present.
Trying to run the [sock shop demo](https://github.com/microservices-demo/microservices-demo) fails since several of the ro rootfs images lack various required bind-mount targets such as `/etc/hosts`. This produces (I've substituted our the long ids and manually wrapper for readability):
```
container_linux.go:265: starting container process caused \
    "process_linux.go:348: container init caused \
    \"rootfs_linux.go:57: mounting \\\"/var/lib/cri-containerd/sandboxes/«sb-id»/hosts\\\" to \
    rootfs \\\"/run/containerd/io.containerd.runtime.v1.linux/k8s.io/«ctr-id»/rootfs\\\" at \
    \\\"/run/containerd/io.containerd.runtime.v1.linux/k8s.io/«ctr-id»/rootfs/etc/hosts\\\" caused
    \\\"open /run/containerd/io.containerd.runtime.v1.linux/k8s.io/«ctr-id»/rootfs/etc/hosts: \
    read-only file system\\\"\""
```
This is `runc` checking if the mount point exists (it doesn't) and then failing to `touch` it to create it (see [runc code](https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L241)).

This example is from `docker.io/library/mongo:latest` but I believe this will be quite common because Docker contains [code](https://github.com/moby/moby/blob/master/daemon/initlayer/setup_unix.go#L19) which creates an "init layer" in order to provide these mount points if they don't exist so it is easy to omit them and not notice.

containerd had some code to produce an initlayer but it was removed in https://github.com/containerd/containerd/pull/1348. @stevvooe is proposing to reinstate them in https://github.com/containerd/containerd/pull/1477 so one (good) option is to make use of that.

An alternative would be to stop using `containerd.WithNewSnapshotView()` for ro containers and always use `containerd.WithNewSnapshot()` but instruct runc to make the rootfs ro in the bundle config. Then runc should be able to make the mount points as necessary before it makes the rootfs readonly. This has the downside of making some files magically appear in the diff of the rootfs, which would be bad in some usecases (e.g. build) but I think is acceptable enough for the cri-containerd usecase (at least for now).

I think the preference should be to reintroduce init layer construction in containerd (i.e. https://github.com/containerd/containerd/pull/1477) and use that, so just mentioning the alternative for completeness/fallback.

NB @crosbymichael tells me that more work would be needed on the `runc` side to have it correctly handle being given a ro rootfs (it handles taking a rw and making it ro correctly though).

/cc @Random-Liu @crosbymichael @stevvooe @dmcgowan

ijc:
This will be made readonly by runc based on spec.Root.Readonly (which we
already set correctly) but defering until then gives runc the chance to make
any missing mount points as it processes the spec.Mount array.

This is necessary because many container images lack mount points for things
like the /etc/hosts which we want to overbind. This is not noticed with e.g.
Docker because it automatically creates an additional layer containing those.
This is something we may want to do here as well eventually but for now using a
writeable snapshot is both necessary and sufficient.

The same does not apply to the sandbox since we never modify its rootfs or want
to mount anything in it etc, add a comment to clarify.

Fixes #220.

//cc @Random-Liu

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -110,12 +110,12 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	opts := []containerd.NewContainerOpts{
 		containerd.WithSnapshotter(c.snapshotter),
 	}
-	// Prepare container rootfs.
-	if config.GetLinux().GetSecurityContext().GetReadonlyRootfs() {
-		opts = append(opts, containerd.WithNewSnapshotView(id, image.Image))
-	} else {
-		opts = append(opts, containerd.WithNewSnapshot(id, image.Image))
-	}
+	// Prepare container rootfs. This is always writeable even if
+	// the container wants a readonly rootfs since we want to give
+	// the runtime (runc) a chance to modify (e.g. to create mount
+	// points corresponding to spec.Mounts) before making the
+	// rootfs readonly (requested by spec.Root.Readonly).
+	opts = append(opts, containerd.WithNewSnapshot(id, image.Image))
 	meta.ImageRef = image.ID
 
 	// Create container root directory.
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -132,6 +132,9 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	}
 	opts := []containerd.NewContainerOpts{
 		containerd.WithSnapshotter(c.snapshotter),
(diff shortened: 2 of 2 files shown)


Random-Liu:
LGTM
Will merge after test passes.

crosbymichael:
LGTM

Issue #224, opened 2017-09-06
Run node e2e test as post submit test.


Random-Liu:
Currently, we could reliably pass 180/182 node e2e test (Non-Slow, Non-Serial, Non-Flaky), and we are running the test everyday.

We should run it more frequently to get more result and catch flakiness.

However, node e2e usually takes a long time, we should run node e2e as post submit test after https://github.com/kubernetes-incubator/cri-containerd/pull/211 is proved to be working after tonight's run.

Random-Liu:
And we don't need to run verification and build for post submit to accelerate it.

PR #228, opened 2017-09-07
Run node e2e for both cron and branch update

Random-Liu:
Fixes: https://github.com/kubernetes-incubator/cri-containerd/issues/224

Node e2e works well. Here is the test log yesterday: http://gcsweb.k8s.io/gcs/cri-containerd_test-e2e-node/0.1.0-179-g7a75a91-20170907-084526/
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/224
This PR will enable node e2e as post-submit test, and expose the failure, so that we'll get a node e2e test result each time after a new PR is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,8 @@ jobs:
       go: 1.8.x
     - stage: E2E Test
       script:
-        - test "${TRAVIS_EVENT_TYPE}" != "cron" && exit 0 || true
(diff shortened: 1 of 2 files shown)


Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/228#pullrequestreview-61641291.

PR #229, opened 2017-09-07
Deploy release for each branch update.

Random-Liu:
Deploy release tarball for each branch update.

The release tarball will be located in `gs://cri-containerd-staging`. They are not publicly accessible, the main purpose is for testing.

I'll add test job in Kubernetes test-infra to consume these tarballs.

For the official release, we could run:
```console
DEPLOY_BUCKET=cri-containerd-release make push
```


Random-Liu:
Also updated containerd to v1.0.0-beta.0 in this PR to unblock @abhinandanpb's work.

abhinandanpb:
@Random-Liu how many tar balls will we keep in the GCS bucket ? How often would we need to update the installer.

Random-Liu:
@abhinandanpb `cri-containerd-staging` is only for testing. We'll have another bucket `cri-containerd-release` for release soon, which will only update per-release.
Our installer could down load from there then.

PR #230, opened 2017-09-08
Ensure the mount point is propagated

miaoyq:
mount with `rshared`, the host path should be shared.
mount with `rslave`, the host pash should be shared or slave.

Based on #246 

/cc @Random-Liu

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/contrib/apparmor"
+	"github.com/docker/docker/pkg/mount"
 	"github.com/golang/glog"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runc/libcontainer/devices"
@@ -551,13 +552,24 @@ func (c *criContainerdService) addOCIBindMounts(g *generate.Generator, mounts []
 		if err != nil {
 			return fmt.Errorf("failed to resolve symlink %q: %v", src, err)
 		}
+
+		mountInfos, err := c.os.GetMounts()
+		if err != nil {
+			return err
+		}
 		options := []string{"rbind"}
 		switch mount.GetPropagation() {
 		case runtime.MountPropagation_PROPAGATION_PRIVATE:
 			options = append(options, "rprivate")
 		case runtime.MountPropagation_PROPAGATION_BIDIRECTIONAL:
+			if err := ensureShared(src, mountInfos); err != nil {
+				return err
+			}
 			options = append(options, "rshared")
 		case runtime.MountPropagation_PROPAGATION_HOST_TO_CONTAINER:
+			if err := ensureSharedOrSlave(src, mountInfos); err != nil {
+				return err
+			}
 			options = append(options, "rslave")
 		default:
 			glog.Warningf("Unknown propagation mode for hostPath %q", mount.HostPath)
@@ -695,4 +707,40 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
 	}
 	spec.Mounts = mounts
 	return spec, nil
+}
+
+// Ensure mount point on which path is mounted, is shared.
+func ensureShared(path string, mountInfos []*mount.Info) error {
+	sourceMount, optionalOpts, err := getSourceMount(path, mountInfos)
+	if err != nil {
+		return err
+	}
+
+	// Make sure source mount point is shared.
+	optsSplit := strings.Split(optionalOpts, " ")
+	for _, opt := range optsSplit {
+		if strings.HasPrefix(opt, "shared:") {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("path %q is mounted on %q but it is not a shared mount", path, sourceMount)
+}
+
+// Ensure mount point on which path is mounted, is either shared or slave.
+func ensureSharedOrSlave(path string, mountInfos []*mount.Info) error {
+	sourceMount, optionalOpts, err := getSourceMount(path, mountInfos)
+	if err != nil {
+		return err
+	}
+	// Make sure source mount point is shared.
+	optsSplit := strings.Split(optionalOpts, " ")
+	for _, opt := range optsSplit {
+		if strings.HasPrefix(opt, "shared:") {
+			return nil
+		} else if strings.HasPrefix(opt, "master:") {
+			return nil
+		}
+	}
+	return fmt.Errorf("path %q is mounted on %q but it is not a shared or slave mount", path, sourceMount)
 }
--- a/pkg/server/container_create_test.go
+++ b/pkg/server/container_create_test.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/docker/docker/pkg/mount"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
@@ -27,6 +28,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
+	ostesting "github.com/kubernetes-incubator/cri-containerd/pkg/os/testing"
 	"github.com/kubernetes-incubator/cri-containerd/pkg/util"
 )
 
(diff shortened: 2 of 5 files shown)


miaoyq:
The CI failed because we have not used the real path. 
Maybe we should add this in kubelet side.
@Random-Liu What do you think for this?

Random-Liu:
@miaoyq Will review this PR after Thursday. BTW, would you mind me mention ZTE is contributing to `cri-containerd` in the moby summit?

miaoyq:
Of course not, i really appreciate that.

miaoyq:
Of course not, i really appreciate that. : )

miaoyq:
I've manually made unit test locally, and it works, but the travis-ci always failed.
@Random-Liu  Does the travis-ci run as non root user?

miaoyq:
@Random-Liu  The travis-ci run unit test as non root user, but mount operation need root user. 

 I have create a pr for this, see #246, PTAL, Thanks :)

pkg/server/container_create_test.go
```diff
@@ -48,8 +50,57 @@ func checkMount(t *testing.T, mounts []runtimespec.Mount, src, dest, typ string,
 	assert.True(t, found, "mount from %q to %q not found", src, dest)
 }
 
-func getCreateContainerTestData() (*runtime.ContainerConfig, *runtime.PodSandboxConfig,
-	*imagespec.ImageConfig, func(*testing.T, string, uint32, *runtimespec.Spec)) {
+func createHostPath(t *testing.T, mounts []*runtime.Mount) {
```
Random-Liu: Don't think we should test this in unit test.
We have replaced or `os` with `FakeOS`. This test won't work.
If we really want to test, we should leverage the `FakeOS` to test.

Random-Liu: I'll add our own integration test framework, we could add this kind of test there.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+
+	sourceMount, optionalOpts, err := getSourceMount(path)
+	if err != nil {
+		return err
+	}
+	// Make sure source mount point is shared.
+	optsSplit := strings.Split(optionalOpts, " ")
+	for _, opt := range optsSplit {
+		if strings.HasPrefix(opt, "shared:") {
+			sharedMount = true
```
Random-Liu: return nil directly.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+			break
+		} else if strings.HasPrefix(opt, "master:") {
+			slaveMount = true
```
Random-Liu: return nil directly.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+			break
+		}
+	}
```
Random-Liu: return error here.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+		if m.Mountpoint == dir {
+			return m
+		}
+	}
+	return nil
+}
+
+func getSourceMount(source string) (string, string, error) {
+	// Ensure any symlinks are resolved.
+	sourcePath, err := filepath.EvalSymlinks(source)
```
Random-Liu: No need to resolve symlink. We've done that in `addOCIBindMounts`.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+	if err != nil {
+		return "", "", err
+	}
+
+	mountinfos, err := mount.GetMounts()
```
Random-Liu: Add `GetMounts` to os package, so that you could use fake os in unit test, and please add unit test based on that.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+	sourceMount, optionalOpts, err := getSourceMount(path)
+	if err != nil {
+		return err
+	}
+
+	// Make sure source mount point is shared.
+	optsSplit := strings.Split(optionalOpts, " ")
+	for _, opt := range optsSplit {
+		if strings.HasPrefix(opt, "shared:") {
+			sharedMount = true
```
Random-Liu: return nil here.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+			break
+		}
+	}
+
```
Random-Liu: return error.

pkg/server/container_create.go
```diff
@@ -646,3 +654,95 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+func getMountInfo(mountinfo []*mount.Info, dir string) *mount.Info {
+	for _, m := range mountinfo {
+		if m.Mountpoint == dir {
+			return m
+		}
+	}
+	return nil
+}
+
+func getSourceMount(source string) (string, string, error) {
```
Random-Liu: Probably move this to helper.

pkg/server/container_create.go
miaoyq: That's better. :+1:

miaoyq:
@Random-Liu  Thank you for review, I'm addressing the comment.

miaoyq:
@Random-Liu  Have addressed, PTAL, Thanks!

pkg/server/container_create.go
```diff
@@ -696,3 +708,39 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+
+	// Make sure source mount point is shared.
+	optsSplit := strings.Split(optionalOpts, " ")
+	for _, opt := range optsSplit {
+		if strings.HasPrefix(opt, "shared:") {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("path %s is mounted on %s but it is not a shared mount", path, sourceMount)
```
Random-Liu: Use `%q` for string.

pkg/server/container_create.go
```diff
@@ -696,3 +708,39 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+	// Make sure source mount point is shared.
+	optsSplit := strings.Split(optionalOpts, " ")
+	for _, opt := range optsSplit {
+		if strings.HasPrefix(opt, "shared:") {
+			return nil
+		} else if strings.HasPrefix(opt, "master:") {
+			return nil
+		}
+	}
+	return fmt.Errorf("path %s is mounted on %s but it is not a shared or slave mount", path, sourceMount)
```
Random-Liu: ditto.

pkg/server/container_create_test.go
```diff
@@ -20,13 +20,15 @@ import (
...
 
+	"github.com/docker/docker/pkg/mount"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
+	testing2 "github.com/kubernetes-incubator/cri-containerd/pkg/os/testing"
```
Random-Liu: `ostesting "github.com/kubernetes-incubator/cri-containerd/pkg/os/testing"`

pkg/server/container_create_test.go
```diff
@@ -610,6 +576,77 @@ func TestPrivilegedBindMount(t *testing.T) {
...
+	}{
+		"HostPath should mount as 'rprivate' if propagation is MountPropagation_PROPAGATION_PRIVATE": {
+			criMount: &runtime.Mount{
+				ContainerPath: "container-path",
+				HostPath:      "host-path",
+				Propagation:   runtime.MountPropagation_PROPAGATION_PRIVATE,
+			},
+			fakeGetMountsFn: nil,
+			optionsCheck:    []string{"rbind", "rprivate"},
+		},
```
Random-Liu: Should add error test case, which host path propagation is not properly set, we should expect an error from `addOCIBindMounts`.

pkg/server/helpers.go
```diff
@@ -384,3 +385,35 @@ func isInCRIMounts(dst string, mounts []*runtime.Mount) bool {
 	}
 	return false
 }
+
```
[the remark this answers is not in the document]
Random-Liu: And please replace with `mount.Loopup()` after https://github.com/kubernetes-incubator/cri-containerd/pull/257 is merged.

miaoyq:
@Random-Liu  Thanks for review, Have addressed.

pkg/server/container_create_test.go
```diff
@@ -610,6 +576,115 @@ func TestPrivilegedBindMount(t *testing.T) {
...
+		return []*mount.Info{
+			{
+				Mountpoint: "host-path",
+				Optional:   "others",
+			},
+		}, nil
+	}
+
+	for desc, test := range map[string]struct {
+		criMount        *runtime.Mount
```
Random-Liu: Add `expectErr bool`.

We could:
1) assert error, if `expectErr==true`.
2) assert no error and check options, if `expectErr==false`.

miaoyq: OK, I will change according to your instructions, thanks!

miaoyq: Done!

Random-Liu:
LGTM

Random-Liu:
The test failure feels like a flake, filed https://github.com/kubernetes-incubator/cri-containerd/issues/259 for it.

Will retrigger the test.

PR #233, opened 2017-09-09
Remove `/run` mount for backward compatibility with docker.

Random-Liu:
For https://github.com/containerd/containerd/issues/1494.

Remove `/run` mount because Docker doesn't do that, and there are images relying on this behavior.

In we really want to make `/run` tmpfs in the future, we may need to handle copyup.


ijc:
LGTM. I wonder if we could construct an overlay with a tmpfs (or perhaps disposable real fs) upperdir rather than doing a copy up?

Random-Liu:
> LGTM. I wonder if we could construct an overlay with a tmpfs (or perhaps disposable real fs) upperdir rather than doing a copy up?

Sorry, I maybe using the wrong term. I just mean create the tmpfs volume on the host, copy content from rootfs to the volume, and mount it into the image.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/233#issuecomment-328293940.

PR #234, opened 2017-09-09
kubernetes + cri-containerd + containerd installer

abhinandanpb:
This commit contains ansible playbooks to install kubernetes, cri-containerd and containerd components as well as supporting configurations and dependencies to facilitate the usage of the kubernetes cluster with containerd.
With this change users can bring up k8 cluster in 2 steps 
1) Run the ansible playbook to install the necessary components
2) Run kubeadm init to initialize the cluster and subsequently run kubeadm join on remaining nodes
I will follow up with another PR that will have detailed set of step by step guide to install kubernetes.
There is a lot of scope for improvements and optimization. This is first cut and will gradually update the playbooks and install methods subsequently.
This has been tested on ubuntu clusters. Though the centos playbooks are attached I havent gotten a change to test it out. I believe the centos playbook is pretty similar to ubuntu dependencies but for few extra configurations which are addressed. 
The goal will be to eventually cover all the supported and tested platforms RHEL, Fedora, Sles, Linuxkit images etc. 

 
### DIFF
--- a/contrib/ansible/cri-containerd.yaml
+++ b/contrib/ansible/cri-containerd.yaml
@@ -1,0 +1,48 @@
+---
+- hosts: all 
+  become: true
+  tasks:
+    - include_vars: vars/vars.yaml # Contains tasks variables for installer
+    - include: tasks/bootstrap_ubuntu.yaml # Contains tasks bootstrap components for ubuntu systems 
+      when: ansible_distribution == "Ubuntu"
+    - include: tasks/bootstrap_centos.yaml # Contains tasks bootstrap components for centos systems
+      when: ansible_distribution == "CentOS"
+    - include: tasks/k8s.yaml # Contains tasks kubernetes component installation
+    - include: tasks/binaries.yaml # Contains tasks for pulling containerd and cri-containerd components
+       
+    - name: "Start Containerd"
+      systemd: name=containerd daemon_reload=yes state=started enabled=yes
+    
+    - name: "Start CRI-Containerd"
+      systemd: name=cri-containerd daemon_reload=yes state=started enabled=yes
+    
+    - name: "Set bridge-nf-call-iptables"
+      lineinfile:
+        line: "net/bridge/bridge-nf-call-iptables = 1"
+        dest: /etc/sysctl.conf
+        insertafter: 'EOF'
+        regexp: '\/net\/bridge\/bridge-nf-call-iptables = 1'
+        state: present
+      ignore_errors: true
+    
+    - name: "Check kubelet args in kubelet config"
+      shell: grep "^Environment=\"KUBELET_EXTRA_ARGS=" /etc/systemd/system/kubelet.service.d/10-kubeadm.conf 
+      ignore_errors: true
+      register: check_args
+
+    - name: "Add runtime args in kubelet conf"
+      lineinfile:
+                dest: "/etc/systemd/system/kubelet.service.d/10-kubeadm.conf"
+                line: "Environment=\"KUBELET_EXTRA_ARGS= --container-runtime=remote --runtime-request-timeout=15m --image-service-endpoint=/var/run/cri-containerd.sock --container-runtime-endpoint=/var/run/cri-containerd.sock\""
+                insertafter: '\[Service\]'
+      when: check_args.stdout == ""
+    
+    - name: "Start Kubelet"
+      systemd: name=kubelet daemon_reload=yes state=started enabled=yes
+    
+    # TODO This needs to be removed once we have consistent concurrent pull results
+    - name: "Pre-pull pause container image"
+      shell: |
+             /usr/local/bin/ctr pull gcr.io/google_containers/pause:3.0
+             /usr/local/bin/crictl --runtime-endpoint /var/run/cri-containerd.sock \
+             pull gcr.io/google_containers/pause:3.0
--- a/contrib/ansible/tasks/binaries.yaml
+++ b/contrib/ansible/tasks/binaries.yaml
@@ -1,0 +1,51 @@
+---
+- name: "Create a directory to download binaries"
+  file: path={{ cri_release_directory }} state=directory
+
+- name: "Get Containerd and CRI-Containerd"
+  get_url:
+    validate_certs: "no"
+    url: "https://storage.googleapis.com/cri-containerd-staging/cri-containerd-{{ cri_containerd_release_version }}.tar.gz"
+    dest: "{{ cri_release_directory }}"
+    mode: 0755
+
+- name: "Unpack Containerd and CRI-Containerd"
+  unarchive:
+    src: "{{ cri_release_directory }}cri-containerd-{{ cri_containerd_release_version }}.tar.gz"
+    dest: "{{ cri_release_directory }}"
+    remote_src: yes
+
+- name: "Install the containerd and cri-containerd binaries" 
+  copy:
+    src: "{{ cri_release_directory }}usr/local/bin/{{ item }}"
+    dest: "{{ local_bin_dir }}{{ item }}"
+    mode: 0755
+    remote_src: yes
+  with_items:
+    - containerd
+    - containerd-stress
+    - containerd-shim
+    - cri-containerd
+    - ctr
+    - crictl
+
+- name: "Install runc"
+  copy:    
+    src: "{{ cri_release_directory }}usr/local/sbin/{{ item }}"
+    dest: "{{ local_sbin_dir }}{{ item }}"
+    mode: 0755
+    remote_src: yes
+  with_items:
+    - runc
+
+- name: "Copy containerd systemd service unit"
+  template: src=../../systemd-units/containerd.service dest=/etc/systemd/system/containerd.service
+
+- name: "Copy cri-containerd systemd service unit"
+  template: src=../../systemd-units/cri-containerd.service dest=/etc/systemd/system/cri-containerd.service
+
+- name: "Create a directory for cni binary"
+  file: path={{ cni_bin_dir }} state=directory
+
+- name: "Create a directory for cni config files"
+  file: path={{ cni_conf_dir }} state=directory
--- a/contrib/ansible/tasks/bootstrap_ubuntu.yaml
+++ b/contrib/ansible/tasks/bootstrap_ubuntu.yaml
@@ -1,0 +1,16 @@
+---
+- name: "Install required packages on Ubuntu"
+  package:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+      - unzip
+      - tar
+      - apt-transport-https
+      - btrfs-tools
+      - libapparmor-dev
+      - libseccomp-dev # Revisit the need and alternatives for all -dev packages
+      - libseccomp2
+      - socat
+      - util-linux
+# TODO: Limited support for trusty for nsenter. Need to handle/verify 
--- a/contrib/ansible/tasks/bootstrap_centos.yaml
+++ b/contrib/ansible/tasks/bootstrap_centos.yaml
@@ -1,0 +1,13 @@
+---
+- name: "Install required packages on CentOS "
+  yum:
+    name: "{{ item }}"
+    state: latest
+    with_items:
+      - unzip
+      - tar
+      - btrfs-progs-devel
+      - libseccomp-devel
+      - util-linux 
+      - socat
+      - libselinux-python
--- a/contrib/ansible/tasks/k8s.yaml
+++ b/contrib/ansible/tasks/k8s.yaml
@@ -1,0 +1,50 @@
+---
+- name: "Add gpg key (Ubuntu)"
+  apt_key:
+    url: https://packages.cloud.google.com/apt/doc/apt-key.gpg
+    state: present
+  when: ansible_distribution == "Ubuntu"
+
+- name: "Add kubernetes source list (Ubuntu)"
+  apt_repository:
+    repo: "deb http://apt.kubernetes.io/ kubernetes-{{ ansible_distribution_release }} main"
+    state: present
+    filename: "kubernetes"
+  when: ansible_distribution == "Ubuntu"
+
+- name: "Update the repository cache (Ubuntu)"
+  apt: 
+    update_cache: yes
+  when: ansible_distribution == "Ubuntu"
+
+- name: "Add Kubernetes repository and install gpg key (CentOS)"
+  yum_repository:
+    name: kubernetes
+    description: Kubernetes repository
+    baseurl: https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
+    gpgcheck: yes
+    enabled: yes
+    repo_gpgcheck: yes
+    gpgkey: https://packages.cloud.google.com/yum/doc/yum-key.gpg
+  when: ansible_distribution == "CentOS"
+
+- name: "Disable SELinux (CentOS)"
+  selinux:
+    state: disabled
+  when: ansible_distribution == "CentOS"
+
+- name: "Install kubelet,kubeadm,kubectl (CentOS)"
+  yum: state=present name={{ item }}
+  with_items:
+    - kubelet
+    - kubeadm
+    - kubectl
+  when: ansible_distribution == "CentOS"
+
+- name: "Install kubelet, kubeadm, kubectl (Ubuntu)"
+  apt: name={{item}} state=installed
+  with_items:
+    - kubelet
+    - kubeadm
+    - kubectl
+  when: ansible_distribution == "Ubuntu"
--- a/contrib/ansible/vars/vars.yaml
+++ b/contrib/ansible/vars/vars.yaml
@@ -1,0 +1,8 @@
+---
+# TODO update official versions once they are available
+cri_containerd_release_version: 0.1.0-234-g55a0887 
+cri_release_directory: /opt/cri-containerd/
+local_bin_dir: /usr/local/bin/
+local_sbin_dir: /usr/local/sbin/
+cni_bin_dir: /opt/cni/bin/
+cni_conf_dir: /etc/cni/net.d/
--- a/contrib/systemd-units/containerd.service
+++ b/contrib/systemd-units/containerd.service
@@ -4,11 +4,13 @@ Documentation=https://containerd.io
 After=network.target
 
 [Service]
-Restart=always
-RestartSec=10
+ExecStartPre=/sbin/modprobe overlay
 ExecStart=/usr/local/bin/containerd
+Restart=always
+RestartSec=5
 Delegate=yes
 KillMode=process
+OOMScoreAdjust=-999
 
 [Install]
 WantedBy=multi-user.target
--- a/contrib/systemd-units/cri-containerd.service
+++ b/contrib/systemd-units/cri-containerd.service
@@ -5,7 +5,7 @@ After=containerd.service
 
 [Service]
 Restart=always
-RestartSec=10
+RestartSec=5
 ExecStart=/usr/local/bin/cri-containerd --logtostderr
 OOMScoreAdjust=-999
 


Random-Liu:
Will review this at the weekend.

install/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,46 @@
+---
+- hosts: all 
+  become: true
+  # Setting the environment variables that might be needed for the entire installation
+  environment:
+    PATH: "{{ ansible_env.PATH }}:/opt/golang/bin:/usr/local/go/bin:/usr/local/bin"
```
Random-Liu: Do we still need gopath? Seems not.

install/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,46 @@
...
+    GOPATH: "/opt/golang"
+    GOBIN: "/opt/golang/bin"
+  tasks:
+    - include_vars: vars/vars.yaml # Contains tasks variables for installer
+    - include: tasks/bootstrap_ubuntu.yaml #Contains tasks bootstrap components for ubuntu systems
```
Random-Liu: nit: Space after `#` and also for the following comments.

install/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,46 @@
...
+    PATH: "{{ ansible_env.PATH }}:/opt/golang/bin:/usr/local/go/bin:/usr/local/bin"
+    GOPATH: "/opt/golang"
+    GOBIN: "/opt/golang/bin"
+  tasks:
+    - include_vars: vars/vars.yaml # Contains tasks variables for installer
+    - include: tasks/bootstrap_ubuntu.yaml #Contains tasks bootstrap components for ubuntu systems 
+      when: ansible_distribution == "Ubuntu"
+    - include: tasks/bootstrap_centos.yaml #Contains tasks bootstrap components for centos systems
+      when: ansible_distribution == "CentOS"
+    - include: tasks/k8.yaml #Contains tasks kubernetes component installation
```
Random-Liu: s/k8/k8s

install/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - include: tasks/build.yaml #Contains tasks for pulling containerd and cri-containerd components
```
Random-Liu: This is not `build`, right? We just download and extract tar.

install/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: "Set bridge-nf-call-iptables"
+      lineinfile:
+        line: "net/bridge/bridge-nf-call-iptables = 1"
+        dest: /etc/sysctl.conf
+        insertafter: 'EOF'
+        regexp: '\/net\/bridge\/bridge-nf-call-iptables = 1'
+        state: present
+      ignore_errors: true
+    
+    - name: "Check kubelet args in kubelet config"
```
Random-Liu: Can we do something similar with cri-o? Just drop in another file https://github.com/cri-o/cri-o-ansible/blob/master/cri-o.yml#L217-L220.

install/ansible/tasks/bootstrap_centos.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: "Install required packages on CentOS "
+  yum:
+    name: "{{ item }}"
+    state: latest
+    with_items:
+      - unzip
```
Random-Liu: Why do we need `unzip`?

install/ansible/tasks/bootstrap_centos.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - tar
+      - btrfs-progs-devel
+      - device-mapper-devel
+      - libseccomp-devel
+      - util-linux 
+      - socat
+      - libselinux-python
```
Random-Liu: Why do we need this?

install/ansible/tasks/bootstrap_ubuntu.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: "Install required packages on Ubuntu"
+  package:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+      - unzip
```
Random-Liu: ditto.

install/ansible/tasks/bootstrap_ubuntu.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: "Install required packages on Ubuntu"
+  package:
+    name: "{{ item }}"
+    state: latest
+  with_items:
```
Random-Liu: What about `nsenter`? And also for centos.

install/ansible/tasks/bootstrap_ubuntu.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - unzip
+      - tar
+      - apt-transport-https
+      - btrfs-tools
+      - libapparmor-dev
+      - libseccomp-dev
+      - libseccomp2
+      - socat
+      - util-linux
```
Random-Liu: when: ansible_distribution == 'Ubuntu'?

install/ansible/tasks/bootstrap_ubuntu.yaml
```diff
@@ -0,0 +1,15 @@
...
+- name: "Install required packages on Ubuntu"
+  package:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+      - unzip
+      - tar
+      - apt-transport-https
+      - btrfs-tools
+      - libapparmor-dev
```
Random-Liu: Do we need `-dev` library? And same question for the other `-dev` library.

install/ansible/tasks/bootstrap_centos.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: "Install required packages on CentOS "
+  yum:
+    name: "{{ item }}"
+    state: latest
+    with_items:
+      - unzip
+      - tar
+      - btrfs-progs-devel
+      - device-mapper-devel
```
Random-Liu: Why do we need device mapper?

install/ansible/tasks/build.yaml
```diff
@@ -0,0 +1,47 @@
...
+- name: "Get Containerd and CRI-Containerd"
+  get_url:
+    validate_certs: "no"
+    url: "https://storage.googleapis.com/cri-containerd-staging/cri-containerd-{{ cri_containerd_release_version }}.tar.gz"
+    dest: "{{ cri_release_directory }}"
+    mode: 0755
+
+- name: "Unpack Containerd and CRI-Containerd"
+  unarchive:
+    src: "{{ cri_release_directory }}cri-containerd-{{ cri_containerd_release_version }}.tar.gz"
```
Random-Liu: copy: false

install/ansible/tasks/build.yaml
```diff
@@ -0,0 +1,47 @@
...
+    dest: "{{ cri_release_directory }}"
+
+- name: "Install the containerd and cri-containerd binaries" 
+  copy:
+    src: "{{ cri_release_directory }}usr/local/bin/{{ item }}"
```
Random-Liu: remote_src: true

install/ansible/tasks/build.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - containerd
+    - containerd-stress
+    - containerd-shim
+    - cri-containerd
+    - ctr
+
+- name: "Install runc"
+  copy:    
+    src: "{{ cri_release_directory }}usr/local/sbin/{{ item }}"
+    dest: "{{ local_sbin_dir }}{{ item }}"
```
Random-Liu: remote_src: true

install/ansible/tasks/k8.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+- name: "Install kubelet,kubeadm,kubectl (CentOS)"
+  yum: state=present name={{ item }}
+  with_items:
+    - kubelet
+    - kubeadm
+    - kubectl
+  when: ansible_distribution == "CentOS"
+
+- name: "Install kubelet,kubeadm,kubectl(Ubuntu)"
```
Random-Liu: nit: add space before (Ubuntu).

Random-Liu:
BTW, we also need this 
https://github.com/kubernetes-incubator/cri-containerd/blob/master/hack/test-e2e-node.sh#L43-L52.

At least add TODO for it.

install/ansible/templates/containerd.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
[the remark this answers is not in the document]
abhinandanpb: typically play books are organized as vars , templates , tasks etc. So I would prefer to keep that structure for anybody trying to hack in.

install/ansible/cleanup.sh
```diff
@@ -0,0 +1,6 @@
+find /var/lib/containerd | xargs sudo umount
```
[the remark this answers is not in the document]
abhinandanpb: contrib would be usually used for extensions is it not ? Like a third party vendor wants to add components to our installer ?

install/ansible/cri-containerd.yaml
abhinandanpb: yes will remove it . I was using it for crictl build.

install/ansible/cri-containerd.yaml
abhinandanpb: well it was more of binary builds being pulled :) Couldn't think of a better term.

install/ansible/cri-containerd.yaml
abhinandanpb: Hmm what about cni configs and stuff ? Does kubelet read from all the config filed in the folder ?

install/ansible/tasks/bootstrap_centos.yaml
abhinandanpb: for the ansible tar , unzip needs to be installed

install/ansible/tasks/bootstrap_ubuntu.yaml
abhinandanpb: nsenter is part of util-linux already ?

install/ansible/tasks/bootstrap_ubuntu.yaml
abhinandanpb: you will using this playbook only when ansible_distribution is ubuntu

install/ansible/cleanup.sh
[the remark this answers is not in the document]
ijc: > I am working on a kubeadm reset like script

I agree with @Random-Liu that it would be better to just fix `kubeadm` to work on CRI based systems.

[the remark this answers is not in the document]
abhinandanpb: yes all kubeadm tools are dependent on docker. So if we don't have a temporary solution. We will leave ppl trying it out in a mess. This is a complete temporary solution.

install/ansible/templates/containerd.service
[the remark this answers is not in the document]
Random-Liu: Hm. Duplication is bad. What about a symlink? Don't know whether a symlink in project is a good idea.

install/ansible/cleanup.sh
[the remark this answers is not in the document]
Random-Liu: I think ansible is only one way to setup cluster, in the future there will be others, e.g. for COS and Ubuntu in our test framework, we need cloud-init script.

I feel like it's more like a extension to our core project. :)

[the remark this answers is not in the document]
Random-Liu: > I am working on a kubeadm reset like script that should use crictl to kill all the pods. This does work and cleanup most of the states. I don't clean up netns because we shouldn't be messing with it without knowing if we created or the user created it. I will update the PR with that script change. That way we can also verify CRI is cleaning up all the states.

`crictl` is a good idea. Let's make it work for us, and may also be useful to kubeadm in the future. :)

install/ansible/cri-containerd.yaml
Random-Liu: * `install-cri-containerd.yaml` ?
* `install.yaml` ?
* `cri-containerd.yaml` ?

Bad at naming. :P But `build` does not properly reflect what the script is doing.

install/ansible/cri-containerd.yaml
Random-Liu: > Hmm what about cni configs and stuff?

What do you mean?

> Does kubelet read from all the config filed in the folder ?

I guess that's how systemd works. Or cri-o installation won't work then. :)

install/ansible/tasks/bootstrap_centos.yaml
Random-Liu: Good to know.

install/ansible/tasks/bootstrap_ubuntu.yaml
Random-Liu: Missed that. Sorry.
If so, it means that this won't support trusty, let's keep in mind that we should document it or add a comment here.

install/ansible/tasks/bootstrap_ubuntu.yaml
Random-Liu: If so, we don't need the corresponding check in centos either

install/ansible/tasks/bootstrap_centos.yaml
abhinandanpb: Somehow was under the impression that device mapper was needed for centos.

install/ansible/tasks/bootstrap_centos.yaml
abhinandanpb: http://docs.ansible.com/ansible/latest/intro_installation.html#managed-node-requirements

install/ansible/tasks/bootstrap_ubuntu.yaml
abhinandanpb: I saw that in our readme so I added this here. Let me know if we need to add a different package

install/ansible/tasks/bootstrap_ubuntu.yaml
abhinandanpb: done

install/ansible/tasks/build.yaml
abhinandanpb: I havent been really sure about copy:false and remote_src when I have used this. I will probably target it for a second round optimization ? 
I know that remote_src is usually used when the src is on a remote server but I want to optimize all these tasks into 1 task in the next iteration instead of pulling , unpacking and copying

install/ansible/cri-containerd.yaml
abhinandanpb: @Random-Liu  I dont see any harm in keeping this way. I am not sure how that works with multiple config files. This seems to be easier if someone peeps into config file for kubelet its all in one place. Unless there is any disadvantage with this approach. Here I have ensured we dont double insert the config as well so if we run the ansible playbook multiple times also it will not insert the configs

contrib/ansible/tasks/bootstrap_ubuntu.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - unzip
+      - tar
+      - apt-transport-https
+      - btrfs-tools
+      - libapparmor-dev
+      - libseccomp-dev
+      - libseccomp2
+      - socat
+      - util-linux
+# TODO: Limited support for trust for nsenter. Need to handle/verify
```
Random-Liu: trusty

contrib/ansible/templates/containerd.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=containerd container runtime
+Documentation=https://containerd.io
+After=network.target
+
+[Service]
```
Random-Liu: At least make it the same with `contrib/systemd-units`. We need systemd to restart containerd for us.
It seems that restart sec could be shorter.
We also need `OOMScoreAdjust=-999` for containerd, please update `contrib/systemd-units`, too.

abhinandanpb:
I will work on the TODO itself now.

contrib/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - name: "Add runtime args in kubelet conf"
+      lineinfile:
+                dest: "/etc/systemd/system/kubelet.service.d/10-kubeadm.conf"
+                line: "Environment=\"KUBELET_EXTRA_ARGS= --container-runtime=remote --runtime-request-timeout=15m --image-service-endpoint=/var/run/cri-containerd.sock --container-runtime-endpoint=/var/run/cri-containerd.sock\""
+                insertafter: '\[Service\]'
+      when: check_args.stdout == ""
+    
+    - name: "Start Kubelet"
+      systemd: name=kubelet daemon_reload=yes state=started enabled=yes
+#TODO This needs to be removed once we have consistent concurrent pull results
```
Random-Liu: indent

contrib/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - name: "Pre-pull pause container image"
+      shell: |
+             /usr/local/bin/ctr pull gcr.io/google_containers/pause:3.0
+             /usr/local/bin/crictl --runtime-endpoint /var/run/cri-containerd.sock \
```
Random-Liu: nit: No need to specify `image-endpoint`.

contrib/ansible/cri-containerd.yaml
```diff
@@ -0,0 +1,48 @@
...
+- hosts: all 
+  become: true
+  tasks:
+    - include_vars: vars/vars.yaml # Contains tasks variables for installer
+    - include: tasks/bootstrap_ubuntu.yaml # Contains tasks bootstrap components for ubuntu systems 
+      when: ansible_distribution == "Ubuntu"
+    - include: tasks/bootstrap_centos.yaml # Contains tasks bootstrap components for centos systems
+      when: ansible_distribution == "CentOS"
+    - include: tasks/k8s.yaml # Contains tasks kubernetes component installation
+    - include: tasks/binaries.yaml # Contains tasks for pulling containerd and cri-containerd components
```
Random-Liu: Hm, this name makes a lot more sense!

Random-Liu:
Don't know why, I could not reply to your comment directly. Something wrong with github. Send my reply here.

> I saw that in our readme so I added this here. Let me know if we need to add a different package

We need them because we need to build the binaries. For actually running the binary, I don't think we need the `-dev` packages. Please at least add TODO to figure out what we need and what not.

contrib/ansible/tasks/binaries.yaml
```diff
@@ -0,0 +1,47 @@
+---
+- name: "Create a directory to download binaries"
+  file: path={{ cri_release_directory }} state=directory
+
+- name: "Get Containerd and CRI-Containerd"
```
Random-Liu: The `remote_src` and `copy` stuffs I mentioned are still required.

contrib/ansible/tasks/bootstrap_ubuntu.yaml
```diff
@@ -0,0 +1,16 @@
...
+- name: "Install required packages on Ubuntu"
+  package:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+      - unzip
+      - tar
+      - apt-transport-https
+      - btrfs-tools
+      - libapparmor-dev
```
Random-Liu: TODO to figure out whether we need the dev libraries or what are the alternatives.

abhinandanpb:
@Random-Liu done. PTAL

contrib/ansible/templates/containerd.service
abhinandanpb: using relative path for now. Lets see if it becomes too much of a hassle, will think about alternative.

Random-Liu:
Will take another look tomorrow.

contrib/ansible/templates/containerd.service
Random-Liu: We should use the systemd units file inside the tarball. I'll do a it in a following up PR.

PR #237, opened 2017-09-12
Create host path is mount source does not exist.

Random-Liu:
Docker is doing this. And Kubelet is replying on this behavior.

Ref https://github.com/kubernetes/kubernetes/issues/52318.

/cc @abhinandanpb 

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -62,20 +62,34 @@ fi
 # For multiple GOPATHs, keep the first one only
 GOPATH=${GOPATH%%:*}
 
+# checkout_repo checks out specified repository
+# and switch to specified  version.
+# Varset:
+# 1) Repo name;
+# 2) Version.
+checkout_repo() {
+  repo=$1
+  version=$2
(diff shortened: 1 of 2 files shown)


abhinandanpb:
LGTM. This fixes the issues seen on the installer

Random-Liu:
@abhinandanpb I've manually tested this PR, and it works. Will merge after the test passes.

miaoyq:
If so, it is necessary to check the propagation of the host path.

Random-Liu:
@miaoyq Please also fix this in #230. I'll have to review your PR later. Have a demo on Thursday. :)

Random-Liu:
I sent another commit to avoid using `go get` to clone dependency repos.

CNI HEAD is broken now, using `go get` will try to build it and cause an error. The new commit use `git clone` instead.

Random-Liu:
Filed an CNI PR for the compile failure https://github.com/containernetworking/plugins/pull/73

miaoyq:
@Random-Liu  OK, I will.

PR #240, opened 2017-09-12
[Do not merge]Removing network namespace of container if its in host network

abhinandanpb:
Need to validate it for containers which expect access for host network namepspace like network plugin containers.


Random-Liu:
@abhinandanpb Do we still need this? I still don't understand there is any difference.

abhinandanpb:
I will confirm after testing it once and close it if not needed.

abhinandanpb:
will open if needed. Havent got a chance to test it yet

Issue #186, opened 2017-08-30
Support ExposedPorts and Volumes in image config.


Random-Liu:
Currently, we are ignoring those fields.
We should:
1) Support Volumes based on the definition in oci image spec.
2) Consider whether we need to support ExposedPorts.

Random-Liu:
The volumes definition in image spec is not very clear, but I assume it's expected behavior is similar with [Dockerfile VOLUME](https://docs.docker.com/engine/reference/builder/#volume).

Actually, we don't use docker volumes in Kubernetes. However, there may be user relying on image volumes (e.g. `mongo:latest` relies on volumes `/data/configdb` and `/data/db`). There 2 options for volumes:
1) Volume directory and content should be in image already, we could ignore it as what we are doing today. However, a problem with this is that if container rootfs is specified as read-only, the volume directories will be read-only, too. This will be a regression.
2) Do what docker does, create local file; copy volume content into the file, and mount the file into container as read-write. There is no regression for this approach, but we need to add helper in containerd to do this, it will be similar with `WithUserID` or `WithUsername` which require pre-access to container root filesystem.

@ijc @crosbymichael Any opinions?

ijc:
> Actually, we don't use docker volumes in Kubernetes

I think it is more accurate to say you don't explicitly use them, it is unfortunately the case that they are implictly relied on though.

Sadly I don't think your no.1 is a viable option, the one [demo](https://github.com/microservices-demo/microservices-demo) I've tried so far relies on them and I expect it to be quite widespread.

So no.2 seems like the only reasonable short term option. Perhaps longer term Kubernetes itself could spot these things and create explicit kube volumes to cover them? AIUI that would be good for accounting etc anyway (I see you filed that as an issue while I wrote the following paragraph!)

On that note the behaviour on no.2 needs to only happen if there is no user bind to the same path. The Docker logs is in [oci_linux.go](https://github.com/moby/moby/blob/master/daemon/oci_linux.go#L498..L514)

Random-Liu:
Volumes are already supported, and I don't think we have plan to support `ExposedPorts`.

PR #241, opened 2017-09-12
Add image volume support.

Random-Liu:
For image volume, currently we only:
1) Create host directory and mount into the image;
2) Change the ownership of the host directory to be the same with mount point inside the image.

What we are going to do in the future:
1) Consider whether we need to copy existing content in the image into the volume. https://github.com/kubernetes-incubator/cri-containerd/issues/186.
2) Consider whether we should add the helper into containerd https://github.com/containerd/containerd/issues/1482.
3) Consider how to properly manage these volumes in kubelet https://github.com/kubernetes/kubernetes/issues/52032.

@ijc 

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -19,6 +19,7 @@ package server
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -36,6 +37,7 @@ import (
 	"golang.org/x/sys/unix"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
+	customopts "github.com/kubernetes-incubator/cri-containerd/pkg/opts"
 	cio "github.com/kubernetes-incubator/cri-containerd/pkg/server/io"
 	containerstore "github.com/kubernetes-incubator/cri-containerd/pkg/store/container"
 	"github.com/kubernetes-incubator/cri-containerd/pkg/util"
@@ -101,26 +103,6 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 		return nil, fmt.Errorf("image %q not found", imageRef)
 	}
 
-	// Generate container runtime spec.
-	mounts := c.generateContainerMounts(getSandboxRootDir(c.rootDir, sandboxID), config)
-	spec, err := c.generateContainerSpec(id, sandboxPid, config, sandboxConfig, image.Config, mounts)
-	if err != nil {
-		return nil, fmt.Errorf("failed to generate container %q spec: %v", id, err)
-	}
-	glog.V(4).Infof("Container spec: %+v", spec)
-
-	// Set snapshotter before any other options.
-	opts := []containerd.NewContainerOpts{
-		containerd.WithSnapshotter(c.snapshotter),
-	}
-	// Prepare container rootfs. This is always writeable even if
-	// the container wants a readonly rootfs since we want to give
-	// the runtime (runc) a chance to modify (e.g. to create mount
-	// points corresponding to spec.Mounts) before making the
-	// rootfs readonly (requested by spec.Root.Readonly).
-	opts = append(opts, containerd.WithNewSnapshot(id, image.Image))
-	meta.ImageRef = image.ID
-
 	// Create container root directory.
 	containerRootDir := getContainerRootDir(c.rootDir, id)
 	if err = c.os.MkdirAll(containerRootDir, 0755); err != nil {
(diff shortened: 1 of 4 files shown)


Random-Liu:
I've verified this PR, ownership of the image volume is set properly now:
```console
# crictl --runtime-endpoint=/var/run/cri-containerd.sock exec -s 37b046da878cca2161049d90b72ee9b95bafad59d16b75b5bccbb354438b9190 cat /proc/self/mountinfo | grep data
314 295 8:1 /var/lib/cri-containerd/sandboxes/53aa9956b4b0eb15331e988d82d130b59664484e04883ea459092e9c579c08cc/hosts /etc/hosts rw,relatime - ext4 /dev/sda1 rw,data=ordered
315 295 8:1 /var/lib/cri-containerd/sandboxes/53aa9956b4b0eb15331e988d82d130b59664484e04883ea459092e9c579c08cc/resolv.conf /etc/resolv.conf rw,relatime - ext4 /dev/sda1 rw,data=ordered
317 295 8:1 /var/lib/cri-containerd/containers/37b046da878cca2161049d90b72ee9b95bafad59d16b75b5bccbb354438b9190/volumes/b3ea0f734c70106472eb183d90ad860686a5bd1722bb45878d03abdb8a8fbd73 /data/configdb rw,relatime - ext4 /dev/sda1 rw,data=ordered
318 295 8:1 /var/lib/cri-containerd/containers/37b046da878cca2161049d90b72ee9b95bafad59d16b75b5bccbb354438b9190/volumes/767a56f132b3e37769e11224dea942759cceee9d7800ff9a00d6eaa0f046e6ca /data/db rw,relatime - ext4 /dev/sda1 rw,data=ordered
# crictl --runtime-endpoint=/var/run/cri-containerd.sock exec -s 37b046da878cca2161049d90b72ee9b95bafad59d16b75b5bccbb354438b9190 ls -al /data
total 16
drwxr-xr-x 4 root    root    4096 Sep  8 08:09 .
drwxr-xr-x 1 root    root    4096 Sep 12 06:08 ..
drwxr-xr-x 2 mongodb mongodb 4096 Sep 12 06:08 configdb
drwxr-xr-x 2 mongodb mongodb 4096 Sep 12 06:08 db

```

ijc:
> 1. Consider whether we need to copy existing content in the image into the volume. #186.

I think we do, I found the sock-shop demo was relying on it (I think to preseed the database within the volume).

pkg/server/container_create.go
```diff
@@ -137,6 +119,39 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
 	}()
 
+	// Create container volumes mounts.
+	// TODO(random-liu): Add cri-containerd integration test for image volume.
+	volumeMounts := generateVolumeMounts(containerRootDir, image.Config)
+
+	// Generate container runtime spec.
+	mounts := c.generateContainerMounts(getSandboxRootDir(c.rootDir, sandboxID), config)
+
+	spec, err := c.generateContainerSpec(id, sandboxPid, config, sandboxConfig, image.Config, append(mounts, volumeMounts...))
```
ijc: I think you need to subtract from `volumeMounts` anything which is also expliclty in `mounts`, since an explicit bind mount is supposed to take precedence over the implicit ones.

ijc: BTW -- this would also mean that as soon as Kubelet starts to understand these things and provide them via `mounts` directly it will magically stop trying to do things here.

Random-Liu:
> I think we do, I found the sock-shop demo was relying on it (I think to preseed the database within the volume).

I'll do this in another PR.

pkg/server/container_create.go
Random-Liu: > I think you need to subtract from volumeMounts anything which is also expliclty in mounts, since an explicit bind mount is supposed to take precedence over the implicit ones.

Done.

> BTW -- this would also mean that as soon as Kubelet starts to understand these things and provide them via mounts directly it will magically stop trying to do things here.

We'll see how CRI will change in that case, and make change our side correspondingly. Added TODO for it.

ijc:
> > > Consider whether we need to copy existing content in the image into the volume. #186.

> > I think we do, I found the sock-shop demo was relying on it (I think to preseed the database within the volume).

> I'll do this in another PR. For this PR, I'll only copy the ownership, and see how things going.

It seems like I was mistaken -- I just fired up sock-shop with this PR and everything looked to be working ok. I was unable to login with any of the [preseeded accounts](https://microservices-demo.github.io/docs/user-accounts.html) but I made a quick patch to add copy up and it didn't seem to make a difference to that functionality. In any case I was able to manually add a new account.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/241#pullrequestreview-62368300.

Will merge after test passes.

ijc:
```
-- FAIL: TestGeneralContainerSpec (0.00s)

	Error Trace:	container_create_test.go:207

	Error:		Received unexpected error failed to set OCI bind mounts \
[«redacted by ijc»]: \
failed to resolve symlink "": lstat host-path-1: no such file or directory
	
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2158fa8]
```

For the `panic` I think `specCheck` should start with `require.NotNil(t, spec)`. Or else one or more `assert.Foo` should be `require.Foo` so the test stops at that point. Perhaps the `assert.NoError(t, err)` after calling `c.generateContainerSpec` (in several places) should instead be `require.NoError(t, err).

The failed lstat appears to be from `resolveSymbolicLink` which appears to have been added by this PR, not sure why it is suddenly showing up as a failure now, I guess these tests run on master+pr rather than just the pr and there was a clash with #242.

ijc:
I fixed the tests carried this in #247.

ijc:
> It seems like I was mistaken -- I just fired up sock-shop with this PR and everything looked to be working ok. I was unable to login with any of the preseeded accounts but I made a quick patch to add copy up and it didn't seem to make a difference to that functionality. In any case I was able to manually add a new account.

Sorry to flip-flop on this, but I tested with docker as the engine and I could login, so the issue here was that my copy-up prototype was somehow incomplete or wrong. I'm looking into why now, but we can move ahead with this PR/#247 in the meantime I think.

Random-Liu:
> Sorry to flip-flop on this, but I tested with docker as the engine and I could login, so the issue here was that my copy-up prototype was somehow incomplete or wrong. I'm looking into why now, but we can move ahead with this PR/#247 in the meantime I think.

No wonder. :p

Random-Liu:
@ijc I'll review and merge your PR instead.

Random-Liu:
Close this one, because this has been included in #247.

PR #242, opened 2017-09-12
Follow symlink for mount host path.

Random-Liu:
This is docker's behavior, we should keep it for backward compatibility.

Also ref https://github.com/kubernetes/kubernetes/issues/52318#issuecomment-328736840.


Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/242#pullrequestreview-62071569

PR #246, opened 2017-09-15
Run unit test with `sudo`

miaoyq:
In some cases, we must use root user to perform some operations in unit tests, such as `mount`.
So I think we should run unit test with `sudo`

Related to #230


Random-Liu:
@miaoyq You should use fake os in unit test, so we don't need `sudo`.

Random-Liu:
@miaoyq Should we close this PR?

miaoyq:
@Random-Liu  Yes, I will close the PR

PR #247, opened 2017-09-15
Add image volume support. 

ijc:
This carries #241 with additional fixes to the test suite from me. From @Random-Liu's original:

For image volume, currently we only:

1. Create host directory and mount into the image;
2. Change the ownership of the host directory to be the same with mount point inside the image.

What we are going to do in the future:

1. Consider whether we need to copy existing content in the image into the volume. #186.
2. Consider whether we should add the helper into containerd containerd/containerd#1482.
3. Consider how to properly manage these volumes in kubelet kubernetes/kubernetes#52032.


Random-Liu:
@ijc Thanks!

PR #248, opened 2017-09-15
return without error when stop nonexist sandbox.

sak0:
return without error when stop non-existent sandbox.

Signed-off-by: CuiHaozhi <cuihz@wise2c.com>

### DIFF
--- a/pkg/server/sandbox_stop.go
+++ b/pkg/server/sandbox_stop.go
@@ -24,6 +24,7 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	"github.com/golang/glog"
+	store "github.com/kubernetes-incubator/cri-containerd/pkg/store"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 )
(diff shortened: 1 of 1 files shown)


Random-Liu:
I think this is expected behavior. :)

sak0:
@Random-Liu Thanks for review.
I'm not very sure for this ,PTAL
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/cri/v1alpha1/runtime/api.proto#L28-L32

Random-Liu:
```
    // StopPodSandbox stops any running process that is part of the sandbox and
    // reclaims network resources (e.g., IP addresses) allocated to the sandbox.
    // If there are any running containers in the sandbox, they must be forcibly
    // terminated.
    // This call is idempotent, and must not return an error if all relevant
    // resources have already been reclaimed. kubelet will call StopPodSandbox
    // at least once before calling RemovePodSandbox. It will also attempt to
    // reclaim resources eagerly, as soon as a sandbox is not needed. Hence,
    // multiple StopPodSandbox calls are expected.
```
Yeah, `StopPodSandbox` won't return error if network namespace is already removed, or sandbox container is removed, which follows `must not return an error if all relevant resources have already been reclaimed`. It doesn't mean that i should not return error when sandbox does not exist. :)

Thanks a lot for sending the PR anyway!

sak0:
i got it, thanks :)

PR #252, opened 2017-09-15
Setting rootfs mount propagation if the mount type is rshared/shared

abhinandanpb:
This is needed by runc to mount volume for containers that expect
biderectional file updates.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -505,10 +505,17 @@ func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount, mountLabel
 		switch mount.GetPropagation() {
 		case runtime.MountPropagation_PROPAGATION_PRIVATE:
 			options = append(options, "rprivate")
+			// Since default root propogation in runc is rprivate ignore
+			// setting the root propagation
 		case runtime.MountPropagation_PROPAGATION_BIDIRECTIONAL:
 			options = append(options, "rshared")
+			g.SetLinuxRootPropagation("rshared") // nolint: errcheck
 		case runtime.MountPropagation_PROPAGATION_HOST_TO_CONTAINER:
 			options = append(options, "rslave")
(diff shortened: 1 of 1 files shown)


pkg/server/container_create.go
```diff
@@ -513,6 +513,13 @@ func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount, mountLabel
 			glog.Warningf("Unknown propagation mode for hostPath %q", mount.HostPath)
 			options = append(options, "rprivate")
 		}
+                // Set the rootfs propogation to rshared (recursive shared) for runc in the spec
```
Random-Liu: We should default root propagation to `rprivate`.

If there is `rslave` bind mount, and rootfspropagation is `rprivate`, then update it to `rslave`.
If there is `rshared` bind mount, and rootfspropagation is not `rshared`, then update it to `rshared`.

And we should put them into the cases.

abhinandanpb: default root propagation is rprivate in runc. Will update the rslave case

pkg/server/container_create.go
```diff
@@ -505,10 +505,14 @@ func addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount, mountLabel
...
 		case runtime.MountPropagation_PROPAGATION_PRIVATE:
 			options = append(options, "rprivate")
+			// Since default root propogation in runc is rprivate ignore
+			// setting the root propagation
 		case runtime.MountPropagation_PROPAGATION_BIDIRECTIONAL:
 			options = append(options, "rshared")
+			g.SetLinuxRootPropagation("rshared") // nolint: errcheck
 		case runtime.MountPropagation_PROPAGATION_HOST_TO_CONTAINER:
 			options = append(options, "rslave")
+			g.SetLinuxRootPropagation("rslave") // nolint: errcheck
```
Random-Liu: This is a loop, we should only set `rslave` when it's not `rshared` originally.

PR #255, opened 2017-09-17
Use config in service.

Random-Liu:
Address a TODO, use config in service, so as to get rid of the long arg list.


Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/255#pullrequestreview-63245773.

PR #256, opened 2017-09-17
Adding crictl to install deps

abhinandanpb:
Having crictl part of the release bundler would help in debugging until we come up with a solution for a common debug tool for both cri-containerd and containerd. 
I am planning to use this installer to address intermittent issues.


hack/install-deps.sh
```diff
@@ -125,3 +126,7 @@ checkout_repo ${CONTAINERD_PKG} ${CONTAINERD_VERSION}
 cd ${GOPATH}/src/${CONTAINERD_PKG}
 make
 ${sudo} make install -e DESTDIR=${CONTAINERD_DIR}
+
+#Install crictl
+go get -u ${CRICTL_PKG}
```
Random-Liu: You may want to make sure we are using the correct version of CRICTL.
Probably using `checkout_repo`.

abhinandanpb: I thought about it. Wasnt sure if this is strictly necessary.

Random-Liu: CRI is not stabilized yet, it may change in the future. If we don't enforce the version here, user may not be able to use this script in older branch, because CRICTL has been updated to newer CRI.

hack/install-deps.sh
```diff
@@ -52,6 +52,8 @@ CNI_CONFIG_DIR=${DESTDIR}/etc/cni/net.d
 RUNC_PKG=github.com/opencontainers/runc
 CNI_PKG=github.com/containernetworking/plugins
 CONTAINERD_PKG=github.com/containerd/containerd
+CRICTL_PKG=github.com/kubernetes-incubator/cri-tools
+CRICTL_VERSION=b3c7579714aedaf9185624ab9ca92faefc73b56b
```
Random-Liu: Use `hack/versions`, so that we don't need to update it everywhere.

You may want to change the version name to `CRITOOLS_VERSION`, so that it includes both critest and crictl.

hack/install-deps.sh
```diff
@@ -125,3 +127,9 @@ checkout_repo ${CONTAINERD_PKG} ${CONTAINERD_VERSION}
 cd ${GOPATH}/src/${CONTAINERD_PKG}
 make
 ${sudo} make install -e DESTDIR=${CONTAINERD_DIR}
+
+#Install crictl
+checkout_repo ${CRICTL_PKG} ${CRICTL_VERSION}
+cd ${GOPATH}/src/${CRICTL_PKG}
+make crictl
+${sudo} cp ${GOPATH}/bin/crictl ${CONTAINERD_DIR}/bin/
```
mikebrow: No need to copy from go/bin to usr/local/bin both should be in the path

Random-Liu: 1) In hack/install-deps.sh we don't check whether `${GOPATH}/bin` is in `PATH`;
2) We also use this to install binaries into `release-stages`

I think that's why we are doing this here. :)

mikebrow: hmm..Just checking my path is /home/mike/go/bin:/usr/local/bin for non root login and /usr/local/bin:/home/mike/go/bin for root.

So if I ran this sh one day then ran make on critools/ or go get etc.. then I'd be busted for root, but work ok when non root?

Am I missing something? Is the containerd_dir always in _output?

Random-Liu: I believe `make crictl` will also install it to your GOPATH, right?
So you'll have the same version in both GOPATH and /usr/local/bin.

mikebrow: The first time you build via make there will be no copy.  Then if you ever run install deps there will be a copy, and for a moment in time it's the same copy.  Then some time after making the copy when you build again you now have two different copies and probably won't be thinking about re-running install deps.

hack/install-deps.sh
abhinandanpb: sure makes sense.

abhinandanpb:
@mikebrow Thanks for pointing it out. Agree with you on that. 
Addressed the comments PTAL

hack/release.sh
```diff
@@ -34,6 +34,8 @@ NOSUDO=true DESTDIR=${destdir} ./hack/install-deps.sh
 # Install cri-containerd into release stage.
 make install -e DESTDIR=${destdir}
 
+# Copy crictl tool into release stage
+cp ${GOPATH}/bin/crictl ${destdir}/usr/local/bin/
```
Random-Liu: Hm, this also works. :)

Random-Liu:
Close and reopen to trigger the test.

Random-Liu:
@abhinandanpb Feel free to merge after test passes.

Random-Liu:
You are importing the new CRI validation test version, which contains test we don't support - seccomp.

mikebrow:
@Random-Liu I pushed this new version of critools with seccomp into our seccomp PR if it passes CI we can unblock this PR... or @abhinandanpb can just add "SeccompProfilePath" to the currently empty skip list over here: https://github.com/kubernetes-incubator/cri-containerd/blob/master/hack/test-cri.sh#L24

hack/release.sh
mikebrow: /LGTM

mikebrow:
@abhinandanpb thanks! I'll yank that skip out when we have merged in our dependencies to make then all work correctly!

PR #260, opened 2017-09-19
Switch to containerd extension

### LINKED ISSUE #251: Switch to containerd extension
Ref https://github.com/containerd/containerd/pull/1378.

We should switch to new containerd extension before 1.0.0-alpha.0, so as to avoid backward compatibility issue.

yanxuean:
fix #251

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/contrib/apparmor"
+	"github.com/containerd/containerd/typeurl"
 	"github.com/docker/docker/pkg/mount"
 	"github.com/golang/glog"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -54,6 +55,11 @@ const (
 	// appArmorEnabled is a flag for globally enabling/disabling apparmor profiles for containers.
 	appArmorEnabled = true // TODO (mikebrow): make these apparmor defaults configurable
 )
+
+func init() {
+	typeurl.Register(&containerstore.Metadata{},
+		"github.com/kubernetes-incubator/cri-containerd/pkg/store/container", "Metadata")
+}
 
 // CreateContainer creates a new container in the given PodSandbox.
 func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.CreateContainerRequest) (_ *runtime.CreateContainerResponse, retErr error) {
@@ -169,14 +175,6 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 		}
 	}()
 
-	metaBytes, err := meta.Encode()
-	if err != nil {
-		return nil, fmt.Errorf("failed to convert sandbox metadata: %+v, %v", meta, err)
-	}
-	labels := map[string]string{
-		containerMetadataLabel: string(metaBytes),
-	}
-
 	var specOpts []containerd.SpecOpts
 	// Set container username. This could only be done by containerd, because it needs
 	// access to the container rootfs. Pass user name to containerd, and let it overwrite
@@ -207,7 +205,8 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	opts = append(opts,
 		containerd.WithSpec(spec, specOpts...),
 		containerd.WithRuntime(defaultRuntime, nil),
-		containerd.WithContainerLabels(labels))
+		containerd.WithContainerLabels(map[string]string{containerKindLabel: containerKindContainer}),
+		containerd.WithContainerExtension(containerMetadataExtension, &meta))
 	var cntr containerd.Container
 	if cntr, err = c.client.NewContainer(ctx, id, opts...); err != nil {
 		return nil, fmt.Errorf("failed to create containerd container: %v", err)
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -95,10 +95,18 @@ const (
 )
 
 const (
-	// sandboxMetadataLabel is label name that identify metadata of sandbox in CreateContainerRequest
-	sandboxMetadataLabel = "io.cri-containerd.sandbox.metadata"
-	// sandboxMetadataLabel is label name that identify metadata of container in CreateContainerRequest
-	containerMetadataLabel = "io.cri-containerd.container.metadata"
+	// criContainerdPrefix is common prefix for cri-containerd
+	criContainerdPrefix = "io.cri-containerd"
+	// containerKindLabel is a label key indicating container is sandbox container or application container
+	containerKindLabel = criContainerdPrefix + ".kind"
+	// containerKindSandbox is a label value indicating container is sandbox container
+	containerKindSandbox = "sandbox"
+	// containerKindContainer is a label value indicating container is application container
+	containerKindContainer = "container"
+	// sandboxMetadataExtension is an extension name that identify metadata of sandbox in CreateContainerRequest
+	sandboxMetadataExtension = criContainerdPrefix + ".sandbox.metadata"
+	// containerMetadataExtension is an extension name that identify metadata of container in CreateContainerRequest
+	containerMetadataExtension = criContainerdPrefix + ".container.metadata"
 )
 
 // makeSandboxName generates sandbox name from sandbox metadata. The name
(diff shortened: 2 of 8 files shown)


yanxuean:
wait to rebase after #209 is merged.

pkg/server/helpers.go
```diff
@@ -97,7 +97,7 @@ const (
 const (
 	// sandboxMetadataLabel is label name that identify metadata of sandbox in CreateContainerRequest
 	sandboxMetadataLabel = "io.cri-containerd.sandbox.metadata"
-	// sandboxMetadataLabel is label name that identify metadata of container in CreateContainerRequest
+	// containerMetadataLabel is label name that identify metadata of container in CreateContainerRequest
```
ijc: it's not a label name any more, so perhaps "...is an extension name..."?

BTW The "sandboxMetadataLabel is label name" should really be "sandboxMetadataLabel is a label name".

yanxuean: done. Thank for review

pkg/server/container_create.go
```diff
@@ -207,7 +204,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	opts = append(opts,
 		containerd.WithSpec(spec, specOpts...),
 		containerd.WithRuntime(defaultRuntime, nil),
-		containerd.WithContainerLabels(labels))
+		containerd.WithContainerExtension(containerMetadataExtension, &meta))
```
Random-Liu: We still need to handle versioning ourselves, I think.

Containerd is using json marshal/unmarshal for non-proto message now. Could we update `Encode`/`Decode` in container/sandbox metadata and container status to `MarshalJSON`, and `UnmarshalJSON`? So that golang will call our customized marshal/unmarshal functions which should automatically handle versioning.

pkg/server/helpers.go
```diff
@@ -95,10 +95,10 @@ const (
...
 
 const (
-	// sandboxMetadataLabel is label name that identify metadata of sandbox in CreateContainerRequest
-	sandboxMetadataLabel = "io.cri-containerd.sandbox.metadata"
-	// sandboxMetadataLabel is label name that identify metadata of container in CreateContainerRequest
-	containerMetadataLabel = "io.cri-containerd.container.metadata"
+	// sandboxMetadataExtension is a extension name that identify metadata of sandbox in CreateContainerRequest
+	sandboxMetadataExtension = "io.cri-containerd.sandbox.metadata"
+	// containerMetadataExtension is a extension name that identify metadata of container in CreateContainerRequest
+	containerMetadataExtension = "io.cri-containerd.container.metadata"
```
Random-Liu: We still need a label to select containers and sandboxes.

Can we add a label `containerKindLabel = "io.cri-containerd.kind"`? And we should put "sandbox" on sandbox container, and "container" on application container.

pkg/server/container_create.go
```diff
@@ -55,6 +56,10 @@ const (
 	appArmorEnabled = true // TODO (mikebrow): make these apparmor defaults configurable
 )
 
+func init() {
+	typeurl.Register(&containerstore.Metadata{}, "github.com/kubernetes-incubator/cri-containerd/pkg/store/container", "Metadata")
```
Random-Liu: nit: Line too long, add new line?

pkg/server/sandbox_run.go
```diff
@@ -35,6 +36,10 @@ import (
 	"github.com/kubernetes-incubator/cri-containerd/pkg/util"
 )
 
+func init() {
+	typeurl.Register(&sandboxstore.Metadata{}, "github.com/kubernetes-incubator/cri-containerd/pkg/store/sandbox", "Metadata")
```
Random-Liu: ditto.

pkg/server/container_create.go
Random-Liu: And please add unit test to make sure `typeurl.Marshal` and `typeurl.Unmarshal` does properly handle the versioning.

pkg/server/container_create.go
yanxuean: done

pkg/server/container_create.go
yanxuean: done

pkg/server/helpers.go
yanxuean: done

pkg/server/sandbox_run.go
yanxuean: done

pkg/server/helpers.go
yanxuean: What do you mean?   Should we use  containerKindLabel  but containerMetadataExtension?

Random-Liu: I mean instead of extension, we still need a label to indicate whether a container is sandbox container or regular container.

We could add a label `io.cri-containerd.kind` on each container, `io.cri-containerd.kind=sandbox` means it is a sandbox container, `io.cri-containerd.kind=container` means it is a regular container.

pkg/store/container/metadata.go
```diff
@@ -34,10 +34,12 @@ const metadataVersion = "v1" // nolint
 // nolint
 type versionedMetadata struct {
 	// Version indicates the version of the versioned container metadata.
-	Version string
-	Metadata
+	Version  string
+	Metadata metadataInternal
```
Random-Liu: Why do we need this?

yanxuean: There is a recursive call  if we use Metadata in versionedMetadata derectly.

pkg/server/helpers.go
yanxuean: Do you mean we shoule add both lable and extension?

yanxuean: For example
```
opts := []containerd.NewContainerOpts{
		containerd.WithSnapshotter(c.config.ContainerdSnapshotter),
		containerd.WithNewSnapshot(id, image.Image),
		containerd.WithSpec(spec, specOpts...),
		containerd.WithContainerLabels(map[string]string{containerKindLabel: "sandbox"})
		containerd.WithContainerExtension(sandboxMetadataExtension, &sandbox.Metadata),
		containerd.WithRuntime(defaultRuntime, nil)}
	container, err := c.client.NewContainer(ctx, id, opts...)
```

pkg/store/container/metadata_test.go
```diff
@@ -37,17 +37,35 @@ func TestMetadataEncodeDecode(t *testing.T) {
...
-	data, err := meta.Encode()
-	assert.NoError(err)
 	newMeta := &Metadata{}
-	assert.NoError(newMeta.Decode(data))
+	newVerMeta := &versionedMetadata{}
+
+	data, err := json.Marshal(meta)
+	assert.NoError(err)
+	assert.NoError(json.Unmarshal(data, newMeta))
+	assert.Equal(meta, newMeta)
```
Random-Liu: We don't know whether `MarshalJSON` or `UnmarshalJSON` is actually called during this.
The test is a little bit confusing.

How about this?
1) Your first case. (verify that data could be stored and recovered)
2) Make sure `json.Marshal(Metadata)` and `json.Marshal(versionedMetadata)` get the same bytes result; (verify that data is stored as versioned data, case1) has proved that versioned data could be recovered)
3) `json.Marshal(unsupportedVersionedMetadata)` and then `json.Unmarshal(data)` should get an error; (verify that unsupported versioned data won't be unmarshalled)

We may want to add some `t.Logf` to make the test more clear.

Random-Liu:
LGTM overall with a comment for the test.

pkg/store/container/metadata.go
Random-Liu: Make sense.

pkg/server/helpers.go
```diff
@@ -95,10 +95,16 @@ const (
 )
 
 const (
-	// sandboxMetadataLabel is label name that identify metadata of sandbox in CreateContainerRequest
-	sandboxMetadataLabel = "io.cri-containerd.sandbox.metadata"
-	// sandboxMetadataLabel is label name that identify metadata of container in CreateContainerRequest
-	containerMetadataLabel = "io.cri-containerd.container.metadata"
+	// containerKindLabel is a label key indicating container is sandbox container or application container
+	containerKindLabel = "io.cri-containerd.kind"
```
abhinandanpb: minor nit: Can we add a constant prefix = "io.cri-containerd"
And then use this for all the constants ?
for eg: containerKindLabel= prefix + "kind"

yanxuean: done

pkg/store/container/metadata_test.go
yanxuean: done

Random-Liu:
LGTM

Issue #238, opened 2017-09-12
TODO CRI Validation / Integration Test List


Random-Liu:
Related to https://github.com/kubernetes-incubator/cri-containerd/pull/237.

Ref https://github.com/kubernetes/kubernetes/issues/52318.

When host path doesn't exist, CRI expects runtime to create it. We should have this tested in CRI validation test.

Random-Liu:
Missing integration test:
Volume copy-up behavior.

Random-Liu:
Missing integration test:
Restart recovery test.

Random-Liu:
We've added restart test for sandbox, trunc index test and some other integration tests.

Move this to next release, we should continue adding full restart test and improve our integration test.

Random-Liu:
We've finished all the integration test listed above. Close this one.

PR #261, opened 2017-09-19
Implement volume copy up.

ijc:
This pulls in and uses github.com/docker/docker/pkg/chrootarchive for the
actual copy up which is some battle hardened code to unpack avoiding things
like symlink traversal security issues.

However it does pull in a pretty huge pile of vendoring, including
github.com/docker/docker/pkg/reexec which we must then call at startup. It's
not immediately clear that this tradeoff is the correct one.

Relates to #186. With this change I can successfully login as one of the [predefined users](https://microservices-demo.github.io/docs/user-accounts.html) on the [sock-shop demo](https://github.com/microservices-demo/microservices-demo), previously this failed.

### DIFF
--- a/pkg/opts/container.go
+++ b/pkg/opts/container.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/containers"
+	"github.com/docker/docker/pkg/chrootarchive"
 	"github.com/docker/docker/pkg/system"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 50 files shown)


pkg/opts/container.go
```diff
@@ -56,14 +57,36 @@ func WithVolumes(volumeMounts map[string]string) containerd.NewContainerOpts {
...
+		return err
+	}
+	if len(volList) > 0 {
+		srcList, err := ioutil.ReadDir(destination)
+		if err != nil {
+			return err
+		}
+		if len(srcList) == 0 {
+			// If the source volume is empty, copies files from the root into the volume
+			if err := chrootarchive.NewArchiver(nil).CopyWithTar(source, destination); err != nil {
```
Random-Liu: Just for my education. Why do we need to chroot archive? I don't quite understand it.

pkg/opts/container.go
```diff
@@ -56,14 +57,36 @@ func WithVolumes(volumeMounts map[string]string) containerd.NewContainerOpts {
...
 
+// copyExistingContents copies from the source to the destination and
+// ensures the ownership is appropriately set.
+func copyExistingContents(source, destination string) error {
+	volList, err := ioutil.ReadDir(source)
+	if err != nil {
+		return err
+	}
+	if len(volList) > 0 {
+		srcList, err := ioutil.ReadDir(destination)
```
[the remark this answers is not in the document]
ijc: It confused me too but I stuck with it to keep in sync with Moby. Will fix.

pkg/opts/container.go
ijc: Mainly just:

> This pulls in and uses github.com/docker/docker/pkg/chrootarchive for the actual copy up which is some battle hardened code to unpack avoiding things like symlink traversal security issues.

This wouldn't be all that hard to implement directly but it does have some slightly tricky corner cases which have already been worked through in that code base. I think the `chroot` aspect of it is just to mitigate the security impact of any bugs.

It also covers a bunch of stuff I wouldn't have thought of myself, such as preserving xattrs and proper handling of hardlinks etc.

Random-Liu: @ijc Good to know. Thanks for explaining this!

Random-Liu:
LGTM

Random-Liu:
We need an integration test for this behavior. Tracked this in #238.

PR #262, opened 2017-09-19
Do not attempt to retrieve IP from host network namespace

### LINKED ISSUE #245: Invalid invocation of nsenter without netns for GetPodNetworkStatus
I was doing some `strace` based debugging and spotted:
```
execve("/usr/bin/nsenter", ["/usr/bin/nsenter", "--net=", "-F", "--", "ip", "-o", "-4", "addr", "show", "dev", "eth0", "scope", "global"], [/* 2 vars */]) = 0
```
Which is invalid syntax due to the lack of the net arguments string option:
```
$ nsenter --net= -F -- echo foo
nsenter: cannot open : No such file or directory
```
And with sufficient logging enabled I see this in the `cri-containerd` logs too:
```
I0914 10:49:54.859715     564 sandbox_status.go:73] GetPodNetworkStatus returns error: Unexpected command output nsenter: cannot open : No such file or directory
 with error: exit status 1
```

Note that I do also see correct invocations of the form:
```
/usr/bin/nsenter --net=/var/run/netns/cni-5e8acebe-810d-c1b9-ced0-47be2f312fa8 -F -- <ip commandline>
```

This seems to come from [`criContainerdService.PodSandboxStatus`](https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/sandbox_status.go#L62...L74).

I suspect it might be due to containers using the host netns?

I'm unsure if the issue is that `sandbox.NetNSPath` is wrong/missing in the first place or if instead the code should be conditional and simply skip this code if there is no `NetNSPath` (hence this is an issue rather than a PR, given some guidance on the expected behaviour I can make a PR though).

ijc:
Since sandboxes which use the host network have no network namespace path this
would result in an invalid invocation of nsenter.

Rework the fetching of the sandbox to take this into account and also avoid
trying to get an IP when the network plugin is not yet ready.

Fixes #245.

### DIFF
--- a/pkg/server/sandbox_status.go
+++ b/pkg/server/sandbox_status.go
@@ -58,24 +58,48 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
 			state = runtime.PodSandboxState_SANDBOX_READY
 		}
 	}
-	config := sandbox.Config
-	podNetwork := ocicni.PodNetwork{
-		Name:         config.GetMetadata().GetName(),
-		Namespace:    config.GetMetadata().GetNamespace(),
-		ID:           id,
-		NetNS:        sandbox.NetNSPath,
-		PortMappings: toCNIPortMappings(config.GetPortMappings()),
-	}
-	ip, err := c.netPlugin.GetPodNetworkStatus(podNetwork)
+
+	ip, err := c.getIP(sandbox)
 	if err != nil {
-		// Ignore the error on network status
-		ip = ""
-		glog.V(4).Infof("GetPodNetworkStatus returns error: %v", err)
+		return nil, fmt.Errorf("failed to get sandbox ip: %v", err)
 	}
 
 	createdAt := sandbox.Container.Info().CreatedAt
 	status := toCRISandboxStatus(sandbox.Metadata, state, createdAt, ip)
 	return &runtime.PodSandboxStatusResponse{Status: status}, nil
+}
+
+func (c *criContainerdService) getIP(sandbox sandboxstore.Sandbox) (string, error) {
+	config := sandbox.Config
+
+	if config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
+		// For sandboxes using the host network we are not
+		// responsible for reporting the IP.
+		return "", nil
+	}
+
+	if err := c.netPlugin.Status(); err != nil {
+		// If the network is not ready then there is nothing to report.
+		glog.V(4).Infof("getIP: unable to get sandbox %q network status: network plugin not ready.", sandbox.ID)
+		return "", nil
+	}
+
+	podNetwork := ocicni.PodNetwork{
+		Name:         config.GetMetadata().GetName(),
+		Namespace:    config.GetMetadata().GetNamespace(),
+		ID:           sandbox.ID,
+		NetNS:        sandbox.NetNSPath,
+		PortMappings: toCNIPortMappings(config.GetPortMappings()),
+	}
+
+	ip, err := c.netPlugin.GetPodNetworkStatus(podNetwork)
+	if err == nil {
+		return ip, nil
+	}
+
+	// Ignore the error on network status
+	glog.V(4).Infof("getIP: failed to read sandbox %q IP from plugin: %v", sandbox.ID, err)
+	return "", nil
 }
 
 // toCRISandboxStatus converts sandbox metadata into CRI pod sandbox status.


pkg/server/sandbox_status.go
```diff
@@ -58,24 +58,50 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
...
+	return &runtime.PodSandboxStatusResponse{Status: status}, nil
+}
+
+func (c *criContainerdService) getIP(sandbox sandboxstore.Sandbox) (string, error) {
 	config := sandbox.Config
+
+	if config.GetLinux().GetSecurityContext().GetNamespaceOptions().GetHostNetwork() {
+		// For sandboxes using the host network we are not
+		// responsible for reporting the IP.
+		glog.V(10).Infof("getIP: %s: sandbox is using host network", sandbox.ID)
```
Random-Liu: This is quite normal, maybe not log it?

pkg/server/sandbox_status.go
```diff
@@ -58,24 +58,50 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
...
+		return "", nil
+	}
+
+	if err := c.netPlugin.Status(); err != nil {
+		// If the network is not ready then there is nothing to report.
+		glog.V(10).Infof("getIP: %s: network is not ready", sandbox.ID)
```
Random-Liu: This should not usually happen, because we should not be running pod which requires pod network when network plugin is not ready.

I think a V(4) log is ok here.

nit: `glog.V(4).Infof("getIP: network %q is not ready", sandbox.ID)`

pkg/server/sandbox_status.go
```diff
@@ -58,24 +58,50 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
...
+		glog.V(10).Infof("getIP: %s: got %q from plugin", sandbox.ID, ip)
+		return ip, nil
 	}
 
-	createdAt := sandbox.Container.Info().CreatedAt
-	status := toCRISandboxStatus(sandbox.Metadata, state, createdAt, ip)
-	return &runtime.PodSandboxStatusResponse{Status: status}, nil
+	// Ignore the error on network status
+	glog.V(4).Infof("getIP: %s: failed to read pod IP from plugin: %v", sandbox.ID, err)
+	return "", err
```
Random-Liu: You are not ignoring it.

And nit: `glog.V(4).Infof("getIP: failed to read sandbox %q IP from plugin: %v", sandbox.ID, err)`

pkg/server/sandbox_status.go
```diff
@@ -58,24 +58,50 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
...
 		PortMappings: toCNIPortMappings(config.GetPortMappings()),
 	}
+
 	ip, err := c.netPlugin.GetPodNetworkStatus(podNetwork)
-	if err != nil {
-		// Ignore the error on network status
-		ip = ""
-		glog.V(4).Infof("GetPodNetworkStatus returns error: %v", err)
+	if err == nil {
+		glog.V(10).Infof("getIP: %s: got %q from plugin", sandbox.ID, ip)
```
Random-Liu: No need to log I think, we'll log the returned pod status with level 5 anyway.

pkg/server/sandbox_status.go
ijc: Yeah, I put it in to validate my code was working then wasn't sure whether to keep it or not, so I just lowered the severity. Will drop this and adjust the others as you suggest.

ijc:
I changed everything as you suggested but I was wondering about a nit WRT the error message in:
```
		// If the network is not ready then there is nothing to report.
		glog.V(4).Infof("getIP: network %q is not ready", sandbox.ID)
		return "", nil
	}
```

The error in this case is that the global network plugin state is not ready, not that something about the specific network is not ready (is that right?). So would "getIP: unable to get network %q status: network plugin not ready." be a more accurate message?

Random-Liu:
> The error in this case is that the global network plugin state is not ready, not that something about the specific network is not ready (is that right?). So would "getIP: unable to get network %q status: network plugin not ready." be a more accurate message?

Sorry, my bad, maybe:
`getIP: unable to get sandbox %q network status: network plugin not ready.`

LGTM please update this, and I'll merge it tomorrow after test passes.

ijc:
Message changed.

Before I pushed these updates the tests were failing but I hadn't looked at why. I'll keep an eye on the results of this new batch and fixup as necessary.

Random-Liu:
@ijc I checked just now. Seems like a flake. We need to stablize the test after all features are finished. :0

PR #265, opened 2017-09-20
Adding container metrics support

abhinandanpb:
Built on top of #257 
Review by commits

1) Implements ListContainerStats and ContainerStats to collect metrics from containerd
2) Added couple of integration tests to verify stat collection

This commit does not include caching. Caching will be implemented if the benchmark results go northwards.

### DIFF
--- a/integration/container_stats_test.go
+++ b/integration/container_stats_test.go
@@ -1,0 +1,322 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+// Test to verify for a container ID
+func TestContainerStats(t *testing.T) {
+	t.Logf("Create a pod config and run sandbox container")
+	sbConfig := PodSandboxConfig("sandbox1", "stats")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+	t.Logf("Create a container config and run container in a pod")
+	containerConfig := ContainerConfig(
+		"container1",
+		pauseImage,
+		WithTestLabels(),
+		WithTestAnnotations(),
+	)
+	cn, err := runtimeService.CreateContainer(sb, containerConfig, sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.RemoveContainer(cn))
+	}()
+	require.NoError(t, runtimeService.StartContainer(cn))
+	defer func() {
+		assert.NoError(t, runtimeService.StopContainer(cn, 10))
+	}()
+
+	t.Logf("Fetch stats for container")
+	var s *runtime.ContainerStats
+	require.NoError(t, Eventually(func() (bool, error) {
+		s, err = runtimeService.ContainerStats(cn)
+		if err != nil {
+			return false, err
+		}
+		if s.GetWritableLayer().GetUsedBytes().GetValue() != 0 &&
+			s.GetWritableLayer().GetInodesUsed().GetValue() != 0 {
+			return true, nil
+		}
+		return false, nil
+	}, time.Second, 30*time.Second))
+
+	t.Logf("Verify stats received for container %q", cn)
+	testStats(t, s, containerConfig)
+}
+
+// Test to verify filtering without any filter
+func TestContainerListStats(t *testing.T) {
+	t.Logf("Create a pod config and run sandbox container")
+	sbConfig := PodSandboxConfig("running-pod", "statsls")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+	t.Logf("Create a container config and run containers in a pod")
+	containerConfigMap := make(map[string]*runtime.ContainerConfig)
+	for i := 0; i < 3; i++ {
+		cName := fmt.Sprintf("container%d", i)
+		containerConfig := ContainerConfig(
+			cName,
+			pauseImage,
+			WithTestLabels(),
+			WithTestAnnotations(),
+		)
+		cn, err := runtimeService.CreateContainer(sb, containerConfig, sbConfig)
+		require.NoError(t, err)
+		containerConfigMap[cn] = containerConfig
+		defer func() {
+			assert.NoError(t, runtimeService.RemoveContainer(cn))
+		}()
+		require.NoError(t, runtimeService.StartContainer(cn))
+		defer func() {
+			assert.NoError(t, runtimeService.StopContainer(cn, 10))
+		}()
+	}
+
+	t.Logf("Fetch all container stats")
+	var stats []*runtime.ContainerStats
+	require.NoError(t, Eventually(func() (bool, error) {
+		stats, err = runtimeService.ListContainerStats(&runtime.ContainerStatsFilter{})
+		if err != nil {
+			return false, err
+		}
+		for _, s := range stats {
+			if s.GetWritableLayer().GetUsedBytes().GetValue() == 0 &&
+				s.GetWritableLayer().GetInodesUsed().GetValue() == 0 {
+				return false, nil
+			}
+		}
+		return true, nil
+	}, time.Second, 30*time.Second))
+
+	t.Logf("Verify all container stats")
+	for _, s := range stats {
+		testStats(t, s, containerConfigMap[s.GetAttributes().GetId()])
+	}
+}
+
+// Test to verify filtering given a specific container ID
+// TODO Convert the filter tests into table driven tests and unit tests
+func TestContainerListStatsWithIdFilter(t *testing.T) {
+	t.Logf("Create a pod config and run sandbox container")
+	sbConfig := PodSandboxConfig("running-pod", "statsls")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+	t.Logf("Create a container config and run containers in a pod")
+	containerConfigMap := make(map[string]*runtime.ContainerConfig)
+	for i := 0; i < 3; i++ {
+		cName := fmt.Sprintf("container%d", i)
+		containerConfig := ContainerConfig(
+			cName,
+			pauseImage,
+			WithTestLabels(),
+			WithTestAnnotations(),
+		)
+		cn, err := runtimeService.CreateContainer(sb, containerConfig, sbConfig)
+		containerConfigMap[cn] = containerConfig
+		require.NoError(t, err)
+		defer func() {
+			assert.NoError(t, runtimeService.RemoveContainer(cn))
+		}()
+		require.NoError(t, runtimeService.StartContainer(cn))
+		defer func() {
+			assert.NoError(t, runtimeService.StopContainer(cn, 10))
+		}()
+	}
+
+	t.Logf("Fetch container stats for each container with Filter")
+	var stats []*runtime.ContainerStats
+	for id := range containerConfigMap {
+		require.NoError(t, Eventually(func() (bool, error) {
+			stats, err = runtimeService.ListContainerStats(
+				&runtime.ContainerStatsFilter{Id: id})
+			if err != nil {
+				return false, err
+			}
+			if len(stats) != 1 {
+				return false, fmt.Errorf("unexpected stats length")
+			}
+			if stats[0].GetWritableLayer().GetUsedBytes().GetValue() != 0 &&
+				stats[0].GetWritableLayer().GetInodesUsed().GetValue() != 0 {
+				return true, nil
+			}
+			return false, nil
+		}, time.Second, 30*time.Second))
+
+		t.Logf("Verify container stats for %s", id)
+		for _, s := range stats {
+			require.Equal(t, s.GetAttributes().GetId(), id)
+			testStats(t, s, containerConfigMap[id])
+		}
+	}
+}
+
+// Test to verify filtering given a specific Sandbox ID. Stats for
+// all the containers in a pod should be returned
+func TestContainerListStatsWithSandboxIdFilter(t *testing.T) {
+	t.Logf("Create a pod config and run sandbox container")
+	sbConfig := PodSandboxConfig("running-pod", "statsls")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+	t.Logf("Create a container config and run containers in a pod")
+	containerConfigMap := make(map[string]*runtime.ContainerConfig)
+	for i := 0; i < 3; i++ {
+		cName := fmt.Sprintf("container%d", i)
+		containerConfig := ContainerConfig(
+			cName,
+			pauseImage,
+			WithTestLabels(),
+			WithTestAnnotations(),
+		)
+		cn, err := runtimeService.CreateContainer(sb, containerConfig, sbConfig)
+		containerConfigMap[cn] = containerConfig
+		require.NoError(t, err)
+		defer func() {
+			assert.NoError(t, runtimeService.RemoveContainer(cn))
+		}()
+		require.NoError(t, runtimeService.StartContainer(cn))
+		defer func() {
+			assert.NoError(t, runtimeService.StopContainer(cn, 10))
+		}()
+	}
+
+	t.Logf("Fetch container stats for each container with Filter")
+	var stats []*runtime.ContainerStats
+	require.NoError(t, Eventually(func() (bool, error) {
+		stats, err = runtimeService.ListContainerStats(
+			&runtime.ContainerStatsFilter{PodSandboxId: sb})
+		if err != nil {
+			return false, err
+		}
+		if len(stats) != 3 {
+			return false, fmt.Errorf("unexpected stats length")
+		}
+		if stats[0].GetWritableLayer().GetUsedBytes().GetValue() != 0 &&
+			stats[0].GetWritableLayer().GetInodesUsed().GetValue() != 0 {
+			return true, nil
+		}
+		return false, nil
+	}, time.Second, 30*time.Second))
+	t.Logf("Verify container stats for sandbox %q", sb)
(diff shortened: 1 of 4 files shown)


cmd/cri-containerd/options/options.go
```diff
@@ -31,6 +31,8 @@ const configFilePathArgName = "config"
 
 // ContainerdConfig contains config related to containerd
 type ContainerdConfig struct {
+	// ContainerdRootDir is the root directory path for containerd.
+	ContainerdRootDir string
```
yanxuean: toml

Random-Liu: Will do in my PR. Thanks!

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,75 @@ limitations under the License.
...
+	metrics := s.(*cgroups.Metrics)
+	cs.Cpu = &runtime.CpuUsage{Timestamp: stats.Timestamp.Unix(), UsageCoreNanoSeconds: &runtime.UInt64Value{Value: metrics.CPU.Usage.Total}}
+	cs.Memory = &runtime.MemoryUsage{Timestamp: stats.Timestamp.Unix(), WorkingSetBytes: &runtime.UInt64Value{metrics.Memory.Usage.Usage}}
+
+	cs.Attributes = &runtime.ContainerAttributes{Id: stats.ID}
+	return nil
+}
+func (c *criContainerdService) buildTaskMetricsRequest(r *runtime.ListContainerStatsRequest) (tasks.MetricsRequest, error) {
+	var req tasks.MetricsRequest
+	if r == nil || r.Filter == nil {
```
Random-Liu: nit: r.GetFilter().GetId() and r.GetFilter().GetPodSandboxId()

These getters could help you avoid the nil check.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,75 @@ limitations under the License.
...
+		return req, nil
+	}
```
Random-Liu: I believe the relationship between different filters in CRI is `AND`.

What about containerd?
1) If different filters in containerd is `OR`, then it's not correct.
2) If different filters in containerd is `AND`, then we'll get no container, because there is no single container has all the ids.

As is mentioned in our discussion, I'm fine with list all containerstats and do filter afterwards, if the filter doesn't work that well :) Up to you to decide.

integration/container_stats_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+const (
+	defaultImage = "gcr.io/google_containers/pause:3.0"
```
Random-Liu: This const is global, let's make the name more specific.

integration/container_stats_test.go
```diff
@@ -0,0 +1,88 @@
...
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+	t.Logf("Fetch stats for sandbox container")
+	s, err := runtimeService.ContainerStats(sb)
+	t.Logf("Verify stats received for sandbox container")
+	require.NoError(t, err)
+	testStats(t, s)
```
Random-Liu: Why this works? We don't store sandbox container id in container store, right?

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,75 @@ limitations under the License.
...
 )
 
 // ListContainerStats returns stats of all running containers.
 func (c *criContainerdService) ListContainerStats(ctx context.Context, in *runtime.ListContainerStatsRequest) (*runtime.ListContainerStatsResponse, error) {
-	return nil, errors.New("not implemented")
+	request, err := c.buildTaskMetricsRequest(in)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build metrics request: %v", err)
+	}
+	resp, err := c.taskService.Metrics(context.Background(), &request)
```
Random-Liu: We may want to add a filter to filter `containerKindLabel=container`. This will also include sandbox containers in the list.

I need to discuss with @yujuhong to finalize whether we should include sandbox container stats in `ListContainerStats`.

integration/container_stats_test.go
```diff
@@ -0,0 +1,88 @@
...
+	stats, err := runtimeService.ListContainerStats(&runtime.ContainerStatsFilter{})
+	require.NoError(t, err)
+	require.NotEmpty(t, stats)
+	t.Logf("Verify all container stats")
+	for _, s := range stats {
+		testStats(t, s)
+	}
+}
+
+func testStats(t *testing.T, s *runtime.ContainerStats) {
```
Random-Liu: Will this be flaky? Is it possible that a container gets 0 cpu seconds? Do we need to use `Eventually` to check this?

Just my question. I'm not sure about the answer. :) I think 0 cpu seconds shouldn't be possible, but you may have some more ideas. :)

pkg/server/container_stats_list.go
Random-Liu: Discussed with @yujuhong, let's exclude the sandbox container for now. We need to figure out a way to handle pod level overhead.

integration/container_stats_test.go
```diff
@@ -0,0 +1,88 @@
...
+	require.NotEmpty(t, s.Attributes)
```
Random-Liu: nit: Use `Getter`, which will save many lines. :)

`Getter` already handles the case that the original object is nil, it will just return zero value.

pkg/server/container_stats_list.go
abhinandanpb: ok makes sense. I wasn't sure about it.

abhinandanpb: Circling back again. If we exclude sandbox container. The aggregated stats for a pod might be misleading right ? Or is the reason that the stats for sandbox containers is negligible and consistent for all pods so it doesnt factor into decision making ?

Random-Liu: @abhinandanpb This is a problem is CRI now.

We have pod level cgroup, and ideally Kubelet should make resource management decision based on resource usage got from pod level cgroup. And the container metrics are only for user who cares about how much resource their workloads (application containers) are using.

However, today Kubelet is not using pod level stats yet. It is still making decision based on container metrics. This means that the problem you mentioned does exist, kubelet is not looking at all resource usage inside the pod (sandbox container, container-shim etc.).

This is an issue needs to be addressed. We should:
1) Either change Kubelet to make resource management decision based on pod level stats, which  include everything inside the pod cgroup.
2) Or change CRI to include pod sandbox extra overhead.

1) is preferable now, and we'll probably do it next Quarter.

For now, we could only provide application container metrics based on current CRI. We'll address this in 1.9. We are going to refactor cadvisor, and pod cgroup metrics will be part of the effort.

abhinandanpb: Got it. Thanks for the explaination.

pkg/server/container_stats.go
```diff
@@ -17,15 +17,34 @@ limitations under the License.
...
-
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 )
 
 // ContainerStats returns stats of the container. If the container does not
 // exist, the call returns an error.
 func (c *criContainerdService) ContainerStats(ctx context.Context, in *runtime.ContainerStatsRequest) (*runtime.ContainerStatsResponse, error) {
-	return nil, errors.New("not implemented")
+	// Validate the stats request
+	if in == nil || in.GetContainerId() == "" {
```
abhinandanpb: will get rid of this

Random-Liu:
@abhinandanpb Dependent PR is merged. Please rebase.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+		Annotations: cnt.Config.GetAnnotations(),
+	}
+	return nil
+}
+
+// buildTaskMetricsRequest constructs a taskMetricsRequest based on
+// the information in the stats request and the containerStore
+func (c *criContainerdService) buildTaskMetricsRequest(r *runtime.ListContainerStatsRequest) (tasks.MetricsRequest, error) {
+	var req tasks.MetricsRequest
+	if r == nil || r.Filter == nil {
```
Random-Liu: nit: `r.GetFilter() == nil`

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+	cs.Attributes = &runtime.ContainerAttributes{
+		Id:          stats.ID,
+		Metadata:    cnt.Config.GetMetadata(),
+		Labels:      cnt.Config.GetLabels(),
+		Annotations: cnt.Config.GetAnnotations(),
+	}
+	return nil
+}
+
+// buildTaskMetricsRequest constructs a taskMetricsRequest based on
```
Random-Liu: nit: task.MetricsRequest.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+		}
+		candidateContainers = append(candidateContainers, c.ID)
+	}
+	for _, id := range candidateContainers {
+		req.Filters = append(req.Filters, "id=="+id)
+	}
+	return req, nil
+}
+
+func matchLabelSelector(selector map[string]string, labels map[string]string) bool {
```
Random-Liu: nit: `selector, labels map[string]string`.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+		Timestamp: stats.Timestamp.Unix(),
+		StorageId: &runtime.StorageIdentifier{
+			Uuid: c.imageFSUUID,
+		},
+		UsedBytes:  &runtime.UInt64Value{usedBytes},
+		InodesUsed: &runtime.UInt64Value{inodesUsed},
+	}
+
+	// Get the container from store and extract the attributes
+	cnt, err := c.containerStore.Get(stats.ID)
```
Random-Liu: It's possible that a container shows up during list, but get deleted before this point. In that case, we should not return error for the whole `ListContainerStats`.

We should either pass in the container config into this function, or distinguish not found error here.

pkg/server/container_stats.go
Random-Liu: nit: `in.GetContainerId() == ""` is enough

integration/container_stats_test.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+		return false, nil
+	}, time.Second, 30*time.Second))
+	t.Logf("Verify container stats for sandbox %q and container %q filter", sb, containers[0])
+	for _, s := range stats {
+		testStats(t, s)
+	}
+}
+
+func testStats(t *testing.T, s *runtime.ContainerStats) {
```
Random-Liu: Pass in container config for verification? Even for `ListContainerStats`, we could find corresponding container config inside the list and pass in.

pkg/server/container_stats.go
```diff
@@ -17,15 +17,34 @@ limitations under the License.
...
+	}
+	request := &tasks.MetricsRequest{Filters: []string{"id==" + containerID}}
+	resp, err := c.taskService.Metrics(ctx, request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch metrics for tasks: %v", err)
+	}
+
+	var cs runtime.ContainerStats
+	if err := c.getContainerMetrics(resp.Metrics[0], &cs); err != nil {
+		return nil, fmt.Errorf("failed to decode container metrics %v", err)
```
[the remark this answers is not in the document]
abhinandanpb: done

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+		UsageCoreNanoSeconds: &runtime.UInt64Value{metrics.CPU.Usage.Total},
+	}
+	cs.Memory = &runtime.MemoryUsage{
+		Timestamp:       stats.Timestamp.Unix(),
+		WorkingSetBytes: &runtime.UInt64Value{metrics.Memory.Usage.Usage},
+	}
+
+	var usedBytes, inodesUsed uint64
+	sn, err := c.snapshotStore.Get(stats.ID)
+	// If snapshotstore doesnt have cached snapshotStore
```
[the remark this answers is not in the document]
abhinandanpb: done

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+	// set WritableLayer to zero
```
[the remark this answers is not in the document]
abhinandanpb: done

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,127 @@ limitations under the License.
...
+	if err == nil {
+		inodesUsed = sn.Size
+		usedBytes = sn.Inodes
+	}
+	cs.WritableLayer = &runtime.FilesystemUsage{
+		Timestamp: stats.Timestamp.Unix(),
```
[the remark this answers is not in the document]
abhinandanpb: done

pkg/server/container_stats_list.go
abhinandanpb: :) done

pkg/server/container_stats_list.go
abhinandanpb: done

pkg/server/container_stats_list.go
abhinandanpb: done

integration/container_stats_test.go
```diff
@@ -0,0 +1,317 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+const (
+	defaultTestImage = "gcr.io/google_containers/pause:3.0"
```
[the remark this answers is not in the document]
abhinandanpb: done

integration/container_stats_test.go
```diff
@@ -0,0 +1,317 @@
...
+			return true, nil
+		}
+		return false, nil
+	}, time.Second, 30*time.Second))
+	t.Logf("Verify container stats for sandbox %q", sb)
+	for _, s := range stats {
+		testStats(t, s)
+	}
+}
+
```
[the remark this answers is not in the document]
abhinandanpb: Thought process here is that its fine not to be conservative with tests especially with this scenario not being exercised by e2e tests. Unit test is good to test the filtering I agree. I would like to add a TODO here to take this off once we have e2e coverage for these functionalities. Always happy to remove code :) Let me know.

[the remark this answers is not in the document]
Random-Liu: I feel like that we should:
1) Add comprehensive unit test for filtering even no matter we test filter or not here.
2) There are many duplicated code in this file, we could easily use a table-driven test pattern to get rid of these duplication.

However, I don't want to block this PR because of test code, so I'm fine with adding TODO for now, and come back later.

abhinandanpb:
@Random-Liu addressed comments. PTAL. I have added to TODO to clean up the tests which I will raise a PR in a day.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,157 @@ limitations under the License.
 package server
 
 import (
-	"errors"
-
+	"fmt"
+	"github.com/golang/glog"
```
Random-Liu: Move to the next group.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,157 @@ limitations under the License.
...
-	return nil, errors.New("not implemented")
+func (c *criContainerdService) ListContainerStats(
+	ctx context.Context,
+	in *runtime.ListContainerStatsRequest,
+) (*runtime.ListContainerStatsResponse, error) {
+	request, candidateContainers, err := c.buildTaskMetricsRequest(in)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build metrics request: %v", err)
+	}
+	resp, err := c.taskService.Metrics(context.Background(), &request)
```
Random-Liu: Pass `ctx` in.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,157 @@ limitations under the License.
...
+	for _, stat := range stats {
+		var cs runtime.ContainerStats
+		if err := c.getContainerMetrics(stat.ID, stat, &cs); err != nil {
+			glog.Errorf("failed to decode container metrics: %v", err)
+			continue
+		}
+		delete(candidateContainers, stat.ID)
+		containerStats.Stats = append(containerStats.Stats, &cs)
+	}
+	// If there is a transient state where containers are dead at the time of query
```
Random-Liu: It's not a transient state. For any dead container, this will happen until we remove the container.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,157 @@ limitations under the License.
...
+	// but present in the containerStore , then check if the writeableLayer information
```
Random-Liu: get rid of the space before `,`.

pkg/server/container_stats_list.go
```diff
@@ -17,14 +17,157 @@ limitations under the License.
...
+	}
+	return containerStats, nil
+}
+
+func (c *criContainerdService) getContainerMetrics(
+	containerID string,
+	stats *types.Metric,
+	cs *runtime.ContainerStats,
+) error {
+
```
Random-Liu: unnecessary empty line.

Random-Liu:
LGTM overall with some nits.

integration/container_stats_test.go
```diff
@@ -0,0 +1,321 @@
...
+	sbConfig := PodSandboxConfig("running-pod", "statsls")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+	t.Logf("Create a container config and run containers in a pod")
+	containerConfigMap := make(map[string]*runtime.ContainerConfig)
+	for i := 0; i < 3; i++ {
```
Random-Liu: We should have a dead container in the test, and for dead container, `testStats` won't work.
Probably add a TODO here, and you may want to manually validate it first.

abhinandanpb: had added on top of testStats coz that needed change. I have manually verified the case for dead containers by introducing a stopContainer and checking the stats.
```
=== RUN   TestContainerListStatsDead
&ContainerStats{Attributes:&ContainerAttributes{Id:6bf3066a1929839b5a5802cf3b3b49c7c292cc53364b3280d2a0a07dc24d9f77,Metadata:&ContainerMetadata{Name:container0,Attempt:0,},Labels:map[string]string{key: value,},Annotations:map[string]string{a.b.c: test,},},Cpu:nil,Memory:nil,WritableLayer:&FilesystemUsage{Timestamp:1506452046974590603,StorageId:&StorageIdentifier{Uuid:c656051a-9391-40fe-8333-57b410884232,},UsedBytes:&UInt64Value{Value:7,},InodesUsed:&UInt64Value{Value:20480,},},}
&ContainerStats{Attributes:&ContainerAttributes{Id:178a2722bd4ecec8728a952f07474b7714896f4afe8d69b3b58989b387e45804,Metadata:&ContainerMetadata{Name:container1,Attempt:0,},Labels:map[string]string{key: value,},Annotations:map[string]string{a.b.c: test,},},Cpu:nil,Memory:nil,WritableLayer:&FilesystemUsage{Timestamp:1506452046972987825,StorageId:&StorageIdentifier{Uuid:c656051a-9391-40fe-8333-57b410884232,},UsedBytes:&UInt64Value{Value:7,},InodesUsed:&UInt64Value{Value:20480,},},}
&ContainerStats{Attributes:&ContainerAttributes{Id:52e891cd436a10d71bc37d71f978d4142c4d49b7922ca7a2442ca75fe15c10b8,Metadata:&ContainerMetadata{Name:container2,Attempt:0,},Labels:map[string]string{key: value,},Annotations:map[string]string{a.b.c: test,},},Cpu:nil,Memory:nil,WritableLayer:&FilesystemUsage{Timestamp:1506452046973408844,StorageId:&StorageIdentifier{Uuid:c656051a-9391-40fe-8333-57b410884232,},UsedBytes:&UInt64Value{Value:7,},InodesUsed:&UInt64Value{Value:20480,},},}
--- PASS: TestContainerListStatsDead (32.88s)
```

Random-Liu: Thanks!

Random-Liu:
LGTM

PR #271, opened 2017-09-21
Should register container/sandbox name after restart.

Random-Liu:
Without this, we'll be able to create sandbox/container with the same name with sandbox/container created before restart.


Random-Liu:
@mikebrow Thanks!

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/271#pullrequestreview-64424725

PR #272, opened 2017-09-22
Improve apparmor and selinux support.

Random-Liu:
Previous behavior:
* If selinux is not supported (not built / OS not support):
  * `EnableSelinux=true`: cri-containerd doesn't report any error, but won't set any selinux options. **(Not Expected)**
  * `EnableSelinux=false`: cri-containerd doesn't set selinux options.
* If apparmor is not supported (not built / OS not support):
  * `EnableAppArmor=true`: cri-containerd always applies default to container and we'll never be able to start a container.**(Not Expected)**
  * `EnableAppArmor=false`: cri-containerd won't apply apparmor profile.

With this PR, for either selinux or apparmor, if the feature is not supported, cri-containerd will:
* `EnableXXX=true`: cri-containerd will log an error at start up, and not set corresponding options. The log will be something like:
```
E0922 04:43:31.108763   17069 cri_containerd.go:74] Selinux is not supported
```
* `EnableXXX=false`: cri-containerd won't set corresponding options.

Another option is to let cri-containerd fail at start up directly if a non-supported feature is enabled, I'm not sure which is better.
@miaoyq @mikebrow 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,22 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-GO ?= go
-EPOCH_TEST_COMMIT ?= f9e02affccd51702191e5312665a16045ffef8ab
+GO := go
+EPOCH_TEST_COMMIT := f9e02affccd51702191e5312665a16045ffef8ab
 PROJECT := github.com/kubernetes-incubator/cri-containerd
-BINDIR ?= ${DESTDIR}/usr/local/bin
-BUILD_DIR ?= _output
+BINDIR := ${DESTDIR}/usr/local/bin
+BUILD_DIR := _output
 # VERSION is derived from the current tag for HEAD plus amends. Version is used
 # to set/overide the criContainerdVersion variable in the verison package for
 # cri-containerd.
 VERSION := $(shell git describe --tags --dirty)
 # strip the first char of the tag if it's a `v`
 VERSION := $(VERSION:v%=%)
-TARBALL ?= cri-containerd-$(VERSION).tar.gz
-ifdef BUILD_TAGS
-BUILD_TAGS := -tags $(BUILD_TAGS)
-endif
-GO_LDFLAGS := -ldflags '-X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)'
+TARBALL := cri-containerd-$(VERSION).tar.gz
+BUILD_TAGS := apparmor
+GO_LDFLAGS := -X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)
 SOURCES := $(shell find . -name '*.go')
 
 all: binaries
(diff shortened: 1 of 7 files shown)


cmd/cri-containerd/cri_containerd.go
```diff
@@ -68,3 +67,19 @@ func main() {
 		glog.Exitf("Failed to run cri-containerd grpc server: %v", err)
 	}
 }
+
+func validateConfig(o *options.CRIContainerdOptions) {
```
miaoyq: `validateConfig` to `validateSecurityConfig`?  Because it only contains the config related to security.

Random-Liu: It may validate other configurations in the future I think, if there are others need to be validated. :)

miaoyq: Makes sense

miaoyq:
LGTM

Makefile
```diff
@@ -12,22 +12,20 @@
...
 VERSION := $(shell git describe --tags --dirty)
 # strip the first char of the tag if it's a `v`
 VERSION := $(VERSION:v%=%)
-TARBALL ?= cri-containerd-$(VERSION).tar.gz
-ifdef BUILD_TAGS
-BUILD_TAGS := -tags $(BUILD_TAGS)
-endif
-GO_LDFLAGS := -ldflags '-X $(PROJECT)/pkg/version.criContainerdVersion=$(VERSION)'
+TARBALL := cri-containerd-$(VERSION).tar.gz
+BUILD_TAGS := apparmor
```
mikebrow: a little nervous forcing the apparmor build tag.. let's talk about this on our call.

cmd/cri-containerd/cri_containerd.go
```diff
@@ -68,3 +67,19 @@ func main() {
...
+	if o.EnableSelinux {
+		if !selinux.GetEnabled() {
+			glog.Errorf("Selinux is not supported")
```
mikebrow: Let's make this a warning..

cmd/cri-containerd/cri_containerd.go
```diff
@@ -68,3 +67,19 @@ func main() {
...
+		}
+	} else {
+		selinux.SetDisabled()
+	}
+
+	if o.EnableAppArmor {
+		if !apparmor.IsEnabled() {
+			glog.Errorf("Apparmor is not supported")
```
mikebrow: same

pkg/server/container_create.go
```diff
@@ -201,7 +202,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 		specOpts = append(specOpts, containerd.WithUsername(username))
 	}
 	// Set apparmor profile, (privileged or not) if apparmor is enabled
-	if c.config.EnableAppArmor {
+	if c.config.EnableAppArmor && runcapparmor.IsEnabled() {
```
mikebrow: I want to rethink this code logic a bit.  For example, we should probably be throwing an error if they request a specific localhost/profile and we ignore that request because enableapparmor is off. I feel like we are abandoning the contract then. Need more detail on the requirement :-)

Makefile
Random-Liu: I think the default is mainly targeting at Kubernetes upstream test environment. Now we are mainly testing on Ubuntu and COS, and both support apparmor.

cmd/cri-containerd/cri_containerd.go
Random-Liu: Will do.

cmd/cri-containerd/cri_containerd.go
Random-Liu: Will do.

pkg/server/container_create.go
Random-Liu: Make a lot of sense. Only when apparmor profile is not specified, we should check whether apparmor is supported and choose to or not to set default profile. Will update.

cmd/cri-containerd/cri_containerd.go
Random-Liu: Done

cmd/cri-containerd/cri_containerd.go
Random-Liu: Done

pkg/server/container_create.go

Random-Liu:
@mikebrow Done

Random-Liu:
@mikebrow Done. Thanks for your suggestion.

PR #274, opened 2017-09-22
Use systemd units in tar

Random-Liu:
@abhinandanpb 
My node could not talk with my VMs, thus I haven't validated this PR yet.
Sent it our first, and I'll verify it again tomorrow.


Random-Liu:
@abhinandanpb I've validated this, it works.

Apply LGTM.

PR #275, opened 2017-09-22
Adds support for configuring the containerd runtime engine

mikebrow:
Enables changing the defaults used by containerd for the location of runtime state (/var/lib/containerd by default), the name of the runc compatible runtime-engine (runc by default), and whether systemd cgroup support is enabled (false by default).

Note: these are specified in the containerd client api as options on each new container. 

### DIFF
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -31,12 +31,22 @@ const configFilePathArgName = "config"
 
 // ContainerdConfig contains config related to containerd
 type ContainerdConfig struct {
-	// ContainerdRootDir is the root directory path for containerd.
-	ContainerdRootDir string `toml:"root"`
-	// ContainerdSnapshotter is the snapshotter used by containerd.
-	ContainerdSnapshotter string `toml:"snapshotter"`
-	// ContainerdEndpoint is the containerd endpoint path.
-	ContainerdEndpoint string `toml:"endpoint"`
+	// RootDir is the root directory path for containerd.
+	RootDir string `toml:"root_dir"`
+	// Snapshotter is the snapshotter used by containerd.
+	Snapshotter string `toml:"snapshotter"`
+	// Endpoint is the containerd endpoint path.
+	Endpoint string `toml:"endpoint"`
+	// Runtime is the runtime to use in containerd. We may support
+	// other runtimes in the future.
+	Runtime string `toml:"runtime"`
+	// RuntimeEngine is the name of the runtime engine used by containerd.
+	// Containerd default should be "runc"
+	// We may support other runtime engines in the future.
+	RuntimeEngine string `toml:"runtime_engine"`
+	// RuntimeRoot is the directory used by containerd for runtime state.
+	// Containerd default should be "/run/containerd/runc"
+	RuntimeRoot string `toml:"runtime_root"`
 }
 
 // CniConfig contains config related to cni
@@ -99,13 +109,18 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 		"/var/run/cri-containerd.sock", "Path to the socket which cri-containerd serves on.")
 	fs.StringVar(&c.RootDir, "root-dir",
 		"/var/lib/cri-containerd", "Root directory path for cri-containerd managed files (metadata checkpoint etc).")
-	fs.StringVar(&c.ContainerdRootDir, "containerd-root-dir",
-		"/var/lib/containerd", "Root directory path where containerd stores persistent data. "+
-			"This should be the same with containerd `root`.")
-	fs.StringVar(&c.ContainerdEndpoint, "containerd-endpoint",
+	fs.StringVar(&c.ContainerdConfig.RootDir, "containerd-root-dir",
+		"/var/lib/containerd", "Root directory path where containerd stores persistent data.")
+	fs.StringVar(&c.ContainerdConfig.Endpoint, "containerd-endpoint",
 		"/run/containerd/containerd.sock", "Path to the containerd endpoint.")
-	fs.StringVar(&c.ContainerdSnapshotter, "containerd-snapshotter",
-		containerd.DefaultSnapshotter, "Snapshotter used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.Snapshotter, "containerd-snapshotter",
+		containerd.DefaultSnapshotter, "The snapshotter used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.Runtime, "containerd-runtime",
+		"io.containerd.runtime.v1.linux", "The runtime used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.RuntimeEngine, "containerd-runtime-engine",
+		"", "Runtime engine used by containerd. (default = \"\" uses containerd default)")
+	fs.StringVar(&c.ContainerdConfig.RuntimeRoot, "containerd-runtime-root",
+		"", "The directory used by containerd for runtime state. (default = \"\" uses containerd default)")
 	fs.BoolVar(&c.PrintVersion, "version",
 		false, "Print cri-containerd version information and quit.")
 	fs.StringVar(&c.NetworkPluginBinDir, "network-bin-dir",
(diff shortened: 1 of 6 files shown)


cmd/cri-containerd/options/options.go
```diff
@@ -35,6 +35,17 @@ type ContainerdConfig struct {
...
+	// other runtimes in the future.
+	DefaultRuntime string `toml:"runtime"`
+	// DefaultRuntimeEngine is the name of the runtime engine used by containerd.
+	// We may support other runtime engines in the future.
+	DefaultRuntimeEngine string `toml:"runtime_engine"`
+	// DefaultRuntimeRoot is the directory used by containerd for runtime state.
+	DefaultRuntimeRoot string `toml:"runtime_root"`
+	// DefaultRuntimeSystemdCgroup is a flag for enabling systemd cgroup support
+	// in the selected runtime engine.
+	DefaultRuntimeSystemdCgroup bool `toml:"runtime_systemd_cgroup"`
```
yanxuean: I'd prefer to remove the prefix "Default".    "Default" indicate only a preferd value. but It is a option  here,we can actually give many value.

Random-Liu:
@mikebrow I sent another PR for the cgroup path support. https://github.com/kubernetes-incubator/cri-containerd/pull/290

Random-Liu:
@mikebrow I'll review other 3 options tomorrow.

cmd/cri-containerd/options/options.go
```diff
@@ -35,6 +35,17 @@ type ContainerdConfig struct {
...
+	// other runtimes in the future.
+	Runtime string `toml:"runtime"`
+	// RuntimeEngine is the name of the runtime engine used by containerd.
+	// We may support other runtime engines in the future.
+	RuntimeEngine string `toml:"runtime_engine"`
+	// RuntimeRoot is the directory used by containerd for runtime state.
+	RuntimeRoot string `toml:"runtime_root"`
+	// RuntimeSystemdCgroup is a flag for enabling systemd cgroup support
+	// in the selected runtime engine.
+	RuntimeSystemdCgroup bool `toml:"runtime_systemd_cgroup"`
```
Random-Liu: I think we could remove this now.

cmd/cri-containerd/options/options.go
```diff
@@ -35,6 +35,17 @@ type ContainerdConfig struct {
 	ContainerdSnapshotter string `toml:"snapshotter"`
 	// ContainerdEndpoint is the containerd endpoint path.
 	ContainerdEndpoint string `toml:"endpoint"`
+	// Runtime is the runtime to use in containerd. We may support
+	// other runtimes in the future.
+	Runtime string `toml:"runtime"`
```
Random-Liu: Follow the pattern? `ContainerdRuntime`?
And the other 3 configs.

pkg/server/container_create.go
```diff
@@ -251,7 +252,10 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 
 	opts = append(opts,
 		containerd.WithSpec(spec, specOpts...),
-		containerd.WithRuntime(defaultRuntime, nil),
+		containerd.WithRuntime(c.config.ContainerdConfig.Runtime,
+			&runcopts.RuncOptions{Runtime: c.config.ContainerdConfig.RuntimeEngine,
+				RuntimeRoot:   c.config.ContainerdConfig.RuntimeRoot,
```
Random-Liu: `c.config.ContainerdRuntimeRoot`, and also for the other configurations.

Random-Liu:
@mikebrow Please see the comments.

cmd/cri-containerd/options/options.go
mikebrow: yeah I'll remove it on the rebase

cmd/cri-containerd/options/options.go
mikebrow: I don't mind prefixing with Containerd .. so many comments on what to prefix variable names with.. :-)

cmd/cri-containerd/options/options.go
mikebrow: well I'll change it to ContainerdSystemdCgroup

cmd/cri-containerd/options/options.go
mikebrow: But I'd rather remove the prefix since we are already in the ContainerdConfig struct

pkg/server/container_create.go
mikebrow: are you asking to move these options out of the containerdconfig struct? and prefix them with containerd instead?

Random-Liu: Nope. `ContainerdConfig` is embedded in Config, you could access its field directly.

cmd/cri-containerd/options/options.go
Random-Liu: See https://github.com/kubernetes-incubator/cri-containerd/pull/275#discussion_r141168532.

mikebrow:
Comments addressed.

cmd/cri-containerd/options/options.go
```diff
@@ -31,12 +31,20 @@ const configFilePathArgName = "config"
...
 // ContainerdConfig contains config related to containerd
 type ContainerdConfig struct {
-	// ContainerdRootDir is the root directory path for containerd.
-	ContainerdRootDir string `toml:"root"`
-	// ContainerdSnapshotter is the snapshotter used by containerd.
-	ContainerdSnapshotter string `toml:"snapshotter"`
-	// ContainerdEndpoint is the containerd endpoint path.
-	ContainerdEndpoint string `toml:"endpoint"`
+	// RootDir is the root directory path for containerd.
+	RootDir string `toml:"containerd_root_dir"`
```
Random-Liu: Remove `containerd_` and for the other configuration.

run `cri-containerd --default-config` you'll know why. :)

cmd/cri-containerd/options/options.go
```diff
@@ -99,13 +107,18 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
+	fs.StringVar(&c.ContainerdConfig.Endpoint, "containerd-endpoint",
 		"/run/containerd/containerd.sock", "Path to the containerd endpoint.")
-	fs.StringVar(&c.ContainerdSnapshotter, "containerd-snapshotter",
+	fs.StringVar(&c.ContainerdConfig.Snapshotter, "containerd-snapshotter",
 		containerd.DefaultSnapshotter, "Snapshotter used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.Runtime, "containerd-runtime",
+		"io.containerd.runtime.v1.linux", "runtime used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.RuntimeEngine, "containerd-runtime-engine",
+		"runc", "runtime engine used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.RuntimeRoot, "containerd-runtime-root",
```
Random-Liu: This seems to be https://github.com/containerd/containerd/blob/fe9d6a46d1e9f4cea358d21bd93bee2ddf9b422d/linux/shim/service.go#L32.

I prefer we set empty default which indicates let containerd use its own default for both RuntimeEngine, and RuntimeRoot.

mikebrow: agree on the comment part if null.. but we should talk about the Engine and Root being empty by default..https://github.com/containerd/containerd/blob/master/linux/runtime.go#L449-L460

cmd/cri-containerd/options/options.go
```diff
@@ -99,13 +107,18 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
-	fs.StringVar(&c.ContainerdEndpoint, "containerd-endpoint",
+	fs.StringVar(&c.ContainerdConfig.RootDir, "containerd-root-dir",
+		"/var/lib/containerd", "Root directory path where containerd stores persistent data.")
+	fs.StringVar(&c.ContainerdConfig.Endpoint, "containerd-endpoint",
 		"/run/containerd/containerd.sock", "Path to the containerd endpoint.")
-	fs.StringVar(&c.ContainerdSnapshotter, "containerd-snapshotter",
+	fs.StringVar(&c.ContainerdConfig.Snapshotter, "containerd-snapshotter",
 		containerd.DefaultSnapshotter, "Snapshotter used by containerd.")
+	fs.StringVar(&c.ContainerdConfig.Runtime, "containerd-runtime",
+		"io.containerd.runtime.v1.linux", "runtime used by containerd.")
```
Random-Liu: s/runtime/Runtime.

cmd/cri-containerd/options/options.go
```diff
@@ -99,13 +107,18 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
+	fs.StringVar(&c.ContainerdConfig.RuntimeEngine, "containerd-runtime-engine",
+		"runc", "runtime engine used by containerd.")
```
Random-Liu: s/runtime/Runtime.

cmd/cri-containerd/options/options.go
Random-Liu: s/the/The.

mikebrow:
Comments addressed.

cmd/cri-containerd/options/options.go
```diff
@@ -119,13 +134,13 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.CgroupPath, "cgroup-path",
 		"", "The cgroup that cri-containerd is part of. By default cri-containerd is not placed in a cgroup.")
 	fs.BoolVar(&c.EnableSelinux, "enable-selinux",
-		false, "Enable selinux support.")
+		false, "Enable selinux support. (default false)")
```
Random-Liu: Actually, `cri-containerd --help` will show default, we don't need to add this ourselves. I'm with fix this in another PR.

Random-Liu:
LGTM

cmd/cri-containerd/options/options.go
Random-Liu: Actually, pflag only print default value if the original string doesn't contain the word `default`, which is a bit not smart. :)

So we do need to add some default value description ourselves if the original description contains the word `default`.

Will send a PR to clean this up a little bit.

mikebrow: It looked to me that false wasn't printing so I added the false ones that I felts needed to be output.

PR #277, opened 2017-09-22
Add net.ipv4.ip_forward=1 and use ansible sysctl.

Random-Liu:
This PR:
1) Set `net.ipv4.ip_forward=1`, which is required by Kubernetes.
2) Use ansible sysctl support instead. Current code has an extra `\/` in front of the regexp, which will never match, thus ansible will put a sysctl entry each time it runs.
3) Format the indent a little bit.

### DIFF
--- a/contrib/ansible/cri-containerd.yaml
+++ b/contrib/ansible/cri-containerd.yaml
@@ -15,16 +15,17 @@
     
     - name: "Start CRI-Containerd"
       systemd: name=cri-containerd daemon_reload=yes state=started enabled=yes
-    
-    - name: "Set bridge-nf-call-iptables"
-      lineinfile:
-        line: "net/bridge/bridge-nf-call-iptables = 1"
-        dest: /etc/sysctl.conf
-        insertafter: 'EOF'
-        regexp: '\/net\/bridge\/bridge-nf-call-iptables = 1'
-        state: present
-      ignore_errors: true
-    
+
+    - name: "Set bridge-nf-call-iptables" 
+      sysctl:
+        name: net.bridge.bridge-nf-call-iptables
+        value: 1
+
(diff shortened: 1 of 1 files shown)


contrib/ansible/cri-containerd.yaml
```diff
@@ -15,26 +15,25 @@
...
-      lineinfile:
-        line: "net/bridge/bridge-nf-call-iptables = 1"
-        dest: /etc/sysctl.conf
-        insertafter: 'EOF'
-        regexp: '\/net\/bridge\/bridge-nf-call-iptables = 1'
-        state: present
-      ignore_errors: true
-    
+
+    - sysctl:
```
abhinandanpb: can u add name for the task

contrib/ansible/cri-containerd.yaml
```diff
@@ -15,26 +15,25 @@
...
+        name: net.bridge.bridge-nf-call-iptables
+        value: 1
+
+    - sysctl:
```
abhinandanpb: same here

contrib/ansible/cri-containerd.yaml
Random-Liu: Actually I wanted to do that, but don't know how to do it... Let me lookup the document.

Random-Liu: Done

contrib/ansible/cri-containerd.yaml
Random-Liu: Done

abhinandanpb:
LGTM

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/277#issuecomment-331583520.

PR #281, opened 2017-09-24
Check seccomp enable and add unit test for seccomp/apparmor.

Random-Liu:
This PR:
1) Use `runc/libcontainer/seccomp` to check whether seccomp is enabled on OS.
2) Move apparmor and seccomp logic into helper functions.
3) Add unit test for apparmor and seccomp logic.

@mikebrow 

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -30,9 +30,7 @@ import (
 	"github.com/docker/docker/pkg/mount"
 	"github.com/golang/glog"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
-	runcapparmor "github.com/opencontainers/runc/libcontainer/apparmor"
 	"github.com/opencontainers/runc/libcontainer/devices"
-	runcseccomp "github.com/opencontainers/runc/libcontainer/seccomp"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/opencontainers/runtime-tools/validate"
@@ -202,7 +200,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	apparmorSpecOpts, err := generateApparmorSpecOpts(
 		securityContext.GetApparmorProfile(),
 		securityContext.GetPrivileged(),
-		runcapparmor.IsEnabled())
+		c.apparmorEnabled)
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate apparmor spec opts: %v", err)
 	}
@@ -213,7 +211,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	seccompSpecOpts, err := generateSeccompSpecOpts(
 		securityContext.GetSeccompProfilePath(),
 		securityContext.GetPrivileged(),
-		runcseccomp.IsEnabled())
+		c.seccompEnabled)
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate seccomp spec opts: %v", err)
 	}
@@ -730,11 +728,23 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
 		mounts = append(mounts, mount)
 	}
 	spec.Mounts = mounts
+
+	// Make sure no default seccomp/apparmor is specified
+	if spec.Process != nil {
+		spec.Process.ApparmorProfile = ""
+	}
+	if spec.Linux != nil {
+		spec.Linux.Seccomp = nil
+	}
 	return spec, nil
 }
 
 // generateSeccompSpecOpts generates containerd SpecOpts for seccomp.
 func generateSeccompSpecOpts(seccompProf string, privileged, seccompEnabled bool) (containerd.SpecOpts, error) {
+	if privileged {
+		// Do not set seccomp profile when container is privileged
+		return nil, nil
+	}
 	// Set seccomp profile
 	if seccompProf == runtimeDefault || seccompProf == dockerDefault {
 		// use correct default profile (Eg. if not configured otherwise, the default is docker/default)
@@ -744,10 +754,6 @@ func generateSeccompSpecOpts(seccompProf string, privileged, seccompEnabled bool
 		if seccompProf != "" && seccompProf != unconfinedProfile {
 			return nil, fmt.Errorf("seccomp is not supported")
 		}
-		return nil, nil
-	}
-	if privileged {
-		// Do not set seccomp profile when container is privileged
 		return nil, nil
 	}
 	switch seccompProf {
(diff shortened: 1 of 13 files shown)


pkg/server/container_create.go
```diff
@@ -768,6 +733,70 @@ func defaultRuntimeSpec() (*runtimespec.Spec, error) {
...
+		// use correct default profile (Eg. if not configured otherwise, the default is docker/default)
+		seccompProf = seccompDefaultProfile
+	}
+	if !seccompEnabled {
+		if seccompProf != "" && seccompProf != unconfinedProfile {
+			return nil, fmt.Errorf("seccomp is not supported")
+		}
+		return nil, nil
+	}
+	if privileged {
```
miaoyq: Maybe we could put the logic to the top of function.

If `privileged==true`, we could not care about the values of  `seccompEnabled` and `seccompProf`.

Random-Liu: Will do, make sense.

pkg/server/container_create.go
```diff
@@ -201,59 +198,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
-			if !strings.HasPrefix(appArmorProf, profileNamePrefix) {
-				return nil, fmt.Errorf("invalid apparmor profile %q", appArmorProf)
-			}
-			specOpts = append(specOpts, apparmor.WithProfile(strings.TrimPrefix(appArmorProf, profileNamePrefix)))
-		}
+
+	apparmorSpecOpts, err := generateApparmorSpecOpts(
+		securityContext.GetApparmorProfile(),
+		securityContext.GetPrivileged(),
+		runcapparmor.IsEnabled())
```
ijc: Perhaps the caller should call this itself?

pkg/server/container_create.go
```diff
@@ -201,59 +198,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
-			// Require and Trim default profile name prefix
-			if !strings.HasPrefix(seccompProf, profileNamePrefix) {
-				return nil, fmt.Errorf("invalid seccomp profile %q", seccompProf)
-			}
-			specOpts = append(specOpts, seccomp.WithProfile(strings.TrimPrefix(seccompProf, profileNamePrefix)))
-		}
+	seccompSpecOpts, err := generateSeccompSpecOpts(
+		securityContext.GetSeccompProfilePath(),
+		securityContext.GetPrivileged(),
+		runcseccomp.IsEnabled())
```
ijc: Perhaps the helper function should call `runccseccomp.IsEnbabled()` itself?

pkg/server/container_create.go
Random-Liu: This is mainly for unit test. And actually, we don't need to call this every time, it does include some syscall.

I'll add global shared fields, and initialize those fields during startup.

pkg/server/container_create.go
Random-Liu: Actually, CRI doesn't define the expected behavior clearly. We are not following docker's behavior.
https://github.com/moby/moby/blob/master/daemon/oci_linux.go#L751
https://github.com/moby/moby/blob/8af4db6f002ac907b6ef8610b237879dfcaa5b7a/daemon/seccomp_linux.go#L20

pkg/server/container_create.go
```diff
@@ -201,59 +198,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
-			}
-		default:
-			// Require and Trim default profile name prefix
-			if !strings.HasPrefix(appArmorProf, profileNamePrefix) {
-				return nil, fmt.Errorf("invalid apparmor profile %q", appArmorProf)
-			}
-			specOpts = append(specOpts, apparmor.WithProfile(strings.TrimPrefix(appArmorProf, profileNamePrefix)))
-		}
+
+	apparmorSpecOpts, err := generateApparmorSpecOpts(
```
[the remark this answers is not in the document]
Random-Liu: > thx for moving these into helpers that was next up on my list of things to do

Just get bored during the weekend, and addressed some TODOs. :)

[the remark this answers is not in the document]
Random-Liu: > but it makes it really hard to review when also changing the logic at the same time

Sorry, messed up my commits, and have to combined them into one at last. :(

pkg/server/container_create.go
```diff
@@ -201,59 +198,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
+		securityContext.GetApparmorProfile(),
+		securityContext.GetPrivileged(),
+		runcapparmor.IsEnabled())
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate apparmor spec opts: %v", err)
```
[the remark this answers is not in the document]
Random-Liu: We prefixed all the errors in `CreateContainer`, and the error returned by the helper is short. I think it's fine to prefix here. :)

pkg/server/container_create.go
```diff
@@ -201,59 +198,27 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
...
-	if seccompProf == runtimeDefault || seccompProf == dockerDefault {
-		// use correct default profile (Eg. if not configured otherwise, the default is docker/default for containers)
-		seccompProf = seccompDefaultContainerProfile
-	}
-	// Unset the seccomp profile, if seccomp is not enabled, unconfined, unset, or the security context is privileged
-	if !seccompEnabled ||
-		seccompProf == unconfinedProfile ||
-		seccompProf == "" ||
-		config.GetLinux().GetSecurityContext().GetPrivileged() {
-		spec.Linux.Seccomp = nil
```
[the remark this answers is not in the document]
Random-Liu: Make sense. Let me see how to add it back.

pkg/server/container_create.go
Random-Liu: ditto.

pkg/server/container_create.go
mikebrow: Well there ya go.. in the absence of a defined spec... use the default behavior of the prior code

Random-Liu: > If privileged==true, we could not care about the values of seccompEnabled and seccompProf.

@miaoyq Done

pkg/server/container_create.go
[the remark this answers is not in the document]
Random-Liu: I updated our `defaultRuntimeSpec()` function to make sure no default apparmor/seccomp is specified by default.

pkg/server/container_create.go

Random-Liu:
@mikebrow Addressed comments.

Random-Liu:
@mikebrow Thanks for reviewing! :)

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/281#pullrequestreview-65068488

PR #282, opened 2017-09-25
Improve ansible playbook.

Random-Liu:
This PR:
1) Add `modprobe br_netfilter`, which is required by kubelet. Without which `net.bridge.bridge-nf-call-iptables` won't work, either.
2) Stop using dev library, which is not needed for running cri-containerd.
3) Add systemd dropin for cri-containerd, previous solution always report an error during installation, which is confusing to user.
4) Update cri-containerd to newest version.

### DIFF
--- a/contrib/ansible/cri-containerd.yaml
+++ b/contrib/ansible/cri-containerd.yaml
@@ -16,6 +16,11 @@
     - name: "Start CRI-Containerd"
       systemd: name=cri-containerd daemon_reload=yes state=started enabled=yes
 
+    - name: "Load br_netfilter kernel module"
+      modprobe:
+        name: br_netfilter
+        state: present
+
     - name: "Set bridge-nf-call-iptables" 
       sysctl:
         name: net.bridge.bridge-nf-call-iptables
@@ -27,14 +32,13 @@
         value: 1
 
     - name: "Check kubelet args in kubelet config"
-      shell: grep "^Environment=\"KUBELET_EXTRA_ARGS=" /etc/systemd/system/kubelet.service.d/10-kubeadm.conf 
-      ignore_errors: true
+      shell: grep "^Environment=\"KUBELET_EXTRA_ARGS=" /etc/systemd/system/kubelet.service.d/10-kubeadm.conf || true
       register: check_args
 
     - name: "Add runtime args in kubelet conf"
       lineinfile:
         dest: "/etc/systemd/system/kubelet.service.d/10-kubeadm.conf"
-        line: "Environment=\"KUBELET_EXTRA_ARGS= --container-runtime=remote --runtime-request-timeout=15m --image-service-endpoint=/var/run/cri-containerd.sock --container-runtime-endpoint=/var/run/cri-containerd.sock\""
+        line: "Environment=\"KUBELET_EXTRA_ARGS= --container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=/var/run/cri-containerd.sock\""
         insertafter: '\[Service\]'
       when: check_args.stdout == ""
     
--- a/contrib/ansible/tasks/bootstrap_centos.yaml
+++ b/contrib/ansible/tasks/bootstrap_centos.yaml
@@ -6,8 +6,8 @@
   with_items:
     - unzip
     - tar
-    - btrfs-progs-devel
-    - libseccomp-devel
+    - btrfs-progs
+    - libseccomp
     - util-linux 
     - socat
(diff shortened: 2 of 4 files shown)


Random-Liu:
@abhinandanpb I've validated this PR. It works, and I've got the sock shop again. :)
http://104.197.75.252:30001/index.html

contrib/ansible/cri-containerd.yaml
```diff
@@ -16,6 +16,11 @@
     - name: "Start CRI-Containerd"
       systemd: name=cri-containerd daemon_reload=yes state=started enabled=yes
 
+    - name: "Load br_netfilter kernel module"
```
abhinandanpb: this should be in bootstrap_ubuntu I feel.

contrib/ansible/cri-containerd.yaml
```diff
@@ -26,17 +31,13 @@
...
-      ignore_errors: true
-      register: check_args
-
-    - name: "Add runtime args in kubelet conf"
-      lineinfile:
-        dest: "/etc/systemd/system/kubelet.service.d/10-kubeadm.conf"
-        line: "Environment=\"KUBELET_EXTRA_ARGS= --container-runtime=remote --runtime-request-timeout=15m --image-service-endpoint=/var/run/cri-containerd.sock --container-runtime-endpoint=/var/run/cri-containerd.sock\""
-        insertafter: '\[Service\]'
-      when: check_args.stdout == ""
+    - name: "cri-containerd systemd dropin for kubeadm"
```
abhinandanpb: personally not a fan of this. Would prefer a single config file for the system unit. Especially if ppl intend to modify KUBELET_EXTRA_ARGS and they don't see it in one config file. They might just add it in that. This will cause more issues than help us.

abhinandanpb: It is not an error :) it is an ignored error. It checks if the regex is present / not . If its present it doesn't add. If it is not then it will add.

contrib/ansible/cri-containerd.yaml
Random-Liu: Why don't we need this for centos? We need this kernel module on any linux distro, right?
Or else there won't even be `net.bridge` sysctl.

contrib/ansible/cri-containerd.yaml
Random-Liu: Let's discuss this tomorrow. :)

contrib/ansible/cri-containerd.yaml
abhinandanpb: my bad. It works. 
bridge                107151  2 br_netfilter,ebtable_broute

Random-Liu:
No error message:
```console
TASK [Check kubelet args in kubelet config] *************************************************************************************************************************************************************************************************
changed: [master-1]
changed: [node-1]

TASK [Add runtime args in kubelet conf] *****************************************************************************************************************************************************************************************************
changed: [master-1]
changed: [node-1]
```

And did work:
```console
$ sudo cat /etc/systemd/system/kubelet.service.d/10-kubeadm.conf 
[Service]
Environment="KUBELET_EXTRA_ARGS= --container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=/var/run/cri-containerd.sock"
Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
Environment="KUBELET_DNS_ARGS=--cluster-dns=10.96.0.10 --cluster-domain=cluster.local"
Environment="KUBELET_AUTHZ_ARGS=--authorization-mode=Webhook --client-ca-file=/etc/kubernetes/pki/ca.crt"
Environment="KUBELET_CADVISOR_ARGS=--cadvisor-port=0"
ExecStart=
ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_SYSTEM_PODS_ARGS $KUBELET_NETWORK_ARGS $KUBELET_DNS_ARGS $KUBELET_AUTHZ_ARGS $KUBELET_CADVISOR_ARGS $KUBELET_EXTRA_ARGS
```

abhinandanpb:
LGTM

PR #285, opened 2017-09-25
Do not include CNI binaries/configs in release tarball.

Random-Liu:
Do not include CNI binaries/configs in release tarball. In real use, user should launch their own CNI daemonset to deploy cni binaries and configs, we should not include one in release tarball.

@abhinandanpb @mikebrow 

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -44,6 +44,12 @@ if ${NOSUDO}; then
   sudo=""
 fi
 
+# INSTALL_CNI indicates whether to install CNI. CNI installation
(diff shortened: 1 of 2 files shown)


Random-Liu:
With this, we could actually tell user to:
```console
$ sudo tar -C / -xzf cri-containerd-XXX.tar.gz
```

abhinandanpb:
definitely LGTM :)

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/285#issue-260448072

PR #292, opened 2017-09-26
Avoid a requirement for /dev/disk/by-uuid

ijc:
This doesn't exist on systems which do not use `udev` (e.g. LinuxKit which uses
`mdev`).

There seem to be no go bindings to libblkid, although we could cgo it, but `blkid` seems widespread enough to rely on (although `blkid` in `busybox` is unsufficient, it needs to be the full `util-linux` variant to support the options we need).

### DIFF
--- a/pkg/os/os.go
+++ b/pkg/os/os.go
@@ -17,11 +17,12 @@ limitations under the License.
 package os
 
 import (
-	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	containerdmount "github.com/containerd/containerd/mount"
 	"github.com/containerd/fifo"
@@ -133,27 +134,9 @@ func (RealOS) LookupMount(path string) (containerdmount.Info, error) {
 // DeviceUUID gets device uuid of a device. The passed in device should be
 // an absolute path of the device.
 func (RealOS) DeviceUUID(device string) (string, error) {
-	const uuidDir = "/dev/disk/by-uuid"
-	if _, err := os.Stat(uuidDir); err != nil {
-		return "", err
-	}
-	files, err := ioutil.ReadDir(uuidDir)
+	output, err := exec.Command("/sbin/blkid", "-s", "UUID", "-o", "value", device).CombinedOutput()
 	if err != nil {
 		return "", err
 	}
-	for _, file := range files {
-		path := filepath.Join(uuidDir, file.Name())
-		target, err := os.Readlink(path)
-		if err != nil {
-			return "", err
-		}
-		dev, err := filepath.Abs(filepath.Join(uuidDir, target))
-		if err != nil {
-			return "", err
-		}
-		if dev == device {
-			return file.Name(), nil
-		}
-	}
-	return "", fmt.Errorf("device not found")
+	return strings.TrimSpace(string(output)), nil
 }


pkg/os/os.go
```diff
@@ -133,27 +134,9 @@ func (RealOS) LookupMount(path string) (containerdmount.Info, error) {
 // DeviceUUID gets device uuid of a device. The passed in device should be
 // an absolute path of the device.
 func (RealOS) DeviceUUID(device string) (string, error) {
-	const uuidDir = "/dev/disk/by-uuid"
-	if _, err := os.Stat(uuidDir); err != nil {
-		return "", err
-	}
-	files, err := ioutil.ReadDir(uuidDir)
+	output, err := exec.Command("/sbin/blkid", "-s", "UUID", "-o", "value", device).CombinedOutput()
```
Random-Liu: I tried an alpine container, and the output will be:
```console
# /sbin/blkid -s UUID -o value /dev/sda1
/dev/sda1: LABEL="cloudimg-rootfs" UUID="70045b4a-dd4c-4e37-889c-f1992da26c82" TYPE="ext4"
```
Did I misunderstand anything?

ijc: Alpine uses `busybox` by default, you'll need `apk add -u blkid` first to get a full fat version.

Random-Liu: Cool. Tried it. Thanks

Random-Liu: But we may want to document the dependency in the README.md
Would you like to send another PR to update README.md?

Random-Liu: BTW, cadvisor is relying on `/dev/disk/by-uuid`, it means that cadvisor will not be available on linuxkit?

Random-Liu:
@ijc LGTM overall, but want to discuss more about https://github.com/kubernetes-incubator/cri-containerd/pull/292#discussion_r141134621 before merging. :)

pkg/os/os.go
ijc: Hrm, that would indeed be a problem but in the kubernetes source tree 9790389a11580d089be2efb6f93a1529e7270d6f:
```
$ git grep -lE 'by-(id|path|label|partuuid|uuid)' | grep -v _test.go$
cluster/gce/configure-vm.sh
cluster/gce/container-linux/configure-helper.sh
cluster/gce/gci/configure-helper.sh
examples/mysql-cinder-pd/README.md
federation/pkg/federation-controller/service/dns/dns.go
pkg/cloudprovider/providers/openstack/metadata.go
pkg/cloudprovider/providers/openstack/openstack_volumes.go
pkg/volume/aws_ebs/aws_util.go
pkg/volume/fc/fc_util.go
pkg/volume/gce_pd/gce_util.go
pkg/volume/iscsi/iscsi_util.go
pkg/volume/photon_pd/photon_util.go
pkg/volume/scaleio/sio_client.go
pkg/volume/util/device_util_linux.go
pkg/volume/vsphere_volume/vsphere_volume_util.go
test/kubemark/resources/start-kubemark-master.sh
vendor/github.com/codedellemc/goscaleio/volume.go
vendor/google.golang.org/api/compute/v0.beta/compute-api.json
vendor/google.golang.org/api/compute/v0.beta/compute-gen.go
vendor/google.golang.org/api/compute/v1/compute-api.json
vendor/google.golang.org/api/compute/v1/compute-gen.go
```
I think those are all particular to specific environments or helper scripts which aren't used by LinuxKit (at least not today).

The function you linked to is `getFsUUIDToDeviceNameMap` which isn't referenced in the kube tree and doesn't appear in `vendor/github.com/google/cadvisor/fs/fs.go` Might be an older cadvisor or a fork perhaps?

ijc: Ah, I was on `release-1.7` I see the function you refer to in `master`.

I think I'll have to fix this on the LinuxKit side.

ijc:
Closing in favour of a fix on the LinuxKit side, although while hacking on that I did also come up with #297 which I think is worthwhile.

Random-Liu:
@ijc That code was added recently. Maybe the cadvisor code itself is making wrong assumption. Let me chat with the team and update there.

Random-Liu:
@ijc I talked with @yguo0905 who added this code in cadvisor recently.
I think we may not be making right assumption, I'll file an cadvisor issue for the mdev/udev thing.
However, it takes time to properly fix cadvisor and patch Kubernetes. Before that, we may have to do some hack in LinuxKit to work around it.

Random-Liu:
@ijc And I'm fine with this PR then. Let's reopen and merge it.

ijc:
I've now (https://github.com/linuxkit/linuxkit/pull/2545) added support for `by-uuid` to LinuxKit. TBH I don't consider that a hack or a workaround, that interface is a standard part of the Linux API these days IMHO for mapping UUID->device, although iterating over it to provide the inverse device->UUID mapping doesn't seem ideal it should work

I think Alpine uses mdev by default (although it does have the eudev fork of udev available too). I'm not sure but I wouldn't be surprised if Gentoo supported mdev at least as an option (likewise other "from source" or "from scratch" type distros). Other than that I would expect mdev to mainly be found in the embedded world.

PR #294, opened 2017-09-27
Cleanup flag default value

Random-Liu:
`pflag` doesn't print default `false` boolean or default empty string, that seems to be how it works.
I'm fine with this behavior.

@mikebrow WDYT? Do you think we should always add default ourselves for all `false` boolean or empty stirng?

### DIFF
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -118,9 +118,9 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.ContainerdConfig.Runtime, "containerd-runtime",
 		"io.containerd.runtime.v1.linux", "The runtime used by containerd.")
 	fs.StringVar(&c.ContainerdConfig.RuntimeEngine, "containerd-runtime-engine",
-		"", "Runtime engine used by containerd. (default = \"\" uses containerd default)")
+		"", "Runtime engine used by containerd. Defaults to containerd's default if not specified.")
 	fs.StringVar(&c.ContainerdConfig.RuntimeRoot, "containerd-runtime-root",
-		"", "The directory used by containerd for runtime state. (default = \"\" uses containerd default)")
+		"", "The directory used by containerd for runtime state. Defaults to containerd's default if not specified.")
 	fs.BoolVar(&c.PrintVersion, "version",
 		false, "Print cri-containerd version information and quit.")
 	fs.StringVar(&c.NetworkPluginBinDir, "network-bin-dir",
@@ -128,19 +128,19 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.NetworkPluginConfDir, "network-conf-dir",
 		"/etc/cni/net.d", "The directory for putting network plugin configuration files.")
 	fs.StringVar(&c.StreamServerAddress, "stream-addr",
-		"", "The ip address streaming server is listening on. Default host interface is used if this is empty.")
+		"", "The ip address streaming server is listening on. The default host interface is used if not specified.")
 	fs.StringVar(&c.StreamServerPort, "stream-port",
 		"10010", "The port streaming server is listening on.")
 	fs.StringVar(&c.CgroupPath, "cgroup-path",
-		"", "The cgroup that cri-containerd is part of. By default cri-containerd is not placed in a cgroup.")
+		"", "The cgroup that cri-containerd is part of. Cri-containerd is not placed in a cgroup if none is specified.")
 	fs.BoolVar(&c.EnableSelinux, "enable-selinux",
-		false, "Enable selinux support. (default false)")
+		false, "Enable selinux support. By default not enabled.")
 	fs.StringVar(&c.SandboxImage, "sandbox-image",
 		"gcr.io/google_containers/pause:3.0", "The image used by sandbox container.")
 	fs.IntVar(&c.StatsCollectPeriod, "stats-collect-period",
 		10, "The period (in seconds) of snapshots stats collection.")
 	fs.BoolVar(&c.SystemdCgroup, "systemd-cgroup",
-		false, "Enables systemd cgroup support. (default false)")
+		false, "Enables systemd cgroup support. By default not enabled.")
 	fs.BoolVar(&c.PrintDefaultConfig, "default-config",
 		false, "Print default toml config of cri-containerd and quit.")
 }


Random-Liu:
```
$ ./cri-containerd --help
Usage of ./cri-containerd:
      --alsologtostderr                    log to standard error as well as files
      --cgroup-path string                 The cgroup that cri-containerd is part of. Empty string means cri-containerd is not placed in a cgroup.
      --config string                      Path to the config file. (default "/etc/cri-containerd/config.toml")
      --containerd-endpoint string         Path to the containerd endpoint. (default "/run/containerd/containerd.sock")
      --containerd-root-dir string         Root directory path where containerd stores persistent data. (default "/var/lib/containerd")
      --containerd-runtime string          The runtime used by containerd. (default "io.containerd.runtime.v1.linux")
      --containerd-runtime-engine string   Runtime engine used by containerd. Empty string means using containerd default.
      --containerd-runtime-root string     The directory used by containerd for runtime state. Empty string means using containerd default.
      --containerd-snapshotter string      The snapshotter used by containerd. (default "overlayfs")
      --default-config                     Print default toml config of cri-containerd and quit.
      --enable-selinux                     Enable selinux support.
      --log_backtrace_at traceLocation     when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                     If non-empty, write log files in this directory
      --logtostderr                        log to standard error instead of files
      --network-bin-dir string             The directory for putting network binaries. (default "/opt/cni/bin")
      --network-conf-dir string            The directory for putting network plugin configuration files. (default "/etc/cni/net.d")
      --root-dir string                    Root directory path for cri-containerd managed files (metadata checkpoint etc). (default "/var/lib/cri-containerd")
      --sandbox-image string               The image used by sandbox container. (default "gcr.io/google_containers/pause:3.0")
      --socket-path string                 Path to the socket which cri-containerd serves on. (default "/var/run/cri-containerd.sock")
      --stats-collect-period int           The period (in seconds) of snapshots stats collection. (default 10)
      --stderrthreshold severity           logs at or above this threshold go to stderr (default 2)
      --stream-addr string                 The ip address streaming server is listening on. Empty string means using default host interface.
      --stream-port string                 The port streaming server is listening on. (default "10010")
      --systemd-cgroup                     Enables systemd cgroup support.
  -v, --v Level                            log level for V logs
      --version                            Print cri-containerd version information and quit.
      --vmodule moduleSpec                 comma-separated list of pattern=N settings for file-filtered logging
```

mikebrow:
I think we should take it on a case by case basis. It's important that the user/admin can determine what the flags do, and what their default is. If that requires telling them flat out (the default is false) I don't see why hiding that fact from them helps. Same with "" if that means use the containerd default we should tell them flat out (the default is "", containerd will use it's own default when "" is used)

cmd/cri-containerd/options/options.go
```diff
@@ -118,29 +118,29 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.ContainerdConfig.Runtime, "containerd-runtime",
 		"io.containerd.runtime.v1.linux", "The runtime used by containerd.")
 	fs.StringVar(&c.ContainerdConfig.RuntimeEngine, "containerd-runtime-engine",
-		"", "Runtime engine used by containerd. (default = \"\" uses containerd default)")
+		"", "Runtime engine used by containerd. Use containerd default if not specified.")
```
mikebrow: Defaults to containerd's default if not specified.

cmd/cri-containerd/options/options.go
```diff
@@ -118,29 +118,29 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&c.ContainerdConfig.RuntimeRoot, "containerd-runtime-root",
-		"", "The directory used by containerd for runtime state. (default = \"\" uses containerd default)")
+		"", "The directory used by containerd for runtime state. Use containerd default if not specified.")
```
mikebrow: same

cmd/cri-containerd/options/options.go
```diff
@@ -118,29 +118,29 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
 	fs.BoolVar(&c.PrintVersion, "version",
 		false, "Print cri-containerd version information and quit.")
 	fs.StringVar(&c.NetworkPluginBinDir, "network-bin-dir",
 		"/opt/cni/bin", "The directory for putting network binaries.")
 	fs.StringVar(&c.NetworkPluginConfDir, "network-conf-dir",
 		"/etc/cni/net.d", "The directory for putting network plugin configuration files.")
 	fs.StringVar(&c.StreamServerAddress, "stream-addr",
-		"", "The ip address streaming server is listening on. Default host interface is used if this is empty.")
+		"", "The ip address streaming server is listening on. Use default host interface if not specified.")
```
mikebrow: The default host interface is used if not specified.

cmd/cri-containerd/options/options.go
```diff
@@ -118,29 +118,29 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&c.StreamServerPort, "stream-port",
 		"10010", "The port streaming server is listening on.")
 	fs.StringVar(&c.CgroupPath, "cgroup-path",
-		"", "The cgroup that cri-containerd is part of. By default cri-containerd is not placed in a cgroup.")
+		"", "The cgroup that cri-containerd is part of. Do not place cri-containerd in any cgroup if not specified.")
 	fs.BoolVar(&c.EnableSelinux, "enable-selinux",
-		false, "Enable selinux support. (default false)")
+		false, "Enable selinux support.")
```
mikebrow: I'd tell them the default here.

cmd/cri-containerd/options/options.go
```diff
@@ -118,29 +118,29 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&c.SandboxImage, "sandbox-image",
 		"gcr.io/google_containers/pause:3.0", "The image used by sandbox container.")
 	fs.IntVar(&c.StatsCollectPeriod, "stats-collect-period",
 		10, "The period (in seconds) of snapshots stats collection.")
 	fs.BoolVar(&c.SystemdCgroup, "systemd-cgroup",
-		false, "Enables systemd cgroup support. (default false)")
+		false, "Enables systemd cgroup support.")
```
mikebrow: same

cmd/cri-containerd/options/options.go
```diff
@@ -118,29 +118,29 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&c.NetworkPluginConfDir, "network-conf-dir",
 		"/etc/cni/net.d", "The directory for putting network plugin configuration files.")
 	fs.StringVar(&c.StreamServerAddress, "stream-addr",
-		"", "The ip address streaming server is listening on. Default host interface is used if this is empty.")
+		"", "The ip address streaming server is listening on. Use default host interface if not specified.")
 	fs.StringVar(&c.StreamServerPort, "stream-port",
 		"10010", "The port streaming server is listening on.")
 	fs.StringVar(&c.CgroupPath, "cgroup-path",
-		"", "The cgroup that cri-containerd is part of. By default cri-containerd is not placed in a cgroup.")
+		"", "The cgroup that cri-containerd is part of. Do not place cri-containerd in any cgroup if not specified.")
```
mikebrow: Cri-containerd is not placed in a cgroup if none is specified.

cmd/cri-containerd/options/options.go
Random-Liu: Golang boolean flag is working like this:
* `cri-containerd` will not enable selinux.
* `cri-containerd --enable-selinux` will enable it.

Given so I feel like this fine here. :)

cmd/cri-containerd/options/options.go

cmd/cri-containerd/options/options.go
Random-Liu: Done

cmd/cri-containerd/options/options.go
Random-Liu: Done

cmd/cri-containerd/options/options.go
Random-Liu: Done

Random-Liu:
@mikebrow Addressed comments and replied the last 2.

cmd/cri-containerd/options/options.go
mikebrow: ah ok not a true false thing.. my bad

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/294#pullrequestreview-65955502

PR #296, opened 2017-09-27
Setting the cpu and memory stats of containers in nanoseconds

abhinandanpb:


Random-Liu:
LGTM

PR #297, opened 2017-09-27
Use stat_t.st_rdev to compare block devices

ijc:
I implemented /dev/disk/by-uuid on my platform but using absolute links (where
udev typically uses relative) which broke the code in `os.DeviceUUID`.

Rather than just patch that up directly instead stat both the target and
candidate devices and pick one with matching major:minor in st_rdev. This saves
manually building paths to resolve symlinks and I think should be more robust
overall.

I also removed the initial stat of /dev/disk/by-uuid, I believe
`ioutil.Readdir` will correctly return an error if the path does not exist.

### DIFF
--- a/pkg/os/os.go
+++ b/pkg/os/os.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	containerdmount "github.com/containerd/containerd/mount"
 	"github.com/containerd/fifo"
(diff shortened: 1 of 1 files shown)


pkg/os/os.go
```diff
@@ -130,28 +131,41 @@ func (RealOS) LookupMount(path string) (containerdmount.Info, error) {
...
 	}
 	for _, file := range files {
 		path := filepath.Join(uuidDir, file.Name())
-		target, err := os.Readlink(path)
-		if err != nil {
-			return "", err
-		}
-		dev, err := filepath.Abs(filepath.Join(uuidDir, target))
+
+		trdev, err := blkrdev(path)
```
Random-Liu: Hm, does `os.Stat` return correct Rdev for symlink of device?

pkg/os/os.go
```diff
@@ -130,28 +131,41 @@ func (RealOS) LookupMount(path string) (containerdmount.Info, error) {
...
 		if err != nil {
-			return "", err
+			continue
```
Random-Liu: Why do we ignore the error here?

Random-Liu:
@ijc If so, we also need to fix cadvisor and kubelet, which may take time. :)

Do you know what distributions are using `mdev`, except busybox and LinuxKit?

pkg/os/os.go
ijc: The idea was to skip over invalid links or other non-block things in the hopes of finding a valid entry later in the directory list. I debated adding a low-severity warning log, I could do that if you want.

ijc:
> Do you know what distributions are using mdev, except busybox and LinuxKit?

I also answered this in https://github.com/kubernetes-incubator/cri-containerd/pull/292#issuecomment-332789966 but since that one is currently closed lets better discuss it here. What I said was:

I think Alpine uses mdev by default (although it does have the eudev fork of udev available too). I'm not sure but I wouldn't be surprised if Gentoo supported mdev at least as an option (likewise other "from source" or "from scratch" type distros). Other than that I would expect mdev to mainly be found in the embedded world.

pkg/os/os.go
Random-Liu: Good to know. Thanks!

pkg/os/os.go
Random-Liu: Make sense.

Random-Liu:
> 
I think Alpine uses mdev by default (although it does have the eudev fork of udev available too). I'm not sure but I wouldn't be surprised if Gentoo supported mdev at least as an option (likewise other "from source" or "from scratch" type distros). Other than that I would expect mdev to mainly be found in the embedded world.

Thanks for the education. :)

PR #301, opened 2017-09-28
Fix container stats.

Random-Liu:
`Inodes` and `UsedBytes` are not set correctly.


abhi:
LGTM

PR #323, opened 2017-10-04
Add node e2e test timeout.

Random-Liu:
Add timeout for node e2e test.

The build is red now, it seems that cri-containerd or containerd gets unresponsive during test running. However,
1) I could not reproduce it in my local machine;
2) because of travis 50m test timeout, it doesn't even get a chance to upload test log, so that there is no way to debug it.

This PR added 40m timeout to node e2e test. Usually it only takes 30-31 minutes. If it timeout, test will abort, and we still get the chance to collect test logs.


abhi:
LGTM

Random-Liu:
@abhi Thanks!

PR #329, opened 2017-10-05
Update `hack/versions`

Random-Liu:
Update Kubernetes to https://github.com/kubernetes/kubernetes/commit/c6a3f26988bce604c035dcd9024ce775ae4aafef.

The previous commit is not correct, it does not include change we need.


abhi:
LGTM

PR #330, opened 2017-10-05
[Do not merge] Skip container stats which have empty cpu or memory stats.

Random-Liu:
This is not defined in CRI today, but this is the expected behavior by kubelet.
See https://github.com/kubernetes/kubernetes/issues/53514.


Random-Liu:
@abhi Please hold this PR for now. We need to figure out https://github.com/kubernetes/kubernetes/issues/53514 first.

Random-Liu:
Close this, and we'll update kubelet to filter out duplicated container stats.

Issue #326, opened 2017-10-04
Use new container update function in containerd client.


Random-Liu:
After https://github.com/containerd/containerd/pull/1590 and https://github.com/containerd/containerd/pull/1595,

1) We don't need to lock around containerd client.
2) We could use `Update` function in containerd client.

PR #332, opened 2017-10-06
Use container update

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/326.

This PR update containerd and uses the new container update function added in https://github.com/containerd/containerd/pull/1590.


vendor.conf
```diff
@@ -37,7 +37,7 @@ github.com/mailru/easyjson d5b7844b561a7bc640052f1b935f7b800330d7e0
 github.com/Microsoft/go-winio v0.4.4
 github.com/opencontainers/go-digest 21dfd564fd89c944783d00d069f33e3e7123c448
 github.com/opencontainers/image-spec v1.0.0
-github.com/opencontainers/runc 593914b8bd5448a93f7c3e4902a03408b6d5c0ce
+github.com/opencontainers/runc 0351df1c5a66838d0c392b4ac4cf9450de844e2d
```
mikebrow: See: https://github.com/containerd/containerd/commit/e789135aa3ffa335b6a49ec52733dc13783747a2

pkg/server/container_update_resources.go
```diff
@@ -90,34 +93,22 @@ func (c *criContainerdService) updateContainerResources(ctx context.Context,
 				glog.Errorf("Failed to marshal spec %+v for container %q: %v", oldSpec, id, err)
 				return
 			}
-			info.Spec = any
-			if _, err := c.client.ContainerService().Update(ctx, info, "spec"); err != nil {
+			if err := cntr.Container.Update(ctx, func(ctx gocontext.Context, client *containerd.Client, c *containers.Container) error {
+				c.Spec = any
```
mikebrow: Nit..  from the marshal to the call to center.Container.Update the code is duplicative with the only exception being which spec is being marshaled.. So, could be made into a update helper passing in the spec and returning one of the marshal error or the possible .Update error if Crosby's .Update is capable of generating an error outside the update function. But that could be done later...

pkg/server/sandbox_list.go
```diff
@@ -55,7 +56,15 @@ func (c *criContainerdService) ListPodSandbox(ctx context.Context, r *runtime.Li
 			state = runtime.PodSandboxState_SANDBOX_READY
 		}
 
-		createdAt := sandboxInStore.Container.Info().CreatedAt
+		info, err := sandboxInStore.Container.Info(ctx)
+		if err != nil {
+			// It's possible that container gets deleted during list.
+			if errdefs.IsNotFound(err) {
+				continue
```
mikebrow: Good!

vendor.conf
Random-Liu: Will update.

Random-Liu: Done.

pkg/server/container_update_resources.go
Random-Liu: Will do.

Random-Liu: Done.

vendor.conf
Random-Liu: We can't update now. The `docker/pkg/signal` package we are using is still using `unix.SIGUNUSED`.

However, the new `x/sys` package removed that https://github.com/containerd/containerd/commit/727fd599fba8218274a4acc3c1014823875e0e05.

Random-Liu: I'll file an issue.

mikebrow: yeah saw some serious changes.. thus the reason for pointing it out. To bad about the docker dependency.

Random-Liu:
@mikebrow Addressed comments.

Random-Liu:
Filed https://github.com/kubernetes-incubator/cri-containerd/issues/333 to track this.

vendor.conf
Random-Liu: Filed #333 to track this.

mikebrow:
some odd "spurious" test result on an image pull..

Random-Liu:
> @Random-Liu you got some odd "spurious?" test result on an image pull.. Maybe restart it..

Seems like a gcr flake. Usually happen to us in Kubernetes test. That's why we prepull image. :)

mikebrow:
if string.Contains(image.name,"gcr.io") retry

PR #334, opened 2017-10-07
Change `Version` to return cri-containerd version instead.

Random-Liu:
Changed `Version` to use cri-containerd version.
We still want to call the runtime `containerd` in the future not to confuse people, but we do need both cri-containerd and containerd version if we don't bundle them well.

This is a problem we need to discuss and fix, but for now just expose cri-containerd version for ease of our development, because we could always infer containerd version from cri-containerd version in our test environment.

### DIFF
--- a/pkg/server/version.go
+++ b/pkg/server/version.go
@@ -17,14 +17,19 @@ limitations under the License.
 package server
 
 import (
-	"fmt"
-
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+
+	"github.com/kubernetes-incubator/cri-containerd/pkg/version"
 )
 
 const (
-	containerName        = "containerd"
+	// For now, containerd and runc are bundled with cri-containerd, cri-containerd
+	// version is more important to us.
+	// TODO(random-liu): Figure out how to package cri-containerd and containerd,
+	// and how to version it. We still prefer calling the container runtime "containerd",
+	// but we care both the cri-containerd version and containerd version.
+	containerName        = "cri-containerd"
 	containerdAPIVersion = "0.0.0"
 	// kubeAPIVersion is the api version of kubernetes.
 	kubeAPIVersion = "0.1.0"
@@ -32,14 +37,10 @@ const (
 
 // Version returns the runtime name, runtime version and runtime API version.
 func (c *criContainerdService) Version(ctx context.Context, r *runtime.VersionRequest) (*runtime.VersionResponse, error) {
-	resp, err := c.client.Version(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get containerd version: %v", err)
-	}
 	return &runtime.VersionResponse{
 		Version:        kubeAPIVersion,
 		RuntimeName:    containerName,
-		RuntimeVersion: resp.Version,
+		RuntimeVersion: version.CRIContainerdVersion,
 		// Containerd doesn't have an api version now.
 		RuntimeApiVersion: containerdAPIVersion,
 	}, nil
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -22,7 +22,8 @@ import (
 	"github.com/blang/semver"
 )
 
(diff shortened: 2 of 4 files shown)


Random-Liu:
New result:
```console
# crictl --runtime-endpoint=/var/run/cri-containerd.sock info
Version:  0.1.0
RuntimeName:  cri-containerd
RuntimeVersion:  1.0.0-alpha.0-27-ge77795b-dirty
RuntimeApiVersion:  0.0.0
```

pkg/server/version.go
```diff
@@ -17,29 +17,30 @@ limitations under the License.
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to get containerd version: %v", err)
-	}
 	return &runtime.VersionResponse{
 		Version:        kubeAPIVersion,
 		RuntimeName:    containerName,
-		RuntimeVersion: resp.Version,
+		RuntimeVersion: version.CRIContainerdVersion,
 		// Containerd doesn't have an api version now.
 		RuntimeApiVersion: containerdAPIVersion,
```
mikebrow: Why don't we use the CRI version here?

pkg/server/version.go
```diff
@@ -17,29 +17,30 @@ limitations under the License.
...
 )
 
 const (
-	containerName        = "containerd"
+	// For now, containerd and runc are bundled with cri-containerd, cri-containerd
+	// version is more important to us.
+	// TODO(random-liu): Figure out how to package cri-containerd and containerd,
+	// and how to version it. We still prefer calling the container runtime "containerd",
+	// but we care both the cri-containerd version and containerd version.
+	containerName        = "cri-containerd"
```
mikebrow: Maybe overload the containerName

if c.config.ContainerdConfig.RuntimeEngine == "" use "runc"

containerName = ... "cri-containerd; containerd; %s", runtimeengine

maybe 
containerName = ... "cri-containerd; containerd %v; %s %v", resp.Version, runtimeengine, runc.version

pkg/server/version.go
```diff
@@ -17,29 +17,30 @@ limitations under the License.
...
 	containerdAPIVersion = "0.0.0"
 	// kubeAPIVersion is the api version of kubernetes.
 	kubeAPIVersion = "0.1.0"
 )
 
 // Version returns the runtime name, runtime version and runtime API version.
 func (c *criContainerdService) Version(ctx context.Context, r *runtime.VersionRequest) (*runtime.VersionResponse, error) {
-	resp, err := c.client.Version(ctx)
```
mikebrow: hate hiding useful information... maybe fix this at the CRI level with an array of versions.. dunno

pkg/version/version.go
```diff
@@ -22,7 +22,8 @@ import (
 	"github.com/blang/semver"
 )
 
-var criContainerdVersion = "UNKNOWN"
+// CRIContainerdVersion is the version of the cri-containerd.
+var CRIContainerdVersion = "UNKNOWN"
```
mikebrow: why change the case here, to make this a public var? curious

pkg/server/version.go
Random-Liu: We need a better way to fix this, only override the runtime name is not enough.

pkg/server/version.go
Random-Liu: Yeah, that's why I filed the TODO. We should expose both containerd and cri-containerd verison if we don't bundle them together, but there is no way to do that in current CRI.

For our current testing, we need cri-containerd version more, so we use cri-containerd version for now. But we should find a way to fix this.

This is a cri-containerd specific problem, because we have 2 daemons out of kubelet. Both dockershim and cri-o don't have this problem, because they either built inside kubelet (dockershim) or only have one daemon (cri-o).

pkg/version/version.go
Random-Liu: To use it in `server/version.go`

pkg/server/version.go
Random-Liu: ? We don't have api version. Currently it's just 0.0.0

Issue #325, opened 2017-10-04
getDeviceUUID doesn't work for ZFS


Random-Liu:
Mount info for ZFS is like this:
```
tmpfs on /var type tmpfs (rw,nosuid,nodev,noexec,relatime,size=1023524k,mode=755)
testpool/nextnas/containerd-vol on /var/lib/containerd/io.containerd.snapshotter.v1.zfs type zfs (rw,xattr,noacl)
testpool on /var/zpool type zfs (rw,xattr,noacl)
testpool/nextnas on /var/zpool/nextnas type zfs (rw,xattr,noacl)
```

The pool is not a file on the disk, thus could not be stated.

We need to figure out how to get device UUID for zfs. Or maybe we need admin to pass the UUID to us for zfs if we could not figure it out.

Random-Liu:
Reopen this issue. According to https://github.com/kubernetes-incubator/cri-containerd/pull/335#issuecomment-336724360, ZFS still doesn't work.

The main problem is that ZFS doesn't have an uuid, because it manages the block/volume itself.

I feel like we should throw out an error and proceeding if we could not get uuid, the only thing won't work is that kubelet can't get image fs capacity. However, we should not block people from trying this out because of that.

Random-Liu:
We added `skip-imagefs-uuid` flag, and will be included in `v1.0.0-beta.1`. If the current uuid based approach doesn't work, user could use that flag to skip that step. It should not prevent user from trying out this.

I'll close this issue. Further discussion will be carried on in the kubernetes issue kubernetes/kubernetes#57356.

PR #335, opened 2017-10-09
Use device number to find uuid

Random-Liu:
Use linux device number from `mount.Lookup` to look for uuid.

This may be able to fix https://github.com/kubernetes-incubator/cri-containerd/issues/325.

And I also see the Ubuntu used by GKE, rootfs is mounted from `/dev/root`, which doesn't have real corresponding device file.

With this fix, we get device number from mount directly, and find uuid with the device number. This doesn't assume that the mount source has to be a existing file.

### DIFF
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -34,6 +34,7 @@ import (
 	runcapparmor "github.com/opencontainers/runc/libcontainer/apparmor"
 	runcseccomp "github.com/opencontainers/runc/libcontainer/seccomp"
 	"golang.org/x/net/context"
+	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 	"k8s.io/kubernetes/pkg/kubelet/server/streaming"
@@ -143,6 +144,7 @@ func NewCRIContainerdService(config options.Config) (CRIContainerdService, error
 	if err != nil {
 		return nil, fmt.Errorf("failed to get imagefs uuid of %q: %v", imageFSPath, err)
 	}
+	glog.V(2).Infof("Get device uuid %q for image filesystem %q", c.imageFSUUID, imageFSPath)
 
 	c.netPlugin, err = ocicni.InitCNI(config.NetworkPluginConfDir, config.NetworkPluginBinDir)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


Random-Liu:
@ijc Could you take a look?

abhi:
Looks fine to me.

pkg/server/service.go
```diff
@@ -243,11 +245,12 @@ func (c *criContainerdService) Stop() {
 
 // getDeviceUUID gets device uuid for a given path.
 func (c *criContainerdService) getDeviceUUID(path string) (string, error) {
-	info, err := c.os.LookupMount(path)
+	mount, err := c.os.LookupMount(path)
```
ijc: `LookupMount` has a comment in it:
> 		// Note that m.{Major, Minor} are generally unreliable for our purpose here
>		// https://www.spinics.net/lists/linux-btrfs/msg58908.html

But we are now using exactly those fields below.

ijc: Actually, we don't seem to rely on these any more or less than we did before, so I guess this is fine.

mikebrow: @Random-Liu @ijc Interesting.. if "generally unreliable" maybe we should do it both ways and compare results then return the best result.  Just thinking out loud and now worrying about what was meant by generally unreliable.

matjohn2:
Thanks for taking this one on @Random-Liu!
Will test if this fixes #325 and report back.

matjohn2:
Hey @Random-Liu 
Sadly this doesn't fix the ZFS issues in #325. For the same reason as the previous code, ZFS manages it's own volumes/filesystems/mounts etc, so the rest of the linux kernel (IIUC) doesn't know which disks are part of a ZFS pool (think LVM/RAID/etc) and which aren't.

If dealing with ZFS, the "truest" way to get a unique identifier which will ALWAYS identify that pool is the GUID property of the pool itself, so for the given mounts (from #325)

```
tmpfs on /var type tmpfs (rw,nosuid,nodev,noexec,relatime,size=1023524k,mode=755)
testpool/nextnas/containerd-vol on /var/lib/containerd/io.containerd.snapshotter.v1.zfs type zfs (rw,xattr,noacl)
testpool on /var/zpool type zfs (rw,xattr,noacl)
testpool/nextnas on /var/zpool/nextnas type zfs (rw,xattr,noacl)
```
the "pool name" is always the first word before the first / for a zfs mount, in our case, testpool.
Running the ZFS userspace tools;

```
zpool get guid testpool
```` 

gives us a UUID which will always identify that pool.
UUID/nextnas would always identify the other mount (as it's in the pool 'testpool').

image

Issue #304, opened 2017-09-28
"latest" tarball


Random-Liu:
We should have "latest" tarball for both `cri-containerd-release` and `cri-containerd-staging`.

Latest `cri-containerd-release` will be useful for user to fetch the latest cri-containerd release version.
Latest `cri-containerd-staging` will be useful for testing and trying out the latest cri-containerd dev version.

Ideas brought by @abhi

Random-Liu:
This is addressed in https://github.com/kubernetes-incubator/cri-containerd/pull/336.

Feel free to file another issue to update ansible playbook to consume it.

PR #336, opened 2017-10-09
Add staging tar for node e2e.

Random-Liu:
This PR:
1) Add separate tar for node e2e test which contains CNI in the tar.
2) Push a file named `latest` in `cri-containerd-staging`. Other tool could read `latest` to know the latest version. /cc @abhi 

Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/304.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,4 +81,6 @@ jobs:
       script:
         - test "${TRAVIS_PULL_REQUEST}" != "false" && exit 0 || true
         - make push
+        # Build a tarball including CNI for node e2e.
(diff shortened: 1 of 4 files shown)


Random-Liu:
In the future, when build release, we should also push with `PUSH_VERSION=true`.

Random-Liu:
Here is the PR add node e2e test in Kubernetes test infra https://github.com/kubernetes/kubernetes/pull/53581

abhi:
LGTM

PR #340, opened 2017-10-10
Disable maligned check.

Random-Liu:
It seems that there is some update on the gometalinter side.
The build is failing because of something like:
```
pkg/store/container/status.go:43:13:warning: struct of size 80 could be 72 (maligned)
```

Personally, I don't think we care about this that much. Code readability is more important for us. For example, we may want to place struct fields in logic order instead of moving them around to make the struct align better.

@abhi @mikebrow WDYT?


abhi:
LGTM. code readability yes. We can turn it off for now.  May be later during one of the code cleanup efforts we can get it back on if it helps better optimize memory usage.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-containerd/pull/340#issuecomment-335599738

PR #345, opened 2017-10-12
check image readiness when recover

### LINKED ISSUE #303: Use containerd image readiness check function to properly recover image during restart.
Currently, during restart we read all image references from containerd image metadata store, and recover the image list.

However, a image reference in containerd image metadata store doesn't mean the image is ready for use, the image could be corrupted because of aborted image pulling.

We should use the newly added `Check` function https://github.com/containerd/containerd/pull/1567 to check content completeness and check top level snapshot existence.

yanxuean:
fix #303

### DIFF
--- a/pkg/server/restart.go
+++ b/pkg/server/restart.go
@@ -26,6 +26,9 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/errdefs"
+	containerdimages "github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	"github.com/containerd/containerd/snapshot"
 	"github.com/containerd/typeurl"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/pkg/system"
@@ -96,7 +99,7 @@ func (c *criContainerdService) recover(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to list images: %v", err)
 	}
-	images, err := loadImages(ctx, cImages, c.client.ContentStore())
+	images, err := loadImages(ctx, cImages, c.client.ContentStore(), c.client.SnapshotService(c.config.ContainerdConfig.Snapshotter))
 	if err != nil {
 		return fmt.Errorf("failed to load images: %v", err)
 	}
@@ -325,7 +328,8 @@ func loadSandbox(ctx context.Context, cntr containerd.Container) (sandboxstore.S
 // loadImages loads images from containerd.
 // TODO(random-liu): Check whether image is unpacked, because containerd put image reference
 // into store before image is unpacked.
-func loadImages(ctx context.Context, cImages []containerd.Image, provider content.Provider) ([]imagestore.Image, error) {
+func loadImages(ctx context.Context, cImages []containerd.Image, provider content.Provider,
+	snapshotter snapshot.Snapshotter) ([]imagestore.Image, error) {
 	// Group images by image id.
 	imageMap := make(map[string][]containerd.Image)
 	for _, i := range cImages {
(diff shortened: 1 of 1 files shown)


pkg/server/restart.go
```diff
@@ -101,6 +103,11 @@ func (c *criContainerdService) recover(ctx context.Context) error {
 		return fmt.Errorf("failed to load images: %v", err)
 	}
 	for _, image := range images {
+		if _, err := c.client.SnapshotService(c.config.ContainerdConfig.Snapshotter).Stat(ctx, image.ChainID); err != nil {
+			glog.Warningf("Failed to check top snapshot %+v: %v", image, err)
```
mikebrow: top snapshot?

yanxuean: Thanks. What do you suggest?

mikebrow: "Failed to stat snapshot for image %+v: %v"
or 
"Failed to locate information for image %+v: %v"

yanxuean: We want to check readiness of top snapshot.  We do it by calling Stat() function now.
Is the error message better to indicating our purpose?

mikebrow: I understand your meaning and the issue description but most won't.  Top snapshot could have many meanings. Images have layers.. the image.chainid points to the top-level layer for an image snapshot and contains references to the rest of the layers. IMO it's better to just say what failed vs. using the issue description of what needed to be done here.  

For example:  "Failed to locate information for snapshot using chainid for image %+v: %v"
As another example: "Failed to stat the top-level snapshot for image %+v: %v"
I think those are more explanatory to the reader of the warnings than: "Failed to check top snapshot %+v: %v"

What do you think?

yanxuean: You are right. You have a lot of experience,we need to learn from you.:)
Thank for your review and guidance.

yanxuean: done

pkg/server/restart.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/errdefs"
+	imagemanage "github.com/containerd/containerd/images"
```
Random-Liu: This name is confusing. I prefer `containerdimages` if we really need another name for the package.

pkg/server/restart.go
```diff
@@ -101,6 +103,12 @@ func (c *criContainerdService) recover(ctx context.Context) error {
 		return fmt.Errorf("failed to load images: %v", err)
 	}
 	for _, image := range images {
+		// Checking existence of top-level snapshot for each image being recovered.
```
Random-Liu: Why not add this logic into `loadImages` function?

pkg/server/restart.go
```diff
@@ -342,6 +350,10 @@ func loadImages(ctx context.Context, cImages []containerd.Image, provider conten
 		// imgs len must be > 0, or else the entry will not be created in
 		// previous loop.
 		i := imgs[0]
+		if ok, _, _, _, err := imagemanage.Check(ctx, provider, i.Target(), platforms.Default()); !ok || err != nil {
+			glog.Warningf("Failed to check image readiness for %q: %v", i.Name(), err)
```
Random-Liu: Let's log different for `!ok` and `err != nil`.

Random-Liu: Failed to check image content readiness for %q: %v

pkg/server/restart.go
yanxuean: Because we need call c.client.SnapshotService.

Random-Liu: Just pass it in. :)

Random-Liu: I feel like logically this should be part of `loadImages`, we load it only when it is ready.

yanxuean: ok

pkg/server/restart.go
yanxuean: done

pkg/server/restart.go
yanxuean: done

Random-Liu:
LGTM

PR #348, opened 2017-10-12
Remove the oom score in systemd unit. cri-containerd sets it by itself.

Random-Liu:
Remove OOMScore from systemd units. The default -999 oom score is set in https://github.com/kubernetes-incubator/cri-containerd/pull/347.


Random-Liu:
--

PR #352, opened 2017-10-19
Add integration test for truncindex

### LINKED ISSUE #343: Add intergration test for truncindex
for  #222

yanxuean:
fix #343

### DIFF
--- a/integration/truncindex_test.go
+++ b/integration/truncindex_test.go
@@ -1,0 +1,157 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	runtimeapi "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+func genTruncIndex(normalName string) string {
+	return normalName[:(len(normalName)+1)/2]
+}
+
+func TestTruncIndex(t *testing.T) {
+	t.Logf("Pull an image")
+	const appImage = "busybox"
+	imgID, err := imageService.PullImage(&runtimeapi.ImageSpec{Image: appImage}, nil)
+	require.NoError(t, err)
+	imgTruncId := genTruncIndex(imgID)
+	defer func() {
+		assert.NoError(t, imageService.RemoveImage(&runtimeapi.ImageSpec{Image: imgTruncId}))
+	}()
+
+	t.Logf("Get image status by truncindex, truncId: %s", imgTruncId)
+	res, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId})
+	require.NoError(t, err)
+	require.NotEqual(t, nil, res)
+	assert.Equal(t, imgID, res.Id)
+
+	// TODO(yanxuean): for failure test case where there are two images with the same truncindex.
+	// if you add n images at least two will share the same leading digit.
+	// "sha256:n" where n is the a number from 0-9 where two images have the same trunc,
+	// for example sha256:9
+	// https://github.com/kubernetes-incubator/cri-containerd/pull/352
+	// I am thinking how I get the two image which have same trunc.
+
+	// TODO(yanxuean): add test case for ListImages
+
+	t.Logf("Create a sandbox")
+	sbConfig := PodSandboxConfig("sandbox", "truncindex")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	sbTruncIndex := genTruncIndex(sb)
+	var hasStopedSandbox bool
+	defer func() {
+		// The 2th StopPodSandbox will fail, the 2th RemovePodSandbox will success.
+		if !hasStopedSandbox {
+			assert.NoError(t, runtimeService.StopPodSandbox(sbTruncIndex))
+		}
+		assert.NoError(t, runtimeService.RemovePodSandbox(sbTruncIndex))
+	}()
+
+	t.Logf("Get sandbox status by truncindex")
+	sbStatus, err := runtimeService.PodSandboxStatus(sbTruncIndex)
+	require.NoError(t, err)
+	assert.Equal(t, sb, sbStatus.Id)
+
+	t.Logf("Forward port for sandbox by truncindex")
+	_, err = runtimeService.PortForward(&runtimeapi.PortForwardRequest{sbTruncIndex, []int32{80}})
+	assert.NoError(t, err)
+
+	// TODO(yanxuean): add test case for ListPodSandbox
+
+	t.Logf("Create a container")
+	cnConfig := ContainerConfig(
+		"containerTruncIndex",
+		appImage,
+		WithCommand("top"),
+	)
+	cn, err := runtimeService.CreateContainer(sbTruncIndex, cnConfig, sbConfig)
+	require.NoError(t, err)
+	cnTruncIndex := genTruncIndex(cn)
+	defer func() {
+		// the 2th RemovePodSandbox will success.
+		assert.NoError(t, runtimeService.RemoveContainer(cnTruncIndex))
+	}()
+
+	t.Logf("Get container status by truncindex")
+	cStatus, err := runtimeService.ContainerStatus(cnTruncIndex)
+	require.NoError(t, err)
+	assert.Equal(t, cn, cStatus.Id)
+
+	t.Logf("Start the container")
+	require.NoError(t, runtimeService.StartContainer(cnTruncIndex))
+	var hasStopedContainer bool
+	defer func() {
+		// The 2th StopPodSandbox will fail
+		if !hasStopedContainer {
+			assert.NoError(t, runtimeService.StopContainer(cnTruncIndex, 10))
+		}
+	}()
+
+	t.Logf("Stats the container")
+	cStats, err := runtimeService.ContainerStats(cnTruncIndex)
+	require.NoError(t, err)
+	assert.Equal(t, cn, cStats.Attributes.Id)
+
+	t.Logf("Update container memory limit after started")
+	err = runtimeService.UpdateContainerResources(cnTruncIndex, &runtimeapi.LinuxContainerResources{
+		MemoryLimitInBytes: 50 * 1024 * 1024,
+	})
+	assert.NoError(t, err)
+
+	t.Logf("Execute cmd in container")
+	execReq := &runtimeapi.ExecRequest{
+		ContainerId: cnTruncIndex,
+		Cmd:         []string{"pwd"},
+		Stdout:      true,
+	}
+	_, err = runtimeService.Exec(execReq)
+	assert.NoError(t, err)
+
+	t.Logf("Execute cmd in container by sync")
+	_, _, err = runtimeService.ExecSync(cnTruncIndex, []string{"pwd"}, 10)
+	assert.NoError(t, err)
+
+	// TODO(yanxuean): add test case for ListContainers
+
+	t.Logf("Get a non exist container status by truncindex")
+	err = runtimeService.StopContainer(cnTruncIndex, 10)
+	assert.NoError(t, err)
+	if err == nil {
+		hasStopedContainer = true
+	}
+	_, err = runtimeService.ContainerStats(cnTruncIndex)
+	assert.Error(t, err)
+	assert.NoError(t, runtimeService.RemoveContainer(cnTruncIndex))
+	_, err = runtimeService.ContainerStatus(cnTruncIndex)
+	assert.Error(t, err)
+
+	t.Logf("Get a non exist sandbox status by truncindex")
+	err = runtimeService.StopPodSandbox(sbTruncIndex)
+	assert.NoError(t, err)
+	if err == nil {
+		hasStopedSandbox = true
+	}
+	assert.NoError(t, runtimeService.RemovePodSandbox(sbTruncIndex))
+	_, err = runtimeService.PodSandboxStatus(sbTruncIndex)
+	assert.Error(t, err)
+}


yanxuean:
@Random-Liu PTAL

Random-Liu:
/ok-to-test

@yanxuean We are in dockercon this week, will probably review a bit late. :) Sorry for the delay.

mikebrow:
I'll review in the morning.

integration/truncindex_test.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+func genTruncIndex(normalName string) string {
+	return normalName[:(len(normalName)+1)/2]
```
mikebrow: It makes sense for the unit tests to do this.. but for integration testing why use a fake TruncIndex? Why not use the vendor package "github.com/docker/docker/pkg/truncindex"?

integration/test_utils.go
```diff
@@ -106,6 +106,14 @@ func WithResources(r *runtime.LinuxContainerResources) ContainerOpts {
 	}
 }
 
+// Add container resource limits.
+func WithCommand(c string, args ...string) ContainerOpts {
```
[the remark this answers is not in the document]
yanxuean: done.

integration/truncindex_test.go
yanxuean: Sorry. I maybe do not understand what you mean.
I need to test if we can get something by its truncated name. so I make genTruncIndex function for getting its truncated name by regular name. The genTruncIndex is only for getting prefix of regular name. So the genTruncIndex is different with "github.com/docker/docker/pkg/truncindex".

mikebrow: I guess I'm trying to understand why the prefix being selected is normalName[:(len(normalName)+1)/2]. It feels like the truncated name is being faked. I see that pretty much anything will work when there are only a couple items in the index. You could probably use normalName[:1] and succeed this bucket :-) The lookup merely needs to be a unique prefix in the patricia tree for the working test and a non-unique prefix for a failing test.  Maybe if we forced a failing test and a successful test this bucket would make more sense?

mikebrow: Is there some kubelet, docker, or patricia tree rule that truncated name is the sliced prefix normalName[:(len(normalName)+1)/2]?

mikebrow: Put another way.. what are the valid and invalid  use cases that we can expect from calls to these services apis.

yanxuean: The genTruncIndex is confused, I will improve it.  And I will a invalid use case.

yanxuean:
PTAL

mikebrow:
Were you going to add an invalid case for the truncated name lookups? For example, two sandboxes with similar names then address by a common prefix?

yanxuean:
Oh. It is difficult to do it in intergration test. Because we can not generate id as expected ,so we can not get two similar name. 
Do you think?
By the way, we have the test case in unit test.

yanxuean:
And the test case look like we are derectly testing trunkindex self  not cri that is the user of truncindex :)

yanxuean:
We can make a test case as below.
A sandbox can be addressed before it be deleted,and cannot be addressed after it be deleted.

mikebrow:
I might be wrong but I thought if you had two similar named pods that they might end up with the truncid depending on how many chars you use for the prefix. Which would cause an error when passing in the common truncid that you are generating based on slice [:3]. 

Eg: pod name "foo1" pod name "foo2" truncid through your :3 slice api is "foo" for both pods lookup status on "foo" should error, no? If not don't worry.

yanxuean:
I will try it.

yanxuean:
I tried it.  It is impossible to construct 2 similar truncID.
Because cri-containerd generate ID by rand. It is not relative with metadata.
```go
func GenerateNonCryptoID() string {
	return generateID(readerFunc(rand.Read))
}
```

the result as below:
```
			truncindex_test.go:61: Create a sandbox,2fc2377e24ad91d338c4498749b035e6100f30883a3afc3df7c2ec2e49e24dc0, &runtime.PodSandboxMetadata{Name:"sandbox", Uid:"30cfd3535531cd0c8427eb625f8bdc0ad9e610d4d9e2b450e6aa9376ea6d65fa", Namespace:"TruncIndex", Attempt:0x0}
	truncindex_test.go:62: Create a sandbox,a981bdbf63af11e0437aade8df3b6709ea2ddf042f768431e8f96848bfd9b678, &runtime.PodSandboxMetadata{Name:"sandbox1", Uid:"8fe6422590e7ead5479c54e226e28cc726660272571c40da7c123d11ece5579c", Namespace:"TruncIndex", Attempt:0x0}
```

yanxuean:
I  add a test case as below.
A sandbox can be addressed before it be deleted,and cannot be addressed after it be deleted.

yanxuean:
@Random-Liu  @mikebrow    PTAL

mikebrow:
I think I might have the answer here.. if the status lookup fails it isn't returning an error. If I remember right that's how @Random-Liu wanted / needed it to work?  Look here:

```
mike@mike-VirtualBox:~/go/src/github.com/kubernetes-incubator/cri-containerd$ sudo git diff
diff --git a/integration/truncindex_test.go b/integration/truncindex_test.go
index bc7e0c6..3ac580a 100644
--- a/integration/truncindex_test.go
+++ b/integration/truncindex_test.go
@@ -30,7 +30,7 @@ func genTruncIndex(normalName string) string {
 
 func TestTruncIndex(t *testing.T) {
        t.Logf("Pull an image")
-       const appImage = "busybox"
+       const appImage = "busybox:1.26"
        img, err := imageService.PullImage(&runtimeapi.ImageSpec{Image: appImage}, nil)
        require.NoError(t, err)
        defer func() {
@@ -39,8 +39,25 @@ func TestTruncIndex(t *testing.T) {
        imgTruncId := genTruncIndex(img)
 
        t.Logf("Get image status by truncindex")
-       _, err = imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId})
+       st, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId})
        require.NoError(t, err)
+       t.Logf("image status: %+v", st)
+
+       t.Logf("Pull another image")
+       const appImage2 = "busybox:1.27"
+       img2, err := imageService.PullImage(&runtimeapi.ImageSpec{Image: appImage2}, nil)
+       require.NoError(t, err)
+       defer func() {
+               assert.NoError(t, imageService.RemoveImage(&runtimeapi.ImageSpec{Image: img2}))
+       }()
+       imgTruncId2 := genTruncIndex(img2)
+
+       t.Logf("Get image2 status truncid")
+       st2, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId2})
+       require.NoError(t, err)
+       t.Logf("image2 status: %+v", st2)
+
+       t.Logf("img & img2: \"%s\",\"%s\" - Truncid & Truncid2: \"%s\",\"%s\"", img, img2, imgTruncId, imgTruncId2)
 
        // TODO(yanxuean): add test case for ListImages
```
And the output:
```
mike@mike-VirtualBox:~/go/src/github.com/kubernetes-incubator/cri-containerd$ sudo make test-integration FOCUS=TestTruncIndex
go test -c github.com/kubernetes-incubator/cri-containerd/integration -o _output/integration.test
1 attempt "sudo /home/mike/go/bin/crictl --runtime-endpoint=/var/run/cri-containerd.sock info"! Trying again in 1 seconds...
=== RUN   TestTruncIndex
--- PASS: TestTruncIndex (7.11s)
	truncindex_test.go:32: Pull an image
	truncindex_test.go:41: Get image status by truncindex
	truncindex_test.go:44: image status: nil
	truncindex_test.go:46: Pull another image
	truncindex_test.go:55: Get image2 status truncid
	truncindex_test.go:58: image2 status: nil
	truncindex_test.go:60: img & img2: "sha256:c30178c5239f2937c21c261b0365efcda25be4921ccb95acd63beeeb78786f27","sha256:54511612f1c4d97e93430fc3d5dc2f05dfbe8fb7e6259b7351deeca95eaf2971" - Truncid & Truncid2: "sha","sha"
	truncindex_test.go:64: Create a sandbox
	truncindex_test.go:74: Get sandbox status by truncindex
	truncindex_test.go:78: Forward port for sandbox by truncindex
	truncindex_test.go:84: Create a container
	truncindex_test.go:97: Get container status by truncindex
	truncindex_test.go:101: Start the container
	truncindex_test.go:107: Stats the container
	truncindex_test.go:111: Update container memory limit after started
	truncindex_test.go:117: Execute cmd in container
	truncindex_test.go:126: Execute cmd in container by sync
	truncindex_test.go:132: Get a non exist sandbox status by truncindex
PASS
```
Note: no error was returned.. and neither was the status, as st was set to nil on the status request.

so.... probably need to revisit this a bit.

Random-Liu:
@yanxuean I'll review this PR this weekend.

yanxuean:
> I think I might have the answer here or at least an issue to address.. if the status lookup fails it isn't returning an error. If I remember right that's how @Random-Liu wanted / needed it to work? Look here:

This is  a problem. I have not noticed it. Tks.

yanxuean:
The image is more complex than container and sandbox:)

yanxuean:
We maybe have to modify criContainerdService.localResolve.
In localResolve, we treat the arg "ref" as truncated imageID first.
if we find image by the "truncID" in imageStore, that is good.
if not, we continue to handle it like originly.
The regular imageID all have the prefix "sha256:",  It can clearly distinguish from the image refence. 
Do you think about it?   @mikebrow @Random-Liu

integration/truncindex_test.go
```diff
@@ -0,0 +1,127 @@
...
+func genTruncIndex(normalName string) string {
+	return normalName[:3]
+}
+
+func TestTruncIndex(t *testing.T) {
+	t.Logf("Pull an image")
+	const appImage = "busybox"
+	img, err := imageService.PullImage(&runtimeapi.ImageSpec{Image: appImage}, nil)
+	require.NoError(t, err)
+	// skip the prifix "sha256:"
```
mikebrow: /s/prifix/prefix/

integration/truncindex_test.go
```diff
@@ -0,0 +1,127 @@
...
+	defer func() {
+		assert.NoError(t, imageService.RemoveImage(&runtimeapi.ImageSpec{Image: imgTruncId}))
+	}()
+
+	t.Logf("Get image status by truncindex, truncId: %s", imgTruncId)
+	res, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId})
+	require.NoError(t, err)
+	require.Equal(t, true, (res != nil) && (res.Id != ""))
+	t.Logf("Image status: %v", res)
+
```
mikebrow: Please add a todo for failure test case where there are two images with the same truncindex.. if you add n images at least two will share the same leading digit.  "sha256:n" where n is the a number from 0-9 where two images have the same trunc, for example `sha256:9`

integration/truncindex_test.go
```diff
@@ -0,0 +1,127 @@
...
+	return normalName[:3]
+}
+
+func TestTruncIndex(t *testing.T) {
+	t.Logf("Pull an image")
+	const appImage = "busybox"
+	img, err := imageService.PullImage(&runtimeapi.ImageSpec{Image: appImage}, nil)
+	require.NoError(t, err)
+	// skip the prifix "sha256:"
+	imgTruncId := img[:len(img)/2]
```
mikebrow: (len(img)+1)/2

yanxuean:
Thanks for review. I will go to sleep. I will do late.

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: The img is from PullImage(), it is regular imageID. so img[:len(img)/2] is sufficient.  :)

integration/truncindex_test.go
yanxuean: @mikebrow 
There are some probability, but I have not good thoughts for handling it. Do you give me some suggestion?
Do you think that we do it in unit test of ImageStore? 
e.g.   in cri-containerd/pkg/store/image/image_test.go

mikebrow: The test is fine for now.. just need to add a todo to cover adding 11 different images to force having two with the same leading x digit. Then loop on the images getting status. I was thinking just do it here, but I suppose you could also do it in the image store unit test. :-)

yanxuean: done

yanxuean:
@random-liu    PTAL

Random-Liu:
@yanxuean #375 is merged. Thanks for working on this!

yanxuean:
rebase + remove confused comment
@Random-Liu @mikebrow  PTAL

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	runtimeapi "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+)
+
+func genTruncIndex(normalName string) string {
+	return normalName[:3]
```
Random-Liu: Can we change this to `normalName[:len(normalName)/2]`? So that we could use the same function for image/container/sandbox id.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NoError(t, err)
+	imgTruncId := img[:len(img)/2]
+	defer func() {
+		assert.NoError(t, imageService.RemoveImage(&runtimeapi.ImageSpec{Image: imgTruncId}))
+	}()
+
+	t.Logf("Get image status by truncindex, truncId: %s", imgTruncId)
+	res, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId})
+	require.NoError(t, err)
+	require.Equal(t, true, (res != nil) && (res.Id != ""))
```
Random-Liu: I prefer separate these 2 checks.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+func genTruncIndex(normalName string) string {
+	return normalName[:3]
+}
+
+func TestTruncIndex(t *testing.T) {
+	t.Logf("Pull an image")
+	const appImage = "busybox"
+	img, err := imageService.PullImage(&runtimeapi.ImageSpec{Image: appImage}, nil)
+	require.NoError(t, err)
+	imgTruncId := img[:len(img)/2]
```
Random-Liu: keep the original id to compare with the id in `ImageStatus` to make sure we get the image status we want.

integration/truncindex_test.go
Random-Liu: We just need to make sure an error is returned, if the trunc index is too short, and causes ambiguity.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	defer func() {
+		assert.NoError(t, imageService.RemoveImage(&runtimeapi.ImageSpec{Image: imgTruncId}))
+	}()
+
+	t.Logf("Get image status by truncindex, truncId: %s", imgTruncId)
+	res, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: imgTruncId})
+	require.NoError(t, err)
+	require.Equal(t, true, (res != nil) && (res.Id != ""))
+	t.Logf("Image status: %v", res)
```
Random-Liu: This will be spammy, don't need to log out.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NoError(t, err)
+	sbTruncIndex := genTruncIndex(sb)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sbTruncIndex))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sbTruncIndex))
+	}()
+
+	t.Logf("Get sandbox status by truncindex")
+	_, err = runtimeService.PodSandboxStatus(sbTruncIndex)
+	require.NoError(t, err)
```
Random-Liu: Compare id.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// if you add n images at least two will share the same leading digit.
+	// "sha256:n" where n is the a number from 0-9 where two images have the same trunc,
+	// for example sha256:9
+	// https://github.com/kubernetes-incubator/cri-containerd/pull/352
+	// I am thinking how I get the two image which have same trunc.
+
+	// TODO(yanxuean): add test case for ListImages
+
+	t.Logf("Create a sandbox")
+	sbConfig := PodSandboxConfig("sandbox", "TruncIndex")
```
Random-Liu: nit: s/TruncIndex/truncindex

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	)
+	cn, err := runtimeService.CreateContainer(sbTruncIndex, cnConfig, sbConfig)
+	require.NoError(t, err)
+	cnTruncIndex := genTruncIndex(cn)
+	defer func() {
+		assert.NoError(t, runtimeService.RemoveContainer(cnTruncIndex))
+	}()
+
+	t.Logf("Get container status by truncindex")
+	_, err = runtimeService.ContainerStatus(cnTruncIndex)
```
Random-Liu: Compare id.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NoError(t, err)
+
+	t.Logf("Start the container")
+	require.NoError(t, runtimeService.StartContainer(cnTruncIndex))
+	defer func() {
+		assert.NoError(t, runtimeService.StopContainer(cnTruncIndex, 10))
+	}()
+
+	t.Logf("Stats the container")
+	_, err = runtimeService.ContainerStats(cnTruncIndex)
```
Random-Liu: Compare id.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+		Stdout:      true,
+	}
+	_, err = runtimeService.Exec(execReq)
+	require.NoError(t, err)
+
+	t.Logf("Execute cmd in container by sync")
+	_, _, err = runtimeService.ExecSync(cnTruncIndex, []string{"pwd"}, 10)
+	require.NoError(t, err)
+
+	// TODO(yanxuean): add test case for ListContainers
```
Random-Liu: Just stop and delete current sandbox, and get it again.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NoError(t, err)
+
+	t.Logf("Execute cmd in container")
+	execReq := &runtimeapi.ExecRequest{
+		ContainerId: cnTruncIndex,
+		Cmd:         []string{"pwd"},
+		Stdout:      true,
+	}
+	_, err = runtimeService.Exec(execReq)
+	require.NoError(t, err)
```
Random-Liu: assert.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	t.Logf("Stats the container")
+	_, err = runtimeService.ContainerStats(cnTruncIndex)
+	require.NoError(t, err)
+
+	t.Logf("Update container memory limit after started")
+	err = runtimeService.UpdateContainerResources(cnTruncIndex, &runtimeapi.LinuxContainerResources{
+		MemoryLimitInBytes: 50 * 1024 * 1024,
+	})
+	require.NoError(t, err)
```
Random-Liu: assert.

integration/truncindex_test.go
```diff
@@ -0,0 +1,133 @@
...
+		assert.NoError(t, runtimeService.RemovePodSandbox(sbTruncIndex))
+	}()
+
+	t.Logf("Get sandbox status by truncindex")
+	_, err = runtimeService.PodSandboxStatus(sbTruncIndex)
+	require.NoError(t, err)
+
+	t.Logf("Forward port for sandbox by truncindex")
+	_, err = runtimeService.PortForward(&runtimeapi.PortForwardRequest{sbTruncIndex, []int32{80}})
+	require.NoError(t, err)
```
Random-Liu: assert

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: I will compare img with the result of ImageStatus.

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: Now we can do it. Now there is sha256 prefix in truncindex.

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: Are there rule for using assert and require?

yanxuean: done

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: There are defer functioin in first RunPodSandbox. If we explicitly stop and remove sandbox, the defer will fail
```
defer func() {
		assert.NoError(t, runtimeService.StopPodSandbox(sbTruncIndex))
		assert.NoError(t, runtimeService.RemovePodSandbox(sbTruncIndex))
	}()
```

integration/truncindex_test.go
yanxuean: We cann't do it. When we use sha256 prefix, digestSet handle it as invalidFormat and imageStore.Get return store.ErrNotExist.

integration/truncindex_test.go
Random-Liu: Sorry. The rule is that, if the assertion doesn't affect following test, we use `assert`.
If we could not move on without the assertion, we should use `require`.

In your case:
* Here, because no following assertion relying on whether `err == nil` or not, you could use assert. So that we could get as much test result as we can.
* However, if you want to use the return value, e.g. compare the id. You may want to use `require.NoError(t, err)`. Because if not, you'll get a panic accessing `nil` return value, and the whole test suite will fail right away.

So in your case, some of assertion should use `assert`, some should use `require`. You may want to update based on the convention above. The goal is to get as much test result as we can in a run. :)

integration/truncindex_test.go
Random-Liu: You can remove the assertion above. That is only for cleanup. In normal test run, it should get here and stop/remove.
You can add assertion here. ;)

If you create a new pod sandbox, you don't even know whether the trunc index has conflict with existing pod sandbox or not, because you haven't tried. So it's still possible that `runtimeService.PodSandboxStatus(sbTruncIndex1)` returns no error.

This is a little bit picky. It's up to you. :)

integration/truncindex_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	t.Logf("Start the container")
+	assert.NoError(t, runtimeService.StartContainer(cnTruncIndex))
+	defer func() {
+		assert.NoError(t, runtimeService.StopContainer(cnTruncIndex, 10))
+	}()
+
+	t.Logf("Stats the container")
+	cStats, err := runtimeService.ContainerStats(cnTruncIndex)
+	assert.NoError(t, err)
```
Random-Liu: e.g. you may want to use `require` here to avoid panic.

Random-Liu:
LGTM overall with final nits.

integration/truncindex_test.go
yanxuean: I see. Thank you.

integration/truncindex_test.go
yanxuean: done

integration/truncindex_test.go
yanxuean: done

yanxuean:
@Random-Liu add a testcase

t.Logf("Get a non exist container status by truncindex")

integration/truncindex_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// I am thinking how I get the two image which have same trunc.
+
+	// TODO(yanxuean): add test case for ListImages
+
+	t.Logf("Create a sandbox")
+	sbConfig := PodSandboxConfig("sandbox", "truncindex")
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	sbTruncIndex := genTruncIndex(sb)
+	var hasStopedSandbox bool
```
Random-Liu: s/Stoped/Stopped.

I won't block the PR for this. Will send another PR to fix this.

PR #358, opened 2017-10-25
Also unpack image during creation.

Random-Liu:
To pull an image in containerd, we need to:
1) Download all contents;
2) Write image reference.
3) Unpack image into snapshotter.

Currently, if 3) fails, we don't rollback 1) and 2). If we don't put the image into our list, containerd will not garbage collect it, and the content will be left there forever.

We have 2 options for this:
1) Rollback 1) and 2) ourselves, which is a bit unreliable, expensive and complex.
2) Only do best effort unpack during image pulling, and always check snapshot existence and unpack image before container creation. This is more reliable, and is also recommended by https://github.com/containerd/containerd/pull/1567#issuecomment-333632696 @stevvooe 

This PR goes with option 2.

### DIFF
--- a/pkg/opts/container.go
+++ b/pkg/opts/container.go
@@ -21,11 +21,40 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/errdefs"
 	"github.com/docker/docker/pkg/chrootarchive"
 	"github.com/docker/docker/pkg/system"
+	"github.com/opencontainers/image-spec/identity"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
+
+// WithImageUnpack guarantees that the image used by the container is unpacked.
+func WithImageUnpack(i containerd.Image) containerd.NewContainerOpts {
+	return func(ctx context.Context, client *containerd.Client, c *containers.Container) error {
+		if c.Snapshotter == "" {
+			return errors.New("no snapshotter set for container")
+		}
+		snapshotter := client.SnapshotService(c.Snapshotter)
+		diffIDs, err := i.RootFS(ctx)
+		if err != nil {
+			return errors.Wrap(err, "get image diff IDs")
+		}
+		chainID := identity.ChainID(diffIDs)
+		_, err = snapshotter.Stat(ctx, chainID.String())
+		if err == nil {
+			return nil
+		}
+		if !errdefs.IsNotFound(err) {
+			return errors.Wrap(err, "stat snapshot")
+		}
+		// Unpack the snapshot.
+		if err := i.Unpack(ctx, c.Snapshotter); err != nil {
+			return errors.Wrap(err, "unpack snapshot")
+		}
+		return nil
+	}
+}
 
 // WithVolumes copies ownership of volume in rootfs to its corresponding host path.
 // It doesn't update runtime spec.
@@ -34,10 +63,10 @@ import (
 func WithVolumes(volumeMounts map[string]string) containerd.NewContainerOpts {
 	return func(ctx context.Context, client *containerd.Client, c *containers.Container) error {
 		if c.Snapshotter == "" {
-			return errors.Errorf("no snapshotter set for container")
+			return errors.New("no snapshotter set for container")
 		}
 		if c.SnapshotKey == "" {
-			return errors.Errorf("rootfs not created for container")
+			return errors.New("rootfs not created for container")
 		}
 		snapshotter := client.SnapshotService(c.Snapshotter)
 		mounts, err := snapshotter.Mounts(ctx, c.SnapshotKey)
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -99,14 +99,20 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
 
 	// TODO(mikebrow): add truncIndex for image id
 	image, err := c.client.Pull(ctx, ref,
-		containerd.WithPullUnpack,
 		containerd.WithSchema1Conversion,
 		containerd.WithResolver(resolver),
-		containerd.WithPullSnapshotter(c.config.ContainerdConfig.Snapshotter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to pull image %q: %v", ref, err)
 	}
+
+	// Do best effort unpack.
+	glog.V(4).Info("Unpack image %q", imageRef)
+	if err := image.Unpack(ctx, c.config.ContainerdConfig.Snapshotter); err != nil {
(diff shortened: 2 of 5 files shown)


mikebrow:
I only see 2 options you said 3? I don't see how option 2, as described above, resolves the issue of not rolling back the download of all contents or rolling back the writing of the image reference. It seems like it just ignores the issue as stated and instead fixes the issue of not validating snapshot existence and fully unpacked image before container creation?

Random-Liu:
@mikebrow Edited.

Previously, a READY image to us is a image has all 1), 2) and 3) done.

With option 2, a READY image is a image has both 1), 2) done. We do best effort 3) during image pulling, so in most cases, we'll have an image with 1), 2) and 3) after pulling. However, if unpack really fails because of some temporary error, we'll unpack it again before container creation.

If unpack failure is not a temporary error, it is a bug, we should fix it.

mikebrow:
ok and the customer must... manually clean up on failures to unpack twice? Sorry, you had me at "containerd will not garbage collect it, and the content will be left there forever." But I do like the description of best effort and second effort before container creation to reduce the windows for failure due to such issues as insufficient storage, memory, bad image, invalid manifest,....

Random-Liu:
> ok and the customer must... manually clean up on failures to unpack twice? 

What do you mean? Nope. Customer doesn't need to do anything. :)

pkg/server/restart.go
```diff
@@ -391,7 +391,7 @@ func loadImages(ctx context.Context, cImages []containerd.Image, provider conten
 		// Checking existence of top-level snapshot for each image being recovered.
 		if _, err := snapshotter.Stat(ctx, image.ChainID); err != nil {
 			glog.Warningf("Failed to stat the top-level snapshot for image %+v: %v", image, err)
-			continue
+			// TODO(random-liu): Consider whether we should try unpack here.
```
yanxuean: Seem like we do not need it here. Because we will do it during CreateContainer.

Random-Liu: Usually, we assume container creation should be fast. We may want to try avoiding unpack before container creation as much as possible. :)

pkg/server/image_pull.go
```diff
@@ -99,14 +99,20 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to pull image %q: %v", ref, err)
 	}
+
+	// Do best effort unpack.
+	glog.V(4).Info("Unpack image %q", imageRef)
+	if err := image.Unpack(ctx, c.config.ContainerdConfig.Snapshotter); err != nil {
+		glog.Errorf("Failed to unpack image %q: %v", imageRef, err)
+		// Do not fail image pulling. Unpack will be retried when container creation.
```
mikebrow: /s/when/during/

pkg/server/image_pull.go
```diff
@@ -99,14 +99,20 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
-		containerd.WithPullSnapshotter(c.config.ContainerdConfig.Snapshotter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to pull image %q: %v", ref, err)
 	}
+
+	// Do best effort unpack.
+	glog.V(4).Info("Unpack image %q", imageRef)
+	if err := image.Unpack(ctx, c.config.ContainerdConfig.Snapshotter); err != nil {
+		glog.Errorf("Failed to unpack image %q: %v", imageRef, err)
```
mikebrow: error, warning, or info?

Suggest making this info.

/s/Failed/Failed on first attempt/

pkg/server/image_pull.go
```diff
@@ -99,14 +99,20 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
...
 		containerd.WithSchema1Conversion,
 		containerd.WithResolver(resolver),
-		containerd.WithPullSnapshotter(c.config.ContainerdConfig.Snapshotter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to pull image %q: %v", ref, err)
 	}
+
+	// Do best effort unpack.
+	glog.V(4).Info("Unpack image %q", imageRef)
```
mikebrow: /s/Unpack/First unpack image attempt/  ?

pkg/server/restart.go
mikebrow: Yeah but on a restart, it probably does not matter if the delay for unpack is attributed to the create should all add up to the same cost?

pkg/opts/container.go
```diff
@@ -21,23 +21,49 @@ import (
...
+		chainID := identity.ChainID(diffIDs)
+		_, err = snapshotter.Stat(ctx, chainID.String())
+		if err == nil {
+			return nil
+		}
+		if !errdefs.IsNotFound(err) {
+			return errors.Wrap(err, "stat snapshot")
+		}
+		// Unpack the snapshot.
+		return i.Unpack(ctx, c.Snapshotter)
```
mikebrow: Maybe wrap the unpack error here with "unpack snapshot"?

pkg/server/restart.go
Random-Liu: @mikebrow Yeah, the cost is the same. We just need to decide whether we want a faster restart or faster container creation.

pkg/opts/container.go
Random-Liu: Will do.

pkg/server/image_pull.go
Random-Liu: We expect this to succeed in most cases. :) If this fails, it is an error to us.

The unpack before creation is just a fall back to handle the error.

pkg/server/image_pull.go
Random-Liu: Will change to `before`.

pkg/opts/container.go
Random-Liu: Done

pkg/server/restart.go
mikebrow: vote for faster container creation..

pkg/server/image_pull.go
mikebrow: But if we handle an error do we want it reported?

pkg/server/restart.go
Random-Liu: I feel like fast restart may be more important. :p Slow container creation only affect one container, but slow restart affect the whole node.

pkg/server/image_pull.go
Random-Liu: Hm, yeah, maybe warning makes more sense then.

pkg/server/restart.go
mikebrow: hmm can we have both faster restart and faster container creation when not restarting :-)

pkg/server/image_pull.go

PR #362, opened 2017-10-26
Add log of generated id for debugging.

Random-Liu:
Add log for generated id. We need these IDs to map container in containerd.log.


pkg/server/container_create.go
```diff
@@ -89,6 +89,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	// the same container.
 	id := util.GenerateID()
 	name := makeContainerName(config.GetMetadata(), sandboxConfig.GetMetadata())
+	glog.V(4).Infof("Generate id %q for container %q", id, name)
```
mikebrow: /s/generate/generated/

Random-Liu: Will do.

PR #366, opened 2017-10-27
Add stack dump.

Random-Liu:
Using the same code with https://github.com/containerd/containerd/pull/292.

### DIFF
--- a/cmd/cri-containerd/cri_containerd.go
+++ b/cmd/cri-containerd/cri_containerd.go
@@ -20,7 +20,10 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"runtime"
+	"syscall"
 
 	"github.com/docker/docker/pkg/reexec"
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


cmd/cri-containerd/cri_containerd.go
```diff
@@ -118,3 +122,28 @@ func validateConfig(o *options.CRIContainerdOptions) {
...
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGUSR1)
+	go func() {
+		for range c {
+			dumpStacks()
+		}
+	}()
+}
+
+func dumpStacks() {
```
mikebrow: from: https://golang.org/src/runtime/debug/stack.go, just a little cleaner with a modification to dump all (true)
```
	buf := make([]byte, 1024)
  	for {
  		n := runtime.Stack(buf, true)
  		if n < len(buf) {
  			return buf[:n]
  		}
  		buf = make([]byte, 2*len(buf))
  	}
```

Random-Liu: Done.

PR #368, opened 2017-10-28
Do not log container output in error log.

Random-Liu:
We should not log container output into our error log.
In my cluster, fluentd keeps logging error message, which fills up the disk. After that cri-containerd will not be able to redirect log for it because the disk is full.

In that case,
1) I see tons of error message including container output in cri-containerd.log, which is too spammy. So in this PR, I removed the container output from the error log.
2) I see `/tmp` directory is filled up very soon, because glog is logging into it by default. So in this PR I use `--logtostderr` instead of `--alsologtostderr` to avoid glog writing to `/tmp`.


Random-Liu:
@abhi Thanks!

PR #371, opened 2017-10-30
Fix removing state recover.

Random-Liu:
In a test, I saw that we could never remove a container any more once containerd container removal returns an error.
```
E1029 23:59:48.015653    1035 container_remove.go:56] failed to reset removing state for container "9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56": failed to checkpoint status to "/var/lib/cri-containerd/containers/9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56/status": open /var/lib/cri-containerd/containers/9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56/.tmp-status551243121: no such file or directory
E1029 23:59:48.015675    1035 instrumented_service.go:176] RemoveContainer for "9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56" failed, error: failed to delete containerd container "9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56": context deadline exceeded: unknown
E1030 00:26:47.270788    1035 instrumented_service.go:176] RemoveContainer for "9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56" failed, error: failed to set removing state for container "9271ad01c1ef091260fca1b432128d98d3b6e40735f8e6b3d8f6d8d10d58ac56": container is already in removing state
```

The reason is that `removing` state could not be reset, because the on-disk container status has been removed, thus the reset function fails.

This PR:
1) Move container status deletion to be in front of container root directory deletion. Status checkpoint file is in container root directory. We may want to atomically remove the status file before we remove the whole root directory.
2) Move container status/root directory deletion after containerd container deletion. Actually it doesn't matter we delete them before or after containerd container. Restart recovery could handle both cases. However, it makes more sense to make sure that a containerd container always has status associated in its whole lifecycle.
3) Address a TODO to distinguish `UpdateSync` and `Update`. Actually in most cases, we just need the code to run in a `Update` transaction to avoid race, and don't actually update the on-disk status. We could use `Update` for this case in the future. **This also fixed the issue I mentioned above. With this PR, we'll only use `Update` to set/reset removing state, which doesn't include any disk operation, and won't be affected by container status/root directory removal.**

### DIFF
--- a/pkg/server/container_remove.go
+++ b/pkg/server/container_remove.go
@@ -31,6 +31,7 @@ import (
 )
 
 // RemoveContainer removes the container.
+// TODO(random-liu): Forcibly stop container if it's running.
 func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.RemoveContainerRequest) (_ *runtime.RemoveContainerResponse, retErr error) {
 	container, err := c.containerStore.Get(r.GetContainerId())
 	if err != nil {
@@ -52,7 +53,6 @@ func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.R
 		if retErr != nil {
 			// Reset removing if remove failed.
 			if err := resetContainerRemoving(container); err != nil {
-				// TODO(random-liu): Do not checkpoint `Removing` state.
 				glog.Errorf("failed to reset removing state for container %q: %v", id, err)
 			}
 		}
@@ -63,10 +63,12 @@ func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.R
 	// kubelet implementation, we'll never start a container once we decide to remove it,
 	// so we don't need the "Dead" state for now.
 
-	containerRootDir := getContainerRootDir(c.config.RootDir, id)
-	if err := system.EnsureRemoveAll(containerRootDir); err != nil {
-		return nil, fmt.Errorf("failed to remove container root directory %q: %v",
-			containerRootDir, err)
+	// Delete containerd container.
+	if err := container.Container.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
+		if !errdefs.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to delete containerd container %q: %v", id, err)
+		}
+		glog.V(5).Infof("Remove called for containerd container %q that does not exist", id, err)
 	}
 
 	// Delete container checkpoint.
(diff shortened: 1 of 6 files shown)


pkg/server/container_remove.go
```diff
@@ -63,23 +63,23 @@ func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.R
...
 	// so we don't need the "Dead" state for now.
 
-	containerRootDir := getContainerRootDir(c.config.RootDir, id)
-	if err := system.EnsureRemoveAll(containerRootDir); err != nil {
-		return nil, fmt.Errorf("failed to remove container root directory %q: %v",
-			containerRootDir, err)
+	// Delete containerd container.
+	if err := container.Container.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
+		if !errdefs.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to delete containerd container %q: %v", id, err)
```
mikebrow: what if we just pop out warnings and continue instead of returning early then return the error after we attempt to delete all the parts, best can do model?

pkg/server/container_remove.go
```diff
@@ -63,23 +63,23 @@ func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.R
...
+		}
+		glog.V(5).Infof("Remove called for containerd container %q that does not exist", id, err)
 	}
 
 	// Delete container checkpoint.
 	if err := container.Delete(); err != nil {
 		return nil, fmt.Errorf("failed to delete container checkpoint for %q: %v", id, err)
```
mikebrow: same

pkg/server/container_remove.go
```diff
@@ -63,23 +63,23 @@ func (c *criContainerdService) RemoveContainer(ctx context.Context, r *runtime.R
...
 
-	// Delete containerd container.
-	if err := container.Container.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
-		if !errdefs.IsNotFound(err) {
-			return nil, fmt.Errorf("failed to delete containerd container %q: %v", id, err)
-		}
-		glog.V(5).Infof("Remove called for containerd container %q that does not exist", id, err)
+	containerRootDir := getContainerRootDir(c.config.RootDir, id)
+	if err := system.EnsureRemoveAll(containerRootDir); err != nil {
+		return nil, fmt.Errorf("failed to remove container root directory %q: %v",
```
mikebrow: same

pkg/server/container_remove.go
Random-Liu: I think if we fail to delete container, we want to keep its status and root directory so that we could still retrieve information about this container.

pkg/server/container_remove.go
Random-Liu: `containerd.Delete` is atomic, we should only use it to delete container status.

If we fail to do that, we may not want to remove container root directory, because that removal is not atomic.

Random-Liu:
In short, we do assume order of those deletion. :)

pkg/server/container_remove.go
mikebrow: but we already changed it's status to removing... I'll ping you on slack..

Random-Liu:
@mikebrow Added TODO.

pkg/server/container_remove.go
mikebrow: ok per discussion, let's add a todo / issue regarding forcibly removing / killing on the remove request. Or getting guidance from the CRI sig-node team on remove expectations.

Issue #194, opened 2017-08-30
Properly configure containerd/cri-containerd cgroup.


Random-Liu:
Both containerd and cri-containerd added the flag to put themselves into a specific cgroup:
* containerd: https://github.com/containerd/containerd/pull/1443
* cri-containerd: https://github.com/kubernetes-incubator/cri-containerd/pull/184

When bootstrap the node, we should put both of them into a runtime cgroup, and pass the cgroup to kubelet via flag.

We should take care of this in our document or any start-up script in the future.

/cc @abhinandanpb

Random-Liu:
@abhinandanpb We should take care of this in our installer after the first version gets merged.

abhinandanpb:
yes I have a list of TODO's . This is on it.

Random-Liu:
Moved to next patch release, not affect feature parity.

PR #374, opened 2017-10-31
Add runtime cgroup and put containerd and cri-containerd into it.

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/194.

Put both `cri-containerd` and `containerd` into a runtime cgroup `/runtime` in ansible playbook. I'll do the same for `kube-up.sh`.


abhi:
LGTM

PR #375, opened 2017-10-31
support get image by truncindex

yanxuean:
refer to #352

We get image by localResolve(). 
The localResolve receive these arguments as below:
1. imageID.   e.g.   sha256:xxxxxx
2. repotag.    e.g.   busybox:1.26
3. repodigest.   e.g.  busybox@sha256:xxxxx

Now we need it to support  the truncindex of imageID.
Because it is imageID that we store into imageStore, so at first the  localResolve() must treat the argument as the truncindex of imageID .

@mikebrow @Random-Liu

### DIFF
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -28,7 +28,6 @@ import (
 	"github.com/containerd/cgroups"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/content"
-	"github.com/containerd/containerd/errdefs"
 	"github.com/docker/distribution/reference"
 	imagedigest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/identity"
@@ -205,34 +204,41 @@ func getRepoDigestAndTag(namedRef reference.Named, digest imagedigest.Digest, sc
 
 // localResolve resolves image reference locally and returns corresponding image metadata. It returns
 // nil without error if the reference doesn't exist.
-func (c *criContainerdService) localResolve(ctx context.Context, ref string) (*imagestore.Image, error) {
-	_, err := imagedigest.Parse(ref)
-	if err != nil {
-		// ref is not image id, try to resolve it locally.
-		normalized, err := util.NormalizeImageRef(ref)
-		if err != nil {
-			return nil, fmt.Errorf("invalid image reference %q: %v", ref, err)
+func (c *criContainerdService) localResolve(ctx context.Context, refOrID string) (*imagestore.Image, error) {
+	getImageID := func(refOrId string) string {
+		if _, err := imagedigest.Parse(refOrID); err == nil {
+			return refOrID
 		}
-		image, err := c.client.GetImage(ctx, normalized.String())
-		if err != nil {
-			if errdefs.IsNotFound(err) {
-				return nil, nil
+
+		return func(ref string) string {
+			// ref is not image id, try to resolve it locally.
+			normalized, err := util.NormalizeImageRef(ref)
+			if err != nil {
+				return ""
 			}
-			return nil, fmt.Errorf("failed to get image from containerd: %v", err)
-		}
-		desc, err := image.Config(ctx)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get image config descriptor: %v", err)
-		}
-		ref = desc.Digest.String()
+			image, err := c.client.GetImage(ctx, normalized.String())
+			if err != nil {
+				return ""
+			}
+			desc, err := image.Config(ctx)
+			if err != nil {
+				return ""
+			}
+			return desc.Digest.String()
+		}(refOrID)
 	}
-	imageID := ref
+
+	imageID := getImageID(refOrID)
+	if imageID == "" {
+		// Try to treat ref as imageID
+		imageID = refOrID
+	}
 	image, err := c.imageStore.Get(imageID)
 	if err != nil {
 		if err == store.ErrNotExist {
 			return nil, nil
 		}
-		return nil, fmt.Errorf("failed to get image %q metadata: %v", imageID, err)
+		return nil, fmt.Errorf("failed to get image %q : %v", imageID, err)
 	}
 	return &image, nil
 }
--- a/pkg/store/image/image.go
+++ b/pkg/store/image/image.go
@@ -20,7 +20,8 @@ import (
 	"sync"
 
 	"github.com/containerd/containerd"
-	"github.com/docker/docker/pkg/truncindex"
+	"github.com/docker/distribution/digestset"
+	godigest "github.com/opencontainers/go-digest"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 
 	"github.com/kubernetes-incubator/cri-containerd/pkg/store"
@@ -47,16 +48,16 @@ type Image struct {
 
 // Store stores all images.
 type Store struct {
-	lock    sync.RWMutex
-	images  map[string]Image
-	idIndex *truncindex.TruncIndex
+	lock      sync.RWMutex
+	images    map[string]Image
+	digestSet *digestset.Set
 }
 
 // NewStore creates an image store.
 func NewStore() *Store {
 	return &Store{
-		images:  make(map[string]Image),
-		idIndex: truncindex.NewTruncIndex([]string{}),
+		images:    make(map[string]Image),
+		digestSet: digestset.NewSet(),
 	}
 }
 
@@ -64,11 +65,11 @@ func NewStore() *Store {
 func (s *Store) Add(img Image) error {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	if _, err := s.idIndex.Get(img.ID); err != nil {
-		if err != truncindex.ErrNotExist {
+	if _, err := s.digestSet.Lookup(img.ID); err != nil {
+		if err != digestset.ErrDigestNotFound {
 			return err
 		}
-		if err := s.idIndex.Add(img.ID); err != nil {
+		if err := s.digestSet.Add(godigest.Digest(img.ID)); err != nil {
 			return err
 		}
 	}
@@ -91,14 +92,14 @@ func (s *Store) Add(img Image) error {
 func (s *Store) Get(id string) (Image, error) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
-	id, err := s.idIndex.Get(id)
+	digest, err := s.digestSet.Lookup(id)
 	if err != nil {
-		if err == truncindex.ErrNotExist {
+		if err == digestset.ErrDigestNotFound {
 			err = store.ErrNotExist
 		}
 		return Image{}, err
 	}
-	if i, ok := s.images[id]; ok {
+	if i, ok := s.images[digest.String()]; ok {
 		return i, nil
 	}
 	return Image{}, store.ErrNotExist
@@ -119,14 +120,14 @@ func (s *Store) List() []Image {
 func (s *Store) Delete(id string) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	id, err := s.idIndex.Get(id)
+	digest, err := s.digestSet.Lookup(id)
 	if err != nil {
 		// Note: The idIndex.Delete and delete doesn't handle truncated index.
 		// So we need to return if there are error.
 		return
 	}
-	s.idIndex.Delete(id) // nolint: errcheck
-	delete(s.images, id)
+	s.digestSet.Remove(digest) // nolint: errcheck
+	delete(s.images, digest.String())
 }
 
 // mergeStringSlices merges 2 string slices into one and remove duplicated elements.
--- a/pkg/store/image/image_test.go
+++ b/pkg/store/image/image_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package image
 
 import (
+	"strings"
 	"testing"
 
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -26,33 +27,33 @@ import (
 )
 
 func TestImageStore(t *testing.T) {
-	images := map[string]Image{
-		"1": {
-			ID:          "1",
+	images := []Image{
+		{
+			ID:          "sha256:1123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
 			ChainID:     "test-chain-id-1",
 			RepoTags:    []string{"tag-1"},
 			RepoDigests: []string{"digest-1"},
 			Size:        10,
 			Config:      &imagespec.ImageConfig{},
 		},
-		"2abcd": {
-			ID:          "2abcd",
+		{
+			ID:          "sha256:2123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
 			ChainID:     "test-chain-id-2abcd",
 			RepoTags:    []string{"tag-2abcd"},
 			RepoDigests: []string{"digest-2abcd"},
 			Size:        20,
 			Config:      &imagespec.ImageConfig{},
 		},
-		"4a333": {
-			ID:          "4a333",
+		{
+			ID:          "sha256:3123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
 			RepoTags:    []string{"tag-4a333"},
 			RepoDigests: []string{"digest-4a333"},
 			ChainID:     "test-chain-id-4a333",
 			Size:        30,
 			Config:      &imagespec.ImageConfig{},
 		},
-		"4abcd": {
-			ID:          "4abcd",
+		{
+			ID:          "sha256:4123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
 			RepoTags:    []string{"tag-4abcd"},
 			RepoDigests: []string{"digest-4abcd"},
 			ChainID:     "test-chain-id-4abcd",
@@ -61,6 +62,7 @@ func TestImageStore(t *testing.T) {
 		},
 	}
 	assert := assertlib.New(t)
+	genTruncIndex := func(normalName string) string { return normalName[:(len(normalName)+1)/2] }
 
 	s := NewStore()
 
@@ -71,11 +73,26 @@ func TestImageStore(t *testing.T) {
 	}
 
 	t.Logf("should be able to get image")
-	genTruncIndex := func(normalName string) string { return normalName[:(len(normalName)+1)/2] }
-	for id, img := range images {
-		got, err := s.Get(genTruncIndex(id))
-		assert.NoError(err)
-		assert.Equal(img, got)
+	for _, v := range images {
+		truncID := genTruncIndex(v.ID)
+		got, err := s.Get(truncID)
+		assert.NoError(err, "truncID:%s, fullID:%s", truncID, v.ID)
+		assert.Equal(v, got)
+	}
+
+	t.Logf("should be able to get image by truncated imageId without algorithm")
+	for _, v := range images {
+		truncID := genTruncIndex(v.ID[strings.Index(v.ID, ":")+1:])
+		got, err := s.Get(truncID)
+		assert.NoError(err, "truncID:%s, fullID:%s", truncID, v.ID)
+		assert.Equal(v, got)
+	}
+
+	t.Logf("should not be able to get image by ambiguous prefix")
+	ambiguousPrefixs := []string{"sha256", "sha256:"}
+	for _, v := range ambiguousPrefixs {
+		_, err := s.Get(v)
+		assert.NotEqual(nil, err)
 	}
 
 	t.Logf("should be able to list images")
@@ -83,8 +100,8 @@ func TestImageStore(t *testing.T) {
 	assert.Len(imgs, len(images))
 
 	imageNum := len(images)
-	for testID, v := range images {
-		truncID := genTruncIndex(testID)
+	for _, v := range images {
+		truncID := genTruncIndex(v.ID)
 		oldRepoTag := v.RepoTags[0]
 		oldRepoDigest := v.RepoDigests[0]
 		newRepoTag := oldRepoTag + "new"
--- a/pkg/server/container_status_test.go
+++ b/pkg/server/container_status_test.go
@@ -30,6 +30,7 @@ import (
 
 func getContainerStatusTestData() (*containerstore.Metadata, *containerstore.Status,
 	*imagestore.Image, *runtime.ContainerStatus) {
+	imageID := "sha256:1123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
 	testID := "test-id"
 	config := &runtime.ContainerConfig{
 		Metadata: &runtime.ContainerMetadata{
@@ -53,7 +54,7 @@ func getContainerStatusTestData() (*containerstore.Metadata, *containerstore.Sta
 		Name:      "test-long-name",
 		SandboxID: "test-sandbox-id",
 		Config:    config,
-		ImageRef:  "test-image-id",
+		ImageRef:  imageID,
 		LogPath:   "test-log-path",
 	}
 	status := &containerstore.Status{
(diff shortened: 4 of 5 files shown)


yanxuean:
The travis is on strike.

pkg/server/helpers.go
```diff
@@ -206,8 +206,12 @@ func getRepoDigestAndTag(namedRef reference.Named, digest imagedigest.Digest, sc
 // localResolve resolves image reference locally and returns corresponding image metadata. It returns
 // nil without error if the reference doesn't exist.
 func (c *criContainerdService) localResolve(ctx context.Context, ref string) (*imagestore.Image, error) {
-	_, err := imagedigest.Parse(ref)
```
mikebrow: not really convinced this is a good idea...

From my skim, sha256: should be a valid imagedigest.Parse(ref), and sha should not. Having a hard time thinking sha should be a valid imageStore.Get lookup. 

Thoughts?

mikebrow: That said, if ```c.client.GetImage(ctx, normalized.String())``` below comes back with is not found... maybe we could do an attempt from c.imageStore.Get(normalized.String()) then check for error and return nil nil if is not found there either?

yanxuean: Now we store sha256:xxx into imageStore .  so "sha" is a valid truncindex when patricia tree have only  one entry.  It is a bit of confused, but it is responsibility of the user. The user should give the truncID unambiguous.
@Random-Liu  What do you think?

mikebrow: Yes, it's a valid truncindex reference, so long as there is only one image in the store, it is also an invalid reference. Won't we have two images in the store by default?  The pause image and one for it's container? Not sure if it's an interesting use case to support looking up sha to get the only image in the store.

yanxuean: Let's put the test aside first. The pr mainly is for supporting truncindex in image.
I think that the image can not support truncindex if we have not the pr.
Do you give me a suggestion about supporting truncindex in image? 
@Random-Liu  I also need your suggestion .

Random-Liu:
Let me think about this. I need to take a look at how docker deals with this.

yanxuean:
docker remove :
1.  imageId with prefix sha256:                   yes      
2.  imageID without  prefix sha256:            yes
3.  trunc  imageID with prefix sha256:        yes
4.  trunc  imageID without prefix sha256:   yes
5.  digest with prefix sha256:                  no
6.  digest without prefix sha256:            no
7.  trunc digest with prefix sha256:        no
8.  trunc digest without prefix sha256:   no

test record:
```
Id: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364  ,
RepoTags:[docker.io/library/busybox:latest],
RepoDigests:[docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce],



root@ubuntu:/home/cloud/test# docker rmi sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Untagged: busybox:latest
Untagged: busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Deleted: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Deleted: sha256:0271b8eebde3fa9a6126b1f2335e170f902731ab4942f9f1914e77016540c7bb


root@ubuntu:/home/cloud/test# docker rmi 9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Untagged: busybox:latest
Untagged: busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Deleted: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Deleted: sha256:0271b8eebde3fa9a6126b1f2335e170f902731ab4942f9f1914e77016540c7bb

root@ubuntu:/home/cloud/test# docker rmi sha256:9d7e
Untagged: busybox:latest
Untagged: busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Deleted: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Deleted: sha256:0271b8eebde3fa9a6126b1f2335e170f902731ab4942f9f1914e77016540c7bb

root@ubuntu:/home/cloud/test# docker rmi 9d7e
Untagged: busybox:latest
Untagged: busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Deleted: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Deleted: sha256:0271b8eebde3fa9a6126b1f2335e170f902731ab4942f9f1914e77016540c7bb


root@ubuntu:/home/cloud/test# docker rmi sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Error response from daemon: No such image: sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce

root@ubuntu:/home/cloud/test# docker rmi 06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Error response from daemon: No such image: sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce


root@ubuntu:/home/cloud/test# docker rmi sha256:06d4
Error response from daemon: No such image: sha256:06d4


root@ubuntu:/home/cloud/test# docker rmi 06d4
Error response from daemon: No such image: 06d4:latest
```

yanxuean:
@mikebrow @Random-Liu 
I think cri-containerd should support 2 and 4.   
The 1 and 3 are not necessary.

docker remove :
1.  imageId with prefix sha256:                   yes      
2.  imageID without  prefix sha256:            yes
3.  trunc  imageID with prefix sha256:        yes
4.  trunc  imageID without prefix sha256:   yes
5.  digest with prefix sha256:                  no
6.  digest without prefix sha256:            no
7.  trunc digest with prefix sha256:        no
8.  trunc digest without prefix sha256:   no

yanxuean:
crictl test record:
```
root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl images
DEBU[0000] ListImagesRequest: &ListImagesRequest{Filter:&ImageFilter{Image:&ImageSpec{Image:,},},} 
DEBU[0000] ListImagesResponse: &ListImagesResponse{Images:[],} 

root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl pull busybox
DEBU[0000] PullImageRequest: &PullImageRequest{Image:&ImageSpec{Image:busybox,},Auth:nil,SandboxConfig:nil,} 
DEBU[0028] PullImageResponse: &PullImageResponse{ImageRef:sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364,} 
Image is update to date for sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364


  
root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl images --digests
DEBU[0000] ListImagesRequest: &ListImagesRequest{Filter:&ImageFilter{Image:&ImageSpec{Image:,},},} 
DEBU[0000] ListImagesResponse: &ListImagesResponse{Images:[&Image{Id:sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364,RepoTags:[docker.io/library/busybox:latest],RepoDigests:[docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce],Size_:720019,Uid:nil,Username:,}],} 
IMAGE                       TAG                 DIGEST                                                                    IMAGE ID            SIZE
docker.io/library/busybox   latest              sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce   9d7e6df8e5ca7       720kB

root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl inspecti 9d
DEBU[0000] ImageStatusRequest: &ImageStatusRequest{Image:&ImageSpec{Image:9d,},} 
DEBU[0000] ImageStatusResponse: &ImageStatusResponse{Image:&Image{Id:sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364,RepoTags:[docker.io/library/busybox:latest],RepoDigests:[docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce],Size_:720019,Uid:nil,Username:,},} 
ID: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Tag: docker.io/library/busybox:latest
Digest: docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Size: 720kB


root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl inspecti sha256:9d
DEBU[0000] ImageStatusRequest: &ImageStatusRequest{Image:&ImageSpec{Image:sha256:9d,},} 
DEBU[0000] ImageStatusResponse: &ImageStatusResponse{Image:&Image{Id:sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364,RepoTags:[docker.io/library/busybox:latest],RepoDigests:[docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce],Size_:720019,Uid:nil,Username:,},} 
ID: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Tag: docker.io/library/busybox:latest
Digest: docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Size: 720kB


root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl inspecti sha256:
DEBU[0000] ImageStatusRequest: &ImageStatusRequest{Image:&ImageSpec{Image:sha256:,},} 
E1103 22:11:10.071061    9499 instrumented_service.go:265] ImageStatus for "sha256:" failed, error: can not resolve "sha256:" locally: invalid image reference "sha256:": invalid reference format
DEBU[0000] ImageStatusResponse: nil                     
FATA[0000] image status request failed: rpc error: code = Unknown desc = can not resolve "sha256:" locally: invalid image reference "sha256:": invalid reference format 
root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl inspecti sha256:9
DEBU[0000] ImageStatusRequest: &ImageStatusRequest{Image:&ImageSpec{Image:sha256:9,},} 
DEBU[0000] ImageStatusResponse: &ImageStatusResponse{Image:&Image{Id:sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364,RepoTags:[docker.io/library/busybox:latest],RepoDigests:[docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce],Size_:720019,Uid:nil,Username:,},} 
ID: sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364
Tag: docker.io/library/busybox:latest
Digest: docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce
Size: 720kB


root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl rmi sha256:9
DEBU[0000] ImageStatusRequest: &ImageStatusRequest{Image:&ImageSpec{Image:sha256:9,},} 
DEBU[0000] ImageStatusResponse: &ImageStatusResponse{Image:&Image{Id:sha256:9d7e6df8e5ca703a7629a6d47f9d6f90c3c4419203913d7b761248469bb0b364,RepoTags:[docker.io/library/busybox:latest],RepoDigests:[docker.io/library/busybox@sha256:06d4020f14549289bf9552dccf99c79c61ef8e664aec2edca614f8ae4af7a4ce],Size_:720019,Uid:nil,Username:,},} 
DEBU[0000] RemoveImageRequest: &RemoveImageRequest{Image:&ImageSpec{Image:sha256:9,},} 
DEBU[0000] RemoveImageResponse: &RemoveImageResponse{}  
Deleted: docker.io/library/busybox:latest


root@yyv:/home/yy/go/src/github.com/kubernetes-incubator/cri-containerd# crictl images
DEBU[0000] ListImagesRequest: &ListImagesRequest{Filter:&ImageFilter{Image:&ImageSpec{Image:,},},} 
DEBU[0000] ListImagesResponse: &ListImagesResponse{Images:[],} 

```

mikebrow:
@yanxuean did you try `docker rmi docker.io/library/busybox@sha256:06d4` ?

pkg/server/helpers.go
mikebrow: I don't think we need to add imagestore.Get above the parse test.. as a failure on parse calls get anyway. 

See below call to imageStore.Get on original lines 229-237.

pkg/store/image/image.go
```diff
@@ -45,6 +46,14 @@ type Image struct {
 	Image containerd.Image
 }
 
+func trimAlgorithmFromImageID(imageID string) string {
```
mikebrow: Curious, how does trimming the prefix help?

Note: there is more than one valid prefix here to the left of the `:`.. so trimming it possibly looses context.

yanxuean: Yes,it possibly looses context, but it only try. If try fail, it send the whole context to Parse.

yanxuean: "Curious, how does trimming the prefix help?"
Becanse we need to support "sha256:xxx" and "xxx" as docker. we store "xxx" in truncindex.
So I must get "xxx" by trimming the prifix

pkg/server/helpers.go
yanxuean: done

pkg/store/image/image.go
mikebrow: I see, so the decision/plan/design is to only store and retrieve the `xxx` portion of the image ref in the truncindex. That makes sense. Then we don't need to store it in both sha256:xxx and xxx.. trees. Sure ok that works. Just a small window where there could be two with different sha values, 256/512 and the same few id values. ok..

yanxuean: yes, there is a small window.
This function is only for crictl. I think it is acceptable.

pkg/server/helpers.go
```diff
@@ -205,34 +204,40 @@ func getRepoDigestAndTag(namedRef reference.Named, digest imagedigest.Digest, sc
...
+			if err != nil {
+				return "", fmt.Errorf("failed to get image from containerd: %v", err)
+			}
+			desc, err := image.Config(ctx)
+			if err != nil {
+				return "", fmt.Errorf("failed to get image config descriptor: %v", err)
+			}
+			return desc.Digest.String(), nil
+		}(ref)
+		if err == nil {
```
mikebrow: Did you want:
```
if err != nil {
   return runtimeImageID, err 
} 
ref = runtimeImageID
```

yanxuean: No.  If it is fail, We need to try it as truncindex.

mikebrow: Then why generate and return all those errors above in the func if you are not going to use them? I presumed from reading func that you only wanted to continue in the nil case.

mikebrow: Then why generate and return all those errors above in func if you are not going to use them? I presumed from reading func that you only wanted to continue in the error == nil case and intended to return the actual errors generated.

yanxuean: Will do

yanxuean: done

yanxuean:
@mikebrow @Random-Liu   PTAL

yanxuean:
@mikebrow  Thank you very much. Thanks for multiple reviews.
I learn  a lot from this pr.

yanxuean:
@random-liu    PTAL

Random-Liu:
@yanxuean Sorry. I didn't quite follow the discussion above.

My question is that, it seems that `crictl` cut off the algorithm from the image id itself, right?
If that is the case, crictl should add the algorithm back.

I think we should still store complete image id in image store, and assume that all image id (truncated or not) passed in should have algorithm prefix.

pkg/store/image/image.go
Random-Liu: >  Just a small window where there could be two with different sha values, 256/512 and the same few id values. And if that happens they should just get a false dup and have to use the full id.

Yeah. I agree. We should not remove the algorithm.

See my new comment.

mikebrow:
So one of the important parts of this is to match prior behavior, since the patricia tree/trunc index stuff already exists with the cri docker shim...

Random-Liu:
Let me take a look at docker implementation. :)

Random-Liu:
Docker is using [`ParseAnyReference`](https://github.com/docker/distribution/blob/545102ea07aa9796f189d82f606b7c27d7aa3ed3/reference/normalize.go#L144).

We might also be able to use it or add something similar.

mikebrow:
Are we only storing the id in the patricia tree / index?  I can see having a hard rule that we must store the full id, and if it's only being stored in the tree then I suppose it makes sense to go ahead and put it in the tree though there will be a lot of s->h->a->2->5->6->:-> traversing.

yanxuean:
I will study docker implementation late.

Random-Liu:
> If it's only being stored in the tree then I suppose it makes sense to go ahead and put it in the tree though there will be a lot of s->h->a->2->5->6->:-> traversing.

Yeah, but comparing with other overhead, this should be trivial. I feel like the truncate index logic is less important compared with the core metadata store logic. I don't think we should make the core logic more complex just for truncate indexing. :)

mikebrow:
> Yeah, but comparing with other overhead, this should be trivial. I feel like the truncate index logic is less important compared to the core metadata store logic. I don't think we should make the core logic more complex and error prone just for truncate indexing. :)

Yeah I agree it's pretty trivial. Just pointing out the only minor negative.

yanxuean:
@mikebrow @Random-Liu 
I studied the docker implementation. I find it is not simple as we think before.:)
The docker implementation is powerful.

The key point is not "ParseAnyReference" but "digestEntry".
1. The data structure as below:
```go
// docker/image/store.go
type store struct {
	sync.RWMutex
	ls        LayerGetReleaser
	images    map[ID]*imageMeta
	fs        StoreBackend
	digestSet *digestset.Set          <------******
	os        string
}

//docker/vendor/github.com/docker/distribution/digestset/set.go
type Set struct {
	mutex   sync.RWMutex
	entries digestEntries           <-----*******
}
type digestEntry struct {             <-----******key point.   The algorithm and val are stored independly
	alg    digest.Algorithm
	val    string
	digest digest.Digest
}

type digestEntries []*digestEntry
```
2. The powerful "Set.Lookup"
```
func (dst *Set) Lookup(d string)
```
The Lookup of set can handle the argument as below:
*  shaxxx:abcd1234
*  abcd1234

3. The control logic of external layer:   GetImageIDAndOS(refOrID string)
When we input truncated imageID (e.g. "abcd1234") without Algorithm,  the ParseAnyReference treat it as imageName.  The ParseAnyReference only can handle full ID and digest self.
Our truncated imageID("abcd1234") is wrap in reference.Named object. But we cann't find it in daemon.referenceStore.  So  it will be pass to daemon.stores[os].imageStore.Search.
```
func (daemon *Daemon) GetImageIDAndOS(refOrID string) (image.ID, string, error) {
	ref, err := reference.ParseAnyReference(refOrID)   
	if err != nil {
		return "", "", validationError{err}
	}
	namedRef, ok := ref.(reference.Named)
	if !ok {
		digested, ok := ref.(reference.Digested)
		if !ok {
			return "", "", errImageDoesNotExist{ref}
		}
		....
		return "", "", errImageDoesNotExist{ref}
	}

	if digest, err := daemon.referenceStore.Get(namedRef); err == nil {
		// Search the image stores to get the operating system, defaulting to host OS.
		imageOS := runtime.GOOS
		id := image.IDFromDigest(digest)
		for os := range daemon.stores {
			if img, err := daemon.stores[os].imageStore.Get(id); err == nil {
				imageOS = img.OperatingSystem()
				break
			}
		}
		return id, imageOS, nil
	}

	// deprecated: repo:shortid https://github.com/docker/docker/pull/799
	if tagged, ok := namedRef.(reference.Tagged); ok {
		.......
	}

       <-----*******The key point is here. The docker search imageStore by "refOrID" ! 
	// Search based on ID
	for os := range daemon.stores {
		if id, err := daemon.stores[os].imageStore.Search(refOrID); err == nil {
			return id, os, nil
		}
	}

	return "", "", errImageDoesNotExist{ref}
}
```

mikebrow:
@yanxuean right.. yeah I skimmed it the other day and saw that as well. Store both but independently to gain both the slight perf enhancement and use of the original stripped alg. prefix is what I saw as well. Not so worried about memory use for ids so much as performance, thus my point above to @Random-Liu that we could store both. But he's also right in so far as it would be nice to keep it simple.

Random-Liu:
@yanxuean Thanks a lot for figuring this out!

yanxuean:
@mikebrow @Random-Liu 
For image the truncindex package have some problem to solve. 
Assume we store fullID with shaxxx in truncindex store.
1. When user input truncated imageID without alg.  
If the alg of image actually is sha384, we will be wrong with adding sha256 in crictl.
In general the user don't input the alg.  I also don't want input it.:)
2. When crictl get the input. we cann't distinguish imagename and truncated imageID. 
For example the input is "deedbeef", I match regular expression [0-9a-z] for digest.:)

I think we should reuse the digestset.Set.

yanxuean:
test record
```
cri-containerd# crictl images
IMAGE                            TAG                 IMAGE ID            SIZE
docker.io/library/busybox        1.25                e02e811dd08fd       670kB
docker.io/library/busybox        1.26                c30178c5239f2       701kB
docker.io/library/busybox        latest              9d7e6df8e5ca7       720kB
gcr.io/google_containers/pause   3.0                 320aedbfaad3d       312kB


cri-containerd# crictl inspecti e02
ID: sha256:e02e811dd08fd49e7f6032625495118e63f597eb150403d02e3238af1df240ba
Tag: docker.io/library/busybox:1.25
Digest: docker.io/library/busybox@sha256:29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912
Size: 670kB


cri-containerd# crictl inspecti sha256:e02
ID: sha256:e02e811dd08fd49e7f6032625495118e63f597eb150403d02e3238af1df240ba
Tag: docker.io/library/busybox:1.25
Digest: docker.io/library/busybox@sha256:29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912
Size: 670kB


cri-containerd# crictl inspecti sha
FATA[0000] no such image present 
                       
                          
cri-containerd# crictl rmi e
Deleted: docker.io/library/busybox:1.25


```

yanxuean:
@mikebrow @Random-Liu  PTAL

yanxuean:
@mikebrow  @random-liu    ptal

pkg/store/image/image_test.go
```diff
@@ -71,20 +73,39 @@ func TestImageStore(t *testing.T) {
...
+		got, err := s.Get(truncID)
+		assert.NoError(err, "truncID:%s, fullID:%s", truncID, v.ID)
+		assert.Equal(v, got)
+	}
+
+	t.Logf("should not be able to get image by ambiguous prefix")
+	ambiguousPrefix := "sha256"
+	for i := range images {
+		if i == 0 {
+			// skip the first item
```
mikebrow: what's up here skipping the first item?

yanxuean: Hmm. It is unnecessary, Maybe I is short-circuted on that time.:)
On that time I think the first item is only one image and it cann't be conflict with the other.
I will modify it.

yanxuean: done

PR #376, opened 2017-11-01
Update cri-containerd version to 1.0.0-alpha.1 in ansible.

Random-Liu:

### DIFF
--- a/contrib/ansible/tasks/binaries.yaml
+++ b/contrib/ansible/tasks/binaries.yaml
@@ -1,8 +1,7 @@
 ---
 - name: "Get Containerd and CRI-Containerd"
   unarchive:
-    # TODO change to cri-containerd-release before official release.
-    src: "https://storage.googleapis.com/cri-containerd-staging/cri-containerd-{{ cri_containerd_release_version }}.tar.gz"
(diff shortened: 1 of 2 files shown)


miaoyq:
@Random-Liu  I think we should add a `cri_containerd_release_version` for centOS. 
I deployed the `cri-container` on my centOS via ansible,  but it didn't work because of not supporting apparmor.

Random-Liu:
The test failure is caused by a containerd issue. 

Random-Liu:
> I deployed the cri-container on my centOS via ansible, but it didn't work because of not supporting apparmor.

Hm? What is the error? I thought it should work on node without apparmor.

miaoyq:
@Random-Liu Error while loading shared libraries: libapparmor.so.1
```
Oct 16 15:02:34 localhost systemd: cri-containerd.service: main process exited, code=exited, status=127/n/a
Oct 16 15:02:34 localhost cri-containerd: /usr/local/bin/cri-containerd: error while loading shared libraries: libapparmor.so.1: cannot open shared object file: No such file or directory
Oct 16 15:02:34 localhost systemd: Unit cri-containerd.service entered failed state.
Oct 16 15:02:34 localhost systemd: cri-containerd.service failed.
```

PR #380, opened 2017-11-01
Do not call `Usage` inside `Walk`.

Random-Liu:
See https://github.com/containerd/containerd/pull/1706#issuecomment-340992262.

Call `Usage` during `Walk` will cause a deadlock in current version of containerd.

### DIFF
--- a/pkg/server/snapshots.go
+++ b/pkg/server/snapshots.go
@@ -18,6 +18,7 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/containerd/containerd/errdefs"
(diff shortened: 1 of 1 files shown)


pkg/server/snapshots.go
```diff
@@ -84,20 +95,16 @@ func (s *snapshotsSyncer) sync() error {
 			Kind:      info.Kind,
 			Timestamp: time.Now().UnixNano(),
 		}
-		usage, err := s.snapshotter.Usage(ctx, info.Name)
+		usage, err := s.snapshotter.Usage(context.Background(), info.Name)
 		if err != nil {
-			if errdefs.IsNotFound(err) {
-				return nil
+			if !errdefs.IsNotFound(err) {
+				glog.Errorf("Failed to get usage for snapshot %q: %v", info.Name, err)
```
mikebrow: good!

pkg/server/snapshots.go
```diff
@@ -68,14 +68,25 @@ func (s *snapshotsSyncer) start() {
...
+	var snapshots []snapshot.Info
+	// Do not call `Usage` directly in collect function, because
+	// `Usage` takes time, we don't want `Walk` to hold read lock
+	// of snapshot metadata store for too long time.
+	// TODO(random-liu): Set timeout for the following 2 contexts.
+	if err := s.snapshotter.Walk(context.Background(), func(ctx context.Context, info snapshot.Info) error {
+		snapshots = append(snapshots, info)
+		return nil
+	}); err != nil {
+		return err
```
mikebrow: log here and/or maybe add context to the error?

Random-Liu: Will add context to the error. Thanks!

PR #387, opened 2017-11-02
Fix our node e2e test.

Random-Liu:
`test/build.sh` generates wrong `${ROOT}`.


Random-Liu:
@mikebrow Thanks!

PR #392, opened 2017-11-05
Add crictl config.

Random-Liu:
Add a default `crictl` confg. @yanxuean 

Also remove a redundant log.

### DIFF
--- a/pkg/server/container_attach.go
+++ b/pkg/server/container_attach.go
@@ -28,14 +28,7 @@ import (
 )
 
 // Attach prepares a streaming endpoint to attach to a running container, and returns the address.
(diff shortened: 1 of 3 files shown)


pkg/server/container_attach.go
```diff
@@ -29,13 +29,6 @@ import (
 
 // Attach prepares a streaming endpoint to attach to a running container, and returns the address.
 func (c *criContainerdService) Attach(ctx context.Context, r *runtime.AttachRequest) (retRes *runtime.AttachResponse, retErr error) {
```
yanxuean: nit:  The name of argument may be removed.  e.g. retRes, retErr

Random-Liu: Will do

Random-Liu: Done.

PR #393, opened 2017-11-06
Adding kube pod and container labels to containerd

abhi:
Currently we have the pod and container labels part of
containerd metadata extensions. However for third party users
like cadvisor that depend on standard kube labels will need
to be aware of the way metadata is stored in containerd to
fetch the labels.

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -143,12 +143,14 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 		specOpts = append(specOpts, seccompSpecOpts)
 	}
 
+	sandboxLabels := buildLabels(config.Labels, containerKindSandbox)
+
 	opts := []containerd.NewContainerOpts{
 		containerd.WithSnapshotter(c.config.ContainerdConfig.Snapshotter),
 		customopts.WithImageUnpack(image.Image),
 		containerd.WithNewSnapshot(id, image.Image),
 		containerd.WithSpec(spec, specOpts...),
-		containerd.WithContainerLabels(map[string]string{containerKindLabel: containerKindSandbox}),
+		containerd.WithContainerLabels(sandboxLabels),
 		containerd.WithContainerExtension(sandboxMetadataExtension, &sandbox.Metadata),
 		containerd.WithRuntime(
 			c.config.ContainerdConfig.Runtime,
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -223,6 +223,7 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	if seccompSpecOpts != nil {
 		specOpts = append(specOpts, seccompSpecOpts)
 	}
+	containerLabels := buildLabels(config.Labels, containerKindContainer)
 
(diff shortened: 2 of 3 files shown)


Random-Liu:
@abhi 
1) Test is failing.
2) What label do we need in cadvisor? Do we control the label key name or does cadvisor have any presumption on it? If it's the former, I prefer we add the labels we need ourselves, instead of passing through labels from kubelet.

abhi:
@Random-Liu yes will correct the test cases.
I am not adding any labels from kubelet. I am only putting the labels passed by the kubelet to CRI into the labels section of containerd. These are the pod_name , pod_namespace and pod_uid labels. Currently we put in all these as a part of metadata and store it using extensions in containerd which is totally fine. But in cadvisor I need to know the type of data stored as metadata extensions to unmarshallany. For this I need to import cri-containerd code into cadvisor. For the purpose of well known labels I feel we could keep labels as easily readable labels by third party user than stored as metadata extensions. cadvisor relies on these labels to identify pod. WDYT ?

Random-Liu:
@abhi 
I mean we can use labels created by kubelet, or we could also apply the required labels ourselves.
I feel like we could create our own label `io.cri-containerd.sandbox.name` or something.

Random-Liu:
@abhi Wait. Does cadvisor in kubelet get information based on the labels? If so, then we have to pass through.

Is that the case?

abhi:
@Random-Liu  Yep something like this - 
https://github.com/kubernetes/kubernetes/blob/440884be9ae54a500139697890889e89dabe25cf/pkg/kubelet/stats/cadvisor_stats_provider.go#L183

pkg/server/sandbox_run.go
```diff
@@ -143,12 +143,18 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 		specOpts = append(specOpts, seccompSpecOpts)
 	}
 
+	podLabels := config.Labels
```
Random-Liu: nit: `sandboxLabels`?

abhi: done

integration/container_stats_test.go
```diff
@@ -306,7 +306,6 @@ func testStats(t *testing.T,
 	require.NotEmpty(t, s.GetAttributes().GetId())
 	require.NotEmpty(t, s.GetAttributes().GetMetadata())
 	require.NotEmpty(t, s.GetAttributes().GetAnnotations())
-	require.Equal(t, s.GetAttributes().GetLabels(), config.Labels)
```
Random-Liu: Why is this changed?

Random-Liu: @abhi I don't quite get it, sorry.

abhi: The container config will have only the added labels. Not the labels added by Cri-containerd. for eg config.Labels will be [key:value] , but the metadata store will have the [key:value cri-containerd.kind:"sandbox"]. I was hoping to take care of these things in the stats test cleanup

abhi: i was thinking we had to store the labels we set for containerd as well. I will make the correction in the labels code above.

Random-Liu:
LGTM

Issue #354, opened 2017-10-23
kubectl exec doesnt have a clean prompt indentation


abhi:
```
root@abhi-k8s-ubuntu-1:~# kubectl exec -it curl-image2-5d76469875-p498b sh
/ #
    / #
        / #
            / #
                / #
                    / #
                        / #
```

crosbymichael:
This would happen if the console settings like OPOST are not synced up with what is being created inthe container and the `kubectl`

PR #394, opened 2017-11-06
Various fixes for container streaming.

Random-Liu:
I've tried cluster e2e in my own cluster, and we could pass most of the test.

There are around 5 test failure I got, this PR fixes 2 of them.

This PR:
1) Add `TERM=xterm` env when do container exec. Sometimes the original container may not have `tty`, thus no `TERM=xterm`. But if we exec with tty, we should add this.
2) Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/354. Moved resize handling after task start, so that we won't miss the first resize request. https://github.com/kubernetes/kubernetes/pull/47991 @abhi Could you verify whether you still have the exec problem after this PR is merged?
3) Fix a case that `Attach` may hang. Previously after stdout/stdin is closed, we close the reader side of stdin. However, that doesn't stop `io.Copy` if the writer side doesn't write any new thing. In this PR, we close the writer side instead.
4) Fix `StdinOnce`. Copy [docker's behavior](https://github.com/moby/moby/blob/v17.05.0-ce/container/stream/attach.go#L104-L114) for `StdinOnce`. 2 kubectl tests are relying on this behavior.

### DIFF
--- a/pkg/server/container_execsync.go
+++ b/pkg/server/container_execsync.go
@@ -26,6 +26,7 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/golang/glog"
+	"github.com/opencontainers/runtime-tools/generate"
 	"golang.org/x/net/context"
 	"golang.org/x/sys/unix"
 	"k8s.io/client-go/tools/remotecommand"
@@ -96,6 +97,11 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 	if err != nil {
 		return nil, fmt.Errorf("failed to load task: %v", err)
 	}
+	if opts.tty {
+		g := generate.NewFromSpec(spec)
+		g.AddProcessEnv("TERM", "xterm")
+		spec = g.Spec()
+	}
 	pspec := spec.Process
 	pspec.Args = opts.cmd
 	pspec.Terminal = opts.tty
@@ -125,12 +131,6 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 		}
 	}()
 
-	handleResizing(opts.resize, func(size remotecommand.TerminalSize) {
-		if err := process.Resize(ctx, uint32(size.Width), uint32(size.Height)); err != nil {
-			glog.Errorf("Failed to resize process %q console for container %q: %v", execID, id, err)
-		}
-	})
-
 	exitCh, err := process.Wait(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to wait for process %q: %v", execID, err)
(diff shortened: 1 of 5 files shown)


pkg/server/container_execsync.go
```diff
@@ -96,6 +97,11 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 	if err != nil {
 		return nil, fmt.Errorf("failed to load task: %v", err)
 	}
+	if opts.tty {
```
Random-Liu: For 1).

pkg/server/container_execsync.go
```diff
@@ -139,6 +139,12 @@ func (c *criContainerdService) execInContainer(ctx context.Context, id string, o
 		return nil, fmt.Errorf("failed to start exec %q: %v", execID, err)
 	}
 
+	handleResizing(opts.resize, func(size remotecommand.TerminalSize) {
```
Random-Liu: For 2)

pkg/server/io/io.go
```diff
@@ -263,56 +275,66 @@ func (c *ContainerIO) Attach(stdin io.Reader, stdout, stderr io.WriteCloser) err
...
+	var stdinRC io.ReadCloser
+	if c.stdinPath != "" && opts.Stdin != nil {
 		f, err := fifo.OpenFifo(c.closer.ctx, c.stdinPath, syscall.O_WRONLY|syscall.O_NONBLOCK, 0700)
 		if err != nil {
 			return err
 		}
+		// Create a wrapper of stdin which could be closed. Note that the
+		// wrapper doesn't close the actual stdin, it only stops io.Copy.
+		// The actual stdin will be closed by stream server.
+		stdinRC = cioutil.NewWrapReadCloser(opts.Stdin)
```
Random-Liu: For 3).

pkg/server/io/io.go
```diff
@@ -263,56 +275,66 @@ func (c *ContainerIO) Attach(stdin io.Reader, stdout, stderr io.WriteCloser) err
...
 			}
 			w.Close()
 			glog.V(2).Infof("Attach stream %q closed", stdinKey)
-			// No matter what, when stdin is closed (io.Copy unblock), close stdout and stderr
-			if stdout != nil {
-				c.stdout.Remove(stdoutKey)
-			}
-			if stderr != nil {
-				c.stderr.Remove(stderrKey)
+			if opts.StdinOnce && !opts.Tty {
```
Random-Liu: For 4).

mikebrow: is stdinonce and not tty a valid scenario?

pkg/server/io/io.go
```diff
@@ -263,56 +275,66 @@ func (c *ContainerIO) Attach(stdin io.Reader, stdout, stderr io.WriteCloser) err
...
+				// Close container stdin if StdinOnce is specified, and tty is false.
+				if err := opts.CloseStdin(); err != nil {
+					glog.Errorf("Failed to close stdin for container %q: %v", c.id, err)
+				}
+			} else {
+				// No matter what, when stdin is closed (io.Copy unblock), close stdout and stderr
```
mikebrow: not following the logic here.. Why only close stdout and stderr if !opts.StdinOnce || opts.Tty ?

Random-Liu: I also don't quite the logic...
This is a weird behavior, but `kubectl` is relying on it. :(

mikebrow: kk
// due to kubectl behavior requirements, when !opts.StdinOnce || opts.Tty we have to hold off closing stdout and stderr until wg.Done() (all streams are closed) this replicates current docker shim behavior.

Random-Liu: Will do.

Random-Liu:
@mikebrow Done.

PR #398, opened 2017-11-06
Remove comment about whether other paths should be read only with root

justincormack:
Since https://github.com/moby/moby/pull/35344 we clarified that this behaviour
was a mistake, and the read only flag should just apply to the actual rootfs,
so it corresponds to the OCI read-only option. Other mounts may be able to be
adjusted by re-specifying them or other means but this is unrelated.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -554,8 +554,6 @@ func setOCIDevicesPrivileged(g *generate.Generator) error {
 }
 
 // addOCIBindMounts adds bind mounts.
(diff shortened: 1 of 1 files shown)


Random-Liu:
/lgtm

@justincormack If we decide to do so, I prefer at least add one comment here to clarify that the behavior will be different from Docker. So that in the future, if any user think this is a bug or regression compared with Docker, we could easily find what is wrong.

I'll add that comment after this is merged.

PR #404, opened 2017-11-08
rename the containerTypeLabel

yanxuean:
consistent with dockershim

### DIFF
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -95,12 +95,12 @@ const (
 const (
 	// criContainerdPrefix is common prefix for cri-containerd
 	criContainerdPrefix = "io.cri-containerd"
-	// containerKindLabel is a label key indicating container is sandbox container or application container
-	containerKindLabel = criContainerdPrefix + ".kind"
-	// containerKindSandbox is a label value indicating container is sandbox container
-	containerKindSandbox = "sandbox"
-	// containerKindContainer is a label value indicating container is application container
-	containerKindContainer = "container"
+	// containerTypeLabelKey is a label key indicating container is sandbox container or application container
+	containerTypeLabelKey = criContainerdPrefix + ".container.type"
+	// containerTypeLabelSandbox is a label value indicating container is sandbox container
+	containerTypeLabelSandbox = "sandbox"
+	// containerTypeLabelContainer is a label value indicating container is application container
+	containerTypeLabelContainer = "container"
 	// sandboxMetadataExtension is an extension name that identify metadata of sandbox in CreateContainerRequest
 	sandboxMetadataExtension = criContainerdPrefix + ".sandbox.metadata"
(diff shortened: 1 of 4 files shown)


pkg/server/helpers.go
```diff
@@ -95,12 +95,12 @@ const (
 const (
 	// criContainerdPrefix is common prefix for cri-containerd
 	criContainerdPrefix = "io.cri-containerd"
-	// containerKindLabel is a label key indicating container is sandbox container or application container
-	containerKindLabel = criContainerdPrefix + ".kind"
-	// containerKindSandbox is a label value indicating container is sandbox container
-	containerKindSandbox = "sandbox"
-	// containerKindContainer is a label value indicating container is application container
-	containerKindContainer = "container"
+	// containerTypeLabelKey is a label key indicating container is sandbox container or application container
```
mikebrow: Found this over in the docker shim:
```
	// Internal docker labels used to identify whether a container is a sandbox
	// or a regular container.
	// TODO: This is not backward compatible with older containers. We will
	// need to add filtering based on names.
	containerTypeLabelKey       = "io.kubernetes.docker.type"
	containerTypeLabelSandbox   = "podsandbox"
	containerTypeLabelContainer = "container"
	containerLogPathLabelKey    = "io.kubernetes.container.logpath"
	sandboxIDLabelKey           = "io.kubernetes.sandbox.id"
```
Why do we need to "align" with these internal docker labels?  Is there something kublet is expecting us to do with them? should we change "sandbox" to "podsandbox" here?

abhi: I had the same question as well .Since this is internal to how cir-containerd programs labels in containerd. It really doesn't have to be at the same parity as dockershim. I came across crio using another label in the cadvisor code as well.

Random-Liu: @yanxuean Agree.
@mikebrow @abhi Thanks for pointing this out. :0

yanxuean: @mikebrow @abhi @Random-Liu 
Thank you all.  I'm glad everyone is here. :)
These constant really only be used in cri-containerd.  It don't need to consistent with dockershim.
I modify it to improve its clarity.
My thoughts are:
1. The name  "containerTypeLabelKey" is more clear than containerKindLabel.
At a glance, we can learn that the containerTypeLabelKey is a key.
containerTypeLabelContainer  vs containerKindContainer
containerTypeLabelSandbox  vs containerKindSandbox

The day before yesterday when I review #393 , I cann't remember  meaning of containerKindContainer.

2. It can reduce out burden if we align it with dockershim.

mikebrow: My preferences:
containerKindKey
sandboxType
containerType
or 
Value instead of Type

yanxuean: Thanks @mikebrow 
Everyone has their own preferences. We cann't agree with each other.
I will close it. Thanks all.

PR #414, opened 2017-11-14
Fix data race.

Random-Liu:
Many integration test failures are caused by this. It's very easy to reproduce by running volume e2e test with parallelism 30.

Original I thought `append(command, args...)` is a read only operation, but it seems not.
When different containers using the same image are created at the same time, they'll access the command and args at the same time, `append(command, args...)` seems to corrupt the underlying memory block.

With this fix, I could not reproduce the failure any more.


abhi:
LGTM

PR #420, opened 2017-11-16
Add simple unit test.

Random-Liu:
Add an unit test.


Random-Liu:
@mikebrow Thanks!

PR #421, opened 2017-11-17
Add new cri log support

Random-Liu:
This depends on https://github.com/kubernetes/kubernetes/pull/55922 and https://github.com/kubernetes-incubator/cri-tools/pull/195.

Support https://github.com/kubernetes/kubernetes/pull/55922.


Random-Liu:
I'll send another PR to update cri-tools.

pkg/server/io/logger.go
```diff
@@ -41,7 +42,7 @@ const (
 	// POSIX.1 says that write less than PIPE_BUF is atmoic.
 	pipeBufSize = 4096
 	// bufSize is the size of the read buffer.
-	bufSize = pipeBufSize - len(timestampFormat) - len(Stdout) - 2 /*2 delimiter*/ - 1 /*eol*/
+	bufSize = pipeBufSize - len(timestampFormat) - len(Stdout) - len(runtime.LogTagPartial) - 3 /*3 delimiter*/ - 1 /*eol*/
```
mikebrow: why 3 for delimiter?

Random-Liu: Because we have one more field now. :)
New log format:
```
TIMESTAMP STREAM TAG CONTENT
2016-10-06T00:17:09.669794202Z stdout P log content 1
2016-10-06T00:17:09.669794203Z stdout F log content 2
```

mikebrow: ok one delimiter for each param not 2 bytes for a char in golang my confusion

Issue #418, opened 2017-11-16
Move `code verification` to  pull-cri-containerd-build


Random-Liu:
Travis is not that stable, and vm seems too small:
```
$ make verify
checking lint
for directory ./cmd ...
for directory ./cmd/cri-containerd ...
for directory ./cmd/cri-containerd/options ...
for directory ./pkg ...
for directory ./pkg/server ...
fatal error: runtime: out of memory
runtime stack:
runtime.throw(0x6961a7, 0x16)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/panic.go:605 +0x95
runtime.sysMap(0xc421560000, 0x100000, 0x0, 0x80e538)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/mem_linux.go:216 +0x1d0
runtime.(*mheap).sysAlloc(0x7f5de0, 0x100000, 0x0)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/malloc.go:470 +0xd7
runtime.(*mheap).grow(0x7f5de0, 0x10, 0x0)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/mheap.go:887 +0x60
runtime.(*mheap).allocSpanLocked(0x7f5de0, 0x10, 0x80e548, 0xc42002e000)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/mheap.go:800 +0x334
runtime.(*mheap).alloc_m(0x7f5de0, 0x10, 0x2b84f6370101, 0x1)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/mheap.go:666 +0x118
runtime.(*mheap).alloc.func1()
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/mheap.go:733 +0x4d
runtime.systemstack(0x2b84f637edd0)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/asm_amd64.s:360 +0xab
runtime.(*mheap).alloc(0x7f5de0, 0x10, 0x2b8400010101, 0x2b84f58080b8)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/runtime/mheap.go:732 +0xa1
runtime.largeAlloc(0x1fe00, 0x2b84f6370101, 0x100000000433a87)
[...]

PR #422, opened 2017-11-18
Move verify to kubernetes test-infra.

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/418.

`make verify` usually OOMs after upgrading to golang 1.9. e.g. https://travis-ci.org/kubernetes-incubator/cri-containerd/jobs/303420847.

This PR configure `gometalinter` to only run 2 linters at the same time to reduce memory usage. By default it's 8, and I tried on my workstation, each linter sometimes take more than 1 GB RES memory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,13 +48,11 @@ jobs:
       script:
         - make install.tools
         - make .gitvalidation
-        - make verify
         - make binaries
       go: 1.9.x
     - script:
         - make install.tools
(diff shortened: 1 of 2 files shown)


Random-Liu:
Timeout is too short. :(

Change concurrency to 4, and extend timeout to 600s. Let's see how it goes.

PR #424, opened 2017-11-20
Should not be able to create existing sandbox/container.

Random-Liu:
Add integration test for duplicated sandbox/container creation.

### DIFF
--- a/integration/duplicate_name_test.go
+++ b/integration/duplicate_name_test.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"testing"
+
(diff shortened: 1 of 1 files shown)


integration/duplicate_name_test.go
```diff
@@ -0,0 +1,51 @@
...
+	sb, err := runtimeService.RunPodSandbox(sbConfig)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, runtimeService.StopPodSandbox(sb))
+		assert.NoError(t, runtimeService.RemovePodSandbox(sb))
+	}()
+
+	t.Logf("Create the sandbox again should fail")
+	_, err = runtimeService.RunPodSandbox(sbConfig)
+	require.Error(t, err)
```
mikebrow: check for specific error, or is that to brittle?

integration/duplicate_name_test.go
```diff
@@ -0,0 +1,51 @@
...
+		"container",
+		pauseImage,
+	)
+	_, err = runtimeService.CreateContainer(sb, cnConfig, sbConfig)
+	require.NoError(t, err)
+
+	t.Logf("Create the container again should fail")
+	_, err = runtimeService.CreateContainer(sb, cnConfig, sbConfig)
+	require.Error(t, err)
+}
```
mikebrow: test for similar names? or is that to pedantic?

integration/duplicate_name_test.go
Random-Liu: There is no specific error defined in CRI, we could only match error message, which could change overtime. :(

integration/duplicate_name_test.go
Random-Liu: Seems not needed. :p

Random-Liu:
@mikebrow Thanks for reviewing!

PR #426, opened 2017-11-21
Add restart test for sandbox recovery.

Random-Liu:
For https://github.com/kubernetes-incubator/cri-containerd/issues/238.

Add integration test to test sandbox state recovery across cri-containerd restart.

### DIFF
--- a/hack/test-utils.sh
+++ b/hack/test-utils.sh
@@ -18,8 +18,13 @@ ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 . ${ROOT}/hack/versions
 # CRI_CONTAINERD_FLAGS are the extra flags to use when start cri-containerd.
 CRI_CONTAINERD_FLAGS=${CRI_CONTAINERD_FLAGS:-""}
+# RESTART_WAIT_PERIOD is the period to wait before restarting cri-containerd/containerd.
+RESTART_WAIT_PERIOD=${RESTART_WAIT_PERIOD:-10}
 
 CRICONTAINERD_SOCK=/var/run/cri-containerd.sock
+
+cri_containerd_pid=
+containerd_pid=
 
 # test_setup starts containerd and cri-containerd.
 test_setup() {
@@ -35,20 +40,40 @@ test_setup() {
     exit 1
   fi
   sudo pkill containerd
-  sudo containerd &> ${report_dir}/containerd.log &
+  keepalive "sudo containerd" ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
+  containerd_pid=$!
   # Wait for containerd to be running by using the containerd client ctr to check the version
   # of the containerd server. Wait an increasing amount of time after each of five attempts
   readiness_check "sudo ctr version"
 
   # Start cri-containerd
-  sudo ${ROOT}/_output/cri-containerd --alsologtostderr --v 4 ${CRI_CONTAINERD_FLAGS} \
-	  &> ${report_dir}/cri-containerd.log &
+  keepalive "sudo ${ROOT}/_output/cri-containerd --alsologtostderr --v 4 ${CRI_CONTAINERD_FLAGS}" \
+	  ${RESTART_WAIT_PERIOD} &> ${report_dir}/cri-containerd.log &
+  cri_containerd_pid=$!
   readiness_check "sudo ${GOPATH}/bin/crictl --runtime-endpoint=${CRICONTAINERD_SOCK} info"
 }
 
 # test_teardown kills containerd and cri-containerd.
 test_teardown() {
+  if [ -n "${containerd_pid}" ]; then
+    kill ${containerd_pid}
+  fi
+  if [ -n "${cri_containerd_pid}" ]; then
+    kill ${cri_containerd_pid}
+  fi
   sudo pkill containerd
+}
+
+# keepalive runs a command and keeps it alive.
+# keepalive process is eventually killed in test_teardown.
+keepalive() {
+  local command=$1
+  echo ${command}
+  local wait_period=$2
+  while true; do
+    ${command}
+    sleep ${wait_period}
+  done
 }
 
 # readiness_check checks readiness of a daemon with specified command.
--- a/integration/test_utils.go
+++ b/integration/test_utils.go
@@ -18,6 +18,8 @@ package integration
 
 import (
 	"errors"
+	"fmt"
+	"os/exec"
 	"time"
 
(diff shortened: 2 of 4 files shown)


integration/test_utils.go
```diff
@@ -48,28 +51,61 @@ var (
 )
 
 func init() {
+	if err := ConnectDaemons(); err != nil {
+		glog.Exitf("Failed to connect daemons: %v", err)
```
mikebrow: Failed to connect with containerd..

integration/test_utils.go
```diff
@@ -48,28 +51,61 @@ var (
...
 	}
 	imageService, err = remote.NewRemoteImageService(sock, timeout)
 	if err != nil {
-		glog.Exitf("Failed to create image service: %v", err)
+		return fmt.Errorf("failed to create image service: %v", err)
+	}
+	// Since CRI grpc client doesn't have `WithBlock` specified, we
+	// need to check whether it is actually connected.
+	// TODO(random-liu): Extend cri remote client to accept extra grpc options.
+	_, err = runtimeService.ListContainers(&runtime.ContainerFilter{})
```
mikebrow: runtimeService.Version

integration/test_utils.go
```diff
@@ -48,28 +51,61 @@ var (
...
-		glog.Exitf("Failed to create runtime service: %v", err)
+		return fmt.Errorf("failed to create runtime service: %v", err)
 	}
 	imageService, err = remote.NewRemoteImageService(sock, timeout)
 	if err != nil {
-		glog.Exitf("Failed to create image service: %v", err)
+		return fmt.Errorf("failed to create image service: %v", err)
+	}
+	// Since CRI grpc client doesn't have `WithBlock` specified, we
+	// need to check whether it is actually connected.
```
mikebrow: what does it mean to have a runtimeService or imageService (grpc service) but it's not actually connected? Will it fail right away? should we do a sleep loop here to wait for it to be actually connected?

mikebrow: is there some grpc api or option we're supposed to be using to demand an actual connection or err?

pkg/client/client.go
```diff
@@ -34,6 +34,7 @@ func NewCRIContainerdClient(endpoint string, timeout time.Duration) (api.CRICont
 		return nil, fmt.Errorf("failed to get dialer: %v", err)
 	}
 	conn, err := grpc.Dial(addr,
+		grpc.WithBlock(),
```
mikebrow: // WithBlock returns a DialOption which makes caller of Dial blocks until the underlying connection is up. Without this, Dial returns immediately and connecting the server happens in background.

pkg/client/client.go
```diff
@@ -34,6 +34,7 @@ func NewCRIContainerdClient(endpoint string, timeout time.Duration) (api.CRICont
...
 		grpc.WithInsecure(),
 		grpc.WithTimeout(timeout),
```
mikebrow: // WithTimeout returns a DialOption that configures a timeout for dialing a ClientConn initially. This is valid if and only if WithBlock() is present. Deprecated: use DialContext and context.WithTimeout instead.

pkg/client/client.go
```diff
@@ -34,6 +34,7 @@ func NewCRIContainerdClient(endpoint string, timeout time.Duration) (api.CRICont
...
 		grpc.WithDialer(dialer),
```
mikebrow: // WithDialer returns a DialOption that specifies a function to use for dialing network addresses. If FailOnNonTempDialError() is set to true, and an error is returned by f, gRPC checks the error's Temporary() method to decide if it should try to reconnect to the network address.

pkg/client/client.go
mikebrow: Note: we're adding something that's been deprecated

hack/test-utils.sh
```diff
@@ -35,22 +40,41 @@ test_setup() {
...
+  fi
   sudo pkill containerd
 }
 
+# keepalive runs a command and keeps it alive.
+keepalive() {
+  local command=$1
+  echo ${command}
+  local wait_period=$2
+  while true; do
```
mikebrow: what am I missing here? doesn't look like keepalive() returns from it's infinite loop

Random-Liu:
Will address comments today.

hack/test-utils.sh
Random-Liu: It is killed eventually.

mikebrow: ah pls add comment to loop :-)

Random-Liu: Done.

integration/test_utils.go
Random-Liu: This tries to connect cri-containerd and containerd, not only containerd.

integration/test_utils.go
Random-Liu: `WithBlock` the problem is that we are not using that in Kubernetes today.

integration/test_utils.go
Random-Liu: There is no `Version` in image service, so let's just use 2 list calls. It doesn't actually matter.

pkg/client/client.go
Random-Liu: Done.

pkg/client/client.go
Random-Liu: Added TODO.

Random-Liu:
@mikebrow Addressed comments.

integration/test_utils.go
mikebrow: "Failed to connect cri-containerd and containerd"?

integration/test_utils.go
mikebrow: Should we file an issue and track here?

integration/test_utils.go
mikebrow: just seems like an odd api to call to check whether it's connected.  Not sure why it matters if imageService doesn't have a version API. But it probably should.

Random-Liu:
> There is no Version in image service

Actually this is a problem... We just haven't hit this yet.
We assume that image and runtime services may be separate, but we don't have separate `Status` and `Version` function.

> WithBlock the problem is that we are not using that in Kubernetes today

I think adding `WithBlock` in kubelet cri remote client may change the original behavior, which might cause some bug. It doesn't add value, but may introduce bug, so I don't think we should do that.
However, we should let kubelet cri remote client accept extra grpc option, so that other users than kubelet could be able to customize it. We could file an issue for that.

Random-Liu:
Filed an issue for `WithBlock` https://github.com/kubernetes/kubernetes/issues/57060.

Issue #164, opened 2017-08-24
Check whether permanent network namespace is removed in SandboxRemove


Random-Liu:
Currently, we stop sandbox container and remove permanent network namespace when stop sandbox.

When remove sandbox, we check whether sandbox container is still running, but we should also check whether network namespace is properly removed.

PR #427, opened 2017-11-21
Do not remove sandbox when netns is not closed.

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/164.
For https://github.com/kubernetes-incubator/cri-containerd/issues/238.

### DIFF
--- a/pkg/server/sandbox_remove.go
+++ b/pkg/server/sandbox_remove.go
@@ -47,13 +47,17 @@ func (c *criContainerdService) RemovePodSandbox(ctx context.Context, r *runtime.
 	id := sandbox.ID
 
 	// Return error if sandbox container is not fully stopped.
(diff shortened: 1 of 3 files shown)


Random-Liu:
@mikebrow Thanks!

pkg/server/sandbox_remove.go
```diff
@@ -56,6 +55,11 @@ func (c *criContainerdService) RemovePodSandbox(ctx context.Context, r *runtime.
 		return nil, fmt.Errorf("sandbox container %q is not fully stopped", id)
 	}
 
+	// Return error if sandbox network namespace is not closed yet.
+	if sandbox.NetNS != nil && !sandbox.NetNS.Closed() {
```
abhi: @Random-Liu qq: so what is the recovery here? How is the network namespace eventually removed and closed ? And if its not closed , what is the behavior if the sandbox is restored ?

PR #440, opened 2017-11-28
Use oci package

dnephin:
https://github.com/containerd/containerd/pull/1796 moved `SpecOpts` into a new package. This PR updates the vendoered version of `containerd`, and updates the code to use the new package.


vendor.conf
```diff
@@ -2,7 +2,7 @@ github.com/blang/semver v3.1.0
 github.com/boltdb/bolt e9cf4fae01b5a8ff89d0ec6b32f0d9c9f79aefdd
 github.com/BurntSushi/toml a368813c5e648fee92e5f6c30e3944ff9d5e8895
 github.com/containerd/cgroups f7dd103d3e4e696aa67152f6b4ddd1779a3455a9
-github.com/containerd/containerd 70e0c8443ff15dcbd2ad8e0d07ed087fc2a83e05
+github.com/containerd/containerd 9e04cff8e9e3a1bf13c088cb3db1c368e93b33ea
```
mikebrow: We have another file over there where we need to modify/keep in sync containerd versions as well:
https://github.com/kubernetes-incubator/cri-containerd/blob/master/hack/versions#L3

dnephin: Updated!

Random-Liu:
/retest

Seems to be something wrong with the bot.

PR #444, opened 2017-11-29
should specify snapshotter for Pull

yanxuean:

Now we don't specify snapshotter for client.Pull. so the containerd use default snapshotter.
When the user don't specify default snapshotter(overlayfs), there will be a problem.


Random-Liu:
@yanxuean Only `Unpack` needs snapshotter, right?

yanxuean:
@Random-Liu Yes, you are correct. I close it.

PR #453, opened 2017-12-01
Add gc.root back in vendored containerd client.

Random-Liu:
For https://github.com/containerd/containerd/issues/1785.

This PR partially revert https://github.com/containerd/containerd/pull/1690 in vendored containerd client.

Note that:
1. Image load logic is not changed, because that never fails in test.
2. A check point related `gc.root` is not added because that is on server side which we are not vendoring. However, we don't use checkpoint anyway.

If after this, the test failure never happens again, we may want to keep this until containerd side is fixed.
If after this, the test still fails, we should revert this PR immediately.

### DIFF
--- a/vendor/github.com/containerd/containerd/client.go
+++ b/vendor/github.com/containerd/containerd/client.go
@@ -238,7 +238,7 @@ func (c *Client) Pull(ctx context.Context, ref string, opts ...RemoteOpt) (Image
 		handler = images.Handlers(append(pullCtx.BaseHandlers, schema1Converter)...)
 	} else {
 		handler = images.Handlers(append(pullCtx.BaseHandlers,
-			remotes.FetchHandler(store, fetcher),
+			remotes.FetchHandler(store, fetcher, desc),
 			images.ChildrenHandler(store, platforms.Default()))...,
 		)
 	}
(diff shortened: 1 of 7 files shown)


Random-Liu:
@yanxuean Could you help me take a look? I want to merge it today, so that we could have some test result over night.

yanxuean:
@Random-Liu    ok

yanxuean:
lgtm
No problem.  We don't use (s *service) writeContent  and withRemappedSnapshotBase.

Random-Liu:
OK. Seems not related.
Got another one https://storage.googleapis.com/kubernetes-jenkins/logs/ci-cri-containerd-e2e-ubuntu-gce/556/build-log.txt
```
event for my-hostname-basic-03b8047e-d667-11e7-9fc0-0a580a3c7a58-7md6n: {kubelet bootstrap-e2e-minion-group-cx69} Failed: Error: failed to create containerd task: OCI runtime create failed: container_linux.go:296: starting container process caused "exec: \"/serve_hostname\": stat /serve_hostname: no such file or directory": unknown
```

yanxuean:
bad news

yanxuean:
Have we any direction?

PR #457, opened 2017-12-01
adding verify back

mikebrow:
Adding verify back. Since we we're having intermittent travis errors for verify with go 1.9, I put it in under a 1.8.x in the build stage which will run in parallel with the 1.9 and tip build tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
     - docker run --rm -v /usr/local/bin:/target jpetazzo/nsenter
     # Pull request test doesn't need google cloud sdk.
     - if [ "$TRAVIS_PULL_REQUEST" = "false" ] && [ ! -d ${HOME}/google-cloud-sdk ]; then
-        rm -rf "${HOME}/google-cloud-sdk"; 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -55,6 +55,12 @@ jobs:
         - make .gitvalidation
         - make binaries
       go: tip
+    - script:
+        - make install.tools
+        - make .gitvalidation
+        - make binaries
+        - make verify
+      go: 1.8.x
```
yanxuean: 1.8.x  ?

mikebrow: we were having timeouts with lint on 1.9 on travis... so verify was moved to just the pull stage. So here I'm just using 1.8 for early detection of lint and boilerplate errors.

Random-Liu:
@mikebrow Given that:
1) verify consumes a lot of resource, it very frequently fails on travis VM.
2) We've got pull-cri-containerd-verify
3) You've fixed boilterplate.

Maybe we could close this one?

mikebrow:
will reopen if / when we get more PR traffic that should be caught before we approve for testing (ok-to-test)

PR #464, opened 2017-12-04
add debug log for rootfs-not-ready

yanxuean:

We should also modify containerd's diff.Apply.

### DIFF
--- a/pkg/server/remove_soon.go
+++ b/pkg/server/remove_soon.go
@@ -48,13 +48,25 @@ func (c *criContainerdService) checkSnapshot(ctx context.Context, snapshot, entr
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(root) // nolint: errcheck
+	removeTemp := true
+	defer func() {
+		if removeTemp {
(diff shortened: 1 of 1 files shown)


yanxuean:
We should also modify diff.Apply in containerd.

Random-Liu:
@yanxuean I don't think it's our `unmount` fails. It is the unmount in containerd fails, it cleans up the snapshot unexpectedly.

IIUC, our removeall will only cleanup the top layer (writable layer), which is fine.

yanxuean:
Yes, we need to modify it in containerd. 
My main thought is printing the "mounts" got from snapshotter.Mounts in checkSnapshot.
It will verify if the parent snapshot is exist.

But Now I think we don't need it again. Because we has been congratulating the 1785 was solve.

yanxuean:
I will close it.

yanxuean:
@Random-Liu  Will we revert the commit about checkSnapshot?  If not I suggest to print the  mounts.

yanxuean:
The checkSnapshot has been reverted. #473 .  Close this.

PR #471, opened 2017-12-05
Improve unmount for snapshot

yanxuean:
ref: https://github.com/containerd/containerd/issues/1785
1. sync Unmount for snapshot
2. best effort to remove temp dir for snapshot

### DIFF
--- a/pkg/containerd/opts/container.go
+++ b/pkg/containerd/opts/container.go
@@ -25,10 +25,11 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/mount"
 	"github.com/docker/docker/pkg/chrootarchive"
 	"github.com/docker/docker/pkg/system"
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
-	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/containerd/opts/container.go
```diff
@@ -71,13 +71,21 @@ func WithVolumes(volumeMounts map[string]string) containerd.NewContainerOpts {
...
+		// We change RemoveAll to Remove so that we either leak a temp dir
+		// if it fails but not RM snapshot data.
+		// refer to https://github.com/containerd/containerd/pull/1868
+		// https://github.com/containerd/containerd/pull/1785
+		defer os.Remove(root) // nolint: errcheck
+		if err := mount.All(mounts, root); err != nil {
+			return errors.Wrap(err, "failed to mount")
 		}
-		defer unix.Unmount(root, 0) // nolint: errcheck
+		defer func() {
```
Random-Liu: I prefer we log the error here in our code.
Or else if `err != nil`, we'll lose the error info.

yanxuean: Yes. It is same with happened in containerd. When i modify it in containerd, I don't know how to aggregate these errors. So...
I will find the way to aggregate it again.

Random-Liu: I think log it is fine. :)

yanxuean: done

crosbymichael:
LGTM

PR #480, opened 2017-12-06
Switch to a tmp GOPATH for `make install.deps`

### LINKED ISSUE #466: Switch to a temporary GOPATH for `make install.deps`.
We may not want to mess up the local GOPATH.
We may probably want to switch to a temporary `GOPATH` to do `make install.deps`.

miaoyq:

fixes #466

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -54,20 +54,16 @@ CONTAINERD_DIR=${DESTDIR}/usr/local
 RUNC_DIR=${DESTDIR}
 CNI_DIR=${DESTDIR}/opt/cni
 CNI_CONFIG_DIR=${DESTDIR}/etc/cni/net.d
+CRICTL_DIR=${DESTDIR}/usr/local/bin
 CRICTL_CONFIG_DIR=${DESTDIR}/etc
 
 RUNC_PKG=github.com/opencontainers/runc
 CNI_PKG=github.com/containernetworking/plugins
 CONTAINERD_PKG=github.com/containerd/containerd
 CRITOOL_PKG=github.com/kubernetes-incubator/cri-tools
-# Check GOPATH
-if [[ -z "${GOPATH}" ]]; then
-  echo "GOPATH is not set"
-  exit 1
-fi
 
-# For multiple GOPATHs, keep the first one only
-GOPATH=${GOPATH%%:*}
+# Create a temporary GOPATH for make install.deps.
+GOPATH=$(mktemp -d /tmp/cri-containerd-install-deps.XXXX)
 
 # checkout_repo checks out specified repository
 # and switch to specified  version.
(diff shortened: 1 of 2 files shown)


mikebrow:
Didn't work for the containerd dependency build.

miaoyq:
@mikebrow Have fixed. :)

miaoyq:
Thanks @mikebrow, but if we clean the tmp dir, `crictl` will also be clean.
Maybe we should add `make install` [here](https://github.com/miaoyq/cri-containerd/blob/baac03de898547ebce8f876593984c76a72919a5/hack/install-deps.sh#L143)?

mikebrow:
I meant tmp/whatever-new-directory-you-pick-for-the-temp-gopath-for-the-install.deps :-)

miaoyq:
@mikebrow For kubecon, this week is a busy week for you :)
So I will add `make install` for `crictl`, it also works for developer mode since `crictl` will be copied to `/usr/local/bin` by `make install`.

miaoyq:
> @miaoyq see "cp: cannot stat '/go/bin/crictl': No such file or directory" we're looking or setting the wrong directory here for the ci case

@mikebrow  It's because release.sh still get the crictl from local `GOPATH` when we set the tmp GOPATH, see:
https://github.com/kubernetes-incubator/cri-containerd/blob/c89e55515016b380ef779f37865cd4ca05b7c5bf/hack/release.sh#L34-L41

Now we have installed `crictl` to `${destdir}/usr/local/bin/` in `install-deps.sh`, so we need delete the following lines in `release.sh`:
```bash
 # Copy crictl tool into release stage 
cp ${GOPATH}/bin/crictl ${destdir}/usr/local/bin/ 
```

hack/install-deps.sh
```diff
@@ -54,20 +54,17 @@ CONTAINERD_DIR=${DESTDIR}/usr/local
...
-# Check GOPATH
-if [[ -z "${GOPATH}" ]]; then
-  echo "GOPATH is not set"
-  exit 1
-fi
 
-# For multiple GOPATHs, keep the first one only
-GOPATH=${GOPATH%%:*}
+# Create a temporary GOPATH for make install.deps.
+GOPATH=/tmp/cri-containerd-install-deps
```
Random-Liu: `mktemp`?

Random-Liu:
@miaoyq LGTM overall other than the comment.

hack/install-deps.sh
miaoyq: Done.

Random-Liu:
```
$ make install.deps
mkdir: cannot create directory ‘/tmp/cri-containerd-install-deps.bJc3’: Permission denied
```
@miaoyq Hm, weird. Did you get this error before you change to `mktemp`?

miaoyq:
@Random-Liu No,  it's OK before this change.  

Maybe we should add `${sudo}` before `checkout_repo` or remove the `${sudo}` before `mktemp`.

PR #484, opened 2017-12-08
add filter for containerd event

yanxuean:
goal:
1. improve cri-containerd performance
2. clean log.
There are many useless info in log.
```
I1208 15:49:43.785508   12909 events.go:73] Received container event timestamp - 2017-12-08 07:49:43.784790728 +0000 UTC, namespace - "k8s.io", topic - "/images/create"
I1208 15:49:43.786397   12909 image_pull.go:111] Unpack image "busybox"
I1208 15:49:43.903693   12909 events.go:73] Received container event timestamp - 2017-12-08 07:49:43.903308963 +0000 UTC, namespace - "k8s.io", topic - "/images/update"
I1208 15:49:43.905175   12909 events.go:73] Received container event timestamp - 2017-12-08 07:49:43.90488295 +0000 UTC, namespace - "k8s.io", topic - "/images/create"
I1208 15:49:43.920262   12909 events.go:73] Received container event timestamp - 2017-12-08 07:49:43.919800908 +0000 UTC, namespace - "k8s.io", topic - "/images/create"
```

### DIFF
--- a/pkg/server/container_status.go
+++ b/pkg/server/container_status.go
@@ -40,7 +40,7 @@ func (c *criContainerdService) ContainerStatus(ctx context.Context, r *runtime.C
 	// * ImageSpec in container config is image ID.
 	// * ImageSpec in container status is image tag.
 	// * ImageRef in container status is repo digest.
-	spec := container.Metadata.Config.GetImage()
+	spec := container.Config.GetImage()
 	imageRef := container.ImageRef
 	image, err := c.imageStore.Get(imageRef)
 	if err != nil {
--- a/pkg/server/events.go
+++ b/pkg/server/events.go
@@ -57,9 +57,13 @@ func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store) *eventMonit
 	}
 }
 
-// subscribe starts subsribe containerd events. We separate subscribe from
+// subscribe starts to subscribe containerd events.
 func (em *eventMonitor) subscribe(client *containerd.Client) {
-	em.ch, em.errCh = client.Subscribe(em.ctx)
+	filters := []string{
+		`topic=="/tasks/exit"`,
+		`topic=="/tasks/oom"`,
+	}
+	em.ch, em.errCh = client.Subscribe(em.ctx, filters...)
 }
 
 // start starts the event monitor which monitors and handles all container events. It returns
@@ -74,7 +78,7 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 		for {
 			select {
 			case e := <-em.ch:
-				logrus.Debugf("Received container event timestamp - %v, namespace - %q, topic - %q", e.Timestamp, e.Namespace, e.Topic)
+				logrus.Debugf("Received containerd event timestamp - %v, namespace - %q, topic - %q", e.Timestamp, e.Namespace, e.Topic)
 				em.handleEvent(e)
 			case err := <-em.errCh:
 				logrus.WithError(err).Error("Failed to handle event stream")


mikebrow:
Will need to use our vendor process to change containerd, and typically with a separate commit. Not so sure the events are useless as they may be important to switch state. But we may be able to filter the logs some when we are not in a debug log mode.

yanxuean:
Seem like  we only handle two events now , e.g. taskExit and taskOOM . 
We receive all events, and Unmarshal them. It is a bit of waste.

yanxuean:
Remove dependency to /containerd/containerd/runtime. It is inappropriate.

Random-Liu:
Actually containerd logs all these events, I think it's fine not to log them on our side.
And today's event log doesn't give us too much information because there is no id associated with each event, we don't know which container/task/image/snapshot/content an event belongs to.

And containerd itself already logs these events in debug log level.

@mikebrow WDYT?

mikebrow:
> Actually containerd logs all these events, I think it's fine not to log them on our side.
> And today's event log doesn't give us too much information because there is no id associated with each event, we don't know which container/task/image/snapshot/content an event belongs to.
> 
> And containerd itself already logs these events in debug log level.
> 
> @mikebrow WDYT?

Shouldn't we be keeping track of our own event monitoring? Maybe need a phone call on this one to discuss active vs. passive patterns for event monitoring under k8s CRI.

Random-Liu:
@mikebrow We only monitor those 2 events. That's our own event monitoring.
We don't care about other events for now.

Random-Liu:
On second thought, I think we may want to keep the event log for now. Debuggability is more important to us know.
Next release, we'll focus on performance. If we do find that this is a performance bottleneck, we could always add the filter then. :)

mikebrow:
I think they are important for "debug" purposes and agree it's a good idea to filter them out for performance reasons... So let's open up and issue to add a filtered events mode (performance) vs debug mode debug (trace events on) switch. I'm thinking if we need an ID for tracing the events to associated objects then let's add them.

yanxuean:
OK, I filed a issue, We close the pr.

Random-Liu:
@yanxuean Can you post your benchmark data here?

pkg/server/container_status.go
```diff
@@ -41,7 +41,7 @@ func (c *criContainerdService) ContainerStatus(ctx context.Context, r *runtime.C
 	// * ImageSpec in container status is image tag.
 	// * ImageRef in container status is repo digest.
 	spec := container.Metadata.Config.GetImage()
-	imageRef := container.ImageRef
+	imageRef := container.Metadata.ImageRef
```
Random-Liu: This doesn't matter I think.

pkg/server/events.go
```diff
@@ -57,9 +57,13 @@ func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store) *eventMonit
 	}
 }
 
-// subscribe starts subsribe containerd events. We separate subscribe from
+// subscribe starts subscribe containerd events. We separate subscribe from
```
Random-Liu: My initial comment is incomplete... Please remove the second sentence.

Random-Liu:
1. Based on @yanxuean's benchmark result, this does reduce memory usage during batch pod creation.
2. We are going to merge cri-containerd and containerd, and containerd already logs the events.

Given so, I think we should filter events here.

@yanxuean Please post your benchmark data here.

/cc @mikebrow

pkg/server/container_status.go
yanxuean: consistent with "container.Metadata.Config.GetImage()"

pkg/server/events.go
yanxuean: done

pkg/server/container_status.go
yanxuean: It is better to remove it.:)

yanxuean:
no filter:
```
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"kubelet" 0.038      72.74                  37.06
"runtime" 0.003      512.56                 18.85
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"runtime" 0.004      516.79                 23.17
"kubelet" 0.089      72.00                  37.94
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"kubelet" 0.072      69.57                  37.10
"runtime" 0.008      514.61                 21.04
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"kubelet" 0.075      69.25                  38.35
"runtime" 0.004      514.23                 20.68
```

add filter:
```
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"kubelet" 0.038      48.90                  38.45
"runtime" 0.005      61.18                  21.88
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"runtime" 0.003      56.61                  17.32
"kubelet" 0.067      46.98                  38.00
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"kubelet" 0.065      45.68                  37.99
"runtime" 0.005      66.34                  27.05
container cpu(cores) memory_working_set(MB) memory_rss(MB)
"kubelet" 0.078      56.87                  38.52
"runtime" 0.016      223.77                 16.88
```

PR #487, opened 2017-12-11
improve image

yanxuean:

1. improve calling for content and remove contentStoreService in criContainerdService
2. remove taskService and imageStoreService from criContainerdService
3. reliably remove image when content missing(e.g. missing image.config)

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -114,8 +114,15 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
 		// Do not fail image pulling. Unpack will be retried before container creation.
 	}
 
+	// Get image information.
+	info, err := getImageInfo(ctx, image)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get image information: %v", err)
+	}
+	imageID := info.id
+
 	repoDigest, repoTag := getRepoDigestAndTag(namedRef, image.Target().Digest, isSchema1)
-	for _, r := range []string{repoTag, repoDigest} {
+	for _, r := range []string{repoTag, repoDigest, imageID} {
 		if r == "" {
 			continue
 		}
@@ -123,16 +130,7 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
 			return nil, fmt.Errorf("failed to update image reference %q: %v", r, err)
 		}
 	}
-	// Get image information.
-	info, err := getImageInfo(ctx, image, c.client.ContentStore())
-	if err != nil {
-		return nil, fmt.Errorf("failed to get image information: %v", err)
-	}
-	imageID := info.id
 
-	if err := c.createImageReference(ctx, imageID, image.Target()); err != nil {
-		return nil, fmt.Errorf("failed to update image reference %q: %v", imageID, err)
-	}
 	glog.V(4).Infof("Pulled image %q with image id %q, repo tag %q, repo digest %q", imageRef, imageID,
 		repoTag, repoDigest)
 	img := imagestore.Image{
@@ -158,7 +156,7 @@ func (c *criContainerdService) PullImage(ctx context.Context, r *runtime.PullIma
 	// by someone else anytime, before/during/after we create the metadata. We should always
 	// check the actual state in containerd before using the image or returning status of the
 	// image.
-	return &runtime.PullImageResponse{ImageRef: img.ID}, err
+	return &runtime.PullImageResponse{ImageRef: img.ID}, nil
 }
 
 // ParseAuth parses AuthConfig and returns username and password/secret required by containerd.
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -199,13 +199,13 @@ func (c *criContainerdService) createImageReference(ctx context.Context, name st
 	}
 	// TODO(random-liu): Figure out which is the more performant sequence create then update or
 	// update then create.
-	_, err := c.imageStoreService.Create(ctx, img)
+	_, err := c.client.ImageService().Create(ctx, img)
 	if err == nil {
 		return nil
 	}
 	if err != nil && !errdefs.IsAlreadyExists(err) {
 		return err
 	}
-	_, err = c.imageStoreService.Update(ctx, img, "target")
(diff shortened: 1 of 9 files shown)


pkg/server/image_remove.go
```diff
@@ -51,15 +51,12 @@ func (c *criContainerdService) RemoveImage(ctx context.Context, r *runtime.Remov
 			}
 			return nil, fmt.Errorf("failed to get image %q: %v", tag, err)
 		}
-		desc, err := cImage.Config(ctx)
```
Random-Liu: I think we should use image id here. Why change it to manifest digest?

Random-Liu:
@yanxuean Please take a look at the comment.

pkg/server/image_pull.go
```diff
@@ -59,6 +59,9 @@ import (
 // if we saw an image without snapshots or with in-complete contents during startup,
 // should we re-pull the image? Or should we remove the entry?
 //
+// yanxuean: We cann't delete image directly, because we don't know if the image
```
Random-Liu: s/cann't/can't

pkg/server/image_remove.go
yanxuean: We can get the manifest digest from the struct directly.  but we only get image id by reading Config from content. If the content of config is missing, we will fail to get image id,we will can't remove the image forever.

pkg/server/image_pull.go
yanxuean: done

pkg/server/image_remove.go
Random-Liu: That's a valid concern.

Did you see that in your environment?

yanxuean:
@Random-Liu I think taskService and imageStoreService fields are unnecessary. We can use client.ImageService directly.
```
 	c := &criContainerdService{
-		taskService:         client.TaskService(),
-		imageStoreService:   client.ImageService(),
-		contentStoreService: client.ContentStore(),
```

pkg/server/image_remove.go
Random-Liu: Logically we should still use image id. If you are concern with corrupted image, can we check not found error instead?

Random-Liu:
@yanxuean That's leaved for legacy reason, feel free to clean it up.

pkg/server/image_remove.go
yanxuean: I saw a problem. But it is ok after we delete dir "/var/lib/containerd":)
So I don't know the reason. just guess it.

The Config() function will read minifest, config from content. It is difficult to check that it is not found.

Random-Liu: The problem is that using manifest digest here is wrong.
Our image is indexed by image ID. Removing an image means removing all references associated with this image id.
It's possible that 2 manifest digests have the same image ID in theory.

Random-Liu: @yanxuean WDYT? If you decide to do this, please add comment in your code to explain we should do that. :)

yanxuean: >It's possible that 2 manifest digests have the same image ID in theory.

Is this from upgraded image. For example, I create a new image by only adding a layer. the new image  will has same config with old image, but their minifest is different.

yanxuean: If 2 manifest digests have the same image ID . There is a new problem.
For example. 
1. image1:    minifest1, imageID1(from config), repoTag1, repoTag2
2. image2:   minifest2, imageID1(from config),  repoTag1, repoTag3
We pull image1 first, then image2.  The repoTag1 in imageService refers to minifest2 now.
We want to remove image1, Our current RemoveImage() will successfully remove the repoTag1 from imageService. because the imageID1 is unmodified.

Random-Liu: @yanxuean `RootFS` is uncompressed. Manifest digest is generated from compressed content.

In theory it's possible that an image is compressed with different algorithm, then they'll have the same uncompressed id - image id, but different ids generated from compressed contents -
 manifest digest.

pkg/server/image_remove.go
```diff
@@ -53,26 +53,39 @@ func (c *criContainerdService) RemoveImage(ctx context.Context, r *runtime.Remov
 		}
 		desc, err := cImage.Config(ctx)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get image %q config descriptor: %v", tag, err)
+			// We can only get image id by reading Config from content.
+			// If the config is missing, we will fail to get image id,
+			// So we will can't remove the image forever,
```
Random-Liu: won't be able to
I'll fix in another PR.

pkg/server/image_remove.go
```diff
@@ -53,26 +53,39 @@ func (c *criContainerdService) RemoveImage(ctx context.Context, r *runtime.Remov
...
+			// and cri-containerd always report the image is ok.
+			// But we also don't check it by manifest,
+			// It's possible that two manifest digests have the same image ID in theory.
+			// In theory it's possible that an image is compressed with different algorithms,
+			// then they'll have the same uncompressed id - image id,
+			// but different ids generated from compressed contents - manifest digest.
+			// So we decide to leave it.
+			// After all, the user can override the repoTag by pulling image again.
+			glog.Errorf("can't remove image,failed to get config for Image tag %q,id %q: %v", tag, image.ID, err)
```
Random-Liu: s/can't/Can't.

I'll do the cleanup in another PR.

pkg/server/image_remove.go
yanxuean: It will waste your many time.
You may wait for my modification. just for a while.
I will be responsible in the end. :)

Random-Liu: Sure. Will wait for your PR then.

pkg/server/image_remove.go
miaoyq: >@yanxuean RootFS is generated from uncompressed layers. Manifest digest is generated from compressed content.

>In theory it's possible that an image is compressed with different algorithms, then they'll have the same uncompressed id - image id, but different ids generated from compressed contents - manifest digest.

@Random-Liu Manifest digest seems to be generated from `manifest`, but it's a json file.
Did you mean that `manifest` contain the info of compressed layers(e.g, `mediaType`, `size`) and layers may be compressed with different algorithms?

Issue #419, opened 2017-11-16
Add containerd/cri-containerd monitor.


Random-Liu:
In our current kube-up.sh integration, we don't have a [`kube-docker-monitor.service`](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/node.yaml#L42) like health monitor for cri-containerd/containerd.

We should add one.

Issue #289, opened 2017-09-26
Integrate with kube-up.sh


Random-Liu:
Although kubespray and kops are more advanced cluster bootstrapping solution, in near term, all Kubernetes upstream test infrastructure and many production workloads will be still using `kube-up.sh`.

We should integrate with `kube-up.sh` and setup cluster e2e test in Kubernetes test infrastructure.

Random-Liu:
This has been mostly done.

We should polish and document this.

PR #490, opened 2017-12-12
Cleanup kube up

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/419.
For https://github.com/kubernetes-incubator/cri-containerd/issues/289.

This PR:
1) Added `cri-containerd-monitor` /cc @miaoyq 
2) Added `cluster/` directory for `kube-up.sh`, and package `cluster/` directory into release tarball.
3) Moved `test/e2e/*.yaml` to `cluster/gce/cloud-init/*.yaml`, and created a symlink at `test/e2e`.

With this PR, user will be able to:
1) Download cri-containerd release tarball and untar.
2) Download kubernetes release tarball and untar.
3) Initialize environment variables `. ${CRI_CONTAINERD_ROOT}/cluster/env`.
4) Create kubernetes cluster `${KUBERNETES_ROOT}/cluster/kube-up.sh`.

I'll add a document for `kube-up.sh` after this PR is merged.


Random-Liu:
New `release-stage/`:
```
$ tree _output/release-stage/
_output/release-stage/
├── etc
│   ├── cni
│   │   └── net.d
│   │       └── 10-containerd-net.conflist
│   ├── crictl.yaml
│   └── systemd
│       └── system
│           ├── containerd.service
│           └── cri-containerd.service
├── opt
│   ├── cni
│   │   └── bin
│   │       ├── bridge
│   │       ├── dhcp
│   │       ├── flannel
│   │       ├── host-local
│   │       ├── ipvlan
│   │       ├── loopback
│   │       ├── macvlan
│   │       ├── portmap
│   │       ├── ptp
│   │       ├── sample
│   │       ├── tuning
│   │       └── vlan
│   └── cri-containerd
│       └── cluster
│           ├── env
│           ├── gce
│           │   ├── cloud-init
│           │   │   ├── master.yaml
│           │   │   └── node.yaml
│           │   └── configure.sh
│           └── health-monitor.sh
└── usr
    └── local
        ├── bin
        │   ├── containerd
        │   ├── containerd-release
        │   ├── containerd-shim
        │   ├── containerd-stress
        │   ├── cri-containerd
        │   ├── crictl
        │   └── ctr
        └── sbin
            └── runc
``

File: cluster/gce/configure.sh [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+set -o pipefail
+
+# CRI_CONTAINERD_HOME is the directory for cri-containerd.
+CRI_CONTAINERD_HOME="/home/cri-containerd"
+cd "${CRI_CONTAINERD_HOME}"
+
+# fetch_metadata fetches metadata from GCE metadata server.
+# Var set:
+# 1. Metadata key: key of the metadata.
+fetch_metadata() {
```
miaoyq: Could we reuse the `fetch_metadata` and other duplicate codes of [test/configure.sh](https://github.com/kubernetes-incubator/cri-containerd/blob/master/test/configure.sh)?

cluster/health-monitor.sh
```diff
@@ -0,0 +1,42 @@
...
+# CHECK_PERIOD is the health check period.
+CHECK_PERIOD=${CHECK_PERIOD:-10}
+# SLEEP_SECONDS is the time to sleep after killing cri-containerd
+# and containerd.
+SLEEP_SECONDS=${SLEEP_SECONDS:-120}
+
+while true; do
+  # Use crictl sandboxes because it requires both containerd and
+  # cri-containerd to be working.
+  if ! timeout ${COMMAND_TIMEOUT} ${CRICTL} sandboxes > /dev/null; then
```
mikebrow: prefer using `version` for health monitor over list `sandboxes`.. could be 100pods.  Version should do the same function but with less memory usage.

mikebrow: That said this is a fairly expensive loop, every ten seconds creates a new instance of crictl, does all the init work of connecting to cri-containerd, grabs a list of all the sandboxes... go to sleep 10sec.

Let's open an issue with CRI for a health check api.. Off the top of my head here's how I would do it:  I'd add a ping / pong service api to the CRI image and runtime services.  I'd add a health check command to crictl that would "ping" the named CRI service every n-seconds, and return with a timeout error when / if the pong is not received before the time out. I'd rather see the pings coming from kubelet than a crictl loop, but I can see that kubelet would not necessarily know how to init cri-containerd/containerd. That's another thing I'd like to fix :-) IMO there should be some way for kubelet to manage the CRI runtime engines.

cluster/health-monitor.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# CRICTL is the path of crictl
+CRICTL=${CRICTL:-"crictl"}
+# COMMAND_TIMEOUT is the timeout for the health check command.
+COMMAND_TIMEOUT=${COMMAND_TIMEOUT:-60}
+# CHECK_PERIOD is the health check period.
+CHECK_PERIOD=${CHECK_PERIOD:-10}
+# SLEEP_SECONDS is the time to sleep after killing cri-containerd
+# and containerd.
+SLEEP_SECONDS=${SLEEP_SECONDS:-120}
```
mikebrow: 120 is a long time for the node to be down.

cluster/health-monitor.sh
miaoyq: >prefer using version for health monitor over list sandboxes.. could be 100pods. Version should do the same function but with less memory usage.

`version` can not seem to prove whether `containerd` is working normally or not. I think it better to add a health check api for CRI. :)

mikebrow: Yeah not sure if the list of pods is proof either :-) A health check api would leave it up to the daemon to pong or otherwise notify via some eventing system when it believes it's important functions are still working which could be based on internal state flags (time stamp values) or whatever health system the daemon wants to use.

cluster/gce/configure.sh
Random-Liu: It's a bit hard. :(

`configure.sh` is passed to the node through GCE metadata. If we have multiple scripts for the initialization, we need to package most of them into the release tar, and refactor `configure.sh`.

I prefer doing that in the future. :) I'll file an issue for this.

cluster/health-monitor.sh
Random-Liu: @mikebrow We do have a `Status` and a `Version` function in CRI. However, it's not sufficient to check whether the cri-containerd/containerd is healthy.

For docker we usually encounter the issue that `docker version` still works, but `docker ps` stucks forever. That's why we are using `docker ps` in the current health monitor. https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/health-monitor.sh#L29

Kubelet does list every second, and we do have benchmark for this behavior, the overhead is acceptable.

Random-Liu: Filed an issue https://github.com/kubernetes-incubator/cri-containerd/issues/491 for this.

PR #496, opened 2017-12-13
[WIP] adds option for setting oci hooks

mikebrow:
To address issue containerd/containerd#6645 I've started a prototype.

User must generate a json file for the hooks struct: https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L114-L130

Hooks explained here: https://github.com/opencontainers/runtime-spec/blob/master/config.md#prestart

### DIFF
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -96,6 +96,9 @@ type Config struct {
 	ProfilingPort string `toml:"profiling_port" json:"profilingPort,omitempty"`
 	// ProfilingAddress is address for profiling via host:port/debug/pprof/
 	ProfilingAddress string `toml:"profiling_addr" json:"profilingAddress,omitempty"`
+	// OCIHooks (optional) is a path to a json file that specifies the OCI
+	// spec Hooks struct.
+	OCIHooks string `toml:"oci_hooks" json:"ociHooks,omitempty"`
 }
 
 // CRIContainerdOptions contains cri-containerd command line and toml options.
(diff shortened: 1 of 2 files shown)


Random-Liu:
@mikebrow Why do we need this feature at cri-containerd level? This should be driven by CRI, right?

mikebrow:
> @mikebrow Why do we need this feature at cri-containerd level? This should be driven by CRI, right?

Yes, it should be driven by CRI. I put it here for prototyping only.

lizrice:
We'd like to be able to configure prestart hooks, and this (or something like this) would give us just what we need. Our case is to hook in a security tool on prestart; our users would have the same hook configured on every node (or at least on every node where they want the tool to run), and the same hook would run for all containers. 

It would be nice to be able to add the hook without a restart, but not essential.

egernst:
@mikebrow @lizrice @Random-Liu  -- I am hopeful that there's some traction/forward progress here.  Any insights on how to drive this forward in containerd?

krsna1729:
Need this functionality to do device provisioning at prestart and poststop stage of container lifecycle. Only cri-o seems to be supporting. It would be great if its part of CRI, or atleast containerd supports.

alban:
I mentioned this issue for discussion in a previous [OCI meeting](https://groups.google.com/a/opencontainers.org/forum/#!forum/dev) ([notes](https://hackmd.io/El8Dd2xrTlCaCG59ns5cwg?view#July-24-2019), [slides](https://docs.google.com/presentation/d/1i8csKAf15j3ZDeHxuUlHBDvHiBI44_UxATdaaAy-pjE/edit)) and I would like to mention it at the next [Kubernetes SIG Node](https://github.com/kubernetes/community/tree/master/sig-node) meeting ([notes](https://docs.google.com/document/d/1Ne57gvidMEWXR70OxxnRkYquAoMpt56o75oZtg-OeBg/edit#heading=h.n6dc6gp6gik8)).

mikebrow:
moving PR to PR containerd/cri#1248 as this branch has been deleted...

PR #498, opened 2017-12-13
Add OS and arch in release tarball.

Random-Liu:
This PR:
1) Add OS and arch in release tarball name.
2) Generate sha1sum for release tarball.
3) Update version to 1.0.0-beta.0 to prepare for the release.

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -63,7 +63,8 @@ CONTAINERD_PKG=github.com/containerd/containerd
 CRITOOL_PKG=github.com/kubernetes-incubator/cri-tools
 
 # Create a temporary GOPATH for make install.deps.
-GOPATH=$(mktemp -d /tmp/cri-containerd-install-deps.XXXX)
+TMPGOPATH=$(mktemp -d /tmp/cri-containerd-install-deps.XXXX)
+GOPATH=${TMPGOPATH}
 
 # checkout_repo checks out specified repository
 # and switch to specified  version.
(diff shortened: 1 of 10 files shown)


hack/install-deps.sh
```diff
@@ -146,5 +146,6 @@ ${sudo} bash -c 'cat >'${CRICTL_CONFIG_DIR}'/crictl.yaml <<EOF
 runtime-endpoint: /var/run/cri-containerd.sock
 EOF'
 
-# Clean the tmp GOPATH dir
-${sudo} rm -r ${GOPATH}
+# Clean the tmp GOPATH dir. Use sudo because runc build generates
+# some privileged files.
+sudo rm -r ${GOPATH}
```
mikebrow: did you mean to remove {} from here and above?

Random-Liu: Added `${sudo}` back, use `-rf` instead.

hack/install-deps.sh
```diff
@@ -146,5 +146,6 @@ ${sudo} bash -c 'cat >'${CRICTL_CONFIG_DIR}'/crictl.yaml <<EOF
 runtime-endpoint: /var/run/cri-containerd.sock
 EOF'
 
-# Clean the tmp GOPATH dir
-${sudo} rm -r ${GOPATH}
+# Clean the tmp GOPATH dir. Use sudo because runc build generates
+# some privileged files.
+${sudo} rm -rf ${GOPATH}
```
mikebrow: scary, but better

Random-Liu: Is this better?

mikebrow: Much!

Issue #385, opened 2017-11-02
Improve and document `hack/update-vendor.sh`


Random-Liu:
We should improve `hack/update-vendor.sh` to compare `vendor.conf` with `hack/versions`. We always forgot to update `hack/versions`. :)

We should also document `hack/update-vendor.sh` in https://github.com/kubernetes-incubator/cri-containerd/blob/master/CONTRIBUTING.md to let people use it.

Random-Liu:
This is done in #502. Close

PR #502, opened 2017-12-14
Compare vendor with hack/versions and update hack/versions

miaoyq:
Part of #385

Compare vendor with hack/versions, update hack/versions if the commit id different from vendor.conf.
If OK, I will update `CONTRIBUTING.md`.

### DIFF
--- a/hack/update-vendor.sh
+++ b/hack/update-vendor.sh
@@ -18,15 +18,37 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# This is a temporary hack, rewrite all `github.com/Sirupsen/logrus` to
-# lower case.
-# TODO(random-liu): Remove this after #106 is resolved.
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 cd ${ROOT}
 
 echo "Sort vendor.conf..."
 sort vendor.conf -o vendor.conf
 
-# TODO(random-liu): Compare vendor with hack/versions.
+echo "Compare vendor with hack/versions..."
+need_update=false
+declare -A map=()
+map["RUNC_VERSION"]="github.com/opencontainers/runc"
+map["CNI_VERSION"]="github.com/containernetworking/cni"
+map["CONTAINERD_VERSION"]="github.com/containerd/containerd"
+map["KUBERNETES_VERSION"]="k8s.io/kubernetes"
+for key in ${!map[@]}
+do
+  vendor_commitid=$(grep ${map[${key}]} vendor.conf | awk '{print $2}')
+  version_commitid=$(grep ${key} hack/versions | awk -F "=" '{print $2}')
+  if [ ${vendor_commitid} != ${version_commitid} ]; then
+    if [ $# -gt 0 ] && [ ${1} = "-only-verify" ]; then
+      need_update=true
+      echo "Need to update the value of ${key} from ${version_commitid} to ${vendor_commitid}."
+    else
+      echo "Updating the value of ${key} from ${version_commitid} to ${vendor_commitid}."
+      sed -i "s/${version_commitid}/${vendor_commitid}/g" hack/versions
+    fi
+  fi
+done
+
+if [ ${need_update} = true ]; then
+  echo "Please update \"hack/versions\" by executing \"hack/update-vendor.sh\"!"
+  exit 1
+fi
 
 echo "Please commit the change made by this file..."
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ help:
 	@echo " * 'uninstall'        - Remove installed binaries from system locations"
 	@echo " * 'version'          - Print current cri-containerd release version"
 
-verify: lint gofmt boiler
+verify: lint gofmt boiler deps-version
 
 version:
 	@echo $(VERSION)
@@ -72,6 +72,10 @@ gofmt:
 boiler:
 	@echo "checking boilerplate"
 	@./hack/verify-boilerplate.sh
+
+deps-version:
+	@echo "checking /hack/versions"
+	@./hack/update-vendor.sh -only-verify
 
 $(BUILD_DIR)/cri-containerd: $(SOURCES)
 	$(GO) build -o $@ \


hack/update-vendor.sh
```diff
@@ -18,15 +18,26 @@ set -o errexit
...
+map["RUNC_VERSION"]="github.com/opencontainers/runc"
+map["CNI_VERSION"]="github.com/containernetworking/cni"
+map["CONTAINERD_VERSION"]="github.com/containerd/containerd"
+map["KUBERNETES_VERSION"]="k8s.io/kubernetes"
+for key in ${!map[@]}
+do
+  vendor_commitid=$(grep ${map[${key}]} vendor.conf | awk '{print $2;}')
+  version_commitid=$(grep ${key} hack/versions | awk -F "=" '{print $2;}')
+  if [ ${vendor_commitid} != ${version_commitid} ]; then
+    echo "Update the value of ${key} from ${version_commitid} to ${vendor_commitid}."
```
mikebrow: /s/Update/Updating/

hack/update-vendor.sh
```diff
@@ -18,15 +18,26 @@ set -o errexit
...
+echo "Compare vendor with hack/versions and update hack/versions..."
+declare -A map=()
+map["RUNC_VERSION"]="github.com/opencontainers/runc"
+map["CNI_VERSION"]="github.com/containernetworking/cni"
+map["CONTAINERD_VERSION"]="github.com/containerd/containerd"
+map["KUBERNETES_VERSION"]="k8s.io/kubernetes"
+for key in ${!map[@]}
+do
+  vendor_commitid=$(grep ${map[${key}]} vendor.conf | awk '{print $2;}')
+  version_commitid=$(grep ${key} hack/versions | awk -F "=" '{print $2;}')
```
mikebrow: are the semicolons necessary here?

mikebrow:
I sort of prefer (2) above but (1) would also solve the issue.. wdyt?

Random-Liu:
@mikebrow The problem with 2) is that cri-tools is not our dependency, it's not in vendor.conf.

mikebrow:
@Random-Liu that's not really a problem in so far as at that point hack/versions would only have cri-tools in it and would only be specified in one location.

Random-Liu:
> @Random-Liu that's not really a problem in so far as at that point hack/versions would only have cri-tools in it and would only be specified in one location.

In our document, we have some link to `hack/versions` to tell users what version of the dependencies we are using, which I think is also useful. :)

LGTM overall.
But as @mikebrow suggested, we should verify this in `make verify`.
You can add an option to the script `-w` or something, and format a temporary vendor.conf
* If `-w` is set, we should rewrite the vendor.conf file, just overwrite with the temporary vendor.conf.
* If `-w` is not set, we should only check whether temporary vendor.conf is the same with vendor.conf.

`-w` is just an example, you can use whichever flag or env you think is more convenient.

hack/update-vendor.sh
miaoyq: No error, but I will remove it. :)

yanxuean:
>In our document, we have some link to hack/versions to tell users what version of the dependencies we are using, which I think is also useful. :)

We may link the link to vndr.conf.  The document only state the dependency of cni, containerd and runc, not include cri-tool.

What do you think we add a function into hack/versions?  The function will get version from vndr.conf to set environment.  Of course it keep  cri-tools from modifying.
Because the hack/version also be a script, it only set some environment variable.

Random-Liu:
@yanxuean `vendor.conf` has a lot of dependencies. :( It's hard to figure out which and which.

yanxuean:
I think the user don't really concern the version of runC etc. They pay more attention to that if the cri-container  will work. 
For developer , we only want to modify it in one place.

miaoyq:
>But as @mikebrow suggested, we should verify this in make verify.
You can add an option to the script -w or something, and format a temporary vendor.conf
>
>If -w is set, we should rewrite the vendor.conf file, just overwrite with the temporary vendor.conf.
If -w is not set, we should only check whether temporary vendor.conf is the same with vendor.conf.
-w is just an example, you can use whichever flag or env you think is more convenient.

I don't understand this clearly.
why do we need to format a temporary vendor.conf?
@Random-Liu

Random-Liu:
@miaoyq We can work on the original vendor.conf, too. As long as we return non-zero exit when vendor is changed, it should be fine.

However, it feels to me that verification should not change the file. :)

miaoyq:
@Random-Liu  This pr is updating the contents of `hack/versions` according to `vendor.conf`, 
Do you mean we should update `vendor.conf` according to `hack/versions`?

miaoyq:
@Random-Liu I will add the verification to `make verify`.

Random-Liu:
@miaoyq Sorry, I thought you were doing the other direction.

Yeah, we need to verify that in our CI so that we won't forget changing that in the future.

miaoyq:
@Random-Liu Make senses, will do. :)

> @Random-Liu that's not really a problem in so far as at that point hack/versions would only have cri-tools in it and would only be specified in one location.

What @mikebrow  means is that `hack/versions` only contains the version of `cri-tools` and the versions of others is get from `vendor.conf` directly, right?  I think it is also a good solution, we can implement it later if necessary. :-)

hack/update-vendor.sh
miaoyq: Done.

hack/update-vendor.sh
miaoyq: Done.

miaoyq:
```
$ hack/update-vendor.sh -only-verify
Sort vendor.conf...
Compare vendor with hack/versions...
Need to update the value of KUBERNETES_VERSION from 164317879bcd81eddws0b97e5ebf1c8df041770f2ff1b to 164317879bcd810b97e5ebf1c8df041770f2ff1b.
Please update "hack/versions" by executing "hack/update-vendor.sh"!
```

```
$ hack/update-vendor.sh                                   
Sort vendor.conf...
Compare vendor with hack/versions...
Updating the value of KUBERNETES_VERSION from 164317879bcfdsfdd810b97e5ebf1c8df041770f2ff1b to 164317879bcd810b97e5ebf1c8df041770f2ff1b.
Please commit the change made by this file...
```

Random-Liu:
Nice!

Issue #509, opened 2017-12-18
open /dev/disk/by-uuid: no such file or directory


tkellen:
Hello,

First, thanks for all your great work on this project!

I am attempting to run cri-containerd v1.0.0-beta.0 on [Scaleway](http://scaleway.com) Ubuntu machines and I'm hitting this:

```root@scaleway-node-1:/etc/systemd/system# /usr/local/bin/cri-containerd --logtostderr
I1218 17:54:40.732642    5698 cri_containerd.go:100] Run cri-containerd &{Config:{ContainerdConfig:{RootDir:/var/lib/containerd Snapshotter:overlayfs Endpoint:/run/containerd/containerd.sock Runtime:io.containerd.runtime.v1.linux RuntimeEngine: RuntimeRoot:} CniConfig:{NetworkPluginBinDir:/opt/cni/bin NetworkPluginConfDir:/etc/cni/net.d} SocketPath:/var/run/cri-containerd.sock RootDir:/var/lib/cri-containerd StreamServerAddress: StreamServerPort:10010 CgroupPath: EnableSelinux:false SandboxImage:gcr.io/google_containers/pause:3.0 StatsCollectPeriod:10 SystemdCgroup:false OOMScore:-999 EnableProfiling:true ProfilingPort:10011 ProfilingAddress:127.0.0.1} ConfigFilePath:/etc/cri-containerd/config.toml}
Error: failed to create CRI containerd service: failed to get imagefs uuid of "/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs": open /dev/disk/by-uuid: no such file or directory
```

...any insight on how to proceed would be greatly appreciated!

Random-Liu:
Are you using `udev` or `mdev`?

Random-Liu:
I filed https://github.com/kubernetes-incubator/cri-containerd/pull/510 for now to unblock people from trying it out.

I'll work with @yujuhong and the other team members to figure out whether we should change CRI to make this more portable.

Random-Liu:
@tkellen Thanks for filing the issue. I've filed https://github.com/kubernetes/kubernetes/issues/57356 to carry on the related api design discussion.

tkellen:
[...]

Issue #399, opened 2017-11-06
/dev/disk/by-uuid not present in lxd container


dhain:
Trying to run cri-containerd in a lxd container, and it refuses to start because /dev/disk/by-uuid is not found:

```
Nov 06 16:46:21 k8s1 cri-containerd[6439]: I1106 16:46:21.612009    6439 cri_containerd.go:130] Run cri-containerd &{Config:{ContainerdConfig:{RootDir:/var/lib/containerd Snapshotter:overlayfs Endpoint:/run/containerd/containerd.sock Runtime:io.containerd.runtime.v1.linux RuntimeEngine: RuntimeRoot:} CniConfig:{NetworkPluginBinDir:/opt/cni/bin NetworkPluginConfDir:/etc/cni/net.d} SocketPath:/var/run/cri-containerd.sock RootDir:/var/lib/cri-containerd StreamServerAddress: StreamServerPort:10010 CgroupPath: EnableSelinux:false SandboxImage:gcr.io/google_containers/pause:3.0 StatsCollectPeriod:10 SystemdCgroup:false OOMScore:-999} ConfigFilePath:/etc/cri-containerd/config.toml}
Nov 06 16:46:21 k8s1 cri-containerd[6439]: Error: failed to create CRI containerd service: failed to get imagefs uuid of "/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs": open /dev/disk/by-uuid: no such file or directory
```

This is a privileged container, and the overlay module is loaded, if that's relevant.

dhain:
```
$ cri-containerd version
1.0.0-alpha.1
```

mikebrow:
Is your root partition mounted by UUID?

dhain:
Here's the fstab on the host. It's just a digitalocean droplet:

```
LABEL=cloudimg-rootfs	/	 ext4	defaults	0 0
```

mikebrow:
thx.. could you try to mark your lxd container as unprivileged and see if the lxd defaults are successful in that case?

Random-Liu:
@dhain Linuxkit had similar problem. /cc @ijc 

1) Does your host have `/dev/disk/by-uuid`?
2) Does LXD mount `/dev` for privileged container?

dhain:
@mikebrow I retried using an unprivileged container. Same result.
@Random-Liu host *does* have it, and here's my mount output in the container (looks like it mounts specific entries, but not the whole enchilada):

```
[...]

PR #510, opened 2017-12-18
Do not error out if uuid is not found.

Random-Liu:
Current uuid handling logic is not perfect. Some known issues have been reported such as https://github.com/kubernetes-incubator/cri-containerd/issues/509, https://github.com/kubernetes-incubator/cri-containerd/issues/399 and https://github.com/kubernetes-incubator/cri-containerd/issues/325.

When this doesn't work, kubelet won't be able to get image filesystem capacity and usage. However, we should not keep user from trying cri-containerd out because of this.

Add a flag to skip this logic, and I'll figure out how we could make this more portable across different environments.

### DIFF
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -96,6 +96,9 @@ type Config struct {
 	ProfilingPort string `toml:"profiling_port" json:"profilingPort,omitempty"`
 	// ProfilingAddress is address for profiling via host:port/debug/pprof/
 	ProfilingAddress string `toml:"profiling_addr" json:"profilingAddress,omitempty"`
+	// SkipImageFSUUID skips retrieving imagefs uuid.
+	// TODO(random-liu): Remove this after we find a generic way to get imagefs uuid.
+	SkipImageFSUUID bool `toml:"skip_imagefs_uuid" json:"skipImageFSUUID,omitempty"`
 }
 
 // CRIContainerdOptions contains cri-containerd command line and toml options.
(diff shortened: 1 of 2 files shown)


cmd/cri-containerd/options/options.go
```diff
@@ -96,6 +96,9 @@ type Config struct {
 	ProfilingPort string `toml:"profiling_port" json:"profilingPort,omitempty"`
 	// ProfilingAddress is address for profiling via host:port/debug/pprof/
 	ProfilingAddress string `toml:"profiling_addr" json:"profilingAddress,omitempty"`
+	// SkipImageFSUUID skips retrieving imagefs uuid.
```
mikebrow: /s/./ for new instance of CRIContainerdService./

cmd/cri-containerd/options/options.go
```diff
@@ -158,6 +161,8 @@ func (c *CRIContainerdOptions) AddFlags(fs *pflag.FlagSet) {
 		defaults.ProfilingPort, "Profiling port for web interface host:port/debug/pprof/.")
 	fs.StringVar(&c.ProfilingAddress, "profiling-addr",
 		defaults.ProfilingAddress, "Profiling address for web interface host:port/debug/pprof/.")
+	fs.BoolVar(&c.SkipImageFSUUID, "skip-imagefs-uuid",
+		defaults.SkipImageFSUUID, "Skip retrieving imagefs uuid. When turned on, kubelet can't get imagefs capacity and imagefs disk eviction can't work.")
```
mikebrow: Skip retrieval of imagefs uuid for new CRIContainerdService instances. When turned on, the kubelet will not be able to get imagefs capacity or request imagefs disk eviction.

pkg/server/service.go
```diff
@@ -134,12 +134,14 @@ func NewCRIContainerdService(config options.Config) (CRIContainerdService, error
 		client:             client,
 	}
 
-	imageFSPath := imageFSPath(config.ContainerdConfig.RootDir, config.ContainerdConfig.Snapshotter)
-	c.imageFSUUID, err = c.getDeviceUUID(imageFSPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get imagefs uuid of %q: %v", imageFSPath, err)
+	if !c.config.SkipImageFSUUID {
```
mikebrow: suggest adding an info logf that it has been skipped..

Random-Liu: Will do.

cmd/cri-containerd/options/options.go
Random-Liu: `CRIContainerdService` is code name, I don't think user cares about that in the flag discription. :)

pkg/server/service.go

Issue #517, opened 2018-01-03
Privileged containers fail on CentOS


pires:
Can't run privileged containers with the following set-up:
- CentOS 7.x, `3.10.0-693.2.2.el7.x86_64 #1 SMP Tue Sep 12 22:26:13 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
- Kubernetes 1.9.0
- containerd 1.0.0
- cri-containerd v1.0.0-beta.0

How to reproduce:
```
$ cat <<EOF | kubectl create -f -
kind: Pod
apiVersion: v1
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    securityContext:
      privileged: true
    image: nginx:1.12.2-alpine
EOF
```

What we're seeing:
```
$ kubectl get po
NAME      READY     STATUS              RESTARTS   AGE
nginx     0/1       ContainerCreating   0          4s

$ kubectl describe po nginx
Name:         nginx
Namespace:    default
Node:         ip-10-10-0-246.us-west-2.compute.internal/10.10.0.246
Start Time:   Wed, 03 Jan 2018 15:22:34 +0000
Labels:       <none>
Annotations:  <none>
Status:       Running
IP:           9.0.3.5
Containers:
  nginx:
    Container ID:   cri-containerd://5337661abdf7b4e538dc3f89d3461af9aeb1203e70939982847f11672bc642b3
    Image:          nginx:1.12.2-alpine
    Image ID:       docker.io/library/nginx@sha256:36b0181554913b471ae33546a9c19cc80e97f44ce5e7234995e307f14da57268
    Port:           <none>
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       StartError
      Message:      failed to create containerd task: OCI runtime create failed: container_linux.go:296: starting container process caused "unknown capability \"CAP_AUDIT_READ\"": unknown
      Exit Code:    128
      Started:      Thu, 01 Jan 1970 00:00:00 +0000
      Finished:     Wed, 03 Jan 2018 15:22:40 +0000
    Ready:          False
    Restart Count:  1
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-vtdsp (ro)
Conditions:
  Type           Status
  Initialized    True
  Ready          False
  PodScheduled   True
Volumes:
  default-token-vtdsp:
    Type:        Secret (a volume populated by a Secret)
[...]

PR #518, opened 2018-01-03
Fix privileged caps

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-containerd/issues/517.

We should enable the `HostSpecific` option, so that the generator will apply host specific cap list https://github.com/opencontainers/runtime-tools/blob/master/generate/generate.go#L1023.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -279,7 +279,7 @@ func (c *criContainerdService) generateContainerSpec(id string, sandboxPid uint3
 	if err != nil {
 		return nil, err
 	}
-	g := generate.NewFromSpec(spec)
+	g := newSpecGenerator(spec)
 
 	// Set the relative path to the rootfs of the container from containerd's
 	// pre-defined directory.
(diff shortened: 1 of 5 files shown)


bmcstdio:
@Random-Liu any chance this lands in a tag anytime soon?

pires:
Sweet! Yeah, a new release would definitely unblock us.

Random-Liu:
@bmcstdio @pires I marked this v1.0.0-beta.1. I'll cut the patch release after we collect several bug fixes, probably next week. Is this OK to you?

We don't have a regular release schedule now, mainly because we plan to do a refactoring to make cri-containerd easier for user to use https://github.com/containerd/containerd/issues/1942. However, we'll start matching Kubernetes release after 1.0.0.

Random-Liu:
@pires @bmcstdio It would be super helpful if you could verify whether this fixes your issue. We don't have a test environment for CentOS today.

pires:
@Random-Liu we are doing so and will report tomorrow morning GMT time.

PR #523, opened 2018-01-04
Sync vendors with containerd

Random-Liu:
For https://github.com/containerd/containerd/issues/1942.
To make cri-containerd a containerd plugin, it is important to keep the vendors in sync with containerd.
This PR:
1) Updated `hack/update-vendor.sh` to sync `vendor.conf` with containerd.
2) Run the script and update vendors out of sync.


hack/update-vendor.sh
```diff
@@ -51,4 +48,48 @@ if [ ${need_update} = true ]; then
...
+  if [ -z "${vendor_in_containerd}" ]; then
+    echo ${vendor} >> ${tmp_vendor}
+    continue
+  fi
+  commit_in_containerd=$(echo ${vendor_in_containerd} | awk '{print $2}')
+  if [ "${commit}" != "${commit_in_containerd}" ]; then
+    echo ${vendor_in_containerd} >> ${tmp_vendor}
+  else
+    echo ${vendor} >> ${tmp_vendor}
+  fi
```
mikebrow: needs checks for optional github repo link.. For example: see "https://github.com/golang/sys" below

Random-Liu: If repo is different, I guess commit id will be different. I think commit id is uniquely reflected the commit history.
If the commit id is the same, I think it means the code should also be exactly the same, even they are in different repo. :)

Random-Liu: I can add a comment here. :)

mikebrow: v1.0 type tags are more risky.. also there may be some odd issue of setup. I'd also expect failures but really the issue was more of a why not check that too while checking version in case someone wanted to link to a personal repo fork for a bit.

mikebrow: I agree odds are low but there could be a match and still a difference in link. Another example, we or containerd may have a superfluous link that should be removed.

Random-Liu:
Yeah you are right, I forgot the tag stuff. Will address  your comment.

hack/update-vendor.sh
Random-Liu: Done.

Random-Liu:
@mikebrow Addressed the comments. Good catch!

miaoyq:
@Random-Liu  
Whether there is this scenario: using a repo in cri-containerd, but using the alias of the repo in repo position in containerd?

Random-Liu:
@miaoyq Yeah, this should still work in that case, right?

miaoyq:
What I mean is this scenario like:
cri-containerd: 
```
   golang.org/x/sys xxxxxx  https://github.com/golang/sys
or golang.org/x/sys xxxxxx
```
containerd: 
```
github.com/golang/sys xxxxxx
```

But it's ok, there is no this scenario currently.

Random-Liu:
@miaoyq I mean in this scenario the script still works. `print $3` will return empty for containerd vendor.conf, and `"" != "https://github.com/golang/sys"`

miaoyq:
@Random-Liu I think `if [ -z "${vendor_in_containerd}" ]; ` will be true in this scenario, and the script will continue [here](https://github.com/kubernetes-incubator/cri-containerd/pull/523/files#diff-23bb1a7aca7a27a516bf76059609d473R68)

Random-Liu:
@miaoyq That should not happen I think. `golang.org/x/sys` should be the package imported in the code I think.

If containerd imports `github.com/golang/sys` in the code, and we import `golang.org/x/sys` but alias to `github.com/golang/sys`, or vise versa, we should fix it.

miaoyq:
@Random-Liu  Got it, Thanks for explanation :)
/lgtm

dnephin:
@Random-Liu is there an issue which has more context about:

> it is important to keep the vendors in sync with containerd.

I don't think I understand why this is necessary.

mikebrow:
@dnephin cri-containerd directly calls the containerd client apis, thus any "common" vendor has to be at the same revision or it won't build. We're merging vendor directories so this should go away soon.

dnephin:
> thus any "common" vendor has to be at the same revision or it won't build.

This is not true. The versions must be compatible, but they don't need to be the same revision. Trying to pin vendor to identical revisions should be unnecessary. If a dependency makes a backwards incompatbile change then CI will fail and the dependency can be updated.

mikebrow:
> This is not true. The versions must be compatible, but they don't need to be the same revision.

Yes, in very limited cases such as comment only changes, the revisions don't have to the be same. Personally I don't like chasing down unnecessary bugs due to unforeseen version incompatibilities.

> then CI will fail..

Yes, that was my point, and many times where it fails is golang's requirement that there only be one package of the same name.

> is there an open issue..

I don't think there is one open in containerd/containerd but we do have one in containerd/cri-containerd.  

Here's the PR 
https://github.com/containerd/containerd/pull/2097/files#diff-4061fcef378a6d912e14e2ce162a1995R45

> I think this is a bad idea.

How do you propose to vendor plugins like CRI? I believe the plan was bring it in and remove the extra vendor parts that are not needed over time.  I see you have already opened an issue in containerd/cri-containerd... 594

dnephin:
> Yes, in very limited cases such as comment only changes, 

Even code changes are fine, as long as they are backwards compatible. The "standard practice" for go packages is supposed to be always preserve backwards compatibility, so theoretically it should be the common case that you don't need to match revision. 

> How do you propose to vendor plugins like CRI?

Yes #594 related to this.

I don't think plugins are a special case. It's exactly the same problem as vendoring any other dependency.

[dep](https://github.com/golang/dep) should make this much easier because it separates version constraints from pinned versions, so it's able to figure out the correct version to vendor.

> I believe the plan was bring it in and remove the extra vendor parts that are not needed over time

Reducing the number of dependencies sounds good, maybe I misunderstood your comment about "merging". If it's just managing transient dependencies that's something we can solve with the right dependency management tools.

dnephin:
>  the issue is whether or not there are a plurality of revisions in gopath when you build the binary.

Could you elaborate? I don't think I've ever seen that issue.

For example:

```
# GIT_DIR=/go/src/github.com/dnephin/glog/.git git rev-parse HEAD
e8d701dc44005ce3b4a2a52e22ad92ce56fe8608
# GIT_DIR=/alt/src/github.com/dnephin/glog/.git git rev-parse HEAD
fca8c8854093a154ff1eb580aae10276ad6b1b5f
# GIT_DIR=/go/src/github.com/dnephin/testing/vendor/github.com/dnephin/glog/.git git rev-parse HEAD
b83197c33c35da605bf2be628a4a3ed7695e21d6
# pwd
/go/src/github.com/dnephin/testing
#
# cat main.go
package main

import (
        "github.com/dnephin/glog"
        "fmt"
)

func main() {
        glog.Warning("warn")
        fmt.Println("ran")
}

# GOPATH=/alt:/go go build .
```

A single dependency (`dnephin/glog`) exists in 3 places in the GOPATH (2 in roots, 1 in vendor). Each instance has a different version, but there is no error.

My understanding is that `GOPATH` is searched (starting with `vendor/`) and the first instance of a dependency is selected. It doesn't matter how many copies you have of that dependency.

Are you referring to the issue where a `vendor` version of a dependency is selected instead of the canonical one? That issue is also resolved by dependency management tools by stripping `vendor/` directories of anything inside `vendor`.

Random-Liu:
@dnephin I still think it would be better to keep vendors in sync between containerd and cri-containerd, because:
1) It is a pain to vendor a package which has different dependency version with you, especially when we use `Godep` in Kubernetes. It always takes 1-2 weeks to update cadvisor version in Kubernetes. So we try to avoid updating dependencies in cadvisor or keeping it in sync with Kubernetes.
2) Just compile is not enough for bug fixes. Sometimes we update vendor for bug fixes, the api may not changed, and the binary may still compile, but if the update doesn't reflect on the other side, the bug still exists.
3) Just compile is not enough for RPCs. Use Kubernetes api as an example, maybe the client interface still compiles, but there might still be change in the RPC, e.g. a new option is added. If the Kubernetes client update doesn't reflect on the other side, the binary may still compile, the RPC may also even work, but the new option may have wrong default thus we could not use the new feature which we want to use.

dnephin:
These are all good points, but personally they sound more like arguments for "keep dependencies updated". I would agree this is a good practice.  I'm not sure there is a need to update them to exactly the version used by something else. However it's likely that will happen anyway if both things are updated to the latest version.

I think we'll find that the problem is a lot easier to manage if we approach it with this perspective (keep things updated) instead of trying to sync to a specific version. `dep` should provide the necessary tooling.

mikebrow:
> used by something else

We in cri-containerd had a desire to sync (in one direction) our vendor with containerd. Containerd/containerd/vendor -> kubernetes-incubator/cri-containerd/vendor.  We are moving cri-containerd to containerd/cri-containerd and continue to keep our vendor in sync. This is in part because we vendor the containerd/containerd/* client code which has dependencies on containerd/containerd/vendor.

Yes as long as we do the vendoring right (flatten) there will be only one used. Don't understand the desire for asking the cri code to be on possibly different vendor dependencies than what containerd uses for it's client binary. Again this is a one way sync we're doing here.

Random-Liu:
@dnephin After containerd starts vendoring cri-containerd, it will be hard to sync vendors with containerd. Because we don't know a dependency is introduced by containerd or cri-containerd.

We can do some trick like https://github.com/containerd/cri-containerd/pull/561, or maybe we can just manually check and enforce that. :(

dnephin:
We should look at using `dep`. It should solve this problem and make updates trivial.

"direct dependencies" (packages that are imported by the project directly) are set as constraints in the `Gopkg.toml`. "transitive dependencies" (packages that are imported by other dependencies) are resolved to a correct version by looking at the projects `Gopkg.toml` and any `Gopkg.toml` of the dependencies. Once the correct version is resolved it gets pinned to a `Gopkg.lock` (equivalent of `vendor.conf`). This setup makes it easy to update because you only set constraints for the things you depend on directly (This is how dependencies are handled in many other languages, and in my experience it makes managing dependencies much easier). 

It looks like a lot of work has been done on the dep docs recently: https://golang.github.io/dep/. I'm sure they will do a better job of explaining things. 

Updating a dependency should be as easy as running a single `dep ensure update` command. No manual work required. Of course it won't be that easy at first. It requires that dependencies use proper git tags and semver, but it's possible to work around any packages that aren't.

We shouldn't need to write any custom tooling. This is a problem faced by every single go project. #561 is not a solution because it assumes that only `containerd` has its own dependencies. In reality every dependency has other dependencies. So just using a version picked by `containerd` can still lead to version conflicts.

Random-Liu:
> We should look at using dep. It should solve this problem and make updates trivial.

> "direct dependencies" (packages that are imported by the project directly) are set as constraints in the Gopkg.toml. "transitive dependencies" (packages that are imported by other dependencies) are resolved to a correct version by looking at the projects Gopkg.toml and any Gopkg.toml of the dependencies. Once the correct version is resolved it gets pinned to a Gopkg.lock (equivalent of vendor.conf). This setup makes it easy to update because you only set constraints for the things you depend on directly (This is how dependencies are handled in many other languages, and in my experience it makes managing dependencies much easier).

I didn't look into `dep` before. If that is what `dep` is doing, it looks very useful to me. :)

However, currently most projects we've vendored are not using `dep` I believe. It takes time to switch.

> This is a problem faced by every single go project.

Yeah. It is a pain everywhere. :(

Issue #524, opened 2018-01-04
cri-containerd cannot create a pod if the runtime removes the IP from the veth device


nlacasse:
Some runtimes create a TAP device linked to the virtual ethernet (veth) device, and then remove the IP address from the veth device.  For example, clear containers takes this approach, as documented [here][1].

On each PodSandboxStatus request, cri-containerd runs [ocicni.GetPodNetworkStatus][2], which joins the pod's network namespace and looks for the IP address of the veth device.  If this IP has been removed, GetPodNetworkStatus returns an error, and cri-containerd returns a status with an empty Network.Ip field.  Kubelet detects this, and stops and re-starts the pod, but the same issue will occur again.

Cri-o does not have this problem, since it only calls GetPodNetworkStatus once (before the IP has been removed from the veth device), and [caches the results][3].

[1]: https://github.com/clearcontainers/runtime/blob/master/docs/architecture/architecture.md#networking
[2]: https://github.com/cri-o/ocicni/blob/fc9c77cc55795c09bf1f74b2afd92078a793a69e/pkg/ocicni/ocicni.go#L353
[3]: https://github.com/kubernetes-incubator/cri-o/blob/b9ffd277b9c708dedfa5ccaf2a9f07ee8c8204ab/server/sandbox_run.go#L497

abhi:
@nlacasse I think what you are asking here makes sense. @Random-Liu WDYT ?

Random-Liu:
Since IP of a pod is considered to be immutable, I think it's reasonable to cache the IP after create the sandbox.

However, I'm not sure that runtime should rely on this behavior or not. They are today, but we might want to find a better solution which doesn't make implication on the cri container runtime implementation detail. :)

Random-Liu:
The fix for v1.0.0-beta.0 has been merged, I'll remove this issue from v1.0.0-beta.0 and keep it open.
We can revisit in the future to figure out how to handle this better.

PR #525, opened 2018-01-05
Caching IP allocated by CNI plugin

abhi:
A Possible fix to address scenario mentioned in kubernetes-incubator/cri-containerd#524
Cache the IP allocated for pods not connected to host network and avoid fetch it on every Sandbox status request.

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -120,6 +120,18 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 				}
 			}
 		}()
+		ip, err := c.netPlugin.GetPodNetworkStatus(podNetwork)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get network status for sandbox %q: %v", id, err)
+		}
+		// Certain VM based solutions like clear containers (Issue kubernetes-incubator/cri-containerd#524)
+		//  rely on the assumption that CRI shim will not be querying the network namespace to check the
+		// network states such as IP.
+		// In furture runtime implementation should avoid relying on CRI shim implementation details.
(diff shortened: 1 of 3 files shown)


pkg/store/sandbox/sandbox.go
```diff
@@ -34,6 +34,8 @@ type Sandbox struct {
 	Container containerd.Container
 	// CNI network namespace client
 	NetNS *NetNS
+	// Pod IP if it is attached to non host network
```
Random-Liu: // IP is the ...
Comment should start with the variable name.

pkg/server/sandbox_status.go
```diff
@@ -57,7 +56,7 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
 		processStatus = taskStatus.Status
 	}
 
-	ip, err := c.getIP(sandbox)
+	ip := c.getIP(sandbox)
 	if err != nil {
```
Random-Liu: err checked not needed.

pkg/server/sandbox_run.go
```diff
@@ -120,6 +120,11 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 				}
 			}
 		}()
+		ip, err := c.netPlugin.GetPodNetworkStatus(podNetwork)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get network status for sandbox %q: %v", id, err)
+		}
+		sandbox.IP = ip
```
Random-Liu: Please add comment to explain that some VM based solution is relying on this behavior, and link the issue. And we should try to eliminate the assumption on cri shim implementation detail in the future.

Random-Liu:
Merging @nlacasse.

PR #535, opened 2018-01-11
Remove reexec usage

Random-Liu:
For https://github.com/containerd/containerd/issues/1966, https://github.com/containerd/cri-containerd/issues/238.

This PR:
1) Use `github.com/docker/docker/daemon/graphdriver/copy` instead of `github.com/docker/docker/pkg/chrootarchive`, because we only do directly directory copy, which doesn't have security issue.
2) Update docker to [77a2bc3e5bbc9](https://github.com/moby/moby/commit/77a2bc3e5bbc9be3fe166ed8321b7cd04e7bd097) and update vendor.
3) Add integration test for volume copy up.

/cc @cpuguy83 @dnephin @ijc @stevvooe 

@dnephin I do have one question about `copy.DirCopy`. We don't need to copy `xattr`, right?


cpuguy83:
Would it make sense to use https://godoc.org/github.com/containerd/containerd/fs#CopyDir instead?
The docker package is currently linux only and the containerd version has the same optimizations for Linux.

dnephin:
I think we probably do want the xattrs. It looks like `containerd/fs.CopyDir()` supports them as well.

pkg/containerd/opts/container.go
```diff
@@ -122,7 +122,7 @@ func copyExistingContents(source, destination string) error {
 			return errors.Errorf("volume at %q is not initially empty", destination)
 		}
 
-		if err := chrootarchive.NewArchiver(nil).CopyWithTar(source, destination); err != nil {
+		if err := copy.DirCopy(source, destination, copy.Content, false); err != nil {
```
dnephin: I believe we can remove `copyOwnership()` (just below this), it's already done by `fs.CopyDir()`

Random-Liu:
> Would it make sense to use https://godoc.org/github.com/containerd/containerd/fs#CopyDir instead?
The docker package is currently linux only and the containerd version has the same optimizations for Linux.

Cool. Don't know containerd has such a util function, will use it.

pkg/containerd/opts/container.go
Random-Liu: Done.

Random-Liu:
Yeah! I need to make my containerd membership public to let the k8s-ci-bot know it.

Random-Liu:
@dnephin Are you ok with the new change?

Random-Liu:
Thanks for reviewing! @cpuguy83 @dnephin

Random-Liu:
Apply LGTM based on https://github.com/containerd/cri-containerd/pull/535#issuecomment-357116269 and  https://github.com/containerd/cri-containerd/pull/535#pullrequestreview-88357029.


/lgtm

PR #537, opened 2018-01-11
Pass PATH for containerd sudo make install.

Random-Liu:
containerd `make install` requires `go` to check OS and ARCH. However, `sudo` doesn't inherit envs usually, so `sudo make install` could not find `go` in `PATH`.

Partially unblock https://github.com/containerd/cri-containerd/pull/485.

@dnephin 


stevvooe:
Odd, I thought we fixed this.

PR #538, opened 2018-01-11
Change hack/update-vendor.sh to sync again with new vendor

Random-Liu:
Update `hack/version` again after update vendor.conf.

Sometimes containerd update also updates the runc version, this could automate that.


Random-Liu:
@mikebrow Yeah, maybe we should do that. Can you file an issue?

mikebrow:
Issue filed... #540

PR #541, opened 2018-01-12
Skip node e2e test in travis.

Random-Liu:
Node e2e test always timeout in travis, because for free account travis has 50min timeout (see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20308)).

Since we've already run node e2e as presubmit and CI in Kubernetes test infra, we don't need to run it in travis now.

I don't remove the code for now, because we may want to reuse it, e.g. in containerd repo in the future.


abhi:
will this be misleading ? Coz on the checks it shows as succeeded. We can remove it and revert it based on history.

Random-Liu:
@abhi I'm fine with removing it. Will do it then.

Random-Liu:
@abhi Done.

Random-Liu:
> Sorry to see it go as it's nice to run these in as many environments as possible. Maybe open up an issue to to split up the node e2e bucket on travis into two parallel test runs?

Will do.

PR #547, opened 2018-01-17
glog to logrus

Random-Liu:
Use `logrus` instead of `glog` to make `cri-containerd` log output consistent with `containerd`.

The log level mapping:
1) glog.Error[f] => logrus.Error[f]
2) glog.Warning[f] => logrus.Warn[f]
3) glog.V(0/2).Info[f] => logrus.Info[f]
4) glog.V(4).Info[f] => logrus.Debug[f]
5) glog.V(5).Info[f] => log.Trace[f]. logrus doesn't support `trace` level, we added a util package `pkg/log` to do that. https://github.com/sirupsen/logrus/issues/663

Note that instead of `": %v", err`, we'll use `logrus.WithError(err)` in the future.

@mikebrow @abhi @yanxuean @miaoyq

### DIFF
--- a/cmd/cri-containerd/cri_containerd.go
+++ b/cmd/cri-containerd/cri_containerd.go
@@ -27,12 +27,13 @@ import (
 	"runtime"
 	"syscall"
 
-	"github.com/golang/glog"
 	"github.com/opencontainers/selinux/go-selinux"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"k8s.io/kubernetes/pkg/util/interrupt"
 
 	"github.com/containerd/cri-containerd/cmd/cri-containerd/options"
+	"github.com/containerd/cri-containerd/pkg/log"
 	"github.com/containerd/cri-containerd/pkg/server"
 	"github.com/containerd/cri-containerd/pkg/version"
 )
@@ -51,11 +52,6 @@ var cmd = &cobra.Command{
 	Use:   "cri-containerd",
 	Short: "A containerd based Kubernetes CRI implementation.",
 	Long:  desc,
-}
-
-// Add golang flags as persistent flags.
-func init() {
-	cmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
 }
 
 func defaultConfigCommand() *cobra.Command {
@@ -93,15 +89,19 @@ func main() {
 		}
 		validateConfig(o)
 
-		glog.V(0).Infof("Run cri-containerd %+v", o)
+		if err := setLogLevel(o.LogLevel); err != nil {
+			return fmt.Errorf("failed to set log level: %v", err)
+		}
+
+		logrus.Infof("Run cri-containerd %+v", o)
 
 		// Start profiling server if enable.
 		if o.EnableProfiling {
-			glog.V(2).Info("Start profiling server")
+			logrus.Info("Start profiling server")
 			go startProfilingServer(o.ProfilingAddress, o.ProfilingPort)
 		}
 
-		glog.V(2).Infof("Run cri-containerd grpc server on socket %q", o.SocketPath)
+		logrus.Infof("Run cri-containerd grpc server on socket %q", o.SocketPath)
 		s, err := server.NewCRIContainerdService(o.Config)
 		if err != nil {
 			return fmt.Errorf("failed to create CRI containerd service: %v", err)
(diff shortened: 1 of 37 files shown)


cmd/cri-containerd/cri_containerd.go
```diff
@@ -93,15 +94,19 @@ func main() {
 		}
 		validateConfig(o)
 
-		glog.V(0).Infof("Run cri-containerd %+v", o)
+		if err := setLogLevel(o.LogLevel); err != nil {
+			return fmt.Errorf("failed to set log level: %v", err)
+		}
+
+		logrus.Infof("Run cri-containerd %+v", o)
```
yanxuean: before setLogLevel？

cmd/cri-containerd/options/options.go
```diff
@@ -105,8 +105,10 @@ type Config struct {
 type CRIContainerdOptions struct {
 	// Config contains cri-containerd toml config
 	Config
-	// Path to the TOML config file.
+	// ConfigFilePath is the path to the TOML config file.
 	ConfigFilePath string `toml:"-"`
+	// LogLevel is the logrus log level.
+	LogLevel string `toml:"-"`
```
yanxuean: Can't we config it in configfile?

pkg/server/container_attach.go
```diff
@@ -62,7 +62,7 @@ func (c *criContainerdService) attachContainer(ctx context.Context, id string, s
 	}
 	handleResizing(resize, func(size remotecommand.TerminalSize) {
 		if err := task.Resize(ctx, uint32(size.Width), uint32(size.Height)); err != nil {
-			glog.Errorf("Failed to resize task %q console: %v", id, err)
+			logrus.WithError(err).Errorf("Failed to resize task %q console", id)
```
yanxuean: Why not use log.G(ctx)?

cmd/cri-containerd/cri_containerd.go
Random-Liu: We want to make sure `setLogLevel` also works for this log line, right?

cmd/cri-containerd/options/options.go
Random-Liu: Hm, we do. Let me change it.

pkg/server/container_attach.go
Random-Liu: Because the log context from containerd is not passed to cri-containerd yet.
It seems that we need to integrate with the [`interceptor`](https://github.com/containerd/containerd/blob/master/server/server.go#L238).

cmd/cri-containerd/options/options.go

Random-Liu:
Added `setGLogLevel` to set glog options based on `log-level`. Fixes https://github.com/containerd/cri-containerd/issues/549.

Random-Liu:
@mikebrow Also added one more commit to remove all glog flags.

cmd/cri-containerd/cri_containerd.go
```diff
@@ -164,6 +164,39 @@ func startProfilingServer(host string, port string) {
...
+	}
+	logrus.SetLevel(lvl)
+	return nil
+}
+
+// TODO(random-liu): Set glog level in plugin mode.
+func setGLogLevel(l logrus.Level) error {
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		return err
+	}
```
mikebrow: alsologtostderr?

cmd/cri-containerd/cri_containerd.go
```diff
@@ -164,6 +164,39 @@ func startProfilingServer(host string, port string) {
...
+		return flag.Set("v", "5")
+	case logrus.DebugLevel:
+		return flag.Set("v", "4")
+	case logrus.InfoLevel:
+		return flag.Set("v", "2")
+	// glog doesn't support following filters.
+	case logrus.WarnLevel:
+	case logrus.ErrorLevel:
+	case logrus.FatalLevel:
+	case logrus.PanicLevel:
```
mikebrow: maybe "0" for these?

Random-Liu: Set "0", means V(0) will still be printed, which is not what we want.

cmd/cri-containerd/cri_containerd.go
Random-Liu: We don't want to log to files, I think. If we want that someday, we can change to `alsologtostderr`.

PR #552, opened 2018-01-18
Use containerd grpc server

Random-Liu:
Let CRI plugin use containerd grpc server.

Previously we had a circular dependency problem:
1) cri plugin becomes one of containerd grpc service.
2) cri plugin itself replies on containerd grpc services to do initialization.

This PR split the CRI GRPC service creation and initialization.
During plugin regstration, cri plugin will:
1) Create the grpc service with `Register` method and return to containerd;
2) Start a goroutine to do initialization asynchronously.
3) All GRPC service request handlers need to return error before initialization is finished.

In this way, we can make cri one of containerd grpc services, and make sure CRI only starts serving request when initialization is done.

The only downside is that after containerd starts, CRI might not be available immediately.
However, we should never assume that grpc service starts serving when containerd process is started. And kubelet can tolerant that.

/cc @containerd/containerd-maintainers

### DIFF
--- a/cmd/cri-containerd/cri_containerd.go
+++ b/cmd/cri-containerd/cri_containerd.go
@@ -110,8 +110,8 @@ func main() {
 		// Pass in non-empty final function to avoid os.Exit(1). We expect `Run`
 		// to return itself.
 		h := interrupt.New(func(os.Signal) {}, s.Stop)
-		if err := h.Run(func() error { return s.Run() }); err != nil {
-			return fmt.Errorf("failed to run cri-containerd grpc server: %v", err)
+		if err := h.Run(func() error { return s.Run(true) }); err != nil {
+			return fmt.Errorf("failed to run cri-containerd with grpc server: %v", err)
 		}
 		return nil
 	}
--- a/cri.go
+++ b/cri.go
@@ -19,11 +19,13 @@ package cri
 import (
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/plugin"
+	"github.com/pkg/errors"
 
 	"github.com/containerd/cri-containerd/cmd/cri-containerd/options"
 	"github.com/containerd/cri-containerd/pkg/server"
 )
 
+// TODO(random-liu): Use github.com/pkg/errors for our errors.
 // Register CRI service plugin
 func init() {
 	plugin.Register(&plugin.Registration{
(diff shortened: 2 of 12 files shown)


pkg/server/service.go
```diff
@@ -156,30 +162,56 @@ func NewCRIContainerdService(config options.Config) (CRIContainerdService, error
...
+	runtime.RegisterImageServiceServer(s, instrumented)
+	api.RegisterCRIContainerdServiceServer(s, instrumented)
+	return nil
 }
 
-// Run starts the cri-containerd service.
-func (c *criContainerdService) Run() error {
+// Run starts the cri-containerd service. startGRPC specifies
+// whether to start grpc server in this function.
+func (c *criContainerdService) Run(startGRPC bool) error {
```
mikebrow: todo remove "startGRPC = true" case when we no longer support cri-containerd stand alone mode

pkg/server/service.go
```diff
@@ -200,24 +232,32 @@ func (c *criContainerdService) Run() error {
...
 		}
-		close(grpcServerCloseCh)
-	}()
+		l, err := net.Listen(unixProtocol, c.config.SocketPath)
+		if err != nil {
+			return fmt.Errorf("failed to listen on %q: %v", c.config.SocketPath, err)
+		}
+		go func() {
+			if err := c.server.Serve(l); err != nil {
+				logrus.WithError(err).Error("Failed to serve grpc grpc request")
```
mikebrow: grpc grpc request?

cri.go
```diff
@@ -48,24 +50,24 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
...
 	// cri service requires containerd to be running.
-	// TODO(random-liu): Resolve the circular dependency.
 	go func() {
-		s, err := server.NewCRIContainerdService(c)
-		if err != nil {
-			log.G(ctx).WithError(err).Fatal("Failed to create CRI service")
-		}
-		if err := s.Run(); err != nil {
+		if err := s.Run(false); err != nil {
 			log.G(ctx).WithError(err).Fatal("Failed to run CRI grpc server")
```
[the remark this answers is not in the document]
Random-Liu: Yeah, sorry, that was old log message, will update.

pkg/server/service.go
Random-Liu: Will do.

pkg/server/service.go
Random-Liu: Will fix.

cmd/cri-containerd/cri_containerd.go
```diff
@@ -110,7 +110,7 @@ func main() {
 		// Pass in non-empty final function to avoid os.Exit(1). We expect `Run`
 		// to return itself.
 		h := interrupt.New(func(os.Signal) {}, s.Stop)
-		if err := h.Run(func() error { return s.Run() }); err != nil {
+		if err := h.Run(func() error { return s.Run(true) }); err != nil {
 			return fmt.Errorf("failed to run cri-containerd grpc server: %v", err)
```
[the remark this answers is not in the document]
Random-Liu: Done.

cri.go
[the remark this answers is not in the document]
Random-Liu: Done.

pkg/server/service.go

pkg/server/service.go

PR #554, opened 2018-01-18
Minor code cleanup.

Random-Liu:
Minor code cleanup after https://github.com/containerd/cri-containerd/pull/546.
/cc @jcvenegas 

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -17,6 +17,8 @@ limitations under the License.
 package annotations
 
 // ContainerType values
+// Following OCI annotations are used by katacontainers now.
+// We'll switch to standard secure pod API after it is defined in CRI.
 const (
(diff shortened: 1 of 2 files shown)


pkg/annotations/annotations.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package annotations
 
 // ContainerType values
+// Following OCI annotations are used by katacontainers now.
+// We'll switch to standard secure pod API after it is defined in CRI.
```
mikebrow: todo() format pls .. do we have an issue / feature to track?

Random-Liu: That is a long term work which may take several quarters to happen. I don't think we need a TODO here. :)

Of course we'll follow all CRI change.

I put this comment here mainly to emphasize that the annotations are used by katacontainers in case people don't know what these annotations are used for.

PR #555, opened 2018-01-18
Move cgroup oom to cmd

Random-Liu:
Based on #552. Only the last commit is new.

Move `SetOOMScore` and `loadCgroup` into `cmd/cri-containerd`.
For plugin mode, containerd will provide those configurations, so we only need to do them in standalone mode.

### DIFF
--- a/cmd/cri-containerd/cri_containerd.go
+++ b/cmd/cri-containerd/cri_containerd.go
@@ -27,6 +27,9 @@ import (
 	"runtime"
 	"syscall"
 
+	"github.com/containerd/cgroups"
+	"github.com/containerd/containerd/sys"
+	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/cri-containerd/cri_containerd.go
```diff
@@ -200,3 +216,23 @@ func setGLogLevel(l logrus.Level) error {
 	}
 	return nil
 }
+
+// loadCgroup loads the cgroup associated with path if it exists and moves the current process into the cgroup. If the cgroup
+// is not created it is created and returned.
```
mikebrow: If the cgroup status for the path is cgroupdeleted an attempt to create the cgroup is made and if successful the current process is moved into the cgroup.

pkg/server/helpers.go
```diff
@@ -288,26 +285,6 @@ func (c *criContainerdService) ensureImageExists(ctx context.Context, ref string
 	return &newImage, nil
 }
 
-// loadCgroup loads the cgroup associated with path if it exists and moves the current process into the cgroup. If the cgroup
-// is not created it is created and returned.
-func loadCgroup(cgroupPath string) (cgroups.Cgroup, error) {
```
mikebrow: why are we moving this helper function to cri-containerd.go? getting ready to delete it when we remove standalone mode?

pkg/util/atomic_boolean.go
```diff
@@ -0,0 +1,54 @@
+/*
```
dnephin: nit: [Bad package names](https://blog.golang.org/package-names#TOC_5.)

> Avoid meaningless package names. Packages named util, ... provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused.

Maybe this could be in `pkg/atomic/` instead? or

Random-Liu: Hm, will put this into a separate package then. :P

Random-Liu: Done in https://github.com/containerd/cri-containerd/pull/552

cmd/cri-containerd/cri_containerd.go
Random-Liu: Will do.

pkg/server/helpers.go
Random-Liu: We'll have to expose this function as public function if we don't move it.

PR #556, opened 2018-01-18
hack/versions: extract most from vendor.conf

### LINKED ISSUE #540: Modify CI to use vendor.conf vs hack/versions
We should pull versions from vendor.conf where possible instead of from hack/versions. Doing so will allow us to remove all but one of the components in hack/versions. Additionally, we will no longer have to keep the two files in sync.

ijc:
This sets `$what_VERSION` and `$what_REPO` for runc, cni, containerd and
kubernetes based on vendor.conf, removing the need to duplicate things in
hack/versions.

With this `update_hack_versions` becomes redundant so remove it and both calls.

Since CONTAINERD_REPO is now unconditionally set we can also simplify the
fetching of vendor.conf in update-vendor.sh a bit, so do so. Further since
`*_REPO` are now unconditionally set we can support alternative clone paths for
all of these repos by adjusting checkout_repo to make the 3rd argument
non-optional and always passing it. Since `CRITOOL_VERSION` is not coming from
`vendor.conf` (since it is not used from Go code) we manually set
`CRITOOL_REPO` for consistency.

The final wrinkle is that `k8s.io/kubernetes` is has a Go specific redirect in
the form of HTML meta headers returned from https://k8s.io/kubernetes/?go-get=1
which point to the real repo to clone. Parsing that in shell is tricky so just
hardcode that.

Fixes #540.

/cc @mikebrow

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -27,6 +27,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+source $(dirname "${BASH_SOURCE[0]}")/utils.sh
+
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 . ${ROOT}/hack/versions
 
@@ -76,15 +78,12 @@ GOPATH=${TMPGOPATH}
 # and switch to specified  version.
 # Varset:
 # 1) Pkg name;
-# 2) Version.
-# 3) Repo name (optional);
+# 2) Version;
+# 3) Repo name;
 checkout_repo() {
   local -r pkg=$1
   local -r version=$2
-  local repo=${3:-""}
-  if [ -z "${repo}" ]; then
-    repo=${pkg}
-  fi
+  local -r repo=$3
   path="${GOPATH}/src/${pkg}"
   if [ ! -d ${path} ]; then
     mkdir -p ${path}
@@ -96,7 +95,7 @@ checkout_repo() {
 }
 
 # Install runc
-checkout_repo ${RUNC_PKG} ${RUNC_VERSION}
+checkout_repo ${RUNC_PKG} ${RUNC_VERSION} ${RUNC_REPO}
 cd ${GOPATH}/src/${RUNC_PKG}
 BUILDTAGS=${BUILDTAGS:-seccomp apparmor}
 make static BUILDTAGS="$BUILDTAGS"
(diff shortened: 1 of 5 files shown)


hack/install-deps.sh
```diff
@@ -76,15 +76,12 @@ GOPATH=${TMPGOPATH}
...
+# 2) Version;
+# 3) Repo name;
 checkout_repo() {
   local -r pkg=$1
   local -r version=$2
-  local repo=${3:-""}
-  if [ -z "${repo}" ]; then
-    repo=${pkg}
-  fi
+  local -r repo=$3
```
ijc: Since the callers are pretty consistent I debated doing something like:
```sh
local -r what=$1
eval `echo local -r pkg=\${${what}_PKG}`
eval `echo local -r version=\${${what}_VERSION}`
eval `echo local -r repo=\${${what}_REPO}`
```
But I figured that might be a bit too tricksy and left it as is.

ijc:
I fixed a typo (`fKUBERNETES_REPO=...`) and a thinko (was using `github.com/containernetworking/cni` where `github.com/containernetworking/plugins` was needed) and now CI is passing ok.

ijc:
I think this sorts the first part of #385 too.

ijc:
Force push to address conflict + moved code to `utils.sh` as suggested. Will look at CI failures if/when they happen again.

There might be scope for reusing some of this stuff in `hack/test-cri.sh`, mostly `checkout_repo` I guess.

ijc:
Travis is happy after the rebase.

/retest
(I suspect i lack privs for that)

ijc:
```
W0207 18:53:28.969] awk: not an option: -e
W0207 18:53:28.970] make: *** [_output/cri-containerd-1.0.0-beta.1-92-g46ee4f3.linux-amd64.tar.gz] Error 2
```

Looks like CI has a different `awk` to me. I have `gawk` (on my Debian box) I suspect CI has `mawk`. I'll try and repro locally with an explicit `mawk`.

ijc:
Force pushed a rebase + `mawk` compatible invocation (tested locally by explicitly calling `mawk`, the error looked similar to above).

hack/install-deps.sh
```diff
@@ -27,6 +27,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+source $(dirname "${BASH_SOURCE[0]}")/utils.sh
```
Random-Liu: I think we can either source this into `hack/versions` and probably renaming to `hack/versions.sh`, or move `hack/versions` into `utils.sh`.
It is weird that we source a dependency of `hack/versions` here.

mikebrow: yeah I moved what was left of hack versions into utils.sh.. so can pick the up with my pr

Random-Liu: Just found that test hasn't run. Will merge after test passes.

Random-Liu: @mikebrow Merged. Can you address this in your PR?

mikebrow: absolutely!

PR #557, opened 2018-01-19
Use containerd config

Random-Liu:
Based on https://github.com/containerd/cri-containerd/pull/555.

Use containerd plugin config. With this PR, we can use contianerd config file to config CRI plugin:
```console
$ cat /etc/containerd/config.toml 
[plugins.linux]
  shim_debug = true
[plugins.cri]
  stream_server_port = "10010"
```

### DIFF
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -35,13 +35,15 @@ const (
 	connectionTimeout = 10 * time.Second
 )
 
-// ContainerdConfig contains config related to containerd
+// ContainerdConfig contains toml config related to containerd
 type ContainerdConfig struct {
 	// RootDir is the root directory path for containerd.
+	// TODO(random-liu): Remove this field when no longer support cri-containerd standalone mode.
 	RootDir string `toml:"root_dir" json:"rootDir,omitempty"`
 	// Snapshotter is the snapshotter used by containerd.
 	Snapshotter string `toml:"snapshotter" json:"snapshotter,omitempty"`
 	// Endpoint is the containerd endpoint path.
+	// TODO(random-liu): Remove this field when no longer support cri-containerd standalone mode.
 	Endpoint string `toml:"endpoint" json:"endpoint,omitempty"`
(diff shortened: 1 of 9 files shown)


cmd/cri-containerd/options/options.go
```diff
@@ -88,19 +84,45 @@ type Config struct {
 	StatsCollectPeriod int `toml:"stats_collect_period" json:"statsCollectPeriod,omitempty"`
 	// SystemdCgroup enables systemd cgroup support.
 	SystemdCgroup bool `toml:"systemd_cgroup" json:"systemdCgroup,omitempty"`
+	// SkipImageFSUUID skips retrieving imagefs uuid.
+	// TODO(random-liu): Remove this after we find a generic way to get imagefs uuid.
+	SkipImageFSUUID bool `toml:"skip_imagefs_uuid" json:"skipImageFSUUID,omitempty"`
+}
+
+// CRIConfig contains toml config related to CRI service.
```
mikebrow: awe thought you were cutting the cord after reading the above edits deleting these  :-)

cmd/cri-containerd/options/options.go
```diff
@@ -88,19 +84,45 @@ type Config struct {
...
+}
+
+// Config contains toml config related cri-containerd daemon.
+type Config struct {
+	CRIConfig `toml:"-"`
+	// CgroupPath is the path for the cgroup that cri-containerd is placed in.
+	CgroupPath string `toml:"cgroup_path"`
 	// OOMScore adjust the cri-containerd's oom score
-	OOMScore int `toml:"oom_score" json:"oomScore,omitempty"`
+	OOMScore int `toml:"oom_score"`
```
mikebrow: why remove the json definitions? I suppose this struct is going away anyhow...

Random-Liu: Initially we have json defination, because we print the config in `Status` info. Now these options are not part of `Status` info. So don't need to have json definition. :)

PR #558, opened 2018-01-19
Report containerd version instead of cri-containerd version.

Random-Liu:
Given that we are going to merge cri-containerd with containerd soon, it makes sense to report containerd version instead.

### DIFF
--- a/pkg/server/version.go
+++ b/pkg/server/version.go
@@ -17,32 +17,30 @@ limitations under the License.
 package server
 
 import (
+	"fmt"
+
 	"golang.org/x/net/context"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
-
-	"github.com/containerd/cri-containerd/pkg/version"
 )
 
 const (
-	// For now, containerd and runc are bundled with cri-containerd, cri-containerd
-	// version is more important to us.
-	// TODO(random-liu): Figure out how to package cri-containerd and containerd,
-	// and how to version it. We still prefer calling the container runtime "containerd",
-	// but we care both the cri-containerd version and containerd version.
-	containerName        = "cri-containerd"
-	containerdAPIVersion = "0.0.0"
+	containerName = "containerd"
 	// kubeAPIVersion is the api version of kubernetes.
+	// TODO(random-liu): Change this to actual CRI version.
 	kubeAPIVersion = "0.1.0"
 )
 
 // Version returns the runtime name, runtime version and runtime API version.
-// TODO(random-liu): Return containerd version since we are going to merge 2 daemons.
 func (c *criContainerdService) Version(ctx context.Context, r *runtime.VersionRequest) (*runtime.VersionResponse, error) {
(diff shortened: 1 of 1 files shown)


pkg/server/version.go
```diff
@@ -17,32 +17,29 @@ limitations under the License.
...
 
 const (
-	// For now, containerd and runc are bundled with cri-containerd, cri-containerd
-	// version is more important to us.
-	// TODO(random-liu): Figure out how to package cri-containerd and containerd,
-	// and how to version it. We still prefer calling the container runtime "containerd",
-	// but we care both the cri-containerd version and containerd version.
-	containerName        = "cri-containerd"
-	containerdAPIVersion = "0.0.0"
+	containerName = "containerd"
```
mikebrow: We should talk about this...
Will our containerd/cri plugin not have it's own version tag? 
Can/Should we change this before we've killed the cri-containerd standalone mode?

I'm not against this direction just dotting i's and crossing t's to make sure one version tag is right answer esp. given that we will be "vendored" in. My gut says while we are vendored we need a version tag and while we have a version tag we probably should report it.

mikebrow: Should we report one version in make version and another in this Version() api?
https://github.com/containerd/cri-containerd/blob/master/Makefile#L63

pkg/server/version.go
```diff
@@ -17,32 +17,29 @@ limitations under the License.
...
-// TODO(random-liu): Return containerd version since we are going to merge 2 daemons.
 func (c *criContainerdService) Version(ctx context.Context, r *runtime.VersionRequest) (*runtime.VersionResponse, error) {
+	resp, err := c.client.Version(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get containerd version: %v", err)
+	}
 	return &runtime.VersionResponse{
 		Version:        kubeAPIVersion,
 		RuntimeName:    containerName,
-		RuntimeVersion: version.CRIContainerdVersion,
```
mikebrow: see..
https://github.com/containerd/cri-containerd/blob/master/Makefile#L31

Note: this is the only use of pkg/version.. so if we don't keep using it we'll want to update the makefile and rm pkg/version/version.go

Random-Liu: @mikebrow For plugin mode, we won't have our own version, containerd will take care of that.

But I agree, we should punt this change until containerd starts vendoring cri plugin.

mikebrow: we may not have our own version via reporting but as long as we have our own github branch we'll at least have a commit version if no longer adding version tags to our release commits..

mikebrow: and because we'll have a vendor'd in version tag/commit.. we may want to figure out how to report that

Random-Liu:
@mikebrow What do you think about https://github.com/containerd/containerd/issues/2174#issuecomment-369439095

It seems that we do still need this PR.

Random-Liu:
Apply LGTM based on https://github.com/containerd/cri-containerd/pull/558#pullrequestreview-101769471

PR #561, opened 2018-01-22
Improve `hack/update-vendor.sh` and `make install.deps`.

Random-Liu:
Current `hack/update-vendor.sh` and `make install.deps` make the assumption that containerd i not vendoring CRI plugin.

However, once containerd starts vendoring CRI plugin, these scripts won't work.

This PR removes cri plugin and its vendors from containerd before compare vendors or cook containerd, which resolves the problem.

With this PR, even after containerd starts vendoring CRI plugin, our script can automatically remove that CRI plugin and replace with newest CRI one. This will be very useful for our CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ jobs:
         - cat /tmp/test-cri/containerd.log
       go: 1.9.x
     - script:
+        - make .install.vndr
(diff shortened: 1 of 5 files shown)


Random-Liu:
Close this one in favor of #615.

With #615 we don't need to play all kinds of tricks anymore. :)

PR #570, opened 2018-01-24
Use channel to pass the stop info instead of polling for container stop

### LINKED ISSUE #569: Do not poll for container stop.
Currently, during container stop, we poll every 0.1s to see whether a container is stopped.

When we stop a batch of containers, this may consume extra CPU.

We should add a channel `stopCh` for this in container metadata. Let event monitor close the channel, and stop function only need to wait for the channel.

@yanxuean @miaoyq

miaoyq:
Fixes #569

/cc @Random-Liu @yanxuean 

### DIFF
--- a/pkg/server/container_stop.go
+++ b/pkg/server/container_stop.go
@@ -28,15 +28,10 @@ import (
 	"golang.org/x/sys/unix"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
 
-	"github.com/containerd/cri-containerd/pkg/store"
 	containerstore "github.com/containerd/cri-containerd/pkg/store/container"
 )
 
 const (
-	// stopCheckPollInterval is the the interval to check whether a container
-	// is stopped successfully.
-	stopCheckPollInterval = 100 * time.Millisecond
-
 	// killContainerTimeout is the timeout that we wait for the container to
 	// be SIGKILLed.
 	killContainerTimeout = 2 * time.Minute
@@ -104,7 +99,7 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
 			}
 		}
 
-		err = c.waitContainerStop(ctx, id, timeout)
+		err = c.waitContainerStop(ctx, container, timeout)
 		if err == nil {
 			return nil
 		}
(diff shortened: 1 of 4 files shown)


Random-Liu:
We need to handle cri-containerd restart.

We should change `containerstore.WithStatus` to stop the channel if `FinishedAt` in the initial status is ste.

pkg/store/container/container.go
```diff
@@ -89,6 +91,16 @@ func (c *Container) Delete() error {
 	return c.Status.Delete()
 }
 
+// Stop close stopCh of the container.
+func (c *Container) Stop() {
```
Random-Liu: nit: Instead of adding a function to `Container`, can we make a type of `stopCh` and add function to it instead?

In this way, it can be reused in Sandbox after my change is in.

pkg/server/container_stop.go
```diff
@@ -138,33 +134,25 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
...
-		}
-		select {
-		case <-ctx.Done():
-			return fmt.Errorf("wait container %q is cancelled", id)
-		case <-timeoutTimer.C:
-			return fmt.Errorf("wait container %q stop timeout", id)
-		case <-ticker.C:
-			continue
+	// Poll once before waiting for stopCheckPollInterval.
+	container, err := c.containerStore.Get(id)
```
Random-Liu: Can we pass `container containerstore.Container` to the `waitContainerStop` function to avoid this?

pkg/store/container/container.go
miaoyq: Will do.

pkg/server/container_stop.go
miaoyq: Yean, it's better, will do.

pkg/server/events.go
```diff
@@ -117,6 +117,8 @@ func (em *eventMonitor) handleEvent(evt *events.Envelope) {
 			// Non-init process died, ignore the event.
 			return
 		}
+
+		cntr.Stop()
```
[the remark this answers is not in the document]
Random-Liu: We shouldn't do for `TaskOOM`. :)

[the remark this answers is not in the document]
Random-Liu: > It should be better to do this after cntr.Status.UpdateSync.
We can ensure that the whole container is over for the user who watch the channel.

Agree.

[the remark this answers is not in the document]
miaoyq: Have changed it yesterday, :-)

pkg/store/container/container.go
miaoyq: Done.

pkg/server/container_stop.go
miaoyq: Done.

pkg/server/container_stop.go
```diff
@@ -138,33 +134,25 @@ func (c *criContainerdService) stopContainer(ctx context.Context, container cont
 
 // waitContainerStop polls container state until timeout exceeds or container is stopped.
```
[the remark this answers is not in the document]
miaoyq: Done.

miaoyq:
@Random-Liu @yanxuean  Have addressed all comments, PTAL.

miaoyq:
Ooh, missed the test cases, will fix it.

pkg/store/container/container.go
```diff
@@ -89,6 +95,19 @@ func (c *Container) Delete() error {
 	return c.Status.Delete()
 }
 
+// StopCh is used to pass the stop information of a container.
```
Random-Liu: s/pass/propagate/g

And please also change the other places.

miaoyq:
> Ooh, missed the test cases, will fix it.

Fixed.

pkg/store/container/container.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	cio "github.com/containerd/cri-containerd/pkg/server/io"
 	"github.com/containerd/cri-containerd/pkg/store"
+	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
```
Random-Liu: Move this to the group above.

Random-Liu:
LGTM with 2 nits.

pkg/store/container/container.go
miaoyq: done.

pkg/store/container/container.go
miaoyq: done.

PR #571, opened 2018-01-25
Avoid containerd access as much as possible.

Random-Liu:
Related to https://github.com/containerd/containerd/issues/2020.

This PR:
1) Cache sandbox status in sandbox store.
2) Update sandbox status based on containerd event.
3) Recover sandbox cache during restart.

This should reduce the containerd/containerd-shim cpu/memory usage.

### DIFF
--- a/integration/sandbox_clean_remove_test.go
+++ b/integration/sandbox_clean_remove_test.go
@@ -18,6 +18,7 @@ package integration
 
 import (
 	"testing"
+	"time"
 
 	"github.com/containerd/containerd"
 	"github.com/stretchr/testify/assert"
@@ -50,9 +51,13 @@ func TestSandboxCleanRemove(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Logf("Sandbox state should be NOTREADY")
-	status, err := runtimeService.PodSandboxStatus(sb)
-	require.NoError(t, err)
-	assert.Equal(t, runtime.PodSandboxState_SANDBOX_NOTREADY, status.GetState())
+	assert.NoError(t, Eventually(func() (bool, error) {
+		status, err := runtimeService.PodSandboxStatus(sb)
+		if err != nil {
+			return false, err
+		}
+		return status.GetState() == runtime.PodSandboxState_SANDBOX_NOTREADY, nil
+	}, time.Second, 30*time.Second), "sandbox state should become NOTREADY")
 
 	t.Logf("Should not be able to remove the sandbox when netns is not closed")
 	assert.Error(t, runtimeService.RemovePodSandbox(sb))
--- a/pkg/server/events.go
+++ b/pkg/server/events.go
@@ -28,12 +28,14 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 
+	"github.com/containerd/cri-containerd/pkg/store"
 	containerstore "github.com/containerd/cri-containerd/pkg/store/container"
 	sandboxstore "github.com/containerd/cri-containerd/pkg/store/sandbox"
 )
 
 // eventMonitor monitors containerd event and updates internal state correspondingly.
-// TODO(random-liu): [P1] Is it possible to drop event during containerd is running?
+// TODO(random-liu): [P1] Figure out is it possible to drop event during containerd
+// is running. If it is, we should do periodically list to sync state with containerd.
 type eventMonitor struct {
 	containerStore *containerstore.Store
 	sandboxStore   *sandboxstore.Store
@@ -106,57 +108,22 @@ func (em *eventMonitor) handleEvent(evt *events.Envelope) {
 		e := any.(*eventtypes.TaskExit)
 		logrus.Infof("TaskExit event %+v", e)
 		cntr, err := em.containerStore.Get(e.ContainerID)
-		if err != nil {
-			if _, err := em.sandboxStore.Get(e.ContainerID); err == nil {
-				return
-			}
+		if err == nil {
+			handleContainerExit(e, cntr)
+			return
+		} else if err != store.ErrNotExist {
 			logrus.WithError(err).Errorf("Failed to get container %q", e.ContainerID)
 			return
 		}
-		if e.Pid != cntr.Status.Get().Pid {
-			// Non-init process died, ignore the event.
+		// Use GetAll to include sandbox in unknown state.
+		sb, err := em.sandboxStore.GetAll(e.ContainerID)
+		if err == nil {
+			handleSandboxExit(e, sb)
+			return
+		} else if err != store.ErrNotExist {
+			logrus.WithError(err).Errorf("Failed to get sandbox %q", e.ContainerID)
 			return
 		}
-		// Attach container IO so that `Delete` could cleanup the stream properly.
-		task, err := cntr.Container.Task(context.Background(),
-			func(*containerdio.FIFOSet) (containerdio.IO, error) {
-				return cntr.IO, nil
-			},
-		)
-		if err != nil {
-			if !errdefs.IsNotFound(err) {
-				logrus.WithError(err).Errorf("failed to stop container, task not found for container %q", e.ContainerID)
-				return
-			}
-		} else {
-			// TODO(random-liu): [P1] This may block the loop, we may want to spawn a worker
-			if _, err = task.Delete(context.Background()); err != nil {
-				// TODO(random-liu): [P0] Enqueue the event and retry.
-				if !errdefs.IsNotFound(err) {
-					logrus.WithError(err).Errorf("failed to stop container %q", e.ContainerID)
-					return
-				}
-				// Move on to make sure container status is updated.
-			}
-		}
-		err = cntr.Status.UpdateSync(func(status containerstore.Status) (containerstore.Status, error) {
-			// If FinishedAt has been set (e.g. with start failure), keep as
-			// it is.
-			if status.FinishedAt != 0 {
-				return status, nil
-			}
-			status.Pid = 0
-			status.FinishedAt = e.ExitedAt.UnixNano()
-			status.ExitCode = int32(e.ExitStatus)
-			return status, nil
-		})
-		if err != nil {
-			logrus.WithError(err).Errorf("Failed to update container %q state", e.ContainerID)
-			// TODO(random-liu): [P0] Enqueue the event and retry.
-			return
-		}
-		// Using channel to propagate the information of container stop
-		cntr.Stop()
 	case *eventtypes.TaskOOM:
 		e := any.(*eventtypes.TaskOOM)
 		logrus.Infof("TaskOOM event %+v", e)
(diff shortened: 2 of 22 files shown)


miaoyq:
golint failed.
```
I0125 01:26:35.928] pkg/store/sandbox/status.go:33:2:warning: don't use underscores in Go names; const PodSandboxState_UNKNOWN should be PodSandboxStateUNKNOWN (golint)
I0125 01:26:35.929] pkg/store/sandbox/status.go:36:2:warning: don't use underscores in Go names; const PodSandboxState_READY should be PodSandboxStateREADY (golint)
I0125 01:26:35.930] pkg/store/sandbox/status.go:42:2:warning: don't use underscores in Go names; const PodSandboxState_NOTREADY should be PodSandboxStateNOTREADY (golint)
```

Random-Liu:
@miaoyq Fixing it. :)

Random-Liu:
@mikebrow @yanxuean @miaoyq PR ready for review.

pkg/store/sandbox/sandbox.go
```diff
@@ -67,9 +76,22 @@ func (s *Store) Add(sb Sandbox) error {
 	return nil
 }
 
-// Get returns the sandbox with specified id. Returns nil
+// Get returns the sandbox with specified id. Returns store.ErrNotExist
 // if the sandbox doesn't exist.
 func (s *Store) Get(id string) (Sandbox, error) {
+	sb, err := s.GetAll(id)
+	if err != nil {
+		return sb, err
```
yanxuean: Sandbox{},err

pkg/server/events.go
```diff
@@ -106,53 +107,20 @@ func (em *eventMonitor) handleEvent(evt *events.Envelope) {
 		e := any.(*eventtypes.TaskExit)
```
yanxuean: Do we not handle sandbox with TaskOOM?

pkg/server/sandbox_stop.go
```diff
@@ -111,5 +118,28 @@ func (c *criContainerdService) stopSandboxContainer(ctx context.Context, contain
...
+
+// waitSandboxStop polls sandbox state until timeout exceeds or sandbox is stopped.
+func (c *criContainerdService) waitSandboxStop(ctx context.Context, sandbox sandboxstore.Sandbox, timeout time.Duration) error {
+	ticker := time.NewTicker(stopCheckPollInterval)
+	defer ticker.Stop()
+	timeoutTimer := time.NewTimer(timeout)
+	defer timeoutTimer.Stop()
+	for {
+		// Poll once before waiting for stopCheckPollInterval.
+		// TODO(random-liu): Use channel with event handler instead of polling.
```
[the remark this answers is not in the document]
miaoyq: OK, will change this after this pr merged.

pkg/store/sandbox/sandbox.go
Random-Liu: It's fine here. I think. :)

pkg/server/events.go
Random-Liu: Yeah, we don't need to do that. Even we handle, what can we do? :)

pkg/server/sandbox_list.go
```diff
@@ -33,52 +27,30 @@ import (
 func (c *criContainerdService) ListPodSandbox(ctx context.Context, r *runtime.ListPodSandboxRequest) (*runtime.ListPodSandboxResponse, error) {
 	// List all sandboxes from store.
 	sandboxesInStore := c.sandboxStore.List()
-
-	response, err := c.client.TaskService().List(ctx, &tasks.ListTasksRequest{})
-	if err != nil {
-		return nil, fmt.Errorf("failed to list sandbox containers: %v", err)
-	}
-
 	var sandboxes []*runtime.PodSandbox
```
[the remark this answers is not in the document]
Random-Liu: Let's do this kind of optimization when we see this is a bottleneck. :)

pkg/server/sandbox_run.go
```diff
@@ -224,32 +227,83 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
+			id, name)
+		// We don't need stdio for sandbox container.
+		task, err := container.NewTask(ctx, containerdio.NullIO)
+		if err != nil {
+			return status, fmt.Errorf("failed to create containerd task: %v", err)
+		}
+		defer func() {
+			if retErr != nil {
+				// Cleanup the sandbox container if an error is returned.
+				if _, err := task.Delete(ctx, containerd.WithProcessKill); err != nil {
```
yanxuean: Now  we do task.Delete also with unknown state in event.handleSandboxExit,  so we should check that if the err is NotFound.  The NotFound is a normal case.

pkg/server/events.go
```diff
@@ -175,3 +143,89 @@ func (em *eventMonitor) handleEvent(evt *events.Envelope) {
...
+		// TODO(random-liu): [P0] Enqueue the event and retry.
+		return
+	}
+}
+
+// handleSandboxExit handles TaskExit event for sandbox.
+func handleSandboxExit(e *eventtypes.TaskExit, sb sandboxstore.Sandbox) {
+	// Don't need to check pid because sandbox container should only have one process.
+	// No stream attached to sandbox container.
+	task, err := sb.Container.Task(context.Background(), nil)
```
yanxuean: I  think we should not do this  on unknown state.  Let RunPodSandbox do it. 
Or else we maybe do task.Delete both in handleSandboxExit and RunPodSandbox.

pkg/server/sandbox_status.go
```diff
@@ -132,10 +112,23 @@ type sandboxInfo struct {
...
+	var pid uint32
+	var processStatus containerd.ProcessStatus
+	if task != nil {
+		taskStatus, err := task.Status(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get task status: %v", err)
+		}
+
+		pid = task.Pid()
+		processStatus = taskStatus.Status
```
miaoyq: Could we also store the `pid` and `processStatus` to `sandboxstore`?

pkg/server/sandbox_run.go
Random-Liu: Changed both `sandbox_run.go` and `container_start.go` to check `IsNotFound`. Good catch!

pkg/server/events.go
Random-Liu: On a second thought, I still prefer put `task.Delete` out of `Update.

`Update` will hold a lock of the container/sandbox, and `List` (which is a critical operation for kubelet) will be blocked until the lock is released.

We should put as little logic as possible into `Update`. The issue you mentioned can be handled by https://github.com/containerd/cri-containerd/pull/571#discussion_r163756316.

pkg/server/sandbox_status.go
Random-Liu: `processState` keeps changing, it's hard to maintain it in our store. I'd prefer not to do that until we have to.

I'll do `pid`.

Random-Liu: Done

pkg/server/sandbox_list.go
```diff
@@ -33,52 +27,30 @@ import (
 func (c *criContainerdService) ListPodSandbox(ctx context.Context, r *runtime.ListPodSandboxRequest) (*runtime.ListPodSandboxResponse, error) {
 	// List all sandboxes from store.
 	sandboxesInStore := c.sandboxStore.List()
-
-	response, err := c.client.TaskService().List(ctx, &tasks.ListTasksRequest{})
```
mikebrow: whoot!!!

mikebrow: Consider adding a todo for running a validate routine to make sure we have stayed in sync and didn't just miss an event.. I like this pattern of working off the push of events, but without some periodic validation there is a risk of an event failing to reach us, low but risk nonetheless? Or are all the events returning successful delivery to containerd?

pkg/server/sandbox_status.go
```diff
@@ -132,10 +112,23 @@ type sandboxInfo struct {
 }
 
 // toCRISandboxInfo converts internal container object information to CRI sandbox status response info map.
-func toCRISandboxInfo(ctx context.Context, sandbox sandboxstore.Sandbox,
-	pid uint32, processStatus containerd.ProcessStatus, verbose bool) (map[string]string, error) {
-	if !verbose {
-		return nil, nil
+func toCRISandboxInfo(ctx context.Context, sandbox sandboxstore.Sandbox) (map[string]string, error) {
```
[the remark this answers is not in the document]
mikebrow: ok

Random-Liu:
Test failure seems like a GCE flake.

pkg/server/sandbox_status.go
```diff
@@ -38,32 +36,14 @@ func (c *criContainerdService) PodSandboxStatus(ctx context.Context, r *runtime.
...
+	status := toCRISandboxStatus(sandbox.Metadata, sandbox.Status.Get(), ip)
+	if !r.GetVerbose() {
+		return &runtime.PodSandboxStatusResponse{Status: status}, nil
 	}
-	createdAt := ctrInfo.CreatedAt
-	status := toCRISandboxStatus(sandbox.Metadata, processStatus, createdAt, ip)
-	info, err := toCRISandboxInfo(ctx, sandbox, pid, processStatus, r.GetVerbose())
+
+	// Generate verbose information.
+	info, err := toCRISandboxInfo(ctx, sandbox)
```
mikebrow: if r.GetVerbose() there is no need to get info here..

pkg/server/sandbox_stop.go
```diff
@@ -19,15 +19,22 @@ package server
...
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
+
+	sandboxstore "github.com/containerd/cri-containerd/pkg/store/sandbox"
 )
 
+// stopCheckPollInterval is the the interval to check whether a sandbox
+// is stopped successfully.
+const stopCheckPollInterval = 100 * time.Millisecond
```
mikebrow: todo make configurable

pkg/server/sandbox_stop.go
```diff
@@ -89,14 +96,18 @@ func (c *criContainerdService) StopPodSandbox(ctx context.Context, r *runtime.St
 		return nil, fmt.Errorf("failed to unmount sandbox files in %q: %v", sandboxRoot, err)
 	}
 
-	if err := c.stopSandboxContainer(ctx, sandbox.Container); err != nil {
-		return nil, fmt.Errorf("failed to stop sandbox container %q: %v", id, err)
+	// Only stop sandbox container when it's running.
```
mikebrow: is no error the the expected CRI behavior if it's not running, seems like we should tell them they can't stop a non running podsandbox ...
At least log a warning?

pkg/store/sandbox/sandbox.go
mikebrow: it's fine..

integration/sandbox_clean_remove_test.go
```diff
@@ -50,9 +51,13 @@ func TestSandboxCleanRemove(t *testing.T) {
...
-	status, err := runtimeService.PodSandboxStatus(sb)
-	require.NoError(t, err)
-	assert.Equal(t, runtime.PodSandboxState_SANDBOX_NOTREADY, status.GetState())
+	assert.NoError(t, Eventually(func() (bool, error) {
+		status, err := runtimeService.PodSandboxStatus(sb)
+		if err != nil {
+			return false, err
+		}
+		return status.GetState() == runtime.PodSandboxState_SANDBOX_NOTREADY, nil
+	}, time.Second, 30*time.Second), "sandbox state should become NOTREADY")
```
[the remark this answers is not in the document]
Random-Liu: I think fixed 30 second for test is fine. This should happen very soon.

pkg/server/sandbox_list.go
Random-Liu: Good point. Will add a TODO for that.

We sometimes see Docker stay in weird state because it missed an event and not in sync with containerd.

pkg/server/sandbox_status.go
Random-Liu: If `r.GetVerbose()` it will return directly above.

pkg/server/sandbox_stop.go
Random-Liu: We'll remove this very soon by adding a stop channel. @yanxuean @miaoyq are going to do that.
See https://github.com/containerd/cri-containerd/pull/570

pkg/server/sandbox_stop.go
Random-Liu: Actually we expect `StopPodSandbox` to be idempotent. Caller should be able to call this many times.

And kubelet does call this function multiple times for one pod sandbox.

pkg/server/sandbox_list.go
Random-Liu: Done.

integration/sandbox_clean_remove_test.go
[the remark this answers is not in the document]
mikebrow: I think it's fine to, but people not wanting to wait for a new pod might not like it..

pkg/server/sandbox_status.go
mikebrow: oh there it is :-)

integration/sandbox_clean_remove_test.go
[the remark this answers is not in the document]
Random-Liu: Hm, what do you mean?

Here, we kill the corresponding task from containerd, and wait for the pod sandbox status to be updated to NOTREADY by cri-containerd. It usually only takes several microseconds I guess?

pkg/server/sandbox_stop.go
mikebrow: ok. just checking, odd behavior IMO Maybe they call stop multiple times cause no one threw an already stopped error that they could easily ignore if they wanted to treat it as idempotent for their use case.

integration/sandbox_clean_remove_test.go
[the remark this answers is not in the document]
mikebrow: NVM I was distracted. Ignore :-)

pkg/server/events.go
```diff
@@ -28,12 +28,14 @@ import (
...
 
+	"github.com/containerd/cri-containerd/pkg/store"
 	containerstore "github.com/containerd/cri-containerd/pkg/store/container"
 	sandboxstore "github.com/containerd/cri-containerd/pkg/store/sandbox"
 )
 
 // eventMonitor monitors containerd event and updates internal state correspondingly.
-// TODO(random-liu): [P1] Is it possible to drop event during containerd is running?
+// TODO(random-liu): [P1] Figure out is it possible to drop event during containerd
+// is running. If it is, we should do periodically list to sync state with containerd.
```
mikebrow: looks good!

Random-Liu:
Rebased.

@miaoyq @yanxuean Are you 2 OK with this PR now? :)

PR #572, opened 2018-01-25
Add runtime cgroup and fix a cli panic.

Random-Liu:
This PR:
1) Add `/runtime` cgroup for containerd and cri-containerd in e2e test. So that our resource usage test in https://k8s-testgrid.appspot.com/sig-node-containerd#node-e2e-serial can get full runtime resource usage. @yanxuean @miaoyq 
2) Fix `cri-containerd default-config` panic. I don't understand why it panics, but this PR does fix it by removing a level of embedded struct `CRIConfig`. See https://github.com/BurntSushi/toml/issues/211.

### DIFF
--- a/cmd/cri-containerd/options/options.go
+++ b/cmd/cri-containerd/options/options.go
@@ -66,7 +66,7 @@ type CniConfig struct {
 }
 
 // PluginConfig contains toml config related to CRI plugin,
-// it is a subset of CRIConfig.
+// it is a subset of Config.
 type PluginConfig struct {
 	// ContainerdConfig contains config related to containerd
 	ContainerdConfig `toml:"containerd" json:"containerd,omitempty"`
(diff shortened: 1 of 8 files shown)


cmd/cri-containerd/options/options.go
```diff
@@ -109,23 +109,20 @@ type CRIConfig struct {
...
-}
-
-// Config contains toml config related cri-containerd daemon.
-type Config struct {
-	CRIConfig `toml:"-"`
+	// TODO(random-liu): Remove following fields when we no longer support cri-containerd
+	// standalone mode.
 	// CgroupPath is the path for the cgroup that cri-containerd is placed in.
-	CgroupPath string `toml:"cgroup_path"`
+	CgroupPath string `toml:"cgroup_path" json:"cgroupPath,omitempty"`
```
mikebrow: :-)

Random-Liu: Yeah, added back because they are part of `info` again with this change. :(

PR #582, opened 2018-02-01
Set registry mirror.

Random-Liu:
Use registry mirror added in https://github.com/containerd/cri-containerd/pull/531.

@abhi 

Let's see whether the test passes.


mikebrow:
It passes! @Random-Liu was there a reason to switch or were you just testing?

Random-Liu:
@mikebrow In Kubernetes GCE environment, we also set the registry mirror. 
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L1058
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/util.sh#L249

This is only for GCE. :)

PR #583, opened 2018-02-01
Put version into metadata so that version won't be changed across reboot.

Random-Liu:
Cloud-init script will run whenever node reboot. Previously we get version from `https://storage.googleapis.com/cri-containerd-release/latest` in the cloud init script.

However, if the latest version changed before reboot, the node will download a new version of cri-containerd during reboot, which might not be compatible with the Kubernetes version in use.

We should fix the version in instance metadata.

### DIFF
--- a/cluster/gce/configure.sh
+++ b/cluster/gce/configure.sh
@@ -43,13 +43,12 @@ DEPLOY_PATH=${DEPLOY_PATH:-"cri-containerd-release"}
 # By default use the release tarball with cni built in.
 PKG_PREFIX=${PKG_PREFIX:-"cri-containerd-cni"}
 
-# VERSION is the cri-containerd version to use. If not specified,
-# the latest version will be used.
+# VERSION is the cri-containerd version to use.
 VERSION_METADATA="version"
 VERSION=$(fetch_metadata "${VERSION_METADATA}")
 if [ -z "${VERSION}" ]; then
-  VERSION=$(curl -f --ipv4 --retry 6 --retry-delay 3 --silent --show-error \
-    https://storage.googleapis.com/${DEPLOY_PATH}/latest)
+  echo "Version is not set."
+  exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


cluster/gce/configure.sh
```diff
@@ -36,9 +36,6 @@ fetch_metadata() {
   fi
 }
 
-# DEPLOY_PATH is the gcs path where cri-containerd tarball is stored.
-DEPLOY_PATH=${DEPLOY_PATH:-"cri-containerd-release"}
-
```
mikebrow: nit: https://storage.googleapis.com/cri-containerd-release is uses a couple times might want to add it back?

cluster/gce/configure.sh
```diff
@@ -48,12 +45,12 @@ PKG_PREFIX=${PKG_PREFIX:-"cri-containerd-cni"}
 VERSION_METADATA="version"
 VERSION=$(fetch_metadata "${VERSION_METADATA}")
 if [ -z "${VERSION}" ]; then
-  VERSION=$(curl -f --ipv4 --retry 6 --retry-delay 3 --silent --show-error \
-    https://storage.googleapis.com/${DEPLOY_PATH}/latest)
+  echo "Version is not set."
```
mikebrow: need to edit comment lines 43-44 above...

cluster/gce/configure.sh
Random-Liu: OK. Will change it to a constant.

cluster/gce/configure.sh
Random-Liu: Will do. Thanks!

cluster/gce/configure.sh
Random-Liu: Done.

cluster/gce/configure.sh
Random-Liu: Done.

PR #585, opened 2018-02-01
vendor: update BurntSushi/toml for MIT license

stevvooe:
Signed-off-by: Stephen J Day <stephen.day@docker.com>


Random-Liu:
@stevvooe In verify script, we enforce that vendors introduced by containerd have to be exactly the same with containerd now.

So we need to merge https://github.com/containerd/containerd/pull/2088 first, and then update containerd.

stevvooe:
@k8s-ci-robot /retest

mikebrow:
Not sure why it didn't restart.. just did it manually...

mikebrow:
ah.. need to update the containerd version in https://github.com/containerd/cri-containerd/blob/master/vendor.conf and https://github.com/containerd/cri-containerd/blob/master/hack/versions to pull in containerd/containerd#2088

stevvooe:
@mikebrow Updated.

Random-Liu:
```
I0202 00:49:59.395] Need to update vendor.conf.
I0202 00:49:59.396] 3c3
I0202 00:49:59.396] < github.com/containerd/cgroups 29da22c6171a4316169f9205ab6c49f59b5b852f
I0202 00:49:59.396] ---
I0202 00:49:59.397] > github.com/containerd/cgroups c0710c92e8b3a44681d1321dcfd1360fc5c6c089
```
@stevvooe Please run `hack/update-vendor.sh` after you update containerd. It will keep all dependencies in sync with containerd automatically.

stevvooe:
> @stevvooe Please run hack/update-vendor.sh after you update containerd, and then of course vndr. It will keep all dependencies in sync with containerd automatically.

ಠ_ಠ

This is why I tend to reject PRs with a bunch of scripts with dependency declarations all over the place. I'll get this fixed for now, but we need a plan that doesn't make contributions such a pain.

Random-Liu:
> This is why I tend to reject PRs with a bunch of scripts with dependency declarations all over the place. I'll get this fixed for now, but we need a plan that doesn't make contributions such a pain.

You only need to run that after you update containerd. :)

Initially, we need to enforce that because we plan to use some vendor trick to automatically update containerd `vendor/` so as to build CI for plugin mode in this repo. That requires the vendors to be exactly the same...
Look at a even more complex script... https://github.com/containerd/cri-containerd/pull/561

However, I think https://github.com/containerd/containerd/pull/1615 will be a much better solution. Let's get that reviewed. :)

Random-Liu:
@stevvooe Actually you can take a look at `kubernetes/hack`, that is even more fun. :)

stevvooe:
@Random-Liu @mikebrow Don't worry. I won't get in the way. :wink: 

The other option is pull up the dependencies into a common project so we don't have the mutual dependencies. Let's get to the result and we can streamline everything once we are where we want to be feature-wise.

PR #586, opened 2018-02-02
Add a separate CLI for cri-containerd `ctrcri`.

Random-Liu:
Add a separate CLI for cri-containerd called `ctrcri`, basically a `ctr` for CRI plugin. I know the name is bad and confusing, but I could not come up with a better name...

We need a separate CLI. Today we use `cri-containerd load`. However, after switching to plugin mode, we don't have a `cri-containerd` binary. We have to support `load` somewhere, so we added `ctrcri`.

This works with both standalone `cri-containerd` and containerd cri plugin.

The CLI is used for image load for now, which is required for cluster bootstrapping today. We may add new subcommand in the future if we have to.

`ctrcri` isn't targeted to be a user facing tool. User should only use `kubectl` or `crictl` in most cases.

```console
$ ctrcri
        __                 _ 
  _____/ /________________(_)
 / ___/ __/ ___/ ___/ ___/ / 
/ /__/ /_/ /  / /__/ /  / /  
\___/\__/_/   \___/_/  /_/   

containerd CRI plugin CLI

Usage:
  ctrcri [command]

Available Commands:
  help        Help about any command
  load        Load an image from a tar archive.
  version     Print ctrcri version information.

Flags:
      --address string     address for containerd's GRPC server. (default "/var/run/cri-containerd.sock")
  -h, --help               help for ctrcri
      --timeout duration   timeout for containerd grpc connection. (default 10s)

Use "ctrcri [command] --help" for more information about a command.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,9 +41,9 @@ help:
 	@echo "Usage: make <target>"
 	@echo
 	@echo " * 'install'          - Install binaries to system locations"
-	@echo " * 'binaries'         - Build cri-containerd"
+	@echo " * 'binaries'         - Build cri-containerd and ctrcri"
 	@echo " * 'plugin'           - Build cri-containerd as a plugin package"
-	@echo " * 'static-binaries   - Build static cri-containerd"
+	@echo " * 'static-binaries   - Build static cri-containerd and ctrcri"
 	@echo " * 'release'          - Build release tarball"
 	@echo " * 'push'             - Push release tarball to GCS"
 	@echo " * 'test'             - Test cri-containerd with unit test"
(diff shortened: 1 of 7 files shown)


Random-Liu:
>Feel weird to have another cli ctrcri if cri-containerd is a plugin. Anyway to tie it to ctr and --endpoint cri-containerd load image ?

It would be good if we can add command directly into `ctr`.
@containerd/containerd-maintainers Can we do that?

AkihiroSuda:
I guess you can eliminate the image loader: https://github.com/containerd/containerd/pull/1602

Random-Liu:
@AkihiroSuda The problem is that in current model, we still have cache in cri plugin. Just loading image to containerd won't update the cache inside cri plugin.

AkihiroSuda:
OK, at least, https://github.com/containerd/containerd/pull/1602 would be beneficial for code deduplication? Could you review the PR when you have a time? (Sorry for off-topic)

crosbymichael:
When this gets merged in, could we add this as `ctr cri` ?

Random-Liu:
> When this gets merged in, could we add this as `ctr cri`.

Sure. I also think that is better. :)

Random-Liu:
Apply LGTM based on the LGTMs above.
And filed issue https://github.com/containerd/cri-containerd/issues/590.

PR #588, opened 2018-02-02
Use registry-1.docker.io as backup

Random-Liu:
@abhi has a good point. We should not use docker.io as fallback in the code. We should use configuration to specify that.


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri-containerd/pull/588#pullrequestreview-93540850

PR #589, opened 2018-02-02
Create cni configuration directory if it does not present.

Random-Liu:
Part of https://github.com/containerd/cri-containerd/issues/573.

Today, `cri-containerd` and `ocicni` fails if cni conf directory does not exist.

However, after `cri-containerd` becomes a plugin of containerd, it doesn't make sense to let user always create `/etc/cni/net.d` before start containerd.

So let `cri-containerd` ensure that itself.

### DIFF
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -149,6 +149,10 @@ func NewCRIContainerdService(config options.Config) (CRIContainerdService, error
 		logrus.Warn("Skip retrieving imagefs UUID, kubelet will not be able to get imagefs capacity or perform imagefs disk eviction.")
 	}
 
+	// ocicni requires the cni config directory to exist, so make sure it is there.
+	if err := os.MkdirAll(config.NetworkPluginConfDir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create cni config directory %q: %v", err, config.NetworkPluginConfDir)
+	}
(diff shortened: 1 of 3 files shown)


pkg/server/service.go
```diff
@@ -149,6 +149,10 @@ func NewCRIContainerdService(config options.Config) (CRIContainerdService, error
 		logrus.Warn("Skip retrieving imagefs UUID, kubelet will not be able to get imagefs capacity or perform imagefs disk eviction.")
 	}
 
+	// ocicni requires the cni config directory to exist, so make sure it is there.
+	if err := os.MkdirAll(config.NetworkPluginConfDir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create cni config directory %q: %v", err, config.NetworkPluginConfDir)
```
ijc: I think you might have the `Errorf` arguments the other way round from what you intended?

abhi:
@Random-Liu we shouldn't be creating this directory right ? Currently dockershim doesn't if I am not mistaken ? The CNI plugins will create this when they launch their daemonset with volume mount.

pkg/server/service.go
Random-Liu: My bad.

Random-Liu:
@abhi I agree. Actually a better solution is not to assume the network config must exist in ocicni.

We should do things right.

I'll file a PR to them, and use my fork for now.

Random-Liu:
Close this in favor of https://github.com/containerd/cri-containerd/pull/591.

PR #599, opened 2018-02-07
Update ocicni to main stream.

Random-Liu:
Update ocicni to main stream. All PRs we need are merged.


Random-Liu:
fixed ocicni. Apply LGTM based on https://github.com/containerd/cri-containerd/pull/599#pullrequestreview-94821629

PR #600, opened 2018-02-07
removes hack versions; splits update vendor into optional sort and sync operations

mikebrow:
Address issue #540

Refactors update vendor into sort and sync with containerd/vendor.conf operations. Makes update vendor optional (for use by contributors who wish to update vendor).


ijc:
I guess this obsoletes #556 ?

mikebrow:
> I guess this obsoletes #556 ?

Don't think so.. I like what you did there, normalizing / ensuring how (and that) we get REPO for all of the dependencies in case they are set in vendor.conf. A little fancy on "forming" the variables :-) But may just need a rebase or this one will need a rebase depending on which one we merge first :-)

Talked to Lantao a bit last night on this pr. I'll tweak this commit to add back in the feature for merging containerd/vendor.conf to cri-containerd/vendor.conf. But will probably do it as an optional tool called from makefile vs forcing the merge on the install.deps command.

hack/install-deps.sh
```diff
@@ -140,12 +139,6 @@ fi
 checkout_repo ${CONTAINERD_PKG} ${CONTAINERD_VERSION} ${CONTAINERD_REPO}
 cd ${GOPATH}/src/${CONTAINERD_PKG}
 if ${COOK_CONTAINERD}; then
-  # Verify that vendor.conf is in sync with containerd before cook containerd,
```
Random-Liu: We do still need this for now.

mikebrow: I understand the desire to have a tool for testing whether they are in sync. Don't see why they have to be in 100% sync before allowing a PR though. Why hard requirement?

Random-Liu: @mikebrow The problem is that in following code, we copy cri-containerd vendor to containerd vendor directory directly.

If there is a go file in older version in containerd, but no such file in newer version in cri-containerd. This file will not be overwritten during the move, and containerd may not build because of this extra go file.

If we don't want to enforce, we need to at least print a warning here, so that if it doesn't build we know it might be caused by vendor not match.

mikebrow: ok... updating the PR to put it back in.. modifying the todo a bit to cover this part as well. Cheers.

PR #601, opened 2018-02-07
Support new cri storage

Random-Liu:
This depends on https://github.com/kubernetes/kubernetes/pull/59475.

This will get rid of the uuid thing, which addresses #509, #399 and #325. 

Return `mountpoint` as image filesystem storage identifier.


Random-Liu:
The PR is ready for review.

Random-Liu:
@yanxuean Thanks!

PR #611, opened 2018-02-14
Do not block on stream server close.

Random-Liu:
In e2e test, I sometimes see that when cri-containerd is starting, it is killed by health-monitor, and then stuck forever:
```
Started Kubernetes containerd CRI shim.
time="2018-02-13T02:26:55Z" level=info msg="Run cri-containerd "
time="2018-02-13T02:26:55Z" level=info msg="Start profiling server"
time="2018-02-13T02:26:55Z" level=info msg="Run cri-containerd grpc server on socket "/var/run/cri-containerd.sock""
time="2018-02-13T02:26:55Z" level=info msg="Get image filesystem path "/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs""
I0213 02:26:55.155330    2490 interface.go:360] Looking for default routes with IPv4 addresses
I0213 02:26:55.155351    2490 interface.go:365] Default route transits interface "eth0"
I0213 02:26:55.155458    2490 interface.go:174] Interface eth0 is up
I0213 02:26:55.155516    2490 interface.go:222] Interface "eth0" has 2 addresses :[10.128.0.5/32 fe80::4001:aff:fe80:5/64].
I0213 02:26:55.155538    2490 interface.go:189] Checking addr  10.128.0.5/32.
I0213 02:26:55.155548    2490 interface.go:196] IP found 10.128.0.5
I0213 02:26:55.155559    2490 interface.go:228] Found valid IPv4 address 10.128.0.5 for interface "eth0".
I0213 02:26:55.155569    2490 interface.go:371] Found active IP 10.128.0.5
time="2018-02-13T02:26:55Z" level=error msg="error updating cni config: Missing CNI default network"
time="2018-02-13T02:26:55Z" level=info msg="Initial CNI setting failed, continue monitoring: Missing CNI default network"
time="2018-02-13T02:26:55Z" level=info msg="Start cri-containerd service"
time="2018-02-13T02:26:55Z" level=info msg="Connect containerd service"
time="2018-02-13T02:26:55Z" level=info msg="Stop cri-containerd service"
time="2018-02-13T02:26:55Z" level=info msg="Start subscribing containerd event"
time="2018-02-13T02:26:55Z" level=info msg="Start recovering state"
time="2018-02-13T02:26:55Z" level=info msg="Start event monitor"
time="2018-02-13T02:26:55Z" level=info msg="Start snapshots syncer"
time="2018-02-13T02:26:55Z" level=info msg="Start streaming server"
time="2018-02-13T02:26:55Z" level=info msg="Start grpc server"
time="2018-02-13T02:26:55Z" level=error msg="Failed to serve grpc request" error="grpc: the server has been stopped"
time="2018-02-13T02:26:55Z" level=info msg="Stop cri-containerd service"
time="2018-02-13T02:26:55Z" level=error msg="Failed to handle event stream" error="rpc error: code = Canceled desc = context canceled"
time="2018-02-13T02:26:55Z" level=info msg="Event monitor stopped"
```

The reason is that there is a race condition between `http.Server.Serve()` and `http.Server.Close()`, which are used by stream server. (The issue https://github.com/golang/go/issues/20239). There is no reliable, race free way to wait for stream server to stop. So we add a timeout here. In most cases, we'll still wait for stream server to stop. However, if we hit the race condition, we'll wait for 2 seconds, throw out an error and continue stopping.

### DIFF
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -262,8 +262,21 @@ func (c *criContainerdService) Run(startGRPC bool) error {
 
 	<-eventMonitorCloseCh
 	logrus.Info("Event monitor stopped")
-	<-streamServerCloseCh
-	logrus.Info("Stream server stopped")
+	// There is a race condition with http.Server.Serve.
+	// When `Close` is called at the same time with `Serve`, `Close`
+	// may finish first, and `Serve` may still block.
+	// See https://github.com/golang/go/issues/20239.
+	// Here we set a 2 second timeout for the stream server wait,
+	// if it timeout, an error log is generated.
+	// TODO(random-liu): Get rid of this after https://github.com/golang/go/issues/20239
+	// is fixed.
+	const streamServerStopTimeout = 2 * time.Second
+	select {
+	case <-streamServerCloseCh:
+		logrus.Info("Stream server stopped")
+	case <-time.After(streamServerStopTimeout):
+		logrus.Errorf("Stream server is not stopped in %q", streamServerStopTimeout)
+	}
 	if startGRPC {
 		// Only wait for grpc server close channel when grpc server is started.
 		<-grpcServerCloseCh
--- a/cluster/gce/cloud-init/master.yaml
+++ b/cluster/gce/cloud-init/master.yaml
@@ -106,7 +106,6 @@ write_files:
       Restart=always
       RestartSec=10
       RemainAfterExit=yes
-      RemainAfterExit=yes
       ExecStartPre=/bin/chmod 544 /home/cri-containerd/opt/cri-containerd/cluster/health-monitor.sh
       ExecStart=/bin/bash -c 'CRICTL=/home/cri-containerd/usr/local/bin/crictl \
       /home/cri-containerd/opt/cri-containerd/cluster/health-monitor.sh'
@@ -175,7 +174,6 @@ write_files:
       [Service]
       Restart=always
       RestartSec=10
-      RemainAfterExit=yes
       RemainAfterExit=yes
       ExecStartPre=/bin/chmod 544 /home/kubernetes/bin/health-monitor.sh
       ExecStart=/home/kubernetes/bin/health-monitor.sh kubelet
--- a/cluster/gce/cloud-init/node.yaml
+++ b/cluster/gce/cloud-init/node.yaml
@@ -106,7 +106,6 @@ write_files:
       Restart=always
       RestartSec=10
       RemainAfterExit=yes
-      RemainAfterExit=yes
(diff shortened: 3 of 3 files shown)


Random-Liu:
Mark P0 because this causes a lot of test flakes.https://k8s-testgrid.appspot.com/sig-node-containerd#e2e-ubuntu

pkg/server/service.go
```diff
@@ -262,8 +262,21 @@ func (c *criContainerdService) Run(startGRPC bool) error {
...
-	logrus.Info("Stream server stopped")
+	// There is a race condition with http.Server.Serve.
+	// When `Close` is called at the same time with `Serve`, `Close`
+	// may finish first, and `Serve` may still block.
+	// See https://github.com/golang/go/issues/20239.
+	// Here we set a 2 second timeout for the stream server wait,
+	// if it timeout, an error log is generated.
+	// TODO(random-liu): Get rid of this after https://github.com/golang/go/issues/20239
+	// is fixed.
+	const streamServerStopTimeout = 2 * time.Second
```
mikebrow: if we called serve after close, they are deadlocked and the amount of time we wait for close is moot...

It seems from reading the golang issue that that we should not be calling serve after close and the golang mitigation will be to give us an error if we do. Maybe wrap it and stop ourselves from serve after close?

Random-Liu: > if we called serve after close, they are deadlocked and the amount of time we wait for close is moot...

1) In most cases, `Close` is called after `Serve`, in that case we do want to wait for stream server to stop here.
2) If we happen to hit the deadlock, there is no way to figure it out. The only solution is to add a deadlock here.

> It seems from reading the golang issue that we should not be calling serve after close and the golang mitigation in 1.11 will be to give us an error if we do.

There are 2 problems:
1) Even we add check before `ListenAndServe`, it's still possible that `Close` is called after `ListAndServe` is called but before the internal `Serve`. There is still race.
2) Even with the proposed golang fix, I still don't think it completely migrate the race condition. `Close` can still be called after `Serve` is called, but before `trackListener`.

I'll also comment there.

Random-Liu: Someone does mentioned what I said:
```
I don't think this fixes the original bug. What happens if the Shutdown call happens between line 2726 (Serve calls shuttingDown) and line 2739 (Serve calls trackListener)? We should add the original example from #20239 as a test: https://play.golang.org/p/ESmgJ6O_WU
```

mikebrow: > If we happen to hit the deadlock, there is no way to figure it out. 

Can we use a mutex to manage the call sequence via state progression/validation and to remove the race condition?

Random-Liu: No. Unless we can add mutex inside golang function.

PR #612, opened 2018-02-14
Add initial wait for health-monitor and use pkill -x.

Random-Liu:
Add `INITIAL_WAIT_SECONDS` for health-monitor. And use `pkill -x` to do exact match. https://gsp.com/cgi-bin/man.cgi?section=1&topic=pkill

### DIFF
--- a/cluster/health-monitor.sh
+++ b/cluster/health-monitor.sh
@@ -19,6 +19,11 @@ set -o pipefail
 
 # CRICTL is the path of crictl
 CRICTL=${CRICTL:-"crictl"}
+# INITIAL_WAIT_ATTEMPTS is the number to attempt, before start
+# performing health check. The problem is that cri-containerd
+# and containerd are started around the same time with health
+# monitor, they may not be ready yet when health-monitor is started.
+INITIAL_WAIT_ATTEMPTS=${INITIAL_WAIT_ATTEMPTS:-5}
 # COMMAND_TIMEOUT is the timeout for the health check command.
 COMMAND_TIMEOUT=${COMMAND_TIMEOUT:-60}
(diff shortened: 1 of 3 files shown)


cluster/health-monitor.sh
```diff
@@ -27,13 +32,14 @@ CHECK_PERIOD=${CHECK_PERIOD:-10}
 # and containerd.
 SLEEP_SECONDS=${SLEEP_SECONDS:-120}
 
+sleep ${INITIAL_WAIT_SECONDS}
```
mikebrow: readiness_check()

cluster/health-monitor.sh
```diff
@@ -27,13 +32,14 @@ CHECK_PERIOD=${CHECK_PERIOD:-10}
...
 while true; do
   # Use crictl pods because it requires both containerd and
   # cri-containerd to be working.
   if ! timeout ${COMMAND_TIMEOUT} ${CRICTL} pods > /dev/null; then
     echo "crictl pods timeout!"
-    pkill containerd
-    pkill cri-containerd
+    pkill -x containerd
```
mikebrow: why kill containerd first?

Random-Liu: Ha, yeah, we may want to kill cri-containerd first. Make sense.

cluster/health-monitor.sh
Random-Liu: Good point. We may not want to let this script rely on another script. So migrate the code here.

cluster/health-monitor.sh
Random-Liu: Done.

PR #616, opened 2018-02-21
Update critools and run critest in parallel.

Random-Liu:
Update cri-tools to 8ec2cbda446377cab11a5ef8ef92e115aa628a91 and run `critest` in parallel.


Random-Liu:
Note that this PR added `FASTBUILD` back for cni plugins. The change was reverted in https://github.com/containerd/cri-containerd/pull/607.

However, it is not the cause of test failure. So it seems safe to add it back to accelerate the build.

PR #618, opened 2018-02-22
Add generic restart test

Random-Liu:
Based on https://github.com/containerd/cri-containerd/pull/617 and another containerd fix (https://github.com/containerd/containerd/pull/2151).

This PR removed 2 original restart tests `TestSandboxAcrossCRIContainerdRestart` and `TestSandboxDeletionAcrossCRIContainerdRestart`, and added a new generic restart test `TestContainerdRestart`.

`TestSandboxAcrossCRIContainerdRestart` and `TestSandboxDeletionAcrossCRIContainerdRestart` assumes that cri-containerd and containerd are running as separate daemons, so there are a lot more corner cases to test.

However, since we are going to merge cri-containerd and containerd into one daemon. We can have a generic and simple restart test.


Random-Liu:
Let's wait for the containerd end to merge.

PR #628, opened 2018-02-28
Handle containerd event reliably

### LINKED ISSUE #434: Handle containerd event reliably
Forked from https://github.com/kubernetes-incubator/cri-containerd/issues/62.

Today, we rely on containerd event to figure out whether a container exits or not.

However, when handling an event, if there is an error, we'll just report error and discard the event, e.g. https://github.com/kubernetes-incubator/cri-containerd/blob/master/pkg/server/events.go#L91-L143.

Once that happens, the container will stay in running forever, although it already exists. We should:
1) Either to periodically relist to consolidate the state;
2) Or requeue the event, and handle it again.

yanxuean:
fix #434 

### DIFF
--- a/pkg/server/events.go
+++ b/pkg/server/events.go
@@ -18,19 +18,29 @@ package server
 
 import (
 	"errors"
+	"fmt"
+	"time"
 
 	eventtypes "github.com/containerd/containerd/api/events"
 	containerdio "github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/events"
 	"github.com/containerd/typeurl"
+	gogotypes "github.com/gogo/protobuf/types"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
+	"k8s.io/apimachinery/pkg/util/clock"
 
 	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
 	"github.com/containerd/cri/pkg/store"
 	containerstore "github.com/containerd/cri/pkg/store/container"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
+)
+
+const (
+	backOffInitDuration        = 1 * time.Second
+	backOffMaxDuration         = 5 * time.Minute
+	backOffExpireCheckDuration = 1 * time.Second
 )
 
 // eventMonitor monitors containerd event and updates internal state correspondingly.
@@ -43,6 +53,23 @@ type eventMonitor struct {
 	errCh          <-chan error
 	ctx            context.Context
 	cancel         context.CancelFunc
+	backOff        *backOff
+}
+
+type backOff struct {
+	queuePool     map[string]*backOffQueue
+	ticker        *time.Ticker
+	minDuration   time.Duration
+	maxDuration   time.Duration
+	checkDuration time.Duration
+	clock         clock.Clock
+}
+
+type backOffQueue struct {
+	events     []interface{}
+	expireTime time.Time
+	duration   time.Duration
+	clock      clock.Clock
 }
 
 // Create new event monitor. New event monitor will start subscribing containerd event. All events
@@ -55,6 +82,8 @@ func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store) *eventMonit
 		sandboxStore:   s,
 		ctx:            ctx,
 		cancel:         cancel,
+		backOff: newBackOff(backOffInitDuration, backOffMaxDuration,
+			backOffExpireCheckDuration, clock.RealClock{}),
 	}
 }
 
@@ -67,6 +96,24 @@ func (em *eventMonitor) subscribe(subscriber events.Subscriber) {
 	em.ch, em.errCh = subscriber.Subscribe(em.ctx, filters...)
 }
 
+func convertEvent(e *gogotypes.Any) (string, interface{}, error) {
+	containerID := ""
+	evt, err := typeurl.UnmarshalAny(e)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to unmarshalany %+v", err)
+	}
+
+	switch evt.(type) {
+	case *eventtypes.TaskExit:
+		containerID = evt.(*eventtypes.TaskExit).ContainerID
+	case *eventtypes.TaskOOM:
+		containerID = evt.(*eventtypes.TaskOOM).ContainerID
+	default:
+		return "", nil, fmt.Errorf("unsupported event")
+	}
+	return containerID, evt, nil
+}
+
 // start starts the event monitor which monitors and handles all container events. It returns
 // a channel for the caller to wait for the event monitor to stop. start must be called after
 // subscribe.
@@ -76,15 +123,40 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 	}
 	closeCh := make(chan struct{})
 	go func() {
+		backOffCheckCh := em.backOff.start()
 		for {
 			select {
 			case e := <-em.ch:
 				logrus.Debugf("Received containerd event timestamp - %v, namespace - %q, topic - %q", e.Timestamp, e.Namespace, e.Topic)
-				em.handleEvent(e)
+				cID, evt, err := convertEvent(e.Event)
+				if err != nil {
+					logrus.WithError(err).Errorf("Failed to convert event %+v", e)
+					break
+				}
+				if em.backOff.isInBackOff(cID) {
+					em.backOff.enBackOff(cID, evt)
+					break
+				}
+				if err := em.handleEvent(evt); err != nil {
+					logrus.WithError(err).Errorf("Failed to handle event %+v for container %s", evt, cID)
+					em.backOff.enBackOff(cID, evt)
+				}
 			case err := <-em.errCh:
 				logrus.WithError(err).Error("Failed to handle event stream")
 				close(closeCh)
 				return
+			case <-backOffCheckCh:
+				cIDs := em.backOff.getExpiredContainers()
+				for _, cID := range cIDs {
+					queue := em.backOff.deBackOff(cID)
+					for i, any := range queue.events {
+						if err := em.handleEvent(any); err != nil {
+							logrus.WithError(err).Errorf("Failed to handle backOff event %+v for container %s", any, cID)
+							em.backOff.reBackOff(cID, queue.events[i:], queue.duration)
+							break
+						}
+					}
+				}
 			}
 		}
 	}()
@@ -94,17 +166,13 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 // stop stops the event monitor. It will close the event channel.
 // Once event monitor is stopped, it can't be started.
 func (em *eventMonitor) stop() {
+	em.backOff.stop()
 	em.cancel()
 }
 
 // handleEvent handles a containerd event.
-func (em *eventMonitor) handleEvent(evt *events.Envelope) {
+func (em *eventMonitor) handleEvent(any interface{}) error {
 	ctx := ctrdutil.NamespacedContext()
-	any, err := typeurl.UnmarshalAny(evt.Event)
-	if err != nil {
-		logrus.WithError(err).Errorf("Failed to convert event envelope %+v", evt)
-		return
-	}
 	switch any.(type) {
 	// If containerd-shim exits unexpectedly, there will be no corresponding event.
 	// However, containerd could not retrieve container state in that case, so it's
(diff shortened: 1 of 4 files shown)


yanxuean:
/cc @Random-Liu   PTAL
I will add test case later.

Random-Liu:
Ref https://github.com/containerd/containerd/issues/2120.

pkg/server/events.go
```diff
@@ -79,11 +92,33 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
...
+				if cID, err := em.handleEvent(any); err != nil && cID != "" {
+					em.enBackOff(cID, any)
+				}
 			case err := <-em.errCh:
 				logrus.WithError(err).Error("Failed to handle event stream")
 				close(closeCh)
 				return
+			case cID := <-em.backOffExpire:
+				for {
+					any := em.deBackOff(cID)
```
Random-Liu: When will `any == nil`? Is that an error?

pkg/server/events.go
```diff
@@ -79,11 +92,33 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
...
+					if any == nil {
+						break
+					}
+					if _, err := em.handleEvent(any); err != nil && cID != "" {
```
Random-Liu: `cID != ""`? You don't even set it based on the return value.

pkg/server/events.go
```diff
@@ -236,9 +267,85 @@ func handleSandboxExit(e *eventtypes.TaskExit, sb sandboxstore.Sandbox) {
...
+		queue.timer = em.newBackOffTimer(key)
+	}
+	em.backOffQueue[key] = queue
+	return evt
+}
+
+// enBackOff start to backOff again and put event to the begin of queue
+func (em *eventMonitor) reBackOff(key string, evt interface{}) {
+	newEvents := []interface{}{evt}
+	queue := em.backOffQueue[key]
```
Random-Liu: Just requeue, and always stop timer as long as it is not `nil`.

No need to check length.

Random-Liu:
@yanxuean I feel like the per container timer logic is unnecessarily complex. And we haven't done real back yet. (See https://github.com/containerd/cri-containerd/pull/628#discussion_r172388420)

I feel like it is very easy to forget starting a timer, or leave an unnecessary timer over.

How about we just have a ticker which ticks every 1 second or so? And every time it ticks, we check whether there is any events which have exceed backoff time. And deal with them if there are any.

This seems more reliable to me. To optimize, we can only start the ticker when there is one event enqueued.

@yanxuean WDYT?

pkg/server/events.go
```diff
@@ -43,6 +45,14 @@ type eventMonitor struct {
 	errCh          <-chan error
 	ctx            context.Context
 	cancel         context.CancelFunc
+	backOffQueue   backOffQueue
```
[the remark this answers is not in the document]
yanxuean: will do

pkg/server/events.go
```diff
@@ -79,11 +92,33 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 			select {
 			case e := <-em.ch:
 				logrus.Debugf("Received containerd event timestamp - %v, namespace - %q, topic - %q", e.Timestamp, e.Namespace, e.Topic)
-				em.handleEvent(e)
+				any, err := typeurl.UnmarshalAny(e.Event)
+				if err != nil {
+					logrus.WithError(err).Errorf("Failed to convert event envelope %+v", e)
+					break
+				}
+				if cID, backOffIng := em.isBackOffIng(any); backOffIng {
```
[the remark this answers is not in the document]
yanxuean: will do

pkg/server/events.go
```diff
@@ -190,33 +222,32 @@ func handleContainerExit(e *eventtypes.TaskExit, cntr containerstore.Container)
...
 	if e.Pid != sb.Status.Get().Pid {
 		// Non-init process died, ignore the event.
-		return
+		return nil
 	}
 	// No stream attached to sandbox container.
 	task, err := sb.Container.Task(context.Background(), nil)
 	if err != nil {
 		if !errdefs.IsNotFound(err) {
 			logrus.WithError(err).Errorf("failed to load task for sandbox %q", e.ContainerID)
```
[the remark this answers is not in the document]
yanxuean: will do

pkg/server/events.go
```diff
@@ -79,11 +92,33 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
...
+				any, err := typeurl.UnmarshalAny(e.Event)
+				if err != nil {
+					logrus.WithError(err).Errorf("Failed to convert event envelope %+v", e)
+					break
+				}
+				if cID, backOffIng := em.isBackOffIng(any); backOffIng {
+					em.enBackOff(cID, any)
+					break
+				}
+				if cID, err := em.handleEvent(any); err != nil && cID != "" {
```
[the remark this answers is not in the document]
yanxuean: To prevent accidents:)  
will do

pkg/server/events.go
yanxuean: No, It is not error.   It is in "for" loop. When we handle over the whole queue, it will be nil.

pkg/server/events.go
yanxuean: will do

pkg/server/events.go
yanxuean: If then, We will stop timer twice when there is only one event.

yanxuean:
Will refactor it.

yanxuean:
@Random-Liu  PTAL, Tks

pkg/server/events.go
```diff
@@ -93,16 +138,12 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
...
 }
 
 // handleEvent handles a containerd event.
-func (em *eventMonitor) handleEvent(evt *events.Envelope) {
-	any, err := typeurl.UnmarshalAny(evt.Event)
-	if err != nil {
-		logrus.WithError(err).Errorf("Failed to convert event envelope %+v", evt)
-		return
-	}
+func (em *eventMonitor) handleEvent(any interface{}) error {
```
Random-Liu: Let's return error with description from this function, and we only need to log the error in `start`.

In this way, we don't need so many logs in this function.

pkg/server/events.go
```diff
@@ -236,10 +278,78 @@ func handleSandboxExit(e *eventtypes.TaskExit, sb sandboxstore.Sandbox) {
...
+	case *eventtypes.TaskOOM:
+		key = evt.(*eventtypes.TaskOOM).ContainerID
+	default:
+		return key, false
+	}
+	return key, len(b.queues[key].events) != 0
+}
+
+// enBackOff start to backOff and put event to the tail of queue
+func (b *backOff) enBackOff(key string, evt interface{}) {
```
Random-Liu:
```golang
queue, ok := b.queues[key]
if ok {
  queue.events = append(queue.events, evt)
  return
}
b.queues[key] = backOffQueue{
  events: []interface{}{evt},
  duration: initialBackoffTime,
  start: time.Now(),
} // Or add a function `newBackoffQueue`
```

pkg/server/events.go
```diff
@@ -236,10 +278,78 @@ func handleSandboxExit(e *eventtypes.TaskExit, sb sandboxstore.Sandbox) {
...
+
+// enBackOff get out the whole queue
+func (b *backOff) deBackOff(key string) backOffQueue {
+	queue := b.queues[key]
+	delete(b.queues, key)
+	return queue
+}
+
+// enBackOff start to backOff again and put [nth:] events to the queue
+func (b *backOff) reBackOff(key string, queue backOffQueue, n int) {
```
Random-Liu:
```golang
func (b *backOff) reBackOff(key string, events []interface{}, oldDuration time.Duration) {
  duration := 2 * oldDuration
  if duration > maxBackoffTime {
    duration = maxBackoffTime
  }
  b.queues[key] = backOffQueue{
    events: events,
    duration: duration,
    start: time.Now(),
  }
}
```

pkg/server/events.go
```diff
@@ -236,10 +278,78 @@ func handleSandboxExit(e *eventtypes.TaskExit, sb sandboxstore.Sandbox) {
...
+
+func (b *backOff) stopBackOff() {
+	b.ticker.Stop()
+}
+
+func (t backOffTime) isExpire(now time.Time) bool {
+	return !now.Before(t.expireTime)
+}
+
+func (t backOffTime) newTime() backOffTime {
```
Random-Liu: I think this function is a little bit unnecessary complex... Simple and straightforward code is preferred. See comments above.

pkg/server/events.go
```diff
@@ -236,10 +278,78 @@ func handleSandboxExit(e *eventtypes.TaskExit, sb sandboxstore.Sandbox) {
...
 	// Using channel to propagate the information of sandbox stop
 	sb.Stop()
+	return nil
+}
+
+func (b *backOff) getExpiredContainers() []string {
+	var containers []string
+	now := time.Now()
+	for c, v := range b.queues {
+		if v.time.isExpire(now) {
```
Random-Liu:
```golang
if time.Since(v.start) > v.duration // or add a `isExpired` function for backOffQueue
```

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+		},
+		"container1": {
+			events: []interface{}{
+				&eventtypes.TaskExit{ContainerID: "container1", ID: "1"},
+				&eventtypes.TaskExit{ContainerID: "container1", ID: "2"},
+			},
+		},
+	}
+
+	t.Logf("Should can backOff a event")
```
Random-Liu: s/can/be able to

And also for following ones.

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+	actual := newEventMonitor(nil, nil).backOff
```
Random-Liu: ? I prefer using the `new` function suggested in https://github.com/containerd/cri/pull/628/files#r174375426.

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+	for k, queue := range inputQueues {
+		for _, event := range queue.events {
+			actual.enBackOff(k, event)
+		}
+	}
+	assert.Equal(t, isQueueListEqual(t, actual.queues, expectedQueues, 1*time.Second), true)
+
+	t.Logf("Should can check if the container is on backOff state")
```
Random-Liu: s/on/in

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
+/*
```
Random-Liu: Use a fake clock for the unit test.
https://github.com/containerd/cri/tree/master/vendor/k8s.io/apimachinery/pkg/util/clock

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+Copyright 2018 The Kubernetes Authors.
```
Random-Liu: containerd Authors

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	t.Logf("Should can backOff a event")
+	actual := newEventMonitor(nil, nil).backOff
+	for k, queue := range inputQueues {
+		for _, event := range queue.events {
+			actual.enBackOff(k, event)
+		}
+	}
+	assert.Equal(t, isQueueListEqual(t, actual.queues, expectedQueues, 1*time.Second), true)
```
Random-Liu: Use fake clock, then you can use `assert.Equal` directly.

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	t.Logf("Should can check if the container is on backOff state")
+	for k, queue := range inputQueues {
```
Random-Liu: Should check an arbitrary container id is not in backoff.

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, e := range queue.events {
+			key, isInBackOff := actual.isInBackOff(e)
+			assert.Equal(t, k, key)
+			assert.Equal(t, isInBackOff, true)
+		}
+	}
+
+	t.Logf("Should can get all keys who are expired for backOff")
```
Random-Liu: s/who/which

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+	time.Sleep(1 * time.Second)
+	var expKeyList []string
+	for k := range inputQueues {
+		expKeyList = append(expKeyList, k)
+	}
+	actKeyList := actual.getExpiredContainers()
+	assert.Equal(t, len(expKeyList), len(actKeyList))
+	for _, expKey := range expKeyList {
+		found := false
```
Random-Liu: assert.Contains

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	t.Logf("Should can get out all backOff events")
+	doneQueues := map[string]backOffQueue{}
+	for k := range inputQueues {
+		actQueue := actual.deBackOff(k)
+		doneQueues[k] = actQueue
+		assert.Equal(t, isQueueEqual(t, actQueue, expectedQueues[k], 1*time.Second), true)
+	}
+
+	t.Logf("Should not get out the event again after having gut out the backOff event")
```
Random-Liu: s/gut/got.

pkg/server/events_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	t.Logf("Should can reBackOff")
+	for k, queue := range doneQueues {
+		failEventIndex := 1
+		actual.reBackOff(k, queue, failEventIndex)
+		actQueue := actual.deBackOff(k)
+		expQueue := backOffQueue{
+			events: queue.events[failEventIndex:],
+		}
+		assert.Equal(t, isQueueEqual(t, actQueue, expQueue, 2*time.Second), true)
```
Random-Liu: Use constant for code and test.

pkg/server/events.go
```diff
@@ -54,6 +72,9 @@ func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store) *eventMonit
 		sandboxStore:   s,
 		ctx:            ctx,
 		cancel:         cancel,
+		backOff: backOff{
```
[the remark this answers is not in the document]
yanxuean: will do

pkg/server/events_test.go
yanxuean: The assert.Equal is very powerful except for slice.

pkg/server/events.go
```diff
@@ -55,6 +82,8 @@ func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store) *eventMonit
 		sandboxStore:   s,
 		ctx:            ctx,
 		cancel:         cancel,
+		backOff: newBackOff(backOffInitDuration, backOffMaxDuration,
```
Random-Liu: I think we can just use the const instead of parameterize it.

pkg/server/events.go
```diff
@@ -76,15 +123,40 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
...
 			select {
 			case e := <-em.ch:
 				logrus.Debugf("Received containerd event timestamp - %v, namespace - %q, topic - %q", e.Timestamp, e.Namespace, e.Topic)
-				em.handleEvent(e)
+				cID, evt, err := convertEvent(e.Event)
+				if err != nil {
+					logrus.WithError(err).Errorf("Failed to convert event %+v", e)
+					break
+				}
+				if em.backOff.isInBackOff(cID) {
```
Random-Liu: Add an info log here.

pkg/server/events.go
```diff
@@ -94,69 +166,73 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
...
-			if _, err := em.sandboxStore.Get(e.ContainerID); err == nil {
-				return
+			if err != store.ErrNotExist {
+				return fmt.Errorf("can't find container for TaskOOM event:%+v", err)
+			}
+			if _, err = em.sandboxStore.Get(e.ContainerID); err != nil {
+				if err != store.ErrNotExist {
+					return fmt.Errorf("can't find sandbox for TaskOOM event:%+v", err)
+				}
+				return nil
```
Random-Liu: I think we can remove this return.

pkg/server/events_test.go
```diff
@@ -0,0 +1,137 @@
...
+			assert.NoError(t, err)
+			assert.Equal(t, k, key)
+			assert.Equal(t, actual.isInBackOff(key), true)
+		}
+	}
+
+	t.Logf("Should be able to check that a container isn't in backOff state")
+	notExistKey := "containerNotExist"
+	assert.Equal(t, actual.isInBackOff(notExistKey), false)
+
```
Random-Liu: Should add a test case that `getExpiredContainers` should be empty when not expired.

pkg/server/events_test.go
```diff
@@ -0,0 +1,137 @@
...
+	expKeyMap := map[string]struct{}{}
+	for k := range inputQueues {
+		expKeyMap[k] = struct{}{}
+	}
+	actKeyList := actual.getExpiredContainers()
+	actKeyMap := map[string]struct{}{} //assert.Equal can't compare slice without order
+	for _, k := range actKeyList {
+		actKeyMap[k] = struct{}{}
+	}
+	assert.Equal(t, actKeyMap, expKeyMap)
```
Random-Liu: We can compare length, and use `assert.Contains`.

Random-Liu:
@yanxuean Thanks a lot! Good job! :D

Issue #626, opened 2018-02-27
Containerd drops support for golang 1.9.x


mikebrow:
See https://github.com/containerd/containerd/pull/2162

They have pulled in a golang 1.10 library.. so now need 1.10 to build.

Random-Liu:
Let's do the same. :)

Random-Liu:
Let's close this issue after containerd build in Kubernetes test infra is also switched to golang 1.10.

Random-Liu:
We can close this now. We've changed Kubernetes test-infra to build cri-containerd with golang 1.10.

PR #629, opened 2018-02-28
moving to go1.10 only

mikebrow:
addressing issue #626

From here on in we can start using 1.10 only golang features and thus will only test on 1.10 and up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,12 +28,6 @@ before_script:
 jobs:
   include:
     - stage: Build
-      script:
-        - make install.tools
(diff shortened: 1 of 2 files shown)


Random-Liu:
Talked with our test-infra team. I'll change cri-containerd build to use golang 1.10, but kubernetes build and test will still use golang 1.9.

This should be good enough for now.

mikebrow:
> Talked with our test-infra team. I'll change cri-containerd build to use golang 1.10, but kubernetes build and test will still use golang 1.9.

Ok.

PR #630, opened 2018-02-28
Add CLI package for `ctr` to use.

Random-Liu:
For https://github.com/containerd/cri-containerd/issues/590.

Add `cli` package for `ctr` to use.

This simplifies the command description a lot to make it similar with `ctr`, e.g. [`remove`](https://github.com/containerd/containerd/blob/master/cmd/ctr/commands/images/images.go#L286).

@mikebrow Are you ok with that? Or do you have any suggestion?

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,0 +1,78 @@
+/*
+Copyright 2018 The containerd Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


mikebrow:
What do you think about modifying the existing ctr image pull command? Maybe add a -withcri option?

Random-Liu:
> What do you think about modifying the existing ctr image pull command? Maybe add a -withcri option?

`ctr image pull` talks with containerd. `ctr cri` talks with CRI plugin. :)

And I don't think we should introduce CRI thing into existing generic ctr commands.

mikebrow:
> And I don't think we should introduce CRI thing into existing generic ctr commands.

yeah let's think it over as I'm not clear which is the more intrusive introduction.. a new ctr command for cri or a new option in image pull for going through cri to pull.. Maybe the cri option should be silent and only used when cri plugin is there I don't think cri has a whole lot of overhead? Might be more confusing to have two different pull/load commands :-)

Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/630#pullrequestreview-103662569

PR #632, opened 2018-03-01
tests: Use path found to containerd.

jcvenegas:
After check if containerd is intalled the command executed using sudo,
/usr/local/bin may not been in the root's PATH, save the path where
containerd was found and use it.

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/hack/test-utils.sh
+++ b/hack/test-utils.sh
@@ -40,13 +40,16 @@ test_setup() {
   fi
 
   # Start containerd
-  if [ ! -x "$(command -v containerd)" ]; then
+  local containerd=$(command -v containerd)
+  if [ ! -x "${containerd}" ]; then
     echo "containerd is not installed, please run hack/install-deps.sh"
     exit 1
   fi
   sudo pkill -x cri-containerd
   sudo pkill -x containerd
-  keepalive "sudo containerd" ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
+  echo "using ${containerd}"
+  echo "containerd version: $(${containerd} --version)"
+  keepalive "sudo ${containerd}" ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
   containerd_pid=$!
   # Wait for containerd to be running by using the containerd client ctr to check the version
   # of the containerd server. Wait an increasing amount of time after each of five attempts


hack/test-utils.sh
```diff
@@ -40,13 +40,14 @@ test_setup() {
   fi
 
   # Start containerd
-  if [ ! -x "$(command -v containerd)" ]; then
```
mikebrow: How about just using (which sudo containerd) here? then in the comment below  mention it's being run in sudo...

"containerd is not installed, please run hack/install-deps.sh [and check that containerd is on your sudo path as containerd must be run as root for now]."

hack/test-utils.sh
```diff
@@ -40,13 +40,14 @@ test_setup() {
...
+  containerd=$(command -v containerd)
+  if [ ! -x "${containerd}" ]; then
     echo "containerd is not installed, please run hack/install-deps.sh"
     exit 1
   fi
   sudo pkill -x cri-containerd
   sudo pkill -x containerd
-  keepalive "sudo containerd" ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
+  keepalive "sudo ${containerd}" ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
```
mikebrow: Agree we should test for and run the same containerd, but a bit concerned someone may have two versions of containerd one on his sudo path and another on his current user path and thus this change might introduce the situation where we are not running the one he thinks we are running. For example, there might be an older version of containerd over in "home/username/go/bin"

Random-Liu: I think this change makes sense.

If you are worried about the 2 containerd case, can we `echo` something after `command -v containerd` to tell developer what containerd is being used? e.g.
```
using /usr/local/bin/containerd
```

mikebrow: echoing the path is nice as is echoing the version used.. 

But would still rather not use a mix of of execution paths.. if we are going to run with the user path use `PATH=$PATH` otherwise IMO it's better to just fix the test condition.

Random-Liu: That also works for me, echo + `PATH=$PATH`.

Random-Liu: @jcvenegas Please echo the path of containerd we are using in the test.
And then I think the PR is good to go. :)

hack/test-utils.sh
jcvenegas: @Random-Liu  @mikebrow updated to echo containerd path and version to be used, I prefer to resolve the path to containerd.

If we use `sudo  which containerd` in the condition, later it may not be possible to  found it because sudo's `secure_path`  might not be included `/usr/loca/bin` ( fedora does not include `/usr/loca/bin`  in sudo `secure_path`)

Alternatively we can modify the line when containerd is executed for something like : 
```bash
keepalive "sudo PATH=$PATH bash -c 'containerd'" ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
```
And add sudo to this conditional as you suggest (because `command` is a shell command it will use the shell PATH env variable and not sudo's `secure_path` ) 
```
if [ ! -x "$(sudo command -v containerd)" ]; then
```

hack/test-utils.sh
jcvenegas: @Random-Liu PR updated, @mikebrow added a comment as alternative to do not mix execution paths. Let me know if you prefer the alternative proposal.

hack/test-utils.sh
mikebrow: prefer your alternative but it's merged.. It's a testutil script not a production one, should be fine :-)

PR #635, opened 2018-03-02
fix travis golang issue

mikebrow:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,14 +31,14 @@ before_script:
 jobs:
   include:
     - stage: Build
-    - script:
+      script:
         - make install.tools
         - make .gitvalidation
         - make binaries
         - make containerd
       go: "1.10"
     - stage: Test
-    - script:
+      script:
         - make install.deps
         - make test
         - make test-integration


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: go
 
+go:
+  - "1.10"
```
Random-Liu: We don't need this after removing `-` I guess.

Random-Liu:
You also need to remove the `-` under the second stage

.travis.yml
```diff
@@ -28,7 +28,7 @@ before_script:
 jobs:
   include:
     - stage: Build
-    - script:
+      script:
```
Random-Liu:
```
     - stage: Test
     - script:
```
->
```
     - stage: Test
       script:
```

mikebrow: yeah tried it that way once and didn't get the right result..

mikebrow: probably because I had not yet removed the -go.. looks good now

mikebrow:
Looks good now...

PR #637, opened 2018-03-02
adds allways to ensure there is a version even if no tag

mikebrow:
If no tag is "found" we should at least fall back to the commit vs setting the version to nil. 

Maybe the root issue is the depth of the git clone, it's set to 50. Setting the clone depth to 150 and leaving the version change in for people that want a shallow clone or don't fetch tags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ BUILD_DIR := _output
 # VERSION is derived from the current tag for HEAD plus amends. Version is used
 # to set/overide the CRIContainerdVersion variable in the verison package for
 # cri-containerd.
-VERSION := $(shell git describe --tags --dirty)
+VERSION := $(shell git describe --tags --dirty --always)
 # strip the first char of the tag if it's a `v`
 VERSION := $(VERSION:v%=%)
 TARBALL_PREFIX := cri-containerd
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+git:
+  depth: 150
+
 language: go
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ BUILD_DIR := _output
 # VERSION is derived from the current tag for HEAD plus amends. Version is used
 # to set/overide the CRIContainerdVersion variable in the verison package for
 # cri-containerd.
-VERSION := $(shell git describe --tags --dirty)
+VERSION := $(shell git describe --tags --dirty --always)
```
Random-Liu: I'm fine with keeping `always` here. But we should keep the unit test unchanged, so that we can catch this issue again in the futrue

mikebrow: Not really an issue more a feature.. shouldn't break a pr or private build due missing tag...

Just have to remember to push a tag when you are going to produce a new release. And if you want to see the new release tag on your clone you have to fetch --tags.

mikebrow: The unit test is just unit testing whether we can semver the version tag it's not unit testing whether the tag is correct or ensuring the version is only a tag.  I'm thinking this is useful for local developers.

mikebrow: At one point unknown was the default.. dunno. I like it this way :-)

Random-Liu: @mikebrow The version will be used by our CI verison of containerd as well. And kubernetes does require the version to be semver.

mikebrow: done.

PR #643, opened 2018-03-05
Fix a potential panic.

Random-Liu:
We should not continue when there is an error. It may cause a panic.
Found this when reviewing https://github.com/containerd/cri-containerd/pull/628.

Ref https://github.com/containerd/cri-containerd/issues/642.


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri-containerd/pull/643#pullrequestreview-101358143

yanxuean:
good catch

PR #644, opened 2018-03-06
[in-process-integration] Start switching to direct function call

Random-Liu:
@crosbymichael @mikebrow Let's start from this PR. Once we are ok with this PR, we can port services one by one without interfering each other.

This PR:
1) Added a `Services` interface in `Client.go`. And added grpcServices implementation and also a directServices implementation. Suggestions for naming are welcome.
2) Added `k8s.io` namespace for all contexts in cri plugin. We can also deal with default namespace in containerd client, but without grpc it requires too many changes. It might be easier to let caller handle namespace if direct services are used.
3) Change CRI plugin to use internal content store.

Please note that **the internal content store doesn't publish events**. Should we fix it? Or we are ok with it?
/cc @containerd/containerd-maintainers

### DIFF
--- a/cri.go
+++ b/cri.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"path/filepath"
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/containerd/plugin"
@@ -28,6 +29,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	criconfig "github.com/containerd/cri-containerd/pkg/config"
+	"github.com/containerd/cri-containerd/pkg/constants"
 	"github.com/containerd/cri-containerd/pkg/server"
 )
 
@@ -83,7 +85,23 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	// Use a goroutine to initialize cri service. The reason is that currently
 	// cri service requires containerd to be initialize.
 	go func() {
-		if err := s.Run(); err != nil {
+		// Connect containerd service here, to get rid of the containerd dependency.
+		// in `NewCRIContainerdService`.
+		log.G(ctx).Info("Connect containerd service")
+		// TODO(random-liu): Create client in `initCRIService` and pass it to
+		// `NewCRIContainerdService` (containerd/containerd#2183)
+		ctrdServices, err := containerd.NewLocalServices(
+			ic.Address,
+			constants.K8sContainerdNamespace,
+		)
+		if err != nil {
+			log.G(ctx).WithError(err).Fatal("Failed to create direct containerd services")
+		}
+		client, err := containerd.NewWithServices(ctrdServices)
+		if err != nil {
+			log.G(ctx).WithError(err).Fatal("Failed to initialize containerd client")
+		}
+		if err := s.Run(client); err != nil {
 			log.G(ctx).WithError(err).Fatal("Failed to run CRI service")
 		}
 		// TODO(random-liu): Whether and how we can stop containerd.
--- a/cri.go
+++ b/cri.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/log"
+	"github.com/containerd/containerd/metadata"
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/containerd/plugin"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -82,6 +83,13 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 		return nil, errors.Wrap(err, "failed to create CRI service")
 	}
 
+	meta, err := ic.Get(plugin.MetadataPlugin)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get metadata plugin")
+	}
+	db := meta.(*metadata.DB)
+	contentStore := db.ContentStore()
+
 	// Use a goroutine to initialize cri service. The reason is that currently
 	// cri service requires containerd to be initialize.
 	go func() {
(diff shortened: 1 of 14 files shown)


vendor/github.com/containerd/containerd/client.go
```diff
@@ -75,71 +68,75 @@ func New(address string, opts ...ClientOpt) (*Client, error) {
...
-		return errors.New("unable to reconnect to containerd, no connector available")
-	}
-	c.conn.Close()
-	conn, err := c.connector()
-	if err != nil {
-		return err
-	}
-	c.conn = conn
-	return nil
+	Services
```
mikebrow: should this be [s]ervices?

vendor/github.com/containerd/containerd/grpc_services.go
```diff
@@ -0,0 +1,170 @@
...
+		conn: conn,
+	}, nil
+}
+
+// Reconnect re-establishes the GRPC connection to the containerd daemon
+func (g *grpcServices) Reconnect() error {
+	if g.connector == nil {
+		return errors.New("unable to reconnect to containerd, no connector available")
+	}
+	g.conn.Close()
```
mikebrow: I see an error may be returned from Close() but the error check wasn't here in the original version. An error indicates close was already called. Not checking indicates we don't need to call close if it's already been called cause we did not care about the error. If it's already been called and we call close it cancels the context (presumably again, note context also uses a mutex sem and returns early if it's already cancelled), locks a mutex semaphore that protects the connection data, tests if conn.conns is nil, if it is nil it releases the mutex and returns ErrClientConnClosing. In short.. we don't care if it's called twice.

suggest adding an error check here if err != ErrClientConnClosing { return ("unable to reconnect, unexpected error while closing connection...

vendor/github.com/containerd/containerd/direct_services.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/containerd/containerd/dialer"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/snapshots"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health/grpc_health_v1"
+)
+
+// directServices is an implementation of Services through direct function call.
```
mikebrow: todo.. make all direct services ... direct, then remove connection.

mlaventure:
Had a quick look, here are my suggestions:
 - Can we make the Services part of the `ClientOpts` ? Then use a `WithServices`?
 - The only difference between `DirectServices` & `GrpcServices` seems to be the access to the content store. Should we have `DirectServices` embedded `GrpcServices` and overload the content store to avoid code duplication
 - I feel like we want to have the events for the internal content store :)
 - For `DirectServices` it feels like that would basically embedded all the services, not just the content store. So would need a better name here to make this obvious. If it were to have all service embedded, I would have proposed `WithLocalServices` but for just one, I'm not sure how to make it clear. Unless we have a `NewService(WithContentStore(...))` that basically return all the services as default but allow overriding each one via options.

cri.go
```diff
@@ -82,6 +83,13 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 		return nil, errors.Wrap(err, "failed to create CRI service")
 	}
 
+	meta, err := ic.Get(plugin.MetadataPlugin)
```
mikebrow: should this new block of code setting the contentStore variable be in the go func()?

Random-Liu:
> Can we make the Services part of the ClientOpts ? Then use a WithServices?

`address` won't make sense anyway, so we still need a separate function for this. :) Unless we make `address` an option as well.

> The only difference between DirectServices & GrpcServices seems to be the access to the content store. Should we have DirectServices embedded GrpcServices and overload the content store to avoid code duplication

We are going to port all services in `DirectServices` to use internal interface. There will be many more differences soon.

> I feel like we want to have the events for the internal content store :)

I agree. Don't know what others think about this.

> For DirectServices it feels like that would basically embedded all the services, not just the content store. So would need a better name here to make this obvious. If it were to have all service embedded, I would have proposed WithLocalServices but for just one, I'm not sure how to make it clear. Unless we have a NewService(WithContentStore(...)) that basically return all the services as default but allow overriding each one via options.

`LocalServices` is a better name. :) We can cleanup the arg list after we make it work. That's why I have this TODO:
```golang
// TODO(random-liu): Design the arg list better, probably
// need some option functions. (containerd/containerd#2183)
```

vendor/github.com/containerd/containerd/client.go
```diff
@@ -75,71 +68,75 @@ func New(address string, opts ...ClientOpt) (*Client, error) {
...
 	return &Client{
-		conn:    conn,
-		runtime: fmt.Sprintf("%s.%s", plugin.RuntimePlugin, runtime.GOOS),
+		Services: services,
+		runtime:  fmt.Sprintf("%s.%s", plugin.RuntimePlugin, runtime.GOOS),
 	}, nil
 }
 
+// Services are all required underlying services by Client.
+type Services interface {
```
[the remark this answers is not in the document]
Random-Liu: This is actually server services for local services.

vendor/github.com/containerd/containerd/grpc_services.go
Random-Liu: This function will be empty for local services.

If you'd like you could send a PR to fix containerd if you think it needs a fix. :)

vendor/github.com/containerd/containerd/direct_services.go
Random-Liu: Will do.

cri.go
Random-Liu: Nope. After all services become local services, we'll pass the client into `NewCRIContainerdService`, which is out of `go func()`.

vendor/github.com/containerd/containerd/client.go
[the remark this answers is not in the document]
mikebrow: Yeah just that services is so overloaded a term.

vendor/github.com/containerd/containerd/direct_services.go
Random-Liu: Done

crosbymichael:
Looks like a good start.  I don't think you need more abstractions like LocalService/DirectServices, just port the code over.  They are all going to be the same, in proc apis, so why make more abstractions? 

LGTM

Random-Liu:
> Looks like a good start. I don't think you need more abstractions like LocalService/DirectServices, just port the code over. They are all going to be the same, in proc apis, so why make more abstractions?

OK. Will clean up the client code after we make it work. :)

PR #650, opened 2018-03-07
Handles 401 in resolver.

Random-Liu:
Fixes https://github.com/containerd/cri-containerd/issues/649.
Ref https://github.com/containerd/cri-containerd/issues/642.

GCR sometimes returns http 401 since this noon, which caused many test failures, e.g. https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/ci-cri-containerd-e2e-gci-gce/2908. 

I've filed an internal issue to them. Before that is fixed let's port the fix https://github.com/containerd/containerd/pull/2171 here, which we need anyway.

@abhi 

### DIFF
--- a/pkg/containerd/resolver/resolver.go
+++ b/pkg/containerd/resolver/resolver.go
@@ -501,7 +501,7 @@ func (r *dockerBase) fetchTokenWithOAuth(ctx context.Context, to tokenOptions) (
 
 	// Registries without support for POST may return 404 for POST /v2/token.
 	// As of September 2017, GCR is known to return 404.
-	if (resp.StatusCode == 405 && r.username != "") || resp.StatusCode == 404 {
+	if (resp.StatusCode == 405 && r.username != "") || resp.StatusCode == 404 || resp.StatusCode == 401 {
 		return r.getToken(ctx, to)
 	} else if resp.StatusCode < 200 || resp.StatusCode >= 400 {
 		b, _ := ioutil.ReadAll(io.LimitReader(resp.Body, 64000)) // 64KB


pkg/containerd/resolver/resolver.go
```diff
@@ -501,7 +501,7 @@ func (r *dockerBase) fetchTokenWithOAuth(ctx context.Context, to tokenOptions) (
 
 	// Registries without support for POST may return 404 for POST /v2/token.
 	// As of September 2017, GCR is known to return 404.
-	if (resp.StatusCode == 405 && r.username != "") || resp.StatusCode == 404 {
+	if (resp.StatusCode == 405 && r.username != "") || resp.StatusCode == 404 || resp.StatusCode == 401 {
```
abhi: should this be if r.username != "" && resp.StatusCode == 401 ?

abhi: can you dump the resp ? It should have some sort of challenge in the header on why it failed coz 401 auth related header.

Random-Liu: This just ports containerd/containerd#2171 to here.

abhi:
Lets open an issue to see if there is a follow up fix needed.

ruediger-maass:
I wasn't aware that a copy of resolver is used here, too. Otherwise, I would have applied the PR https://github.com/containerd/containerd/pull/2171 also here in cri-containerd. Please keep the change for handling 401 in the resolver of cri-containerd even if GCR would not respond with 401's anymore - we need it for Artifactory. Thanks!

PR #653, opened 2018-03-07
[in-process-integration] Support event publish and use internal image store and container store.

Random-Liu:
This PR:
1) Added a new plugin `ServicePlugin`. The extra layer is especially useful for services which require more than one existing plugins, such as images, tasks etc.
2) Add internal service implementation for content and snapshot which have proper events published.
3) Add internal image service.
4) Add internal container service.


cri.go
```diff
@@ -127,6 +135,37 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	return s, nil
 }
 
+// getServices get service instances from plugin context.
+func getServices(ic *plugin.InitContext) (map[string]interface{}, error) {
```
dmcgowan: Could this function just return the service options? This would prevent haven't to handle the list of services in 2 places

vendor/github.com/containerd/containerd/services/snapshots/snapshotters.go
```diff
@@ -0,0 +1,98 @@
...
+func (s *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...snapshots.Opt) ([]mount.Mount, error) {
+	mounts, err := s.Snapshotter.Prepare(ctx, key, parent, opts...)
+	if err != nil {
+		return nil, err
+	}
+	if err := s.publisher.Publish(ctx, "/snapshot/prepare", &eventstypes.SnapshotPrepare{
+		Key:    key,
+		Parent: parent,
+	}); err != nil {
+		return nil, err
```
dmcgowan: Was the conclusion here that we want to change these to error logs?

Random-Liu: It seems that there will be error only when there is a bug, e.g. failed to `Unmarshal` or something.

So I think it is fine to return an error here. :)

cri.go
Random-Liu: Let me see. Need to a table to map the service to corresponding option and type.

Random-Liu: Done.

Random-Liu:
Close this in favor of https://github.com/containerd/cri-containerd/pull/658

Issue #574, opened 2018-01-25
Allow define runtime for host privileged operations


jcvenegas:
### Problem statement ###

Some CRI compatible runtimes may not support provileged                                                                                                                                                                                       
operations. Specifically hypervisor based runtimes  (like kata-containers,                                                                                                                                                                     
cc-runtime and runv) do not support privileged operations like:                                                                                                                                                                                      
                                                                                                                                                                                                                                              
- Provide access to the host namespaces                                                                                                                                                                                                       
- Create fully privileged containers with access to  host devices.                                                                                                                                                                      
                                                                                                                                                                                                                                              
Hypervisor based runtimes create  container workloads within virtual machines. When running host                                                                                                                                                                        
privileged containers with them,  they wont  provide support to requested the `privileged opertations`.                                                                                                                                                                                                                    
 
### Proposed solution ###                                                                                                                                                                                                                         
To solve this problem a proposed solution is to  allow CRI implementations                                                                                                                                                                       
like cri-containerd to define two runtimes:                                                                                                                                                                                                   
                                                                                                                                                                                                                                              
- `Host privileged runtime` :   Used when a privileged container  is requested.                                                                                           
- `Default runtime` : for non-privileged workloads.                                                                                                                                                                                         
                                                                                                                                                                                                                                              
When cri-containerd implementation gets a request for creating a sandbox. It should                                                                                                                                                                    
classify the sandbox as :                                                                                                                                                                                                                     
                                                                                                                                                                                                                                              
- `trusted/privileged` Sandbox                                                                                                                                                                                                                
                                                                                                                                                                                                                                              
Witch has either the privileged flag set or access to at least one of the                                                                                                                                                                     
host namespaces (PID, IPC or networking).                                                                                                                                                                                                     
                                                                                                                                                                                                                                              
- `untrusted/non-privileged` Sandbox:                                                                                                                                                                                                         
                                                                                                                                                                                                                                              
Any other Sandbox that is not a `trusted/privileged` Sandbox.                                                                                                                                                                                   
                                                                                                                                                                                                                                              
Then cri-containerd will check if a `Host privileged runtime` is defined and use
it to create `trusted/privileged` Sanbox. It will use the                                                                                                                                                                               
`default runtime`([RuntimeEngine](https://github.com/containerd/cri-containerd/blob/dca05358dccc6581467be200fcc15d408bd18444/cmd/cri-containerd/options/options.go#L52)) if the host privileged one is not defined.

### Expected result ###

[...]

PR #657, opened 2018-03-12
[WIP]config: Allow to define trusted runtime

jcvenegas:
This PR adds:
- Trusted runtime configuration option.
- Allow to change contaienrd configuration for testing.

Some CRI compatible runtimes may not support privileged operations.
Specifically hypervisor based runtimes (like kata-containers, cc-runtime
and runv) do not support privileged operations like:

- Provide access to the host namespaces
- Create fully privileged containers with access to host devices

Hypervisor based runtimes create container workloads within virtual machines.
When a running host privileged containers using them,
they wont provide support to requested the privileged opertations.

This commits add the new options to define two runtimes:

Trusted runtime : Used when a privileged container is requested.
Default runtime : for non-privileged workloads.

A container that belongs to a privileged pod will inherent this property
an will be created with the trusted runtime.

- Add options to define trusted runtime
- Add logic to decide if a sanbox is trusted
- Export annotation containers below to a trusted sandbox

## Note 
This feature is implemented solution to address gaps described in #574. For long term solution will be defined as part of CRI API for `secure containers `

Fixes: #574 

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/hack/test-utils.sh
+++ b/hack/test-utils.sh
@@ -18,6 +18,13 @@ source $(dirname "${BASH_SOURCE[0]}")/utils.sh
 
 # RESTART_WAIT_PERIOD is the period to wait before restarting containerd.
 RESTART_WAIT_PERIOD=${RESTART_WAIT_PERIOD:-10}
+CONTAINERD_CONFIG="--log-level=debug "
+
+# Use a configuration file for containerd.
+CONTAINERD_CONFIG_FILE=${CONTAINERD_CONFIG_FILE:-""}
+if [ -f "${CONTAINERD_CONFIG_FILE}" ]; then
+	CONTAINERD_CONFIG+="--config $CONTAINERD_CONFIG_FILE"
+fi
 
 CONTAINERD_SOCK=/run/containerd/containerd.sock
 
@@ -32,7 +39,7 @@ test_setup() {
     exit 1
   fi
   sudo pkill -x containerd
-  keepalive "sudo ${ROOT}/_output/containerd --log-level=debug" \
+  keepalive "sudo ${ROOT}/_output/containerd ${CONTAINERD_CONFIG}" \
     ${RESTART_WAIT_PERIOD} &> ${report_dir}/containerd.log &
   containerd_pid=$!
   # Wait for containerd to be running by using the containerd client ctr to check the version
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -18,20 +18,24 @@ package config
 
 import "github.com/containerd/containerd"
 
+// Runtime  struct to contain the type(ID), engine, and root variables for a default and a privileged runtime
+type Runtime struct {
+	//Type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
+	Type string `toml:"runtime_type" json:"runtimeType,omitempty"`
+	// Engine is the name of the runtime engine used by containerd.
+	Engine string `toml:"runtime_engine" json:"runtimeEngine,omitempty"`
+	// Root is the directory used by containerd for runtime state.
+	Root string `toml:"runtime_root" json:"runtimeRoot,omitempty"`
+}
+
 // ContainerdConfig contains toml config related to containerd
 type ContainerdConfig struct {
 	// Snapshotter is the snapshotter used by containerd.
 	Snapshotter string `toml:"snapshotter" json:"snapshotter,omitempty"`
-	// Runtime is the runtime to use in containerd. We may support
-	// other runtimes in the future.
-	Runtime string `toml:"runtime" json:"runtime,omitempty"`
-	// RuntimeEngine is the name of the runtime engine used by containerd.
-	// Containerd default should be "runc"
-	// We may support other runtime engines in the future.
-	RuntimeEngine string `toml:"runtime_engine" json:"runtimeEngine,omitempty"`
-	// RuntimeRoot is the directory used by containerd for runtime state.
-	// Containerd default should be "/run/containerd/runc"
-	RuntimeRoot string `toml:"runtime_root" json:"runtimeRoot,omitempty"`
+	// DefaultRuntime is the runtime to use in containerd.
+	DefaultRuntime Runtime `toml:"default_runtime" json:"defaultRuntime,omitempty"`
+	// PrivilegedRuntime is a non-secure runtime used only to run trusted workloads on it
+	PrivilegedRuntime Runtime `toml:"privileged_runtime" json:"privilegedRuntime,omitempty"`
 }
 
 // CniConfig contains toml config related to cni
@@ -101,10 +105,17 @@ func DefaultConfig() PluginConfig {
 			NetworkPluginConfDir: "/etc/cni/net.d",
 		},
 		ContainerdConfig: ContainerdConfig{
-			Snapshotter:   containerd.DefaultSnapshotter,
-			Runtime:       "io.containerd.runtime.v1.linux",
-			RuntimeEngine: "",
-			RuntimeRoot:   "",
+			Snapshotter: containerd.DefaultSnapshotter,
+			DefaultRuntime: Runtime{
+				Type:   "io.containerd.runtime.v1.linux",
+				Engine: "",
+				Root:   "",
+			},
+			PrivilegedRuntime: Runtime{
+				Type:   "io.containerd.runtime.v1.linux",
+				Engine: "",
+				Root:   "",
+			},
 		},
 		StreamServerAddress: "",
 		StreamServerPort:    "10010",
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -87,6 +87,9 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	}
 	sandboxPid := s.Pid()
 
+	trusted := sandbox.Config.Annotations[annotations.PrivilegedSandbox] == "true"
+	containerRuntime := c.getRuntime(trusted)
+
 	// Generate unique id and name for the container and reserve the name.
 	// Reserve the container name to avoid concurrent `CreateContainer` request creating
 	// the same container.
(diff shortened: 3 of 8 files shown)


jessfraz:
oh yay! I am excited to see this and pop out my existing thing with this :) thanks so much

Random-Liu:
Thanks! @jcvenegas I'll get this in before the coming release.

pkg/server/sandbox_run.go
```diff
@@ -150,6 +176,15 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
...
 	}
 
+	trusted := c.privilegedSandbox(r)
+	containerRuntime := c.getRuntime(trusted)
+
+	if sandbox.Config.Annotations == nil {
+		sandbox.Config.Annotations = make(map[string]string)
+	}
+
+	sandbox.Config.Annotations[annotations.PrivilegedSandbox] = fmt.Sprintf("%v", trusted)
```
mikebrow: let's remove the above alloc and move this assignment to the bottom of generateSandboxContainerSpec() along with the other g.AddAnnotation() calls..

hack/test-utils.sh
```diff
@@ -18,6 +18,13 @@ source $(dirname "${BASH_SOURCE[0]}")/utils.sh
 
 # RESTART_WAIT_PERIOD is the period to wait before restarting containerd.
 RESTART_WAIT_PERIOD=${RESTART_WAIT_PERIOD:-10}
+CONTAINERD_CONFIG=""
+
+# Use a configuration file for containerd.
+CONTAINERD_CONFIG_FILE=${CONTAINERD_CONFIG_FILE:-none}
```
Random-Liu: Let's use `""` as default. Although unlikely, but is possible that there is a `none` file.

And then you can check `-z` or `-f`.

hack/test-utils.sh
```diff
@@ -18,6 +18,13 @@ source $(dirname "${BASH_SOURCE[0]}")/utils.sh
 
 # RESTART_WAIT_PERIOD is the period to wait before restarting containerd.
 RESTART_WAIT_PERIOD=${RESTART_WAIT_PERIOD:-10}
+CONTAINERD_CONFIG=""
```
Random-Liu: nit: Move `--log-level=debug` here
and append in the underlying if clause.

pkg/config/config.go
```diff
@@ -32,6 +32,24 @@ type ContainerdConfig struct {
 	// RuntimeRoot is the directory used by containerd for runtime state.
 	// Containerd default should be "/run/containerd/runc"
 	RuntimeRoot string `toml:"runtime_root" json:"runtimeRoot,omitempty"`
+
+	// Trusted runtime options.
```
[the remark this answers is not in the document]
Random-Liu: @mikebrow Good point.

This also confused to me...
Does `TrustedRuntime` mean the runtime is secure so we trust the runtime (katacontainer), or the runtime is insecure so we only run trusted workloads on it (runc)? Based on your code and comment, it seems to be the latter... But I got it wrong until I read the following code and comment.

And I also agree adding a runtime struct. :)

pkg/server/sandbox_run.go
```diff
@@ -150,6 +176,15 @@ func (c *criContainerdService) RunPodSandbox(ctx context.Context, r *runtime.Run
 		sandbox.IP = ip
 	}
 
+	trusted := c.privilegedSandbox(r)
+	containerRuntime := c.getRuntime(trusted)
```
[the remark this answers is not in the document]
Random-Liu: Agree. Although the terminologies are in a mess now, but we do agree on that:
* runc is a `runtime`.
* containerd, cri-o is a `container runtime`.

pkg/server/sandbox_run.go
```diff
@@ -48,6 +48,32 @@ func init() {
 		"github.com/containerd/cri-containerd/pkg/store/sandbox", "Metadata")
 }
 
+// privilegedSandbox returns true if the sandbox configuration
+// requires additional host privileges for the sandbox.
+func (c *criContainerdService) privilegedSandbox(req *runtime.RunPodSandboxRequest) bool {
```
Random-Liu: We need a simple unit test for this to make sure the behavior is correct, and no future change will break this.
And I don't think this is necessarily need to be a member function of `criContainerdService`, it is not used. :)

pkg/server/service.go
```diff
@@ -253,3 +253,29 @@ func (c *criContainerdService) Close() error {
...
+type runtimeInfo struct {
+	name   string
+	engine string
+	root   string
+}
+
+// getRuntime returns the runtime configuration
+// If the container is privileged, it will return
+// the privileged runtime else not.
+func (c *criContainerdService) getRuntime(privileged bool) (runtime runtimeInfo) {
```
Random-Liu: nit: I prefer moving this function to `helper.go` :)

pkg/server/service.go
```diff
@@ -253,3 +253,29 @@ func (c *criContainerdService) Close() error {
 func imageFSPath(rootDir, snapshotter string) string {
 	return filepath.Join(rootDir, fmt.Sprintf("%s.%s", plugin.SnapshotPlugin, snapshotter))
 }
+
+// runtimeInfo is the information about a runtime for container creation
+type runtimeInfo struct {
+	name   string
```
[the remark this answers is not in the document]
Random-Liu: nit: Move this struct to config.go

pkg/server/sandbox_run.go
Random-Liu: I don't think we should change the configuration we get from CRI. We want it to reflect what we get from kubelet.

Just treat the sandbox as trusted when `Annotations` is nil.

mikebrow: I think I remember hearing them say they are using this annotation.. but you make a good point we won't be using the privileged runtime unless it's privileged so if they are called it's unnecessary to have the extra flag. Maybe at one point they were being called for both privileged and unprivileged cases and needed the flag to propagate?

Random-Liu:
@jcvenegas We plan to cut a release mid of next week. Can you address the comments before that? Thanks!

jcvenegas:
@mikebrow @mikebrow  I will work on this today.

pkg/config/config.go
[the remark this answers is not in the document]
jcvenegas: @Random-Liu yes, it is the second case, wil ladd the struct and a  comment to make it clear.

pkg/server/sandbox_run.go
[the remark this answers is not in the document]
jcvenegas: I see `runtime` is used to refer to "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2" what should be the new name when it is imported?

pkg/server/container_create.go
```diff
@@ -88,6 +88,9 @@ func (c *criContainerdService) CreateContainer(ctx context.Context, r *runtime.C
 	}
 	sandboxPid := s.Pid()
 
+	trusted := sandbox.Config.Annotations[annotations.PrivilegedSandbox] == "true"
```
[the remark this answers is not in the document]
jcvenegas: PrivilegedSandbox should be exported in annotations pkg to be used from here.

pkg/server/service.go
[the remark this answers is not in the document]
jcvenegas: Done

pkg/server/sandbox_run.go
jcvenegas: Done :)

Random-Liu:
The test failure is a flaky test. I've disabled it.

pkg/server/helpers.go
```diff
@@ -407,3 +409,18 @@ func getPodCNILabels(id string, config *runtime.PodSandboxConfig) map[string]str
 		"IgnoreUnknown":              "1",
 	}
 }
+
+// getRuntime returns the runtime configuration
+// If the container is privileged, it will return
```
Random-Liu: If the workload is trusted, it will return the privileged runtime else not.

I'll take care of this.

pkg/server/helpers_test.go
```diff
@@ -19,10 +19,10 @@ package server
 import (
 	"testing"
 
+	criconfig "github.com/containerd/cri/pkg/config"
```
Random-Liu: Moving cri-containerd dependencies to its own group. I'll do it.

pkg/server/helpers_test.go
```diff
@@ -142,3 +142,64 @@ func TestBuildLabels(t *testing.T) {
 	assert.Empty(t, configLabels[containerKindLabel], "should not add new labels into original label")
 	assert.Equal(t, "b", configLabels["a"], "change in new labels should not affect original label")
 }
+
+func Test_criService_getRuntime(t *testing.T) {
```
Random-Liu: Follow naming convention. `TestGetRuntime` or `TestCRIServiceGetRuntime`.

pkg/server/sandbox_run.go
```diff
@@ -130,6 +156,15 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
...
 	}
 
+	privileged := privilegedSandbox(r)
+	containerRuntime := c.getRuntime(privileged)
+
+	if sandbox.Config.Annotations == nil {
+		sandbox.Config.Annotations = make(map[string]string)
+	}
+
+	sandbox.Config.Annotations[annotations.PrivilegedSandbox] = fmt.Sprintf("%v", privileged)
```
Random-Liu: We should not update CRI config. This will be checkpointed and returned in `PodSandboxStatus`.

Random-Liu:
Let's merge this and iterate fast.

I'll take care of my own comments. :)

jessfraz:
woohoo! when will the release be :)

Random-Liu:
@jessfraz Today. :)

PR #663, opened 2018-03-14
Moving to use go-cni library from containerd

abhi:
This PR aims to use the cni library form containerd. The library avoids the usage of nsenter by making use of the result provided by the cni plugins.


abhi:
I realised that I may need to make changes to the fake cni plugin . Will update it tomorrow. Keeping this open now for any early comments.

pkg/server/status.go
```diff
@@ -40,10 +41,18 @@ func (c *criContainerdService) Status(ctx context.Context, r *runtime.StatusRequ
 		Type:   runtime.NetworkReady,
 		Status: true,
 	}
-	if err := c.netPlugin.Status(); err != nil {
-		networkCondition.Status = false
-		networkCondition.Reason = networkNotReadyReason
-		networkCondition.Message = fmt.Sprintf("Network plugin returns error: %v", err)
+	if c.netPlugin == nil {
```
Random-Liu: Why `netPlugin` can be nil here?

There is a race condition of `netPlugin` here? I prefer a `Status` or `Sync` call from go-cni instead of always `New`.

pkg/server/service.go
```diff
@@ -129,11 +134,13 @@ func NewCRIContainerdService(config criconfig.Config, client *containerd.Client)
 	c.imageFSPath = imageFSPath(config.ContainerdRootDir, config.ContainerdConfig.Snapshotter)
 	logrus.Infof("Get image filesystem path %q", c.imageFSPath)
 
-	c.netPlugin, err = ocicni.InitCNI(config.NetworkPluginConfDir, config.NetworkPluginBinDir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize cni plugin: %v", err)
+	c.netPlugin, err = cni.New(cni.WithMinNetworkCount(2),
```
Random-Liu: Can we make `2` a const, and explain why it is 2?

And are we sure that for different CNI plugins, only 2 are needed?

pkg/server/sandbox_run.go
```diff
@@ -498,14 +487,36 @@ func (c *criContainerdService) unmountSandboxFiles(rootDir string, config *runti
 	return nil
 }
 
+// setupPod setups up the network for a pod
+func (c *criContainerdService) setupPod(id string, path string, config *runtime.PodSandboxConfig) (string, error) {
+	labels := map[string]string{
+		"K8S_POD_NAMESPACE":          config.GetMetadata().GetNamespace(),
+		"K8S_POD_NAME":               config.GetMetadata().GetName(),
+		"K8S_POD_INFRA_CONTAINER_ID": id,
+		"IgnoreUnknown":              "1",
```
Random-Liu: For my education, what does `IgnoreUnknown` mean?

pkg/server/sandbox_run.go
```diff
@@ -498,14 +487,36 @@ func (c *criContainerdService) unmountSandboxFiles(rootDir string, config *runti
 	return nil
 }
 
+// setupPod setups up the network for a pod
+func (c *criContainerdService) setupPod(id string, path string, config *runtime.PodSandboxConfig) (string, error) {
+	labels := map[string]string{
```
Random-Liu: Let's make these string constants.
And let's make the label generation a helper function. The logic is duplicated at least twice.

pkg/server/sandbox_run.go
```diff
@@ -498,14 +487,36 @@ func (c *criContainerdService) unmountSandboxFiles(rootDir string, config *runti
...
+		cni.WithLabels(labels),
+		cni.WithCapabilityPortMap(toCNIPortMappings(config.GetPortMappings())))
+	if err != nil {
+		return "", err
+	}
+	// Check if the default interface has IP config
+	if configs, ok := result.Interfaces[defaultIfName]; ok && len(configs.IPConfigs) > 0 {
+		return configs.IPConfigs[0].IP.String(), nil
+	}
+	return "", fmt.Errorf("failed to find network info for sandbox %q", id)
```
Random-Liu: Do we need to remove on error?

pkg/server/service.go
```diff
@@ -45,6 +45,11 @@ import (
 	snapshotstore "github.com/containerd/cri/pkg/store/snapshot"
 )
 
+const (
+	// defaultIfName is the default network interface for the pods
+	defaultIfName = "eth0"
```
Random-Liu: Put this into helpers.go? Now all constants like this are in helpers.

pkg/server/service.go
```diff
@@ -129,11 +134,13 @@ func NewCRIContainerdService(config criconfig.Config, client *containerd.Client)
...
 	logrus.Infof("Get image filesystem path %q", c.imageFSPath)
 
-	c.netPlugin, err = ocicni.InitCNI(config.NetworkPluginConfDir, config.NetworkPluginBinDir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize cni plugin: %v", err)
+	c.netPlugin, err = cni.New(cni.WithMinNetworkCount(2),
+		cni.WithPluginConfDir(config.NetworkPluginConfDir),
+		cni.WithPluginDir([]string{config.NetworkPluginBinDir}),
+		cni.WithLoNetwork())
+	if err != nil && !cni.IsCNINotInitialized(err) {
```
Random-Liu: I'm not very comfortable with leaving `netPlugin` as `nil` if CNI is not initialized.
1) There will be race condition as I mentioned in https://github.com/containerd/cri/pull/663/files#r174359472.
2) If functions other than `Status` is called, this may cause a panic. Especially kubelet assume that host network pod could run even when network is not ready, and this assume that `netPlugin` will not be accessed at all if pod is using host network.
3) Even if someone try to run normal pod when network plugin is not ready, containerd should not panic.

Anyway, I feel like it is better to have netPlugin here, but add another function to return whether the netPlugin is ready or not.

pkg/server/service.go
abhi: 2 is - 1 for loop back and 1 for default network . You ar right about const. Willl do

pkg/server/sandbox_run.go
abhi: not needed. will remove

pkg/server/sandbox_run.go
abhi: The failure here would mean the cni plugin is not allocating the IP and hence missing in the result.

pkg/server/service.go
abhi: will do

pkg/server/service.go
abhi: In what scenario does the kublet would use netplugin for host network ? We can definitely add a check if thats the case. 
The reason we stuck to New is for better error handling instead of ignoring the errors. I am not sure about the race. We can discuss about this.

pkg/server/sandbox_run.go
Random-Liu: I mean, do we need a `defer teardown` on error?

pkg/server/service.go
Random-Liu: CRI is a grpc service, all functions could be called at anytime as long as client sends a request.
There are several functions using `netPlugin`. We need a lock to protect it if we want to use it in one function, and set it in another function.

We don't necessarily need to ignore the error. We just need to define error better. .To me, missing CNI plugins/config is not an error, it is a state of cni library. We can just return the `netPlugin` in `UnInitialized` state. As for other errors, of course, we can return error.

abhi: I think the issue with the way it is. We return the cni struct and error. essentially
```
cni:=CNI{......}
err:= filnetworkConfig(....)
if err != nil{
    return cni,err
}
return cni,nil

Comment (abhi) [reply pkg/server/service.go]: This is what it will end up looking. If we need to lock it , we can lock it. I was of the opinion same as yours. But after some discussion with @crosbymichael I feel that in order to make it work in cri , we end up return an uninitialized state as above ^.

Comment (crosbymichael) [reply pkg/server/service.go]: My argument is that CRI should not impose bad API design for libs that it consumes.  A lib's `New` should either return a fully initialized struct or error.  Not something inbetween that its incapable of doing work.  It just adds complexity and bugs.  Before, if there are invalid configs,etc. all the load errors are discarded.  How is this ever good in any universe?

File: pkg/server/sandbox_run.go (hunk 5 above)
Comment (abhi) [reply pkg/server/sandbox_run.go]: it is takes care of by the called RunPodSandbox

File: pkg/server/sandbox_run.go (hunk 3 above)
Comment (abhi) [reply pkg/server/sandbox_run.go]: It is needed to ignore the arguments that are not known by the plugin. More details : https://github.com/containernetworking/cni/pull/158 and https://github.com/kubernetes/kubernetes/pull/24983

File: pkg/server/sandbox_run.go (hunk 4 above)
Comment (abhi) [reply pkg/server/sandbox_run.go]: done

File: pkg/server/sandbox_run.go (hunk 5 above)
Comment (Random-Liu) [reply pkg/server/sandbox_run.go]: No, that defer is after this error.

Comment (abhi) [reply pkg/server/sandbox_run.go]: ok i se ewhat u mean

File: integration/test_utils.go [hunk 8]
```diff
@@ -91,6 +91,10 @@ func ConnectDaemons() error {
 	if err != nil {
 		return fmt.Errorf("failed to connect cri plugin: %v", err)
 	}
+	_, err = runtimeService.Status()
```
Random-Liu: I think we don't need this after we added `Load` during bootstrap. Or we should check readiness here.

Just check `Status` error doesn't make too much sense to me.

pkg/server/service.go
```diff
@@ -129,11 +129,22 @@ func NewCRIContainerdService(config criconfig.Config, client *containerd.Client)
...
+		cni.WithPluginDir([]string{config.NetworkPluginBinDir}))
 	if err != nil {
-		return nil, fmt.Errorf("failed to initialize cni plugin: %v", err)
+		return nil, fmt.Errorf("failed to initialize cni: %v", err)
 	}
 
+	// Try to load the config if it exists. Just log the error if load fails
+	// This is not disruptive for containerd to panic
+	if err := c.netPlugin.Load(cni.WithLoNetwork(), cni.WithDefaultConf()); err != nil {
+		logrus.WithError(err).Error("Failed to load cni during init")
```
Random-Liu: Add something like:
```
Failed to load cni during init, please check CRI plugin status before using pod network.
```

pkg/server/status.go
```diff
@@ -40,10 +41,14 @@ func (c *criContainerdService) Status(ctx context.Context, r *runtime.StatusRequ
 		Type:   runtime.NetworkReady,
 		Status: true,
 	}
+	// Check the status of the cni initialization
 	if err := c.netPlugin.Status(); err != nil {
```
Random-Liu: Check `ErrCNINotInitialized`? If it is not the error, we should just return the error in message.

vendor.conf
```diff
@@ -8,11 +8,11 @@ github.com/containerd/containerd 3013762fc58941e33ba70e8f8d9256911f134124
 github.com/containerd/continuity d8fb8589b0e8e85b8c8bbaa8840226d0dfeb7371
 github.com/containerd/fifo fbfb6a11ec671efbe94ad1c12c2e98773f19e1e6
 github.com/containerd/go-runc 4f6e87ae043f859a38255247b49c9abc262d002f
+github.com/containerd/go-cni 575b6a4762a00f6c10d71eb300f07a9d4b5ac695 https://github.com/abhi/go-cni
 github.com/containerd/typeurl f6943554a7e7e88b3c14aad190bf05932da84788
 github.com/containernetworking/cni v0.6.0
-github.com/containernetworking/plugins v0.6.0
+github.com/containernetworking/plugins v0.7.0
```
Random-Liu: If DAD disable doesn't work, I prefer updating cni plugins later, because the benefit is not clear, and it may introduce new bugs. Kubernetes hasn't validated 0.7.0 plugins yet.

If DAD disable does work, we can try update plugin version to get rid of the mess on our side.

Random-Liu:
LGTM overall. The main question is about cni verison.

pkg/server/status.go
abhi: status only returns the not initialized error

integration/test_utils.go
abhi: yes will add that check

vendor.conf
abhi: sure. did you try with 0.6.0 ? or earlier ?

pkg/server/status.go
Random-Liu: OK. Then let's just keep your current code. :)

vendor.conf
Random-Liu: If you try with 0.6.0 without DAD disabled, pod creation will be super slow.

abhi: Cool ! I am going to make a separate PR for DAD instead of dirtying this

abhi:
@Random-Liu thanks for reviewing! All comments addressed. I will send follow up PRs on DAD and netns for scout next.

integration/test_utils.go
Random-Liu: It seems that you removed this. I'm also fine with that. :)

Random-Liu:
/lgtm

One less `nsenter`!

PR #670, opened 2018-03-15
Enable promisc mode because Kubernetes network model needs it.

Random-Liu:
We need promisc mode to handle the case that one pod talks to its own service.


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/670#pullrequestreview-104412783

PR #647, opened 2018-03-06
makes copyright year optional

mikebrow:
Addressing #646

With this change the YEAR is optional for copyright lines if YEAR is not specified in the boiler text.  

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

PR #671, opened 2018-03-15
Ctrcri to ctr cri

Random-Liu:
Fixes https://github.com/containerd/cri/issues/590

This PR:
1) Upgrade containerd to https://github.com/containerd/containerd/commit/3c1ef1a714cf5b0104f340f76d539802fc24c75f.
2) Use `ctr cri load` instead.


Random-Liu:
@mikebrow Can you take a look at this one?

cmd/ctr/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 The Kubernetes Authors.
+Copyright 2018 The containerd Authors.
```
mikebrow: note: once #647 is pushed the year is no longer necessary. The containerd guys don't want the year.

pkg/version/version.go
```diff
@@ -1,43 +0,0 @@
...
-import (
-	"fmt"
-
-	"github.com/blang/semver"
-)
-
-// CRIContainerdVersion is the version of the cri plugin.
-var CRIContainerdVersion = "UNKNOWN"
-
-func validateSemver(sv string) error {
```
mikebrow: does containerd already have a validateSemver unit test?

Random-Liu: No, they don't. I feel like this can be a CRI validation test instead. Unless kubelet stops relying on it.

Random-Liu: Filed an issue for this.
https://github.com/containerd/cri/issues/672

cmd/ctr/main.go
Random-Liu: Will review that PR later. :P Need to finish the integration first. :)

PR #674, opened 2018-03-16
Removing DAD config and updating plugins to v0.7.0

abhi:
After testing out the v0.7.0 bridge plugin, it seems like the latency issue of pod bring up is fixed. We are still looking at the exact commit which is helping fix the latency.

Fixes https://github.com/containerd/cri/issues/566


mikebrow:
latency issue is back.. @MikeZappa87

Issue #665, opened 2018-03-14
Streaming server should set TLS config


Random-Liu:
To avoid regression, streaming server should generate a random public/private key and set TLS config to make the connection encrypted.

Random-Liu:
Close this one. Since it is done in #681.

PR #681, opened 2018-03-19
adds tls certificate to tls config

mikebrow:
Addresses issue #665

Adds a tls config comprising a newly generated x509 certificate from a newly generated public/private key pair.

### DIFF
--- a/pkg/server/streaming.go
+++ b/pkg/server/streaming.go
@@ -17,18 +17,36 @@ limitations under the License.
 package server
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
 	"io"
 	"math"
+	"math/big"
 	"net"
+	"os"
+	"time"
 
 	"github.com/pkg/errors"
 	k8snet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/remotecommand"
+	k8scert "k8s.io/client-go/util/cert"
 	"k8s.io/kubernetes/pkg/kubelet/server/streaming"
 	"k8s.io/utils/exec"
 
 	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
+)
+
+const (
+	// OrganizationName is is the name of this organization, used for certificates etc.
+	OrganizationName = "containerd"
+	// CRIName is the common name of the CRI plugin
+	CRIName = "cri"
 )
 
 func newStreamServer(c *criService, addr, port string) (streaming.Server, error) {
@@ -42,6 +60,14 @@ func newStreamServer(c *criService, addr, port string) (streaming.Server, error)
 	config := streaming.DefaultConfig
 	config.Addr = net.JoinHostPort(addr, port)
 	runtime := newStreamRuntime(c)
+	tlsCert, err := newTLSCert()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate tls certificate for stream server")
+	}
+	config.TLSConfig = &tls.Config{
+		Certificates:       []tls.Certificate{tlsCert},
+		InsecureSkipVerify: true,
+	}
 	return streaming.NewServer(config, runtime)
 }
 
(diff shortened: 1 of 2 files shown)


Random-Liu:
Can we use https://github.com/kubernetes/client-go/blob/master/util/cert/cert.go#L63?

mikebrow:
> Can we use https://github.com/kubernetes/client-go/blob/master/util/cert/cert.go#L63?

> It seems that we only need cert.NewPrivateKey + cert.NewSelfSignedCACert + tls.X509KeyPair.
I searched pem.EncodeToMemory in Kubernetes code and found this library.

Yes, if/when we want to generate a "self signed" Certificate Authority (CA) certificate to use when creating our self signed user certificate, thus creating the CA chain I mentioned over chat, we could use this code, or something close to it, to do that. 

As for using these cert. apis "now" they are just thin wrappers with a few predefined constants that we may or may not want. 

If we did use this cert. api we'd also be using cert.NewSignedCert to consume our new CACert when generating the cert that we need for the streaming encryption. But we'd still have almost as much code to add to fill in the config they use and then do the rest of what we need.

Let's leave this question as a potential route to go depending on what auth requirements get set for the CRI API.  

We should open an auth design issue for that here in containerd/cri and another over in the kubernetes CRI api.

pkg/server/streaming.go
```diff
@@ -112,3 +130,80 @@ func handleResizing(resize <-chan remotecommand.TerminalSize, resizeFunc func(si
...
+
+	// Generate new private key
+	privKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return fail(errors.Wrap(err, "private key cannot be created"))
+	}
+
+	// Generate pem block using the private key
+	keyPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
```
Random-Liu: Use a const.

pkg/server/streaming.go
```diff
@@ -112,3 +130,80 @@ func handleResizing(resize <-chan remotecommand.TerminalSize, resizeFunc func(si
...
+	}
+
+	cert, err := x509.CreateCertificate(rand.Reader, &tml, &tml, &privKey.PublicKey, privKey)
+	if err != nil {
+		return fail(errors.Wrap(err, "certificate cannot be created"))
+	}
+
+	// Generate a pem block with the certificate
+	certPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
```
Random-Liu: ditto.

pkg/server/streaming.go
```diff
@@ -112,3 +130,80 @@ func handleResizing(resize <-chan remotecommand.TerminalSize, resizeFunc func(si
...
+		return fail(errors.Wrap(err, "failed to get host IP addresses"))
+	}
+
+	// Configure and create new certificate
+	tml := x509.Certificate{
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(years, 0, 0),
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			CommonName:   fmt.Sprintf("system:node:%s", hostName),
```
Random-Liu: Let's use something else. This is for kubelet I think.

`containerd`?

pkg/server/streaming.go
```diff
@@ -112,3 +130,80 @@ func handleResizing(resize <-chan remotecommand.TerminalSize, resizeFunc func(si
 		}
 	}()
 }
+
+func newTLSCert() (tls.Certificate, error) {
```
Random-Liu: Add comment to explain what we are doing here, and add TODO.

pkg/server/streaming.go
mikebrow: done

pkg/server/streaming.go
mikebrow: done

pkg/server/streaming.go
mikebrow: done

pkg/server/streaming.go
mikebrow: done

mikebrow:
comments addressed @Random-Liu

pkg/server/streaming.go
```diff
@@ -17,20 +17,38 @@ limitations under the License.
...
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/remotecommand"
+	k8scert "k8s.io/client-go/util/cert"
 	"k8s.io/kubernetes/pkg/kubelet/server/streaming"
 	"k8s.io/utils/exec"
 
 	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
 )
 
+const (
```
Random-Liu: Prefer private const. I can take care of that.

PR #685, opened 2018-03-20
Update cri-tools and build critest into release tarball again.

Random-Liu:
Build `critest` into release tarball again, because we've got rid of the dependency on the source code. https://github.com/kubernetes-incubator/cri-tools/pull/261

Now `critest` binary is usable itself.


Random-Liu:
The test failure is caused by https://github.com/kubernetes/kubernetes/pull/61443.

I'll retrigger the test.


--

Random-Liu:
Filed https://github.com/kubernetes/test-infra/pull/7354 to skip summary api test and unblock the presubmit test for now.

PR #688, opened 2018-03-21
Address comments for privileged runtime code.

Random-Liu:
Address some comments from https://github.com/containerd/cri/pull/657.

Changes need input:
1) I changed `privilegedSandbox` to `hostPrivilegedSandbox`. The reason is that `privileged` already has special meaning today, and `trusted` is still a bit confusing to me. I feel that `hostPrivileged` is more accurate, but a bit long.
2) I changed `containerRuntime` and `getRuntime` to `ociRuntime` and `getOCIRuntime` to make it more clear.
3) I changed sandbox to use privileged runtime if `annotations.HostPrivilegedSandbox` is "true". This behavior is not included in the original PR, I'm not sure whether it is a mistake.

@jcvenegas Thanks a lot for contributing this! :)
/cc @mikebrow @jcvenegas 

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -32,6 +32,7 @@ const (
 	// SandboxID is the sandbox ID annotation
 	SandboxID = "io.kubernetes.cri.sandbox-id"
 
-	// PrivilegedSandbox is the privileged annotation
-	PrivilegedSandbox = "io.kubernetes.cri.privileged-sandbox"
+	// UntrustedWorkload is the sandbox annotation for untrusted workload. Untrusted
+	// workload can only run on dedicated runtime for untrusted workload.
+	UntrustedWorkload = "io.kubernetes.cri.untrusted-workload"
 )
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -18,9 +18,10 @@ package config
 
 import "github.com/containerd/containerd"
 
-// Runtime  struct to contain the type(ID), engine, and root variables for a default and a privileged runtime
+// Runtime struct to contain the type(ID), engine, and root variables for a default runtime
+// and a runtime for untrusted worload.
 type Runtime struct {
-	//Type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
+	// Type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
 	Type string `toml:"runtime_type" json:"runtimeType"`
 	// Engine is the name of the runtime engine used by containerd.
 	Engine string `toml:"runtime_engine" json:"runtimeEngine"`
@@ -34,8 +35,8 @@ type ContainerdConfig struct {
 	Snapshotter string `toml:"snapshotter" json:"snapshotter"`
 	// DefaultRuntime is the runtime to use in containerd.
 	DefaultRuntime Runtime `toml:"default_runtime" json:"defaultRuntime"`
-	// PrivilegedRuntime is a non-secure runtime used only to run trusted workloads on it
-	PrivilegedRuntime Runtime `toml:"privileged_runtime" json:"privilegedRuntime"`
+	// UntrustedWorkloadRuntime is a runtime to run untrusted workloads on it.
+	UntrustedWorkloadRuntime Runtime `toml:"untrusted_workload_runtime" json:"untrustedWorkloadRuntime"`
 }
 
 // CniConfig contains toml config related to cni
@@ -107,11 +108,6 @@ func DefaultConfig() PluginConfig {
 		ContainerdConfig: ContainerdConfig{
 			Snapshotter: containerd.DefaultSnapshotter,
 			DefaultRuntime: Runtime{
-				Type:   "io.containerd.runtime.v1.linux",
-				Engine: "",
-				Root:   "",
-			},
-			PrivilegedRuntime: Runtime{
 				Type:   "io.containerd.runtime.v1.linux",
 				Engine: "",
 				Root:   "",
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -87,9 +87,6 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	}
 	sandboxPid := s.Pid()
 
-	trusted := sandbox.Config.Annotations[annotations.PrivilegedSandbox] == "true"
-	containerRuntime := c.getRuntime(trusted)
-
 	// Generate unique id and name for the container and reserve the name.
 	// Reserve the container name to avoid concurrent `CreateContainer` request creating
 	// the same container.
@@ -124,6 +121,17 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	if image == nil {
 		return nil, errors.Errorf("image %q not found", imageRef)
 	}
+
+	// Run container using the same runtime with sandbox.
+	sandboxInfo, err := sandbox.Container.Info(ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get sandbox %q info", sandboxID)
+	}
+	ociRuntime, err := getRuntimeConfigFromContainerInfo(sandboxInfo)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get OCI runtime")
+	}
+	logrus.Debugf("Use OCI %+v for container %q", ociRuntime, id)
 
 	// Create container root directory.
 	containerRootDir := getContainerRootDir(c.config.RootDir, id)
@@ -230,10 +238,10 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	opts = append(opts,
 		containerd.WithSpec(spec, specOpts...),
 		containerd.WithRuntime(
-			containerRuntime.Type,
+			ociRuntime.Type,
 			&runctypes.RuncOptions{
-				Runtime:       containerRuntime.Engine,
-				RuntimeRoot:   containerRuntime.Root,
+				Runtime:       ociRuntime.Engine,
+				RuntimeRoot:   ociRuntime.Root,
 				SystemdCgroup: c.config.SystemdCgroup}), // TODO (mikebrow): add CriuPath when we add support for pause
 		containerd.WithContainerLabels(containerLabels),
 		containerd.WithContainerExtension(containerMetadataExtension, &meta))
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -37,6 +37,7 @@ import (
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 
 	"github.com/containerd/cri/pkg/annotations"
+	criconfig "github.com/containerd/cri/pkg/config"
 	customopts "github.com/containerd/cri/pkg/containerd/opts"
 	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
 	"github.com/containerd/cri/pkg/log"
@@ -47,32 +48,6 @@ import (
 func init() {
 	typeurl.Register(&sandboxstore.Metadata{},
 		"github.com/containerd/cri/pkg/store/sandbox", "Metadata")
-}
-
-// privilegedSandbox returns true if the sandbox configuration
-// requires additional host privileges for the sandbox.
-func privilegedSandbox(req *runtime.RunPodSandboxRequest) bool {
-	securityContext := req.GetConfig().GetLinux().GetSecurityContext()
-	if securityContext == nil {
-		return false
-	}
-
-	if securityContext.Privileged {
-		return true
-	}
-
-	namespaceOptions := securityContext.GetNamespaceOptions()
-	if namespaceOptions == nil {
-		return false
-	}
-
-	if namespaceOptions.Network == runtime.NamespaceMode_NODE ||
-		namespaceOptions.Pid == runtime.NamespaceMode_NODE ||
-		namespaceOptions.Ipc == runtime.NamespaceMode_NODE {
-		return true
-	}
-
-	return false
 }
 
 // RunPodSandbox creates and starts a pod-level sandbox. Runtimes should ensure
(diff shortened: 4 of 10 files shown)


Random-Liu:
@jcvenegas I can't assign you as reviewer, but could you help me take a look at this PR? Thanks!

pkg/server/container_create.go
```diff
@@ -87,8 +87,7 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	}
 	sandboxPid := s.Pid()
 
-	trusted := sandbox.Config.Annotations[annotations.PrivilegedSandbox] == "true"
-	containerRuntime := c.getRuntime(trusted)
+	ociRuntime := c.getOCIRuntime(hostPrivilegedSandbox(sandbox.Config))
```
jcvenegas: Do you expect that all the runtimes registered in the daemon are an OCI runtime?. I ask because as far I understand containerd could have different runtime plugins and some of them may not be an OCI compatible plunging (depending on `ociRuntime.Type` value).  

From my perspective, it make it more clear that the underling runtime is an OCI runtime (not obvious when I see  `io.containerd.runtime.v1.linux`), but not sure is there a plan support any other type of runtime plugin in the future.

Random-Liu: Good question. Let me confirm it.

pkg/annotations/annotations.go
```diff
@@ -32,6 +32,8 @@ const (
 	// SandboxID is the sandbox ID annotation
 	SandboxID = "io.kubernetes.cri.sandbox-id"
 
-	// PrivilegedSandbox is the privileged annotation
-	PrivilegedSandbox = "io.kubernetes.cri.privileged-sandbox"
+	// HostPrivilegedSandbox is the host privileged annotation. Please note
+	// that host privileged != privileged in CRI. It means that the sandbox
+	// requires host privilege, and can't run on a hypervisor based runtime.
```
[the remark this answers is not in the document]
jcvenegas: @mikebrow Yeah this is mainly to know if the sandbox that is created is privileged or not and then make sure we create containers in this pod with the same runtime.  So yes is more a flag more than a annotation, so you suggest to have a flag for this?. And yes this should be a short-term solution, longer term is being discused at k8s level https://docs.google.com/document/d/1QQ5u1RBDLXWvC8K3pscTtTRThsOeBSts_imYEoRyw8A/edit#heading=h.ypyhxoaw8f95

pkg/config/config.go
```diff
@@ -34,7 +34,7 @@ type ContainerdConfig struct {
 	Snapshotter string `toml:"snapshotter" json:"snapshotter,omitempty"`
 	// DefaultRuntime is the runtime to use in containerd.
 	DefaultRuntime Runtime `toml:"default_runtime" json:"defaultRuntime,omitempty"`
-	// PrivilegedRuntime is a non-secure runtime used only to run trusted workloads on it
+	// PrivilegedRuntime is a non-secure runtime used only to run trusted workloads on it.
```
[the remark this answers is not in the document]
jcvenegas: Agree, what about "is a runtime used only to run trusted workloads on it."
then the questions should be:
```
What is a "privilegedworkload?"
Workload with privileged flag or access to host namespaces. 
```
where should be the best place to document this?

pkg/server/sandbox_run.go
```diff
@@ -49,32 +49,6 @@ func init() {
 		"github.com/containerd/cri/pkg/store/sandbox", "Metadata")
 }
 
-// privilegedSandbox returns true if the sandbox configuration
```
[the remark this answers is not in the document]
Random-Liu: I can't help moving helper function to bottom... Sorry... :P

pkg/server/container_create.go
Random-Liu: Hm. It seems that containerd could potentially support non OCI runtimes.

However, I think for cri-containerd, we do only care about OCI runtime for now. :)

pkg/annotations/annotations.go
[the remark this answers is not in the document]
Random-Liu: @jcvenegas Do we expect user specify this annotation in pod spec of Kubernetes api?

If we do, I think we should still keep this annotation. If we don't, then I think it is fine to remove it. :)

I guess we do.

Random-Liu:
@mikebrow @jcvenegas I'm still not very comfortable with my previous PR. I updated this PR.

There are 3 concepts "secure runtime", "untrusted workload" and "host privilege". The relationship between them:
1) **Secure runtime.** A runtime which is secure that can run untrusted workload. Secure runtime doesn't support host privilege.
2) **Untrusted workload.** The workload is untrusted, it should run in a secure runtime.
3) **Host privilege.** Workload run in secure runtime shouldn't request host privilege, because the runtime doesn't support it.

I think this is more clear than before. And the PR is based on the assumption above.

For `cri-o`, the relationship is even more complex and confusing, because they support "trust level" support.
IIUC, if "trust level" is "untrusted", `cri-o` will run containers in secure runtime by default, unless the container is "Host Privileged", or explicitly set trusted.

I don't think we should support that for now, because it is Kubernetes should do that level of control. We should only provide basic mechanism.
For example, Kubernetes user should set pod security policy saying that all pods in namespace XX is untrusted by default, and the admission handler should apply "untrusted" label by default.

Random-Liu:
Note that this PR also added runtime config into `ContainerStatus` and `PodSandboxStatus` verbose return, e.g.:
```golang
# crictl inspect 80
{
  "status": {
    "id": "8075f66dd12f07bc862dc067df072d5392cf0123255c0eb5b928a0e66eb79cc5",
    "metadata": {
      "attempt": 0,
      "name": "busybox"
    },
    "state": "CONTAINER_RUNNING",
    "createdAt": "2018-03-20T00:32:28.304028075Z",
    "startedAt": "2018-03-20T00:32:34.630970729Z",
    "finishedAt": "1970-01-01T00:00:00Z",
    "exitCode": 0,
    "image": {
      "image": "docker.io/library/busybox:latest"
    },
    "imageRef": "docker.io/library/busybox@sha256:2107a35b58593c58ec5f4e8f2c4a70d195321078aebfadfbfb223a2ff4a4ed21",
    "reason": "",
    "message": "",
    "labels": null,
    "annotations": null,
    "mounts": null,
    "logPath": ""
  },
  "info": {
    "sandboxID": "8945fe343ea50bcb4b2d75c6001255a3b095b508b9c7c829c5d7424b3635c27a",
    "pid": 21356,
    "removing": false,
    "snapshotKey": "8075f66dd12f07bc862dc067df072d5392cf0123255c0eb5b928a0e66eb79cc5",
    "snapshotter": "overlayfs",
    "runtime": {
      "runtimeType": "io.containerd.runtime.v1.linux",
      "runtimeEngine": "",
      "runtimeRoot": ""
    },
...
```

pkg/config/config.go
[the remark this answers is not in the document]
mikebrow: once we get comfortable with the naming and comments, and a demonstrable / tested pattern we'll generate a doc about it and link that off the front page readme..

mikebrow:
Still think we're going about this the wrong way.   IMO we are discussing a pattern with n-types of runtime engines and rules for selecting between them when creating pods.  So I recommend an array of types, along with an array of rules. We can start with just two types and one hard coded rule for selecting between them. Then progress out into the various types runtime engines and make the rules configurable.

jcvenegas:
@Random-Liu the PR looks good I will tests it in my k8s clear/kata setup.

Random-Liu:
@mikebrow We'll change the whole thing once Kubernetes secure pod api is finalized.

Let's extend it when we have more requirement, for now we only need a "default runtime" and a "secure runtime", which is also kind of proved by cri-o. :)

mikebrow:
sigh.. how about "default runtime" and "other runtime" or "vm runtime"

I don't want to go to my boss (customers) and have to explain why containers are not secure unless they are hosted in a vm :-)

pkg/server/sandbox_run.go
```diff
@@ -553,3 +525,48 @@ func toCNIPortMappings(criPortMappings []*runtime.PortMapping) []cni.PortMapping
...
+
+	return false
+}
+
+// getSandboxRuntime returns the runtime configuration for sandbox.
+// If the sandbox contains untrusted workload, secure runtime will be returned,
+// or else default runtime will be returned.
+func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig) (criconfig.Runtime, error) {
+	secureRuntimeRequired := false
+	if untrustedWorkload(config) {
```
jcvenegas: untrustedWorkload will check for UntrustedWorkload being set. But because the sandbox is not created, this was never set before (is not set by k8s today). 
Here we need to  check:
```
if is a *not* a privileged sandbox;then
set UntrustedWorkload
set secureRuntimeRequired
fi
```

```diff
+func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig, sandbox *sandboxstore.Sandbox) (criconfig.Runtime, error) {
        secureRuntimeRequired := false
-       if untrustedWorkload(config) {
-               // TODO(random-liu): Figure out we should return error or not.
-               if hostPrivilegedSandbox(config) {
-                       return criconfig.Runtime{}, errors.New("untrusted workload with host privilege is not allowed")
-               }
+       if !hostPrivilegedSandbox(config) {
+               sandbox.Config.Annotations[annotations.UntrustedWorkload] = "true"
                secureRuntimeRequired = true
        }
 ```

File: pkg/server/sandbox_run.go [hunk 6]
```diff
@@ -553,3 +525,48 @@ func toCNIPortMappings(criPortMappings []*runtime.PortMapping) []cni.PortMapping
...
+		// TODO(random-liu): Figure out we should return error or not.
+		if hostPrivilegedSandbox(config) {
+			return criconfig.Runtime{}, errors.New("untrusted workload with host privilege is not allowed")
+		}
+		secureRuntimeRequired = true
+	}
+
+	if secureRuntimeRequired {
+		if c.config.ContainerdConfig.SecureRuntime.Type == "" {
+			return criconfig.Runtime{}, errors.New("no secure runtime is configured")
```
jcvenegas: Intead of fail because secure runtime was not set. Lets use the default runtime.
```diff
        if secureRuntimeRequired {
                if c.config.ContainerdConfig.SecureRuntime.Type == "" {
-                       return criconfig.Runtime{}, errors.New("no secure runtime is configured")
+                       logrus.Debugf("secure runtime not defined, using default")
+                       return c.config.ContainerdConfig.DefaultRuntime, nil
                }
+               logrus.Debugf("DEBUG: using secure runtime to create Sandbox")
                return c.config.ContainerdConfig.SecureRuntime, nil
        }
        return c.config.ContainerdConfig.DefaultRuntime, nil
 }

```

pkg/server/container_create.go
```diff
@@ -125,6 +122,17 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 		return nil, errors.Errorf("image %q not found", imageRef)
 	}
 
+	// Run container using the same runtime with sandbox.
+	sandboxInfo, err := sandbox.Container.Info(ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get sandbox %q info", sandboxID)
+	}
+	ociRuntime, err := getRuntimeConfigFromContainerInfo(sandboxInfo)
```
jcvenegas: sweet :)

pkg/config/config.go
```diff
@@ -34,8 +34,8 @@ type ContainerdConfig struct {
 	Snapshotter string `toml:"snapshotter" json:"snapshotter"`
 	// DefaultRuntime is the runtime to use in containerd.
 	DefaultRuntime Runtime `toml:"default_runtime" json:"defaultRuntime"`
-	// PrivilegedRuntime is a non-secure runtime used only to run trusted workloads on it
-	PrivilegedRuntime Runtime `toml:"privileged_runtime" json:"privilegedRuntime"`
+	// SecureRuntime is a secure runtime used to run untrusted workloads on it.
+	SecureRuntime Runtime `toml:"secure_runtime" json:"secureRuntime"`
```
jcvenegas: As @mikebrow points, to avoid confusion about what is secure or not , lets rename it to something like `runtime_untrusted_workload` (name used by cri-o). IMO SecureRuntime is a an easy name but lets keep it generic to reduce any potential confusion.

Random-Liu:
> sigh.. how about "default runtime" and "other runtime" or "vm runtime"

> I don't want to go to my boss (customers) and have to explain why containers are not secure unless they are hosted in a vm :-) For one, I'm not convinced runc containers are always insecure and two i'm not convinced VMs are always secure. Again to me this is the equivalent of stating one runtime is the "preferred" runtime for containerd/kubernetes. Customers will not understand why they want insecure container runtimes.

I totally understand what you mean. :) How about `runtime_untrusted_workload` suggested by @jcvenegas ?

pkg/server/sandbox_run.go
Random-Liu: @jcvenegas Can we chat? I think I may have misunderstood something here.

Random-Liu: As an example, if user create a pod like this:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  annotations:
    io.kubernetes.cri.untrusted-workload: true
spec:
  containers:
  - name: nginx
    image: nginx
```

We'll run it in kata.

mikebrow:
runtime_untrusted_workload suggestion looks good to me

Random-Liu:
@mikebrow Done.

Random-Liu:
@jcvenegas Are you OK with https://github.com/containerd/cri/pull/688#discussion_r176585175?

jcvenegas:
@Random-Liu I would like to have parity with cri-o about the this logic. I tested this approach and works well. But the final wanted approach is make  `runtime_untrusted_workload`   optional.

When `runtime_untrusted_workload` is configured : 
```go
runtime_untrusted_workload.engine = "kata-runtime"
```
The cri containerd plugin will start to evaluate the sandbox to be created.
If the sandbox is evaluated as untrusted, it  will be created with runtime_untrusted_workload ( If defined, else will use the default runtime to not break previous behavior)
```go
CreateSadnbox(){
if  !Privileged(sandbox) {
     ociRuntime = runtime_untrusted_workload
     // mark as untrusted to create containers from this pod with same runtime.
     sandbox.annotation[untrusted] = true
  }
}
```

Having that logic by default all the containers that are not privileged will be created with the `runtime_untrusted_workload`.

jcvenegas:
@Random-Liu  If the logic I explained does not make sense for cri-containerd, we could merge this PR and open an issue to define if we want to  use by default the runtime_untrusted_workload if is defined (expect when the workload is privileged).

Random-Liu:
@jcvenegas Are you on kubernetes slack channel? I pinged you there. :)

Random-Liu:
> I would like to have parity with cri-o about the this logic. I tested this approach and works well. But the final wanted approach is make runtime_untrusted_workload optional.

@jcvenegas Actually I believe the logic you mention will be done at Kubernetes level. Kubernetes will evaluate user defined security policy and pod configuration to make a decision saying that whether the pod is trusted or not.

At runtime level, we just need to provide the simple logic to support trusted and untrusted workload.

Random-Liu:
Offline discussed with @jcvenegas. Let's go with current PR. :)

More advanced policy should be enforced at Kubernetes level.

PR #691, opened 2018-03-22
Getting rid of nsenter and socat

abhi:
This commit removes the usage of nsenter and uses netns
to perform socat operation.
Fixes https://github.com/containerd/cri/issues/144

### DIFF
--- a/pkg/server/sandbox_portforward.go
+++ b/pkg/server/sandbox_portforward.go
@@ -23,12 +23,12 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 
-	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 )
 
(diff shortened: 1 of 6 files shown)


Random-Liu:
@abhi Can we remove socat as well? I think we just need a goroutine to a tcp connection to the port and forward traffics.

abhi:
@Random-Liu yes. Just raising the PR for e2e.

Random-Liu:
@abhi I see. Will review it when it is ready. :)

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,47 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
-	// Check following links for meaning of the options:
-	// * socat: https://linux.die.net/man/1/socat
-	// * nsenter: http://man7.org/linux/man-pages/man1/nsenter.1.html
-	args := []string{"-t", fmt.Sprintf("%d", pid), "-n", socat,
-		"-", fmt.Sprintf("TCP4:localhost:%d", port)}
-
-	nsenter, err := exec.LookPath("nsenter")
-	if err != nil {
-		return errors.Wrap(err, "failed to find nsenter")
+	if s.NetNS == nil {
```
Random-Liu: || s.NetNS.Closed()

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,47 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
+		}()
+		wg.Add(1)
+		go func() {
+			if _, err := io.Copy(stream, client); err != nil {
+				logrus.WithError(err).Errorf("Failed to copy port forward output for %q port %d", id, port)
+			}
+			wg.Done()
+		}()
+		wg.Wait()
+		logrus.Infof("Finish copy port forward input for %q port %d: %v", id, port)
```
Random-Liu: Add a log for both input copy and output copy, and move them into goroutine?

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,47 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
-
-	if err := cmd.Run(); err != nil {
-		return errors.Errorf("nsenter command returns error: %v, stderr: %q", err, stderr.String())
+		defer client.Close()
+
+		wg.Add(1)
+		go func() {
+			if _, err := io.Copy(client, stream); err != nil {
+				logrus.WithError(err).Errorf("Failed to copy port forward input from %q port %d", id, port)
+			}
```
Random-Liu: client.Close()?

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,47 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
+			wg.Done()
+		}()
+		wg.Add(1)
+		go func() {
+			if _, err := io.Copy(stream, client); err != nil {
+				logrus.WithError(err).Errorf("Failed to copy port forward output for %q port %d", id, port)
+			}
+			wg.Done()
+		}()
```
Random-Liu: stream.Close()?

Are these 2 `Close()` are necessary?
If `dst` is closed, `src` is open, but there is no new input, will `io.Copy` stop?
* If it does, we don't need the 2 `Close()`.
* If it doesn't, I think we'll probably need them.

pkg/server/sandbox_portforward.go
```diff
@@ -17,24 +17,22 @@ limitations under the License.
 package server
```
Random-Liu: Update dependencies in `README.md`.

pkg/server/sandbox_portforward.go
abhi: thats already checked by netns.Do

pkg/server/sandbox_portforward.go
abhi: I have added a defer on top

pkg/server/sandbox_portforward.go
abhi: sure

abhi:
@Random-Liu done.

Random-Liu:
@abhi We need to start 2 goroutines to copy stream. However, the 2 goroutines may not be in the same network namespace. We are not sure whether it will work or not. The test result shows that it works, but we need to double check with the golang team.

I don't think we have time to figure that out in the coming `rc.0` release. @abhi Let's only remove `nsenter` and still use `socat` in this PR. We can remove `socat` in `rc.1` release.

abhi:
@Random-Liu yes I am of the same opinion as well. I will remove the socat commits.

Random-Liu:
@abhi also pointed me to another potential issue https://github.com/containernetworking/plugins/issues/98.

Given so, let's move this issue to next release, so that we have more time to investigate this.

It is safe to move this PR to next release, because there is no actual functionality change, we just get rid of 2 binary dependencies `nsenter` and `socat` in this PR.

Random-Liu:
We plan to cut an early `rc.1` release today to carry several fix and improvement. Move this to `rc.2`.

Random-Liu:
@abhi Here is a piece of code:
```golang
package main

import (
	"fmt"
	"io"
	"os"
	"sync"
	"time"
)

func main() {
	f, err := os.OpenFile("io", os.O_RDWR|os.O_CREATE, 0666)
	if err != nil {
		panic(err)
	}
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		io.Copy(os.Stdout, f)
		fmt.Println("file -> stdout done")
		wg.Done()
	}()
	go func() {
		io.Copy(f, os.Stdin)
		fmt.Println("stdin -> file done")
		wg.Done()
	}()
	time.Sleep(5 * time.Second)
	f.Close()
	wg.Wait()
	fmt.Println("all done")
}
```

You will see, after 5 seconds, the problem won't exit even though `f` is closed. Because there is no input from `os.Stdin` to drive the `io.Copy` to write once, and `io.Copy` won't know `f` has been closed. Once you input something into stdin, the problem will exit.

This is why I said, you need to close the other side when one `io.Copy` is done.

pkg/server/sandbox_portforward.go
Random-Liu: Our `Closed` is different https://github.com/containerd/cri/blob/master/pkg/store/sandbox/netns.go#L108.

We have the `restore` support.

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,48 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
-	// * socat: https://linux.die.net/man/1/socat
-	// * nsenter: http://man7.org/linux/man-pages/man1/nsenter.1.html
-	args := []string{"-t", fmt.Sprintf("%d", pid), "-n", socat,
-		"-", fmt.Sprintf("TCP4:localhost:%d", port)}
-
-	nsenter, err := exec.LookPath("nsenter")
-	if err != nil {
-		return errors.Wrap(err, "failed to find nsenter")
+	if s.NetNS == nil {
+		return errors.Errorf("failed to find network namespace fo sandbox %q in store", id)
```
Random-Liu: network namespace for sandbox %q is closed

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,48 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
-		return errors.Wrap(err, "failed to create stdin pipe")
-	}
-	go func() {
-		if _, err := io.Copy(in, stream); err != nil {
-			logrus.WithError(err).Errorf("Failed to copy port forward input for %q port %d", id, port)
+	err = s.NetNS.GetNs().Do(func(_ ns.NetNS) error {
+		var wg sync.WaitGroup
+		client, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
+		if err != nil {
+			return errors.Wrap(err, "failed to dial")
```
Random-Liu: nit: "failed to dial %q", port

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,48 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
-		logrus.Debugf("Finish copy port forward input for %q port %d", id, port)
-	}()
-
-	if err := cmd.Run(); err != nil {
-		return errors.Errorf("nsenter command returns error: %v, stderr: %q", err, stderr.String())
+		wg.Add(1)
+		go func() {
+			defer client.Close()
+			if _, err := io.Copy(client, stream); err != nil {
+				logrus.WithError(err).Errorf("Failed to copy port forward input from %q port %d", id, port)
```
Random-Liu: s/from/for?

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,48 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
+			}
+			logrus.Infof("Finish copy port forward input for %q port %d: %v", id, port)
```
Random-Liu: remove %v

pkg/server/sandbox_portforward.go
```diff
@@ -46,69 +44,48 @@ func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequ
...
+			wg.Done()
+		}()
+		wg.Add(1)
+		go func() {
+			defer stream.Close()
+			if _, err := io.Copy(stream, client); err != nil {
+				logrus.WithError(err).Errorf("Failed to copy port forward output for %q port %d", id, port)
+			}
+			logrus.Infof("Finish copy port forward output for %q port %d: %v", id, port)
```
Random-Liu: ditto.

Issue #686, opened 2018-03-20
Prevent mounting under module directory


dmcgowan:
After running the critest I noticed a mount in the module directory

```
shm on /var/lib/containerd/io.containerd.grpc.v1.cri/sandboxes/aa3f30975daf5225994ef38d618678899ded0e3c1b313206825195d2ecf17cf2/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)
```

Mounts should be done under `/run/containerd` if they are tied to a pod/container or in `/var/lib/containerd/tmpmounts` if they are used directly by the containerd process

Random-Liu:
This requires fs layout change, let's do it before the release.

PR #692, opened 2018-03-22
adds volatile state directory to the fs plan for cntrs/pods/fifo

mikebrow:
Addressing issue: #686

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -38,8 +38,11 @@ The explanation and default value of each configuration item are as follows:
       # runtime_engine is the name of the runtime engine used by containerd.
       runtime_engine = ""
 
-      # runtime_root is the directory used by containerd for runtime state.
+      # runtime_root is the directory used by containerd for persistent runtime state.
       runtime_root = ""
+
+      # runtime_state is the directory used by containerd for volatile runtime state.
+      runtime_state = ""
 
     # "plugins.cri.containerd.untrusted_workload_runtime" is a runtime to run untrusted workloads on it.
     [plugins.cri.containerd.untrusted_workload_runtime]
@@ -49,8 +52,11 @@ The explanation and default value of each configuration item are as follows:
       # runtime_engine is the name of the runtime engine used by containerd.
       runtime_engine = ""
 
-      # runtime_root is the directory used by containerd for runtime state.
+      # runtime_root is the directory used by containerd for persistent runtime state.
       runtime_root = ""
+
+      # runtime_state is the directory used by containerd for volatile runtime state.
+      runtime_state = ""
 
   # "plugins.cri.cni" contains config related to cni
   [plugins.cri.cni]
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -26,6 +26,7 @@ FOCUS=${FOCUS:-""}
 REPORT_DIR=${REPORT_DIR:-"/tmp/test-integration"}
 
 CRI_ROOT="/var/lib/containerd/io.containerd.grpc.v1.cri"
+CRI_STATE="/run/containerd/io.containerd.grpc.v1.cri"
 
 mkdir -p ${REPORT_DIR}
 test_setup ${REPORT_DIR}
@@ -33,7 +34,8 @@ test_setup ${REPORT_DIR}
 # Run integration test.
 sudo ${ROOT}/_output/integration.test --test.run="${FOCUS}" --test.v \
   --cri-endpoint=${CONTAINERD_SOCK} \
-  --cri-root=${CRI_ROOT}
+  --cri-root=${CRI_ROOT} \
+  --cri-state=${CRI_STATE}
 
 test_exit_code=$?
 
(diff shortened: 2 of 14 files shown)


mikebrow:
rebased

docs/config.md
```diff
@@ -38,9 +38,12 @@ The explanation and default value of each configuration item are as follows:
       # runtime_engine is the name of the runtime engine used by containerd.
       runtime_engine = ""
 
-      # runtime_root is the directory used by containerd for runtime state.
+      # runtime_root is the directory used by containerd for persistent runtime state.
       runtime_root = ""
 
+      # runtime_state is the directory used by containerd for volatile runtime state.
+      runtime_state = ""
```
Random-Liu: Why do we need this? We are not using it.

Random-Liu: There is no such thing in runc options, either.
https://github.com/containerd/containerd/blob/856ea51c8b5923b88ae41c9df501909c07776ea7/linux/runctypes/runc.pb.go#L40

hack/test-integration.sh
```diff
@@ -26,14 +26,16 @@ FOCUS=${FOCUS:-""}
...
 
 mkdir -p ${REPORT_DIR}
 test_setup ${REPORT_DIR}
 
 # Run integration test.
 sudo ${ROOT}/_output/integration.test --test.run="${FOCUS}" --test.v \
   --cri-endpoint=${CONTAINERD_SOCK} \
-  --cri-root=${CRI_ROOT}
+  --cri-root=${CRI_ROOT} \
+  --cri-state=${CRI_STATE}
```
Random-Liu: We don't need this for now, volume should be in persistent state.

integration/volume_copy_up_test.go
```diff
@@ -70,7 +70,7 @@ func TestVolumeCopyUp(t *testing.T) {
 	assert.Equal(t, "test_content\n", string(stdout))
 
 	t.Logf("Check host path of the volume")
-	hostCmd := fmt.Sprintf("ls %s/containers/%s/volumes/*/test_file | xargs cat", *criRoot, cn)
+	hostCmd := fmt.Sprintf("ls %s/containers/%s/volumes/*/test_file | xargs cat", *criState, cn)
```
Random-Liu: Volume should be in persistent state.

How about we add "volatileStateRoot" and only change logic for `shm` and `fifos`, and keep other original logic.

pkg/server/container_create.go
```diff
@@ -134,26 +134,35 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
...
+	}
 	defer func() {
 		if retErr != nil {
-			// Cleanup the container root directory.
+			// Cleanup the container root directories.
 			if err = c.os.RemoveAll(containerRootDir); err != nil {
 				logrus.WithError(err).Errorf("Failed to remove container root directory %q",
 					containerRootDir)
 			}
+			if err = c.os.RemoveAll(persistentContainerRootDir); err != nil {
```
Random-Liu: Each defer cleanup should be right after the creation.
e.g. `containerRootDir` cleanup won't run when `c.os.MkdirAll(persistentContainerRootDir, 0755)` returns error, which is not what we want.

pkg/server/sandbox_run.go
```diff
@@ -189,19 +189,28 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
...
 	}
+	persistentSandboxRootDir := getSandboxRootDir(c.config.RootDir, id)
+	if err := c.os.MkdirAll(persistentSandboxRootDir, 0755); err != nil {
+		return nil, errors.Wrapf(err, "failed to create persistent sandbox root directory %q",
+			persistentSandboxRootDir)
+	}
 	defer func() {
 		if retErr != nil {
-			// Cleanup the sandbox root directory.
+			// Cleanup the sandbox root directories.
```
Random-Liu: Same defer issue.

pkg/server/sandbox_run.go
```diff
@@ -189,19 +189,28 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		}
 	}()
 
-	// Create sandbox container root directory.
-	sandboxRootDir := getSandboxRootDir(c.config.RootDir, id)
+	// Create sandbox container root directories.
+	sandboxRootDir := getSandboxRootDir(c.config.StateDir, id)
```
Random-Liu: `hosts` and `resolv.conf` should also be persistent.

Random-Liu:
I think at least till now, we have more things in "persistent" directly. Let's still call it `rootDir`, and add a `volatileStateRootDir`. We only need fifos and `shm` in `volatileStateRootDir`.

Random-Liu:
@mikebrow It seems that you are not online today. We need to get PR merged today. :)

I carried on this PR in https://github.com/containerd/cri/pull/697.

docs/config.md
mikebrow: yeah it's a todo..

PR #701, opened 2018-03-27
Fix event monitor panic.

Random-Liu:
Fix the event monitor panic.

We should only stop ticker when it is not nil.

Since `start` and `stop` can be called at the same time, we need a lock to protect it.

### DIFF
--- a/pkg/server/events.go
+++ b/pkg/server/events.go
@@ -17,6 +17,7 @@ limitations under the License.
 package server
 
 import (
+	"sync"
 	"time"
 
 	eventtypes "github.com/containerd/containerd/api/events"
@@ -56,7 +57,9 @@ type eventMonitor struct {
 }
 
 type backOff struct {
-	queuePool     map[string]*backOffQueue
+	queuePool map[string]*backOffQueue
+	// tickerMu is mutex used to protect the ticker.
+	tickerMu      sync.Mutex
 	ticker        *time.Ticker
 	minDuration   time.Duration
 	maxDuration   time.Duration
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

stevvooe:
LGTM

pkg/server/events.go
```diff
@@ -56,7 +57,9 @@ type eventMonitor struct {
 }
 
 type backOff struct {
-	queuePool     map[string]*backOffQueue
+	queuePool map[string]*backOffQueue
```
yanxuean: go fmt

Random-Liu: That is `gofmt` result.

pkg/server/events.go
```diff
@@ -366,12 +369,18 @@ func (b *backOff) reBackOff(key string, events []interface{}, oldDuration time.D
...
 	return b.ticker.C
 }
 
 func (b *backOff) stop() {
-	b.ticker.Stop()
+	b.tickerMu.Lock()
+	defer b.tickerMu.Unlock()
+	if b.ticker != nil {
+		b.ticker.Stop()
+	}
```
mikebrow: else...logrus.WithError(err).Error("attempt to stop ticker before start has completed allocating the ticker")

note: should be impossible for this to happen now that you have a mutex... if the log is impossible we don't need the nil check..

pkg/server/events.go
```diff
@@ -120,8 +123,8 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 		return nil, errors.New("event channel is nil")
 	}
 	closeCh := make(chan struct{})
+	backOffCheckCh := em.backOff.start()
```
mikebrow: good!

pkg/server/events.go
mikebrow: user interrupts during init of the plugins... interesting thx

PR #702, opened 2018-03-27
Disable TLS config because of unknown hang issue.

Random-Liu:
Disable TLS config because containerd hangs in containerd test because of this.

Please hold this PR until we get newest test result from https://github.com/containerd/containerd/pull/2233. We will get more data to decide whether to do this.


Random-Liu:
Close this one, because the issue in https://github.com/containerd/containerd/pull/2233 has been resolved.

PR #703, opened 2018-03-27
Make release tarballs for multiple architectures

nitkon:
Signed-off-by: Nitesh Konkar niteshkonkar@in.ibm.com


Makefile
```diff
@@ -148,12 +150,22 @@ uninstall:
 	rm -f $(BINDIR)/ctr
 
 $(BUILD_DIR)/$(TARBALL): static-binaries vendor.conf
-	@BUILD_DIR=$(BUILD_DIR) TARBALL=$(TARBALL) VERSION=$(VERSION) ./hack/release.sh
+	@BUILD_DIR=$(BUILD_DIR) TARBALL=$(join $(TARBALL),-$(ARCH).tar.gz) VERSION=$(VERSION) ./hack/release.sh
 
-release: $(BUILD_DIR)/$(TARBALL)
+release: all-release
```
mkumatag: we don't need `all-release` target here, we can use `$(addprefix sub-release-,$(ALL_ARCH))` here.

Makefile
```diff
@@ -148,12 +150,22 @@ uninstall:
...
 
-push: $(BUILD_DIR)/$(TARBALL)
-	@BUILD_DIR=$(BUILD_DIR) TARBALL=$(TARBALL) VERSION=$(VERSION) ./hack/push.sh
+sub-release-%:
+	$(MAKE) --no-print-directory ARCH=$* $(BUILD_DIR)/$(TARBALL)
+
+all-release: $(addprefix sub-release-,$(ALL_ARCH))
+
+push: all-push
```
mkumatag: we don't need all-push target here, we can use $(addprefix sub-release-push-,$(ALL_ARCH)) here.

Makefile
```diff
@@ -128,6 +129,7 @@ clean:
 
 binaries: $(BUILD_DIR)/containerd $(BUILD_DIR)/ctr
 
+static-binaries: GOARCH=$(ARCH)
```
mkumatag: This will not work somebody run `make static-binaries` directly as we are not assigning anything to ARCH by default.

Makefile
```diff
@@ -25,13 +25,14 @@ VERSION := $(shell git describe --tags --dirty --always)
 # strip the first char of the tag if it's a `v`
 VERSION := $(VERSION:v%=%)
 TARBALL_PREFIX := cri-containerd
-TARBALL := $(TARBALL_PREFIX)-$(VERSION).$(GOOS)-$(GOARCH).tar.gz
+TARBALL := $(TARBALL_PREFIX)-$(VERSION).$(GOOS)
```
mkumatag: Just see we can keep this as is and change the tar filename based on the GOARCH we supply runtime.

Makefile
mkumatag: Just check whether we can directly pass GOARCH instead of ARCH while running make targets.

mkumatag: Just check whether we can use GOARCH itself instead of ARCH while calling make targets.

Random-Liu:
I don't think we support so many architectures now.

@containerd/containerd-maintainers What do we support? amd and arm?

mikebrow:
Which architectures: 

I presume we are shooting to match docker ce support of 64bit architectures https://docs.docker.com/install/#docker-ce


x86_64 / amd64, ARM64 / AARCH64, IBM Power (ppc64le), and IBM Z (s390x)

stevvooe:
@mikebrow @nitkon Feel free to add more cross-builds to the travis configuration on containerd. We can't test them, so support will be best effort, but if they run and work okay, I see no reason we can't build them.

nitkon:
@Random-Liu : Do I need to do anything more for this patch to get merged? Any comments? Thanks.

Random-Liu:
@nikhita Thanks for contributing this. I'll try to get it reviewed today.

nitkon:
@Random-Liu : Any inputs/review shall be appreciated. Thanks.

Random-Liu:
Sorry for being late. Marked 1.11, will get this in soon.

Random-Liu:
The `containerd` and `ctr` binary built in this repository is just for testing.

In the release tarball, we have several binaries, containerd, ctr, crictl, cni, runc etc. You need to build multi-arch for all of them.

Given so, I think it is more useful to add multi-arch build in containerd instead.

Random-Liu:
After thinking about this. I think it is more important to add the multi-arch support in containerd.

`cri` is just a plugin, the Makefile here is just for development.

We may want to support multi-arch release tarball, but that need multi-arch support in containerd, cni, runc, crictl. Please feel free to update `hack/install.deps` to support multi-arch.

dims:
@nitkon are you going to pick this up again?

wilmardo:
The containerd arm build is almost a fact!

https://github.com/containerd/containerd/pull/3242

PR #704, opened 2018-03-28
Use systemd service cgroup and oom score adj.

Random-Liu:
Use systemd oom score and service cgroup for GCE, because we do run containerd as a systemd service.


Random-Liu:
The test fails because we need to get the test-infra side change in https://github.com/kubernetes/test-infra/pull/7444.

I'll get that in after this PR is LGTMed.

Random-Liu:
--

PR #711, opened 2018-03-31
Drop libapparmor dependency from build docs

tklauser:
As of opencontainers/runc@db093f6 runc no longer depends on libapparmor
thus libapparmor-dev no longer needs to be installed to build it or
anythind that depends on it (like containerd or cri-containerd). Adjust
the documentation accordingly.

containerd/containerd#2238 did the same for containerd.

Signed-off-by: Tobias Klauser <tklauser@distanz.ch>


mikebrow:
checked dependencies.. we have the correct version of runc and containerd to enable this change :-)

Issue #712, opened 2018-04-02
TLS Error on exec


DefunctExodus:
Good Morning,

I have recently upgraded my cluster to 1.10 and the latest release of this plugin,  I noticed that `exec`, `attach` and `portforward` are now encrypted.

I am sure I am missing something but with these now being encrypted how exactly do i get around the TLS error?

mikebrow:
Not aware of a TLS error. Can you elaborate.

DefunctExodus:
Sure sorry, did that on my phone this morning

I am seeing this following error `TLS handshake error from XX.XX.XX.XX:55048: remote error: tls: bad certificate`

Initially when i ran into this error on Friday i figured it was something with my setup and the upgrade to 1.10, i suppose it could still be however this only happens when this plugin. I eventually ended up switching back to using docker to see if i was able to `exec` into a test container and it worked correctly.

Using this plugin i see no errors during startup or  looking at the logs of a container. It only happens when i am trying to `exec` in.

mikebrow:
exec in using crictl?  

If crictl need to get you to move up to the latest version. If you are running from developer environment just need to make install.deps..

DefunctExodus:
Sorry my bad i am talking about getting into the container through Kubernetes

mikebrow:
kube 1.10? with containerd 1.10?

DefunctExodus:
yeah

mikebrow:
ok will have to check it out... thx

Random-Liu:
@mikebrow We may want to make the encryption optional. In GKE, we don't validate certificate, but I guess in other setup, people may do that.

Let's make the encryption opt-in.

mikebrow:
done.. let's push #714 when it passes ..

Random-Liu:
@DefunctExodus I can't reproduce this in my cluster, `kubectl exec` works in my cluster.

How did you setup your cluster?

DefunctExodus:
@Random-Liu 

Yep it is, I actually took this node out of the cluster and rebuilt it using TLS Bootstrapping, The node successfully joins the cluster and works as you would except.

The only thing that is not working right now is `kubectl exec` 

The rest of my nodes are still running 1.9.2 and CRI beta 1 and it works correctly, and like i said switching to using docker as the run-time engine does not produce this error.

Random-Liu:
@DefunctExodus 

[...]

PR #714, opened 2018-04-02
adds a new flag to enable TLS support (insecure for now)

mikebrow:
Addresses issue  #712 by disabling TLS support by default. Some clients do not support insecure TLS.


Random-Liu:
/lgtm

Not sure whether disabling TLS can resolve #712, it seems so because `1.0.0-beta.1` does work for @DefunctExodus.

Anyway, I do think that there should be a way to disable TLS.

PR #720, opened 2018-04-03
Update containerd once before release.

Random-Liu:
Update containerd vendor once to avoid surprise.


Random-Liu:
@mikebrow Our containerd vendors are in sync with containerd HEAD. We need to sync our vendors back to containerd when I update the vendor in containerd later.

mikebrow:
> @mikebrow Our containerd vendors are in sync with containerd HEAD. We need to sync our vendors back to containerd when I update the vendor in containerd later.

Yes.. I was just reminding :-)

PR #724, opened 2018-04-07
Add cni config template

Random-Liu:
This PR:
1) Updates containerd to v1.1.0-rc.1.
2) Added a new config option `plugins.cri.cni.conf_template`.
  * If it is specified, containerd will get pod CIDR from kubelet through [`UpdateRuntimeConfig`](https://github.com/containerd/cri/blob/master/pkg/server/update_runtime_config.go), and generate a cni config file `10-containerd-net.conflist`. This is useful for cluster which only needs simple cni network setup.
  * If it is not specified, containerd will still wait for system admin or pod network addons to drop the cni config file. This is still the default behavior.

For GCE, we need a p2p config:
```
{
  "name": "k8s-pod-network",
    "cniVersion": "0.3.1",
    "plugins": [
    {
      "type": "ptp",
      "mtu": 1460,
      "ipam": {
	"type": "host-local",
	"subnet": "{{.PodCIDR}}",
	"routes": [
	{"dst": "0.0.0.0/0"}
	]
      }
    },
    {
      "type": "portmap",
      "capabilities": {
	"portMappings": true
      },
      "noSnat": true
    }
    ]
}
```
For other vendors, they can specify their own template.

@yujuhong @dchen1107 @jingax10 @abhi


pkg/server/update_runtime_config.go
```diff
@@ -17,13 +17,54 @@ limitations under the License.
...
 // UpdateRuntimeConfig updates the runtime config. Currently only handles podCIDR updates.
-// TODO(random-liu): Figure out how to handle pod cidr in the cri plugin.
 func (c *criService) UpdateRuntimeConfig(ctx context.Context, r *runtime.UpdateRuntimeConfigRequest) (*runtime.UpdateRuntimeConfigResponse, error) {
+	podCIDR := r.GetRuntimeConfig().GetNetworkConfig().GetPodCidr()
+	if podCIDR == "" {
+		return &runtime.UpdateRuntimeConfigResponse{}, nil
+	}
+	if err := c.netPlugin.Status(); err == nil {
+		// network is ready, no need to generate cni config.
+		return &runtime.UpdateRuntimeConfigResponse{}, nil
```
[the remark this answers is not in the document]
Random-Liu: Good catch

pkg/server/update_runtime_config.go
```diff
@@ -17,13 +17,54 @@ limitations under the License.
...
+	}
+	confTemplate := c.config.NetworkPluginConfTemplate
+	if confTemplate == "" {
+		// no config template is specified, no need to
+		// generate cni config.
```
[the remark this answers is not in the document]
Random-Liu: Done

cluster/gce/cloud-init/master.yaml
```diff
@@ -38,7 +38,8 @@ write_files:
         enable_tls_streaming = true
       [plugins.cri.cni]
         bin_dir = "/home/containerd/opt/cni/bin"
```
abhi: any reason this still needs to be under ```/home/containerd``` Why not keep it consistent ? :)

pkg/server/update_runtime_config.go
```diff
@@ -17,13 +17,60 @@ limitations under the License.
...
+
 // UpdateRuntimeConfig updates the runtime config. Currently only handles podCIDR updates.
-// TODO(random-liu): Figure out how to handle pod cidr in the cri plugin.
 func (c *criService) UpdateRuntimeConfig(ctx context.Context, r *runtime.UpdateRuntimeConfigRequest) (*runtime.UpdateRuntimeConfigResponse, error) {
+	podCIDR := r.GetRuntimeConfig().GetNetworkConfig().GetPodCidr()
+	if podCIDR == "" {
+		return &runtime.UpdateRuntimeConfigResponse{}, nil
+	}
+	confTemplate := c.config.NetworkPluginConfTemplate
+	if err := c.netPlugin.Status(); err == nil {
```
abhi: Shouldn't it load and then check status ?

cluster/gce/env
```diff
@@ -13,7 +13,7 @@ export KUBE_NODE_EXTRA_METADATA="user-data=${GCE_DIR}/cloud-init/node.yaml,conta
 export KUBE_CONTAINER_RUNTIME="remote"
 export KUBE_CONTAINER_RUNTIME_ENDPOINT="/run/containerd/containerd.sock"
 export KUBE_LOAD_IMAGE_COMMAND="/home/containerd/usr/local/bin/ctr cri load"
-export NETWORK_POLICY_PROVIDER="calico"
+export NETWORK_PROVIDER=""
```
abhi: shouldn't this be an option for gce ?

cluster/gce/cni.template
```diff
@@ -0,0 +1,24 @@
...
+	{"dst": "0.0.0.0/0"}
+	]
+      }
+    },
+    {
+      "type": "portmap",
+      "capabilities": {
+	"portMappings": true
+      },
+      "noSnat": true
```
abhi: What would be the podCidr passed here? Since snat is turned off ,Is it a globally routable IP ?

cluster/gce/cloud-init/master.yaml
Random-Liu: Because we place all cni binaries there.

If I remember correctly, `/opt/cni/bin` on COS image is readonly. That's why we can't put it there.

cluster/gce/cni.template
Random-Liu: Yeah, in GCE it is globally routable IP.

cluster/gce/env
Random-Liu: Yeah, this is an option for gce. :)
This is `cluster/gce/env`.

pkg/server/update_runtime_config.go
Random-Liu: Will do.

Random-Liu: Done.

Random-Liu:
@abhi Addressed comments, and update the document to address your concern.

yujuhong:
lgtm

Random-Liu:
Apply LGTM based on the LGTMs above.

PR #725, opened 2018-04-09
Fix resolver race

Random-Liu:
Port https://github.com/containerd/containerd/pull/2276 to the resolver package we are using.

Fixes https://github.com/containerd/cri/issues/722.

### DIFF
--- a/pkg/containerd/resolver/resolver.go
+++ b/pkg/containerd/resolver/resolver.go
@@ -27,6 +27,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/containerd/containerd/images"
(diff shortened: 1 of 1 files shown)


pkg/containerd/resolver/resolver.go
```diff
@@ -250,12 +251,12 @@ func (r *containerdResolver) Pusher(ctx context.Context, ref string) (remotes.Pu
...
 
-	client   *http.Client
-	useBasic bool
-	username string
-	secret   string
+	client           *http.Client
+	useBasic         bool
+	username, secret string
+	token            string
+	mu               sync.Mutex
```
mikebrow: // lock used to protect access to token..

pkg/containerd/resolver/resolver.go
```diff
@@ -300,6 +301,23 @@ func (r *containerdResolver) base(refspec reference.Spec) (*dockerBase, error) {
 	}, nil
 }
 
+func (r *dockerBase) getToken() string {
```
mikebrow: // getToken returns token value protected by mu mutex

pkg/containerd/resolver/resolver.go
```diff
@@ -300,6 +301,23 @@ func (r *containerdResolver) base(refspec reference.Spec) (*dockerBase, error) {
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	return r.token
+}
+
+func (r *dockerBase) setToken(token string) bool {
```
mikebrow: ditto

Random-Liu:
@mikebrow I prefer keeping the change identical to containerd/containerd#2276 to make it easy to port future changes. :)

We should consolidate this in the future anyway https://github.com/containerd/cri/issues/723.

PR #729, opened 2018-04-11
Add socat back.

Random-Liu:
`io.Copy` doesn't work for us. The main problem is that:
1) When one side (either pod side or user side) of portforward is closed, we should stop port forwarding.
2) When one side is closed, the `io.Copy` use that side as source will close, but the `io.Copy` use that side as dest won't. (`io.Copy` doesn't stop when dest is closed until there is new input from source)
3) However, we can't simply forcibly stop the hanging `io.Copy` by closing the source. Because TCP is bi-direction, even though one direction is closed (we know because one `io.Copy` stops), the other direction may not, and that `io.Copy` might be working. There is no way for us to know it now.

There should be some solution for this problem, and socat must have implemented that. We need more time to investigate it.

So let's switch back to `socat` for this release.

@abhi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ install:
     - sudo apt-get install btrfs-tools
     - sudo apt-get install libseccomp2/trusty-backports
     - sudo apt-get install libseccomp-dev/trusty-backports
+    - sudo apt-get install socat
 
 before_script:
     - export PATH=$HOME/gopath/bin:$PATH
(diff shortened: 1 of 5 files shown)


AkihiroSuda:
> There should be some solution for this problem, and socat must have implemented that. We need more time to investigate it.

Could you open a github issue for tracking this?

Random-Liu:
@AkihiroSuda Done. https://github.com/containerd/cri/issues/730

Apply LGTM based on https://github.com/containerd/cri/pull/729#pullrequestreview-111076363

PR #742, opened 2018-04-18
Change node e2e to use cni template.

Random-Liu:
Node e2e should also use cni template.


Random-Liu:
@@jingax10 @yujuhong This is very interesting... After switching from `bridge` to `p2p`, the container cpu usage even reduces. :)

Random-Liu:
This should be fixed by https://github.com/kubernetes/kubernetes/pull/62761.

Random-Liu:
Close this in favor of https://github.com/containerd/cri/pull/740

PR #743, opened 2018-04-18
Fix sandbox stop race condition.

Random-Liu:
`sandbox_stop.go` should only kill the sandbox container, and event monitor will handle `TaskExit` event and `Delete` the task.

Errors caused by this race condition:
```
time="2018-04-18T00:53:09Z" level=error msg="Failed to handle event &TaskExit{ContainerID:e162c2c3f61deeebc88a9f12219fc574eb2b3a47a5bd9d0e1a79f2fc643a37c4,ID:e162c2c3f61deeebc88a9f12219fc574eb2b3a47a5bd9d0e1a79f2fc643a37c4,Pid:21642,ExitStatus:137,ExitedAt:2018-04-18 00:53:09.292640191 +0000 UTC,} for container e162c2c3f61deeebc88a9f12219fc574eb2b3a47a5bd9d0e1a79f2fc643a37c4" error="failed to handle sandbox TaskExit event: failed to stop sandbox: container must be created: failed precondition"
```

### DIFF
--- a/pkg/server/sandbox_stop.go
+++ b/pkg/server/sandbox_stop.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
+	"golang.org/x/sys/unix"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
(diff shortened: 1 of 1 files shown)


pkg/server/sandbox_stop.go
```diff
@@ -105,10 +106,10 @@ func (c *criService) stopSandboxContainer(ctx context.Context, sandbox sandboxst
 		return errors.Wrap(err, "failed to get sandbox container")
 	}
 
-	// Delete the sandbox container from containerd.
-	_, err = task.Delete(ctx, containerd.WithProcessKill)
+	// Kill the sanbox container.
```
mikebrow: /s/san/sand/

mikebrow: Please update function comment on line 98 to:
// stopSandboxContainer kills the sandbox container.
// note: task.Delete is not called here because it will be called when the event monitor handles the TaskExit event

Random-Liu: Will do.

Random-Liu: Done.

PR #748, opened 2018-04-21
Add `unix://` prefix for socket addresses used by CRI remote client.

Random-Liu:
Kubelet and `crictl` prefers socket address with `unix://` prefix, or else it prints out warning. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/util/util_unix.go#L75

For example:
```
# crictl ps
W0420 17:20:40.130286  237131 util_unix.go:75] Using "/run/containerd/containerd.sock" as endpoint is deprecated, please consider using full url format "unix:///run/containerd/containerd.sock".
CONTAINER ID        IMAGE               CREATED             STATE               NAME                ATTEMPT
```

This PR adds `unix://` prefix to all containerd socket addresses except the ones passed to containerd client, because containerd client doesn't work with the `unix://` prefix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ sudo containerd
 ```
 2. From the Kubernetes project directory startup a local cluster using `containerd`:
 ```bash
-CONTAINER_RUNTIME=remote CONTAINER_RUNTIME_ENDPOINT='/run/containerd/containerd.sock' ./hack/local-up-cluster.sh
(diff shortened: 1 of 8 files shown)


Random-Liu:
This PR also updates cri-tools to include the fix https://github.com/kubernetes-incubator/cri-tools/pull/291

PR #754, opened 2018-04-26
os.Unmount: do not consult mountinfo

kolyshkin:
Currently, Unmount() call takes a burden to parse the whole nine yards of /proc/self/mountinfo to figure out whether the given mount point is mounted or not (and returns an error in case parsing fails somehow).

Instead, let's just call umount() and ignore EINVAL, which results in the same behavior, but much better performance.

Note that EINVAL is returned from umount(2) not only in the case when target is not mounted, but also for invalid flags. To avoid ignoring a valid error, the check for valid flags is added.

Note that MNT_EXPIRE is not allowed as it requires a special handling.

While at it, remove the 'containerdmount' alias from the package.

### DIFF
--- a/pkg/os/os.go
+++ b/pkg/os/os.go
@@ -22,9 +22,8 @@ import (
 	"os"
 	"path/filepath"
 
-	containerdmount "github.com/containerd/containerd/mount"
+	"github.com/containerd/containerd/mount"
 	"github.com/containerd/fifo"
-	"github.com/docker/docker/pkg/mount"
 	"github.com/docker/docker/pkg/symlink"
 	"golang.org/x/net/context"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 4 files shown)


kolyshkin:
The only thing I am not sure is if `panic()` is the best thing to do in there...

Random-Liu:
/ok-to-test

@kolyshkin Thanks for this improvement!

2 things worth mentioning:
1) Actually we do only use `unix.MNT_DETACH` in `cri`. I'm fine with not exposing flag from the util function for now to simplify the error handling, just like what moby is doing today. https://github.com/moby/moby/blob/53982e3fc1642f2605189cea48bc6de7b9318611/pkg/mount/mount.go#L92
2) The previous unmount implementation will return `nil` if the mount point does not exist, the behavior will be changed in this PR. Fortunately, the only 2 places calling this function both check `os.IsNotExist`, so this change should not cause problem.
* https://github.com/containerd/cri/blob/master/pkg/server/sandbox_run.go#L506
* https://github.com/containerd/cri/blob/master/pkg/store/sandbox/netns.go#L96

But could you clarify that "this function returns 'not exist error' if the mount point does not exist"?

kolyshkin:
> The previous unmount implementation will return nil if the mount point does not exist

this totally eluded from my radar, thanks for pointing it out!

Please see the updated patch. I have dropped the `flags` argument (and therefore the check), and documented the ENOENT returned.

Random-Liu:
/lgtm

@kolyshkin Thanks for the contribution!

PR #755, opened 2018-04-27
Always mount sysfs as `rw`.

Random-Liu:
Similar with https://github.com/moby/moby/pull/36808.

In `cri`, we don't change mount `ro/rw` based on `ReadonlyRootfs`, and strictly respect whatever CRI specifies. https://github.com/Random-Liu/cri-containerd/blob/42a22ae1991dc4ef354d81b6b48a1efb6387c339/pkg/server/container_create.go#L644-L648 

[Default mounts](https://github.com/containerd/containerd/blob/master/oci/spec_unix.go#L111) won't be affected by `ReadonlyRootfs`, so we should be able to safely set `sysfs` as `rw` for `Privileged=true` regardless of `ReadonlyRootfs`.


justincormack:
Yes this is correct (we changed Docker recently to match this too).

PR #758, opened 2018-05-02
Use crictl installed in kube-up.sh

Random-Liu:
Depends on https://github.com/kubernetes/kubernetes/pull/63357.

kube-up.sh will install crictl on GCE. We don't need to install it our own.

### DIFF
--- a/cluster/gce/configure.sh
+++ b/cluster/gce/configure.sh
@@ -133,13 +133,13 @@ else
     # Download and untar the release tar ball.
     curl -f --ipv4 -Lo "${TARBALL}" --connect-timeout 20 --max-time 300 --retry 6 --retry-delay 10 "${TARBALL_GCS_PATH}"
     tar xvf "${TARBALL}"
+    # Use crictl installed by kube-up.sh.
+    # TODO(random-liu): Create release tarball for GCE, and do not package crictl.
+    rm "${CONTAINERD_HOME}/usr/local/bin/crictl"
+    rm "${CONTAINERD_HOME}/etc/crictl.yaml"
     rm -f "${TARBALL}"
   fi
 fi
-
-# Configure containerd.
-# Copy crictl config.
-cp "${CONTAINERD_HOME}/etc/crictl.yaml" /etc
 
 # Generate containerd config
 config_path="${CONTAINERD_CONFIG_PATH:-"/etc/containerd/config.toml"}"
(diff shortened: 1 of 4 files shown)


Random-Liu:
https://github.com/kubernetes/kubernetes/pull/63357 is merged.

This PR is unblocked.

yujuhong:
Do we need to worry about non-kube-up cluster setup?

Random-Liu:
@yujuhong The bootstrapping thing in this repo is only for `kube-up.sh` now.

cluster/gce/configure.sh
```diff
@@ -133,14 +133,14 @@ else
     # Download and untar the release tar ball.
     curl -f --ipv4 -Lo "${TARBALL}" --connect-timeout 20 --max-time 300 --retry 6 --retry-delay 10 "${TARBALL_GCS_PATH}"
     tar xvf "${TARBALL}"
+    # Use crictl installed by kube-up.sh.
+    # TODO(random-liu): Create release tarball for GCE, and do not package crictl.
+    rm "${CONTAINERD_HOME}/usr/local/bin/crictl"
+    rm "${CONTAINERD_HOME}/etc/crictl.yaml"
```
mikebrow: We good here? Just checking:
1) will GCE health monitor look for containerd not docker? 
2) How do we ensure now, that the right version of crictl has been installed for a particular install of containerd/cri? Maybe a sanity check here, and report if the version is out of date?

cluster/health-monitor.sh
```diff
@@ -1,51 +0,0 @@
-#!/bin/bash
```
mikebrow: Seems odd to remove this... was/is this only useful for GCE?

cluster/gce/configure.sh
Random-Liu: `crictl` will ship with Kubernetes on GCE now kubernetes/kubernetes#63357. It has to work with that version of Kubernetes.

If `crictl` works with that version of Kubernetes, but not `containerd`, it means we are using a wrong `containerd` version.

cluster/health-monitor.sh
Random-Liu: Currently in the code base, only GCE is using this.

However, I think what you said makes sense, there might be someone using this script. Given so, I think I'll leave the file here for one more release, and file an issue to remove it next release.

Random-Liu: Done.

PR #761, opened 2018-05-05
Add log max size

Random-Liu:
Fixes https://github.com/containerd/cri/issues/283.

This PR added a configuration `max_container_log_size` for configuring max container log line size. Log line longer than the size limit will be split into multiple lines.

@yujuhong

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -27,6 +27,11 @@ The explanation and default value of each configuration item are as follows:
   # enable_tls_streaming enables the TLS streaming support.
   enable_tls_streaming = false
 
+  # max_container_log_line_size is the maximum log line size in bytes for a container.
+  # Log line longer than the limit will be split into multiple lines. -1 means no
+  # limit.
+  max_container_log_line_size = 16384
+
   # "plugins.cri.containerd" contains config related to containerd
   [plugins.cri.containerd]
 
--- a/pkg/ioutil/write_closer_test.go
+++ b/pkg/ioutil/write_closer_test.go
@@ -17,9 +17,16 @@ limitations under the License.
 package ioutil
 
 import (
+	"io/ioutil"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestWriteCloseInformer(t *testing.T) {
@@ -46,4 +53,56 @@ func TestWriteCloseInformer(t *testing.T) {
 	default:
 		assert.Fail(t, "write closer not closed")
 	}
+}
+
+func TestSerialWriteCloser(t *testing.T) {
+	const (
+		// Test 10 times to make sure it always pass.
+		testCount = 10
+
+		goroutine = 10
+		dataLen   = 100000
+	)
+	for n := 0; n < testCount; n++ {
+		testData := make([][]byte, goroutine)
+		for i := 0; i < goroutine; i++ {
+			testData[i] = []byte(repeatNumber(i, dataLen) + "\n")
+		}
+
+		f, err := ioutil.TempFile("/tmp", "serial-write-closer")
+		require.NoError(t, err)
+		defer os.RemoveAll(f.Name())
+		defer f.Close()
+		wc := NewSerialWriteCloser(f)
+		defer wc.Close()
+
+		// Write data in parallel
+		var wg sync.WaitGroup
+		wg.Add(goroutine)
+		for i := 0; i < goroutine; i++ {
+			go func(id int) {
+				n, err := wc.Write(testData[id])
+				assert.NoError(t, err)
+				assert.Equal(t, dataLen+1, n)
+				wg.Done()
+			}(i)
+		}
+		wg.Wait()
+		wc.Close()
+
+		// Check test result
+		content, err := ioutil.ReadFile(f.Name())
+		require.NoError(t, err)
+		resultData := strings.Split(strings.TrimSpace(string(content)), "\n")
+		require.Len(t, resultData, goroutine)
+		sort.Strings(resultData)
+		for i := 0; i < goroutine; i++ {
+			expected := repeatNumber(i, dataLen)
+			assert.Equal(t, expected, resultData[i])
+		}
+	}
+}
+
+func repeatNumber(num, count int) string {
+	return strings.Repeat(strconv.Itoa(num), count)
 }
--- a/pkg/server/container_start.go
+++ b/pkg/server/container_start.go
@@ -18,6 +18,7 @@ package server
 
 import (
 	"io"
+	"os"
 	"time"
 
 	"github.com/containerd/containerd"
@@ -29,6 +30,7 @@ import (
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 
 	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
+	cioutil "github.com/containerd/cri/pkg/ioutil"
 	cio "github.com/containerd/cri/pkg/server/io"
 	containerstore "github.com/containerd/cri/pkg/store/container"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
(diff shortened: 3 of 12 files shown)


thaJeztah:
I know user configurable buffer size has been a hot topic on dockerd; configuring a huge size can easily lead to memory issues if there's congestion; is that an issue here as well?

/cc @cpuguy83

Random-Liu:
@thaJeztah The read buffer size is 4096 fixed. If there is no long line, the memory usage will be 4096 (fixed read buffer size); if there is, the memory usage grows once for the log line.

Since docker is a user facing tool, I agree that we may not want user to care about and configure this directly.

However, to Kubernetes, `containerd` and `cri` are just building block. And the configuration is just for system admin. They should be able to configure the behavior according to their system environment. For example, in GCP, stackdriver has a 100kb log length limit, we want the log generated by container runtime to match that limit.

Another solution of this is to do this in fluentd. We've thought about it, and https://github.com/fluent-plugins-nursery/fluent-plugin-concat can probably be used. However, since not everyone is using fluentd, we think that it is still better to make this configurable in container runtime instead

pkg/server/container_start.go
```diff
@@ -142,24 +134,36 @@ func (c *criService) startContainer(ctx context.Context,
...
-		if stdout, err = cio.NewCRILogger(logPath, cio.Stdout); err != nil {
-			return nil, nil, errors.Wrap(err, "failed to start container stdout logger")
+		f, err := os.OpenFile(logPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0640)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "failed to open log file")
 		}
 		defer func() {
 			if err != nil {
-				stdout.Close()
+				f.Close()
```
mikebrow: in the error case I don't think we need to close "f" ...

pkg/ioutil/write_closer_test.go
```diff
@@ -47,3 +54,51 @@ func TestWriteCloseInformer(t *testing.T) {
 		assert.Fail(t, "write closer not closed")
 	}
 }
+
+func TestSerialWriteCloesr(t *testing.T) {
```
mikebrow: /s/Cloesr/Closer/

pkg/server/container_start.go
```diff
@@ -142,24 +134,36 @@ func (c *criService) startContainer(ctx context.Context,
...
+			stderr, stderrCh = cio.NewCRILogger(logPath, wc, cio.Stderr, c.config.MaxContainerLogSize)
 		}
+		go func() {
+			if stdoutCh != nil {
+				<-stdoutCh
+			}
+			if stderrCh != nil {
+				<-stderrCh
+			}
+			logrus.Debugf("Finish redirecting log file %q, close it", logPath)
```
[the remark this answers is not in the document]
mikebrow: though it might be better to just remove as this may just clutter up the debug log .. no?

Random-Liu:
@mikebrow Thanks for reviewing! Will address comments today.

docs/config.md
```diff
@@ -27,6 +27,11 @@ The explanation and default value of each configuration item are as follows:
   # enable_tls_streaming enables the TLS streaming support.
   enable_tls_streaming = false
 
+  # max_container_log_size is the maximum log line size in bytes for a container.
+  # Log line longer than the limit will be split into multiple lines. Negative
+  # value means no limit.
```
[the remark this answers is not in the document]
Random-Liu: Done.

pkg/server/container_start.go
Random-Liu: Why not?

pkg/ioutil/write_closer_test.go
Random-Liu: Will do.

Random-Liu: Done.

pkg/ioutil/write_closer_test.go
```diff
@@ -47,3 +54,51 @@ func TestWriteCloseInformer(t *testing.T) {
 		assert.Fail(t, "write closer not closed")
 	}
 }
+
+func TestSerialWriteCloser(t *testing.T) {
+	const (
+		// Test 100 times to make sure it always pass.
```
Random-Liu: 100 --> 10

pkg/server/container_start.go
Random-Liu: That error is checked before this defer.

mikebrow: exactly and the only way the function can return error (atm) is if it failed to open.. see "return" statements :-)

Random-Liu:
@mikebrow @yujuhong Thanks for reviewing! Will try to address comments today.

Random-Liu:
Distracted by other things. Will address comments today.

pkg/server/container_start.go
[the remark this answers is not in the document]
Random-Liu: The log here is to indicate the goroutine stops.

pkg/server/io/logger.go
```diff
@@ -38,11 +36,8 @@ const (
 	eol = '\n'
 	// timestampFormat is the timestamp format used in CRI logging format.
 	timestampFormat = time.RFC3339Nano
-	// pipeBufSize is the system PIPE_BUF size, on linux it is 4096 bytes.
-	// POSIX.1 says that write less than PIPE_BUF is atmoic.
-	pipeBufSize = 4096
 	// bufSize is the size of the read buffer.
-	bufSize = pipeBufSize - len(timestampFormat) - len(Stdout) - len(runtime.LogTagPartial) - 3 /*3 delimiter*/ - 1 /*eol*/
+	bufSize = 4096
```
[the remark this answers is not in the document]
Random-Liu: Done

pkg/server/io/logger.go
```diff
@@ -51,46 +46,103 @@ func NewDiscardLogger() io.WriteCloser {
...
+			length += len(l)
 		}
-		tagBytes := fullBytes
-		if isPrefix {
-			tagBytes = partialBytes
+		var tag, line []byte
+		switch {
+		case maxLen < 0:
+			// Always try reading the full line, if no max length limit.
+			fallthrough
```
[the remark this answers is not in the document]
Random-Liu: Simplified and fixed the logic here.

pkg/server/io/logger.go
```diff
@@ -51,46 +46,103 @@ func NewDiscardLogger() io.WriteCloser {
...
+		timestamp := time.Now().AppendFormat(nil, timestampFormat)
+		data := bytes.Join([][]byte{timestamp, stream, tag, line}, delimiter)
 		data = append(data, eol)
-		if _, err := wc.Write(data); err != nil {
-			logrus.WithError(err).Errorf("Fail to write %q log to log file %q", stream, path)
+		if _, err := w.Write(data); err != nil {
+			logrus.WithError(err).Errorf("Fail to write %q log to log file %q", s, path)
 		}
 		// Continue on write error to drain the input.
+		if stop {
```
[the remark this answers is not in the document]
Random-Liu: We've already drain the input. I just mean we should not return error if we fail to write one line, because we should continue draining the container output.

I put the comment in a wrong place. Updated.

Random-Liu:
@yujuhong @mikebrow Addressed comments.

pkg/server/container_start.go
mikebrow: ok.. yeah that's true that it could become useful in the future..

pkg/ioutil/write_closer_test.go
```diff
@@ -47,3 +54,51 @@ func TestWriteCloseInformer(t *testing.T) {
...
+		wc.Close()
+
+		// Check test result
+		content, err := ioutil.ReadFile(f.Name())
+		require.NoError(t, err)
+		resultData := strings.Split(strings.TrimSpace(string(content)), "\n")
+		require.Len(t, resultData, goroutine)
+		sort.Strings(resultData)
+		for i := 0; i < goroutine; i++ {
+			expected := strings.Repeat(strconv.Itoa(i), dataLen)
```
yujuhong: nit: make `strings.Repeat(strconv.Itoa(i), dataLen)` a function to reuse.

pkg/ioutil/write_closer_test.go
```diff
@@ -47,3 +54,51 @@ func TestWriteCloseInformer(t *testing.T) {
...
 	}
 }
+
+func TestSerialWriteCloser(t *testing.T) {
+	const (
+		// Test 10 times to make sure it always pass.
+		testCount = 10
+
+		goroutine = 10
+		dataLen   = 100000 // larger than PIPE_BUF=4096
```
yujuhong: Does the same test fail if non-serial writer is used?

pkg/ioutil/write_closer_test.go
Random-Liu: Will do.

Random-Liu: Done.

pkg/ioutil/write_closer_test.go
Random-Liu: https://stackoverflow.com/questions/29981050/concurrent-writing-to-a-file

Random-Liu: It only has problem with old kernel. I can't reproduce this in my desktop. :)

PR #766, opened 2018-05-11
Fix workingset memory calculation.

Random-Liu:
Fix container workingset memory calculation.

In kubernetes and cadvisor, workingset is defined as "usage - total_inactive_file".
See https://github.com/google/cadvisor/blob/08f0c2397cbca790a4db0f1212cb592cc88f6e26/container/libcontainer/handler.go#L520.

/cc @dashpole @abhi 

### DIFF
--- a/pkg/server/container_stats_list.go
+++ b/pkg/server/container_stats_list.go
@@ -109,8 +109,10 @@ func (c *criService) getContainerMetrics(
 		}
 		if metrics.Memory != nil && metrics.Memory.Usage != nil {
 			cs.Memory = &runtime.MemoryUsage{
-				Timestamp:       stats.Timestamp.UnixNano(),
-				WorkingSetBytes: &runtime.UInt64Value{Value: metrics.Memory.Usage.Usage},
+				Timestamp: stats.Timestamp.UnixNano(),
+				WorkingSetBytes: &runtime.UInt64Value{
+					Value: getWorkingSet(metrics.Memory),
+				},
 			}
 		}
 	}
@@ -166,4 +168,18 @@ func matchLabelSelector(selector, labels map[string]string) bool {
 		}
 	}
 	return true
+}
+
+// getWorkingSet calculates workingset memory from cgroup memory stats.
+// The caller should make sure memory is not nil.
+// workingset = usage - total_inactive_file
+func getWorkingSet(memory *cgroups.MemoryStat) uint64 {
+	if memory.Usage == nil {
+		return 0
+	}
+	var workingSet uint64
+	if memory.TotalInactiveFile < memory.Usage.Usage {
+		workingSet = memory.Usage.Usage - memory.TotalInactiveFile
+	}
+	return workingSet
 }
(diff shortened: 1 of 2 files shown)


pkg/server/container_stats_list.go
```diff
@@ -107,10 +107,12 @@ func (c *criService) getContainerMetrics(
 				UsageCoreNanoSeconds: &runtime.UInt64Value{Value: metrics.CPU.Usage.Total},
 			}
 		}
-		if metrics.Memory != nil && metrics.Memory.Usage != nil {
+		if metrics.Memory != nil {
```
dashpole: since we dont check `metrics.Memory.Usage != nil` anymore, missing memory metrics result in a adding a value with WorkingSet = 0, instead of `cs.Memory = nil`.  I don't know the codebase, so I don't know if this matters.

Random-Liu: Hm, good catch. I'll keep the behavior unchanged.

pkg/server/container_stats_list.go
```diff
@@ -167,3 +169,19 @@ func matchLabelSelector(selector, labels map[string]string) bool {
...
 }
+
+// getWorkingSet calculates workingset memory from cgroup memory stats.
+// The caller should make sure memory is not nil.
+// workingset = usage - total_inactive_file
+func getWorkingSet(memory *cgroups.MemoryStat) uint64 {
+	if memory.Usage == nil {
+		return 0
+	}
+	workingSet := memory.Usage.Usage
```
mikebrow: should be set to zero here

pkg/server/container_stats_list.go
```diff
@@ -167,3 +169,19 @@ func matchLabelSelector(selector, labels map[string]string) bool {
...
+	if memory.TotalInactiveFile < workingSet {
```
mikebrow: should compare with memory.Usage.Usage here

pkg/server/container_stats_list.go
```diff
@@ -167,3 +169,19 @@ func matchLabelSelector(selector, labels map[string]string) bool {
...
+		workingSet -= memory.TotalInactiveFile
+	} else {
```
mikebrow: then remove else case

pkg/server/container_stats_list.go
```diff
@@ -167,3 +169,19 @@ func matchLabelSelector(selector, labels map[string]string) bool {
...
+// getWorkingSet calculates workingset memory from cgroup memory stats.
+// The caller should make sure memory is not nil.
+// workingset = usage - total_inactive_file
+func getWorkingSet(memory *cgroups.MemoryStat) uint64 {
+	if memory.Usage == nil {
+		return 0
+	}
+	workingSet := memory.Usage.Usage
+	if memory.TotalInactiveFile < workingSet {
+		workingSet -= memory.TotalInactiveFile
```
mikebrow: workingSet = memory.Usage.Usage - memory.TotalInactiveFile..

Random-Liu:
@mikebrow Done

PR #776, opened 2018-05-19
Disable TLS streaming to work with new kubelet streaming proxy.

Random-Liu:
Disable TLS streaming to work with new kubelet streaming proxy added in https://github.com/kubernetes/kubernetes/pull/64006.


Random-Liu:
Will get this merged once https://github.com/kubernetes/kubernetes/pull/64006 is approved.

PR #785, opened 2018-05-24
containerd vendor update

ehazlett:
This is a containerd vendor update for https://github.com/containerd/containerd/pull/2353.  Since there is a circular vendor dependency between the projects I'll do this in two stages.  First this with my fork then a follow up once it's merged on the containerd side to go to master (thx @Random-Liu for the help!)


ehazlett:
Do I need to squash the commits to get the vendor change?

ehazlett:
Cool thanks.  I'm not sure either.  I updated the commit and ran `vndr`.  Should I update a different way?

mikebrow:
sometimes I rm the vendor/ tree first.

ehazlett:
CI is failing the tests because of `install-deps` (they work fine locally).  I have my fork of containerd in the vendor.conf and it looks like https://github.com/containerd/cri/blob/master/hack/install/utils.sh#L49 is adding the `https` prefix.  If I remove that then `vndr` fails.  I can update the utils script but I'm not sure what that might break.

Random-Liu:
We need to get rid of `https` prefix here https://github.com/containerd/cri/blob/67908190c5117e8889751a3eeab9326081078f13/hack/utils.sh#L84.

ehazlett:
OK I _think_ this is ready now :)  PR on the containerd side was merged (https://github.com/containerd/containerd/pull/2353).  I will squash if it looks good.

hack/utils.sh
```diff
@@ -81,18 +81,19 @@ from-vendor() {
                   BEGIN { rc=1 }                        # Assume we did not find what we were looking for.
                   // {
                      if ($1 == REPO) {
-                        if ($3 != "") { REPO = $3 };    # Override repo.
+                        if ($3 != "") { gsub(/http.*\/\//, "", $3); REPO = $3 };    # Override repo.
                         printf("%s_VERSION=%s; %s_REPO=%s\n", WHAT, $2, WHAT, REPO);
                         rc=0;                           # Note success for use in END block.
-                        exit                            # No point looking further.
+                        #exit                            # No point looking further.
```
Random-Liu: Remove if not needed?

hack/utils.sh
```diff
@@ -81,18 +81,19 @@ from-vendor() {
...
                      }
                  }
                  END { exit rc }                        # Exit with the desired code.
                  ' $vendor)
     if [ $? -ne 0 ] ; then
         echo "failed to get version of $repo from $vendor" >&2
-        exit 1
+        #exit 1
```
Random-Liu: ditto.

hack/utils.sh
```diff
@@ -81,18 +81,19 @@ from-vendor() {
...
     fi
+    echo $setvars
```
Random-Liu: Add description to the log, or remove?

Random-Liu:
LGTM with nits. Please squash. :)

PR #786, opened 2018-05-25
Remove useless error-check in createImageReference

fuweid:
Remove the useless error check during creating reference in containerd.

ping @Random-Liu

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -199,7 +199,7 @@ func (c *criService) createImageReference(ctx context.Context, name string, desc
 	if err == nil {
 		return nil
 	}
-	if err != nil && !errdefs.IsAlreadyExists(err) {
+	if !errdefs.IsAlreadyExists(err) {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/server/image_pull.go
```diff
@@ -199,7 +199,7 @@ func (c *criService) createImageReference(ctx context.Context, name string, desc
 	if err == nil {
 		return nil
 	}
-	if err != nil && !errdefs.IsAlreadyExists(err) {
+	if !errdefs.IsAlreadyExists(err) {
```
mikebrow: nit.. can we make this an else case so it's explicit that we meant it so.. :-)

fuweid: it is ok w/o `else` here I think...both are explicit, lol

Random-Liu: Since last `if` return directly, it is fine not to add an else. :)

mikebrow: I suppose if one does not like if else chains a switch is more elegant. But yeah was just a nit :-)

PR #789, opened 2018-05-30
Make DEPLOY_PATH configurable.

Random-Liu:
Make `DEPLOY_PATH` configurable.
This is useful for branch test. See https://github.com/kubernetes/test-infra/pull/8158


Random-Liu:
Cherrypick this, so that we can use `cri` `release/1.0` to build and test corresponding `containerd` `release/1.1`.

PR #793, opened 2018-05-31
Port docker resolver fix #2364.

Random-Liu:
Port docker resolver fix https://github.com/containerd/containerd/pull/2364.

We should really get rid of the fork. https://github.com/containerd/cri/issues/723

/cc @abhi @dmcgowan 

### DIFF
--- a/pkg/containerd/resolver/fetcher.go
+++ b/pkg/containerd/resolver/fetcher.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


Random-Liu:
We should cherrypick this, and catch on the containerd 1.1.1 release train.

pkg/containerd/resolver/fetcher_test.go
```diff
@@ -0,0 +1,94 @@
+/*
+Copyright 2018 The Containerd Authors.
+
```
mikebrow: Was going to say.. boiler plate mismatch.. see you already fixed it.

pkg/containerd/resolver/fetcher.go
```diff
@@ -114,6 +116,30 @@ func (r dockerFetcher) open(ctx context.Context, u, mediatype string, offset int
...
 	}
+	if offset > 0 {
+		cr := resp.Header.Get("content-range")
+		if cr != "" {
+			if !strings.HasPrefix(cr, fmt.Sprintf("bytes %d-", offset)) {
+				return nil, errors.Errorf("unhandled content range in response: %v", cr)
+
+			}
+		} else {
+			// TODO: Should any cases where use of content range
```
mikebrow: todo format..

Random-Liu: This is copied from the original containerd fix. Wanna to keep change minimal so as to make it easier to maintain the fork or consolidate in the future.

mikebrow: yes I saw it was a copy.. todo comment would not be much of a merge :-) but already approved skip the nit .. merge away!

PR #794, opened 2018-05-31
Generate fatal error when cri plugin fail to start.

Random-Liu:
Fixes https://github.com/containerd/containerd/issues/2371.

This PR:
1) Let event monitor and streaming server return an error channel.
2) Let CRI plugin `Run` return error if an error is received from the error channels of event monitor or streaming server.
3) Ignore `http.ErrServerClosed` which is a normal error generated when http server is Shutdown/Closed.

@t3hmrman @crosbymichael 

### DIFF
--- a/pkg/server/events.go
+++ b/pkg/server/events.go
@@ -116,15 +116,16 @@ func convertEvent(e *gogotypes.Any) (string, interface{}, error) {
 }
 
 // start starts the event monitor which monitors and handles all container events. It returns
-// a channel for the caller to wait for the event monitor to stop. start must be called after
-// subscribe.
-func (em *eventMonitor) start() (<-chan struct{}, error) {
+// an error channel for the caller to wait for stop errors from the event monitor.
+// start must be called after subscribe.
+func (em *eventMonitor) start() <-chan error {
+	errCh := make(chan error)
 	if em.ch == nil || em.errCh == nil {
-		return nil, errors.New("event channel is nil")
+		panic("event channel is nil")
 	}
-	closeCh := make(chan struct{})
 	backOffCheckCh := em.backOff.start()
 	go func() {
+		defer close(errCh)
 		for {
 			select {
 			case e := <-em.ch:
@@ -144,8 +145,11 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 					em.backOff.enBackOff(cID, evt)
 				}
 			case err := <-em.errCh:
-				logrus.WithError(err).Error("Failed to handle event stream")
-				close(closeCh)
+				// Close errCh in defer directly if there is no error.
+				if err != nil {
+					logrus.WithError(err).Errorf("Failed to handle event stream")
+					errCh <- err
+				}
 				return
 			case <-backOffCheckCh:
 				cIDs := em.backOff.getExpiredContainers()
@@ -162,7 +166,7 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 			}
 		}
 	}()
-	return closeCh, nil
+	return errCh
 }
 
 // stop stops the event monitor. It will close the event channel.
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -19,6 +19,7 @@ package server
 import (
 	"fmt"
 	"io"
+	"net/http"
 	"path/filepath"
 	"time"
 
@@ -179,10 +180,7 @@ func (c *criService) Run() error {
 
 	// Start event handler.
 	logrus.Info("Start event monitor")
-	eventMonitorCloseCh, err := c.eventMonitor.start()
-	if err != nil {
-		return errors.Wrap(err, "failed to start event monitor")
-	}
+	eventMonitorErrCh := c.eventMonitor.start()
 
 	// Start snapshot stats syncer, it doesn't need to be stopped.
 	logrus.Info("Start snapshots syncer")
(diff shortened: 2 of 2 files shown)


Random-Liu:
Log when fail to start streaming server:
```
INFO[0000] Connect containerd service                   
INFO[0000] Get image filesystem path "/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs" 
INFO[0000] loading plugin "io.containerd.grpc.v1.introspection"...  type=io.containerd.grpc.v1
INFO[0000] Start subscribing containerd event           
INFO[0000] Start recovering state                       
INFO[0000] serving...                                    address="/run/containerd/containerd.sock"
INFO[0000] containerd successfully booted in 0.005218s  
INFO[0000] Start event monitor                          
INFO[0000] Start snapshots syncer                       
INFO[0000] Start streaming server                       
ERRO[0000] Failed to start streaming server              error="stayUp=false is not yet implemented"
INFO[0000] Stop CRI service                             
INFO[0000] Event monitor stopped                        
INFO[0000] Stream server stopped                        
FATA[0000] Failed to run CRI service                     error="stream server error: stayUp=false is not yet implemented"
```

t3hmrman:
Hey @Random-Liu thanks so much for the quick fix! This change will definitely improve my setup and hopefully it helps others too.

As always thanks to the team for the work on `containerd` (and of course `cri-containerd`)

pkg/server/events.go
```diff
@@ -118,13 +118,14 @@ func convertEvent(e *gogotypes.Any) (string, interface{}, error) {
 // start starts the event monitor which monitors and handles all container events. It returns
 // a channel for the caller to wait for the event monitor to stop. start must be called after
 // subscribe.
-func (em *eventMonitor) start() (<-chan struct{}, error) {
+func (em *eventMonitor) start() <-chan error {
+	errCh := make(chan error)
 	if em.ch == nil || em.errCh == nil {
-		return nil, errors.New("event channel is nil")
+		logrus.Fatal("event channel is nil")
```
crosbymichael: If this case is a programmer error, not initializing channels, then you should `panic()` not log fatal

pkg/server/events.go
```diff
@@ -144,8 +145,10 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 					em.backOff.enBackOff(cID, evt)
 				}
 			case err := <-em.errCh:
-				logrus.WithError(err).Error("Failed to handle event stream")
-				close(closeCh)
+				if err != nil {
+					logrus.WithError(err).Errorf("Failed to handle event stream")
+					errCh <- err
```
crosbymichael: How do we know the difference between an bootup error and a normal runtime error?  Do we care?

pkg/server/events.go
```diff
@@ -118,13 +118,14 @@ func convertEvent(e *gogotypes.Any) (string, interface{}, error) {
 // start starts the event monitor which monitors and handles all container events. It returns
 // a channel for the caller to wait for the event monitor to stop. start must be called after
 // subscribe.
-func (em *eventMonitor) start() (<-chan struct{}, error) {
+func (em *eventMonitor) start() <-chan error {
```
mikebrow: cool.. Should probably change the function declaration comments to reflect the change in return value. For example, "It returns an error return channel to the caller. Callers should wait for errors to be returned over the error return channel, such as Failed to handle event stream."

pkg/server/events.go
```diff
@@ -144,8 +145,10 @@ func (em *eventMonitor) start() (<-chan struct{}, error) {
 					em.backOff.enBackOff(cID, evt)
 				}
 			case err := <-em.errCh:
-				logrus.WithError(err).Error("Failed to handle event stream")
-				close(closeCh)
+				if err != nil {
```
mikebrow: comment here would be nice // if errCH is nil just return (and close errCh) do not report error with event stream

pkg/server/service.go
```diff
@@ -179,10 +180,7 @@ func (c *criService) Run() error {
 
 	// Start event handler.
 	logrus.Info("Start event monitor")
-	eventMonitorCloseCh, err := c.eventMonitor.start()
-	if err != nil {
-		return errors.Wrap(err, "failed to start event monitor")
-	}
+	eventMonitorErrCh := c.eventMonitor.start()
```
mikebrow: you can still check if eventMonitorErrCh == nil and leave the old return message if it does, though the make chan error should always work...

pkg/server/service.go
```diff
@@ -195,28 +193,32 @@ func (c *criService) Run() error {
...
-	case <-streamServerCloseCh:
+	case eventMonitorErr = <-eventMonitorErrCh:
+	case streamServerErr = <-streamServerErrCh:
 	}
 	if err := c.Close(); err != nil {
 		return errors.Wrap(err, "failed to stop cri service")
 	}
-
-	<-eventMonitorCloseCh
+	if err := <-eventMonitorErrCh; err != nil {
```
mikebrow: ? not sure what's going on here. Should we be overwriting eventMonitorErr if it was set in the above select?

pkg/server/service.go
```diff
@@ -227,18 +229,28 @@ func (c *criService) Run() error {
 	// is fixed.
 	const streamServerStopTimeout = 2 * time.Second
 	select {
-	case <-streamServerCloseCh:
+	case err := <-streamServerErrCh:
+		if err != nil {
+			streamServerErr = err
```
mikebrow: same comment as above ..

pkg/server/events.go
Random-Liu: Will do. Thought panic and fatal are same thing.

pkg/server/events.go
Random-Liu: Since there is no grpc in between now, `Subscribe` either generates a real error https://github.com/containerd/containerd/blob/master/events/exchange/exchange.go#L153, https://github.com/containerd/containerd/blob/master/events/exchange/exchange.go#L194, or a error which should never happen https://github.com/containerd/containerd/blob/master/events/exchange/exchange.go#L178.

pkg/server/events.go
crosbymichael: Same outcome yes, but panic means "I messed up in the code" and a log message with an exit(1) usually just means an application/runtime error, not developer error

Random-Liu: Thanks, :D Got it.

pkg/server/events.go
Random-Liu: Done.

pkg/server/events.go
Random-Liu: Done

pkg/server/service.go
Random-Liu: Yeah, don't think we need to check. :) If it is nil, it is our problem.

pkg/server/service.go
Random-Liu: The above select only waits for one of event monitor and stream server to stop, we don't know which one.

pkg/server/service.go
Random-Liu: ditto.

pkg/server/events.go
Random-Liu: Done.

pkg/server/service.go
mikebrow: kk cool

crosbymichael:
LGTM

t3hmrman:
Hey how would I find out when this PR will make it into `containerd`?  I just ran into this issue today on containerd `1.1.2` so I'm assuming it's not in yet?

PR #803, opened 2018-06-05
Select ipv4 first if there is one.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/783.

Select ipv4 address if there is one.
The expected behavior is not clearly defined in Kubernetes, but this is the implementation for all existing kubernetes network plugins. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/network/plugins.go#L251

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -524,7 +524,7 @@ func (c *criService) setupPod(id string, path string, config *runtime.PodSandbox
 	}
 	// Check if the default interface has IP config
 	if configs, ok := result.Interfaces[defaultIfName]; ok && len(configs.IPConfigs) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/server/sandbox_run.go
```diff
@@ -550,6 +550,16 @@ func toCNIPortMappings(criPortMappings []*runtime.PortMapping) []cni.PortMapping
 	return portMappings
 }
 
+// selectPodIP select an ip from the ip list. It prefers ipv4 more than ipv6.
```
mikebrow: that works

Random-Liu:
@mikebrow @abhi Thanks for reviewing! Will merge after test passes.

PR #808, opened 2018-06-08
Erase ambient capabilities.

Random-Liu:
This is not a huge problem to `containerd/cri`, because our default ambient list is empty. https://github.com/containerd/containerd/blob/master/oci/spec_unix.go#L97

However, we add ambient cap if user explicitly adds it. This is not a CVE, but a behavior change to me. We decided to follow docker's behavior for backward compatibility. And we can carry on discussion about how to better handle this in https://github.com/kubernetes/kubernetes/issues/56374.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -372,6 +372,11 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 				securityContext.GetCapabilities())
 		}
 	}
+	// Clear all ambient capabilities. The implication of non-root + caps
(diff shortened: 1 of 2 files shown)


dmcgowan:
Do you want this in 1.1.1? It is a behavior change, but are you more concerned with being consistent with Docker or 1.1.10? I don't quite understand how such a change might affect users though, is this a commonly used feature which anyone would be relying on if we changed it in 1.1.1?

Random-Liu:
@dmcgowan Will cherry-pick. :) We'll maintain 1.1.x for a long time, let's make sure it is working correctly.

PR #804, opened 2018-06-06
Use pod ip instead of localhost in pod netns for portforward.

Random-Liu:
Ref https://github.com/containerd/cri/issues/524.

This should be equivalent to `netns.Do + localhost`, but this works with kata and gvisor. Because they steal ip from veth.

@abhi @jingax10

Signed-off-by: Lantao Liu <lantaol@google.com>

PR #810, opened 2018-06-11
Revert "Use pod ip instead of localhost in pod netns for portforward."

Random-Liu:
This reverts commit dd886bc281c764c3a40f1c5e1eadca3337f847e4.

### DIFF
--- a/pkg/server/sandbox_portforward.go
+++ b/pkg/server/sandbox_portforward.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


Random-Liu:
#804 broke Kubernetes cluster e2e tests:
* [sig-cli] Kubectl Port forwarding [k8s.io] With a server listening on localhost [k8s.io] that expects NO client request should support a client that connects, sends DATA, and disconnects
* [sig-cli] Kubectl Port forwarding [k8s.io] With a server listening on localhost [k8s.io] that expects a client request should support a client that connects, sends DATA, and disconnects
* [sig-cli] Kubectl Port forwarding [k8s.io] With a server listening on localhost [k8s.io] that expects a client request should support a client that connects, sends NO DATA, and disconnects

The reason is that Kubernetes portforward supports forwarding port from pod network namespace, even when the application listens on localhost inside the network namespace. In this case, we can only use nsenter instead of pod ip.

@fvoznika We need find another way to solve the problem.

PR #812, opened 2018-06-11
Erase ambient capabilities.

Random-Liu:
Cherrypick https://github.com/containerd/cri/pull/808.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -372,6 +372,11 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 				securityContext.GetCapabilities())
 		}
 	}
+	// Clear all ambient capabilities. The implication of non-root + caps
+	// is not clearly defined in Kubernetes.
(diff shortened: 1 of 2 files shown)


pkg/server/container_create.go
```diff
@@ -372,6 +372,11 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 				securityContext.GetCapabilities())
 		}
 	}
+	// Clear all ambient capabilities. The implication of non-root + caps
+	// is not clearly defined in Kubernetes.
+	// See https://github.com/kubernetes/kubernetes/issues/56374
+	// Keep docker's behavior for now.
+	g.Spec().Process.Capabilities.Ambient = []string{}
```
mikebrow: If there are ambient caps, let's log that ambient caps are not supported for containers in Kubernetes pods, but that the container will be created with the ambient caps masked.

Random-Liu:
Will fix the presubmit test.

PR #816, opened 2018-06-15
Fix double /dev/shm mount.

Random-Liu:
We have double `/dev/shm` mount now, one from containerd default runtime spec, one from our own `/dev/shm` mount.
```
# crictl exec 0c mount | grep shm
shm on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)
shm on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)
```

This PR fixes it:
```
# crictl exec 9 mount | grep shm
shm on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)
```

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -777,6 +777,10 @@ func defaultRuntimeSpec(id string) (*runtimespec.Spec, error) {
 		if mount.Destination == "/run" {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


Random-Liu:
We should cherrypick this.

PR #818, opened 2018-06-15
Fix double /dev/shm mount.

Random-Liu:
Cherrypick #816.


Random-Liu:
The test failure seems to be a test-infra bug. It should not use Kubernetes 1.12 to test.

Random-Liu:
@mikebrow I'm so sorry.... I thought you've LGTMed.......

PR #820, opened 2018-06-18
Update github.com/opencontainers/runtime-tools to v0.6.0

filbranden:
Also add new dependencies on github.com/xeipuuv/gojson* (brought up by new runtime-tools) and adapt the containerd/cri code to replace the APIs that were removed by runtime-tools.

In particular, add new helpers to handle the capabilities, since runtime-tools now split them into separate sets of functions for each capability set.

cc @Random-Liu


mikebrow:
Need to run make test.. see build issue:
```
# github.com/containerd/cri/pkg/server
pkg/server/container_create_test.go:429:20: not enough arguments in call to generate.New
	have ()
	want (string)
pkg/server/container_create_test.go:429:20: multiple-value generate.New() in single-value context
pkg/server/container_create_test.go:623:20: not enough arguments in call to generate.New
	have ()
	want (string)
pkg/server/container_create_test.go:623:20: multiple-value generate.New() in single-value context
pkg/server/container_create_test.go:731:20: not enough arguments in call to generate.New
	have ()
	want (string)
pkg/server/container_create_test.go:731:20: multiple-value generate.New() in single-value context
FAIL	github.com/containerd/cri/pkg/server [build failed]
ok  	github.com/containerd/cri/pkg/server/io	1.083s
?   	github.com/containerd/cri/pkg/server/testing	[no test files]
?   	github.com/containerd/cri/pkg/store	[no test files]
ok  	github.com/containerd/cri/pkg/store/container	1.046s
ok  	github.com/containerd/cri/pkg/store/image	1.029s
ok  	github.com/containerd/cri/pkg/store/sandbox	1.023s
ok  	github.com/containerd/cri/pkg/store/snapshot	1.024s
ok  	github.com/containerd/cri/pkg/util	1.031s
make: *** [test] Error 2
```

Random-Liu:
@filbranden There is a compile error.

filbranden:
Indeed I had missed that... Fixed now, PTAL.

Random-Liu:
```

I0619 19:07:36.025] pkg/server/container_create.go:379:2:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
I0619 19:07:36.025] pkg/server/container_create.go:410:9:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
I0619 19:07:36.026] pkg/server/container_create.go:536:10:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
I0619 19:07:36.026] pkg/server/container_create.go:568:10:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
I0619 19:07:36.026] pkg/server/container_create.go:643:7:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
I0619 19:07:36.026] pkg/server/container_create.go:644:5:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
I0619 19:07:36.026] pkg/server/container_create.go:677:10:warning: g.Spec is deprecated: Replace with generator.Config.  (SA1019) (megacheck)
```

mikebrow:
We typically split out the vendor code into a separate commit.

pkg/server/container_create.go
```diff
@@ -597,7 +597,12 @@ func setOCIDevicesPrivileged(g *generate.Generator) error {
 // addOCIBindMounts adds bind mounts.
 func (c *criService) addOCIBindMounts(g *generate.Generator, mounts []*runtime.Mount, mountLabel string) error {
 	// Mount cgroup into the container as readonly, which inherits docker's behavior.
-	g.AddCgroupsMount("ro") // nolint: errcheck
+	g.AddMount(runtimespec.Mount{
+		Source:      "cgroup",
+		Destination: "/sys/fs/cgroup",
+		Type:        "cgroup",
+		Options:     []string{"nosuid", "noexec", "nodev", "relatime", "ro"},
```
mikebrow: pretty big delta here.. do you have a link to this being the new expected default behavior options for cgroup? 

Will need to modify the above comment to reflect the additional changes, eg. and with nosuid to block the operation of suid and sgid bits (filesystem cannot contain set userid files), and...

pkg/server/container_create.go
```diff
@@ -635,8 +640,8 @@ func (c *criService) addOCIBindMounts(g *generate.Generator, mounts []*runtime.M
 				return err
 			}
 			options = append(options, "rslave")
-			if g.Spec().Linux.RootfsPropagation != "rshared" &&
-				g.Spec().Linux.RootfsPropagation != "rslave" {
+			if g.Config.Linux.RootfsPropagation != "rshared" &&
```
mikebrow: can we add a comment here explaining and specifying where this change comes from..

mikebrow:
@Random-Liu will need some release notes for this one..

pkg/server/container_create.go
filbranden: Actually, I just copied what the previous implementation of `g.AddCgroupsMount("ro")` was actually doing... So this should be a no-op. I don't see how the comment is not relevant, since it's not inaccurate to say the cgroup fs is being mounted read-only...

pkg/server/container_create.go
filbranden: Sure, I'll update the commit description. It's because of the deprecation.

pkg/server/container_create.go
mikebrow: Just took a look at the old code sitting on my master .. it looks to me like it would have returned early with nil? Might have been a coding error..
```
// AddCgroupsMount adds a cgroup mount into g.spec.Mounts.
func (g *Generator) AddCgroupsMount(mountCgroupOption string) error {
	switch mountCgroupOption {
	case "ro":
	case "rw":
	case "no":
		return nil
	default:
		return fmt.Errorf("--mount-cgroups should be one of (ro,rw,no)")
	}

	mnt := rspec.Mount{
		Destination: "/sys/fs/cgroup",
		Type:        "cgroup",
		Source:      "cgroup",
		Options:     []string{"nosuid", "noexec", "nodev", "relatime", mountCgroupOption},
	}
	g.initSpec()
	g.spec.Mounts = append(g.spec.Mounts, mnt)

	return nil
}
```

filbranden: Go is a [weird](https://stackoverflow.com/questions/40821855/do-go-switch-cases-fallthrough-or-not) language... :disappointed:

mikebrow: :-) all good. So my read is that func was only able to check if the mount option was one of "ro,rw, or no" I presume no is for one of nosuid, nodev, noexec but we're not truncating so not sure what's up there. And for our case we just received a nil for the error. With the func skipping over the dunsel code by not having a break or continue but having a default. So we probably don't want to change up the options we were using unless there is a valid reason to change them.

mikebrow: yeah nvm..

pkg/server/container_create.go
mikebrow: nvm looks good

Random-Liu:
LGTM

PR #821, opened 2018-06-19
Fix snapshotter nil panic.

Random-Liu:
The specified snapshotter can be `nil`. When that happens, it is better to log a clear fatal error than panic for nil pointer access.

/cc @dmcgowan 

### DIFF
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -128,6 +128,10 @@ func NewCRIService(config criconfig.Config, client *containerd.Client) (CRIServi
 		selinux.SetDisabled()
 	}
 
+	if client.SnapshotService(c.config.ContainerdConfig.Snapshotter) == nil {
+		return nil, errors.Errorf("failed to find snapshotter %q", c.config.ContainerdConfig.Snapshotter)
+	}
+
 	c.imageFSPath = imageFSPath(config.ContainerdRootDir, config.ContainerdConfig.Snapshotter)
 	logrus.Infof("Get image filesystem path %q", c.imageFSPath)
 


pkg/server/service.go
```diff
@@ -184,9 +184,13 @@ func (c *criService) Run() error {
 
 	// Start snapshot stats syncer, it doesn't need to be stopped.
 	logrus.Info("Start snapshots syncer")
+	snapshotter := c.client.SnapshotService(c.config.ContainerdConfig.Snapshotter)
+	if snapshotter == nil {
+		return errors.Errorf("snapshotter %q does not exist", c.config.ContainerdConfig.Snapshotter)
```
mikebrow: c.eventMonitor.stop() ???

Random-Liu: I'll move this to earlier instead.

Random-Liu: Moved to `NewCRIService`.

Random-Liu:
I feel like this should be more a plugin initialization error. With the new change, it won't panic containerd if the configured snapshotter does not exist.
```
INFO[0000] loading plugin "io.containerd.grpc.v1.cri"...  type=io.containerd.grpc.v1
INFO[0000] Start cri plugin with config {PluginConfig:{ContainerdConfig:{Snapshotter:overlayfs DefaultRuntime:{Type:io.containerd.runtime.v1.linux Engine: Root:} UntrustedWorkloadRuntime:{Type: Engine: Root:}} CniConfig:{NetworkPluginBinDir:/opt/cni/bin NetworkPluginConfDir:/etc/cni/net.d NetworkPluginConfTemplate:} Registry:{Mirrors:map[docker.io:{Endpoints:[https://registry-1.docker.io]}]} StreamServerAddress: StreamServerPort:10010 EnableSelinux:false SandboxImage:k8s.gcr.io/pause:3.1 StatsCollectPeriod:10 SystemdCgroup:false EnableTLSStreaming:false MaxContainerLogLineSize:16384} ContainerdRootDir:/var/lib/containerd ContainerdEndpoint:/run/containerd/containerd.sock RootDir:/var/lib/containerd/io.containerd.grpc.v1.cri StateDir:/run/containerd/io.containerd.grpc.v1.cri} 
INFO[0000] Connect containerd service                   
WARN[0000] failed to load plugin io.containerd.grpc.v1.cri  error="failed to create CRI service: snapshotter "overlayfs" does not exist"
```

pkg/server/service.go
```diff
@@ -128,6 +128,10 @@ func NewCRIService(config criconfig.Config, client *containerd.Client) (CRIServi
 		selinux.SetDisabled()
 	}
 
+	if c.client.SnapshotService(c.config.ContainerdConfig.Snapshotter) == nil {
+		return nil, errors.Errorf("snapshotter %q does not exist", c.config.ContainerdConfig.Snapshotter)
```
mikebrow: can we normalize the errors in this func? maybe defer a wrap for  "Failed to initialize cni"? Because some say cni failed some say what failed..

Random-Liu:
Let's cherrypick this to avoid panic.

PR #822, opened 2018-06-19
Change to keep in sync with latest cni config

abhi:
This commit contains change to pick the latest cni config from the configured CNIConfDir.
With this change any changes made to the cni config file will be picked up on the kubelet's runtime status check call.
Ofcourse this would lead to undefined behavior when the cni config change is made in parallel during pod creation. However its reasonable to assume that the operator is aware of the need to drain the nodes of pods before making cni configuration change. The behavior is currently not defined in kubernetes. However I see that similar approach being adopted in the upstream kubernetes with dockershim. Keeping the behavior consistent for now.

### DIFF
--- a/pkg/server/status.go
+++ b/pkg/server/status.go
@@ -22,6 +22,7 @@ import (
 	goruntime "runtime"
 
 	cni "github.com/containerd/go-cni"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 )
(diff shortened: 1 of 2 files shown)


pkg/server/status.go
```diff
@@ -41,14 +42,16 @@ func (c *criService) Status(ctx context.Context, r *runtime.StatusRequest) (*run
 		Type:   runtime.NetworkReady,
 		Status: true,
 	}
+
+	// Load the latest cni configuration to be in sync with the latest network configuration
+	if err := c.netPlugin.Load(cni.WithLoNetwork(), cni.WithDefaultConf()); err != nil {
```
mikebrow: seems like an expensive operation if there has been no change..?

mikebrow: any of the moby repos using a fsnotify package?

pkg/server/status.go
```diff
@@ -41,14 +42,16 @@ func (c *criService) Status(ctx context.Context, r *runtime.StatusRequest) (*run
...
+		logrus.WithError(err).Errorf("Failed to load cni configuration")
+	}
```
mikebrow: might want to use an else if in this case as we don't care much about status() being successful for a prior load if we can't load properly with the current config.

If so we can cp the networkCondition assignments below to the if case and modify the else case to indicate which failure case we are in.. load or get status.

pkg/server/status.go
abhi: yes it reads the config file on every status check. I can add some intelligence in the Load on the go-cni side to only load if the checksum/date has changed. Keeping this simple and similar to kube upstream for now.

mikebrow: yeah simple and low memory/fast

Random-Liu:
This breaks the test.

Random-Liu:
@abhi Why you have a branch in upstream?

abhi:
@Random-Liu my bad. Will close the PR and reopen

PR #849, opened 2018-07-12
Replace stringid call with simple random reader

dmcgowan:
Prevents importing of `docker/docker/pkg/stringid` which sets a seed in the init function. We do not want this seed being set and especially doing a blocking read for crypto on startup. Note the ID function used was a very simple math rand read of 32 bytes followed by a hex encode. There was additionally a check to ensure the first 12 characters were [not all numeric](https://github.com/moby/moby/issues/3869), this was because Docker seemed to have a use case where this function was used to generate a hostname. If you have that same use case I can add that check back in, but seems very odd to use here.

See https://github.com/containerd/containerd/issues/2451


Random-Liu:
@dmcgowan The containerd and ctr binary in this repo is just for testing.

We need a `rand.Seed` in containerd.

dmcgowan:
https://github.com/containerd/containerd/pull/2461

PR #850, opened 2018-07-13
Update crictl to v1.11.1.

Random-Liu:
crictl v1.11.1 has some useful bug fix.

See https://github.com/kubernetes-incubator/cri-tools/releases/tag/v1.11.1

We should cherrypick this to the release/1.11 branch.


Random-Liu:
There is a bug in v1.11.1, although the shipped binaries are fixed, the code is still problematic.
We are going to cut v1.11.2 soon. Let's wait for that.

Issue #855, opened 2018-07-20
untrusted runtime should support privileged


egernst:
Being able to run an untrusted runtime, such as kata-runtime, as privileged should be allowed.  In this case, all device nodes, etc, would be made available to it.  Note, none of the host available features would be available.

If a workload explicitly is marked as untrusted but has --privileged, allow the configuration.

I still think we should return an error if they are asking for namespaces of the host (the node).

This should address the TODO @ [1] in the short term while we wait for runtimeClass to be implemented.

[1]  https://github.com/containerd/cri/blob/master/pkg/server/sandbox_run.go#L610

egernst:
I'll put together a PR accordingly.  Want to follow similar logic as to what is introduced with https://github.com/kubernetes-incubator/cri-o/pull/1701

Random-Liu:
@egernst Current integration is experimental.

@tallclair is working on the RuntimeClass. https://github.com/kubernetes/features/issues/585 And that will define what features will be supported by a runtime, which includes `privileged`. I prefer we make the behavior consistent with that.

@tallclair Do you think we should support privileged for sandbox container?

egernst:
Sounds good @Random-Liu.  I expect this to become more clear once runtime class is supported.  In the meantime there are many use cases where having privileged support in the sandboxed runtime (ie, kata in my scenario) makes sense (device node access, update the guest kernel, etc).  

I didn't want to wait until 1.12 to start using these patterns.

PR #857, opened 2018-07-20
sandbox: separate host accessing workload and privileged

egernst:
VM isolated runtimes can support privileged workloads. In this
scenario, access to the guest VM is provided instead of the host.
Based on this, allow untrusted runtimes to run privileged workloads.

If the workload is specifically asking for node PID/IPC/network, etc.,
then continue to require the trusted runtime.

This commit repurposes the hostPrivilegedSandbox utility function to
only check for node namespace checking.

Fixes: #855

Signed-off-by: Eric Ernst <eric.ernst@intel.com>

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -583,13 +583,10 @@ func untrustedWorkload(config *runtime.PodSandboxConfig) bool {
 	return config.GetAnnotations()[annotations.UntrustedWorkload] == "true"
 }
 
-// hostPrivilegedSandbox returns true if the sandbox configuration
-// requires additional host privileges for the sandbox.
-func hostPrivilegedSandbox(config *runtime.PodSandboxConfig) bool {
+// hostAccessingSandbox returns true if the sandbox configuration
+// requires additional host access for the sandbox.
+func hostAccessingSandbox(config *runtime.PodSandboxConfig) bool {
 	securityContext := config.GetLinux().GetSecurityContext()
-	if securityContext.GetPrivileged() {
-		return true
-	}
 
 	namespaceOptions := securityContext.GetNamespaceOptions()
 	if namespaceOptions.GetNetwork() == runtime.NamespaceMode_NODE ||
@@ -607,9 +604,13 @@ func hostPrivilegedSandbox(config *runtime.PodSandboxConfig) bool {
 func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig) (criconfig.Runtime, error) {
 	untrusted := false
 	if untrustedWorkload(config) {
-		// TODO(random-liu): Figure out we should return error or not.
-		if hostPrivilegedSandbox(config) {
-			return criconfig.Runtime{}, errors.New("untrusted workload with host privilege is not allowed")
+		//  If the untrusted workload is requesting access to the host/node, this request will fail.
+		//
+		//  Note: If the workload is marked untrusted but requests privileged, this can be granted, as the
+		// runtime may support this.  For example, in a virtual-machine isolated runtime, privileged
+		// is a supported option, granting the workload to access the entire guest VM instead of host.
+		if hostAccessingSandbox(config) {
+			return criconfig.Runtime{}, errors.New("untrusted workload with host access is not allowed")
 		}
 		untrusted = true
 	}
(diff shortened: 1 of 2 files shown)


egernst:
Hate to plead ignorance, but I don't see/understand the travis failure.

Random-Liu:
```
pkg/server/sandbox_run_test.go:516:14: undefined: hostPrivilegedSandbox
```

pkg/server/sandbox_run.go
```diff
@@ -583,13 +583,10 @@ func untrustedWorkload(config *runtime.PodSandboxConfig) bool {
 	return config.GetAnnotations()[annotations.UntrustedWorkload] == "true"
 }
 
-// hostPrivilegedSandbox returns true if the sandbox configuration
+// hostAccessingandbox returns true if the sandbox configuration
```
Random-Liu: hostAccessingSandbox

pkg/server/sandbox_run.go
```diff
@@ -583,13 +583,10 @@ func untrustedWorkload(config *runtime.PodSandboxConfig) bool {
...
 // requires additional host privileges for the sandbox.
```
Random-Liu: s/privileges/access

pkg/server/sandbox_run.go
```diff
@@ -607,8 +604,13 @@ func hostPrivilegedSandbox(config *runtime.PodSandboxConfig) bool {
 func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig) (criconfig.Runtime, error) {
 	untrusted := false
 	if untrustedWorkload(config) {
-		// TODO(random-liu): Figure out we should return error or not.
-		if hostPrivilegedSandbox(config) {
+		//  If the workload is requesting access to the host/node, this request will fail when using
+		// an untrusted runtime. Return error in this case.
```
Random-Liu: The cri-o trusted/untrusted runtime is confusing.

`If the untrusted workload is requesting access`, and remove `untrusted runtime`.

pkg/server/sandbox_run.go
```diff
@@ -607,8 +604,13 @@ func hostPrivilegedSandbox(config *runtime.PodSandboxConfig) bool {
...
+		//
+		//  Note: If the workload is marked untrusted but requests privileged, this can be granted, as the
```
Random-Liu: as the runtime may support this.

pkg/server/sandbox_run.go
```diff
@@ -607,8 +604,13 @@ func hostPrivilegedSandbox(config *runtime.PodSandboxConfig) bool {
...
+		// untrusted runtime may support this.  For example, in a virtual-machine isolated runtime, privileged
+		// is a supported option, granting the workload to access the entire guest VM instead of host.
+		if hostAccessingSandbox(config) {
 			return criconfig.Runtime{}, errors.New("untrusted workload with host privilege is not allowed")
```
Random-Liu: s/host privilege/host access

Random-Liu:
/ok-to-test

And let's cherry-pick to 1.11, so that we'll have this in containerd 1.2

pkg/server/sandbox_run.go
egernst: ack

pkg/server/sandbox_run.go
egernst: ack

pkg/server/sandbox_run.go
egernst: Agreed.  I better understand the containerd approach now, and prefer this explicit behavior.

pkg/server/sandbox_run.go
egernst: ack

pkg/server/sandbox_run.go
egernst: ack

pkg/server/sandbox_run_test.go
```diff
@@ -474,21 +474,14 @@ func TestTypeurlMarshalUnmarshalSandboxMeta(t *testing.T) {
 	}
 }
 
-func TestHostPrivilegedSandbox(t *testing.T) {
+func TestHostAccessingSandbox(t *testing.T) {
```
Random-Liu: Let's keep both privleged and nonPrivleged, and both of them should return true. This proves that the change takes effect.

egernst:
Updated.  Thanks @Random-Liu

Random-Liu:
@egernst Can you send a cherrypick PR to `release/1.11` branch? So that in the containerd 1.2 version we'll have this.

PR #858, opened 2018-07-20
Serve streaming on localhost by default to match k8s 1.11 default.

Random-Liu:
Ref https://github.com/containerd/cri/issues/777.

Change the default config to match Kubernetes 1.11 default behavior.

Initially, I thought we should keep default unchanged so that the default config can work with both K8s 1.10 and 1.11.

However, I talked with @dmcgowan today, and we found users don't like arbitrary fixed port listening. Given that we don't need a fixed port anymore for K8s 1.11, I think we should change the default to match the k8s 1.11 behavior to get rid of this fixed port.

We need to mention in the release note that, if you want to use containerd 1.2 with K8s 1.10, please set `stream_server_address=""` and `stream_server_port` to a fixed port if you need a fixed one.


Random-Liu:
> Agree arbitrary not as good as specified port.. more secure.

Just to clarify, with this PR, we are going to use arbitrary (non-fixed) port on localhost. Because it is localhost, there shouldn't be security problem ideally.

mikebrow:
Sry was typing fast. I meant Agree with arbitrary point, but specified is more secure.  Release note and maybe a note in the config.md will suffice.

mikebrow:
Which reminds me.. can we update the comment for this up on line 105- and the config.md?

Random-Liu:
@mikebrow Make sense... Will do! Thanks for reminding!

Random-Liu:
@mikebrow Done.

PR #864, opened 2018-07-26
Unpack image during import.

Random-Liu:
We should unpack image during import, so that the lease can help us hold the resource.

Similar issue with https://github.com/containerd/containerd/pull/2331.

### DIFF
--- a/hack/verify-lint.sh
+++ b/hack/verify-lint.sh
@@ -29,6 +29,7 @@ for d in $(find . -type d -a \( -iwholename './pkg*' -o -iwholename './cmd*' \) 
 		 --disable=aligncheck \
 		 --disable=gotype \
 		 --disable=gas \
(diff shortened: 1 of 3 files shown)


Random-Liu:
This PR also disables `gsec`, because `gas` is renamed to `gsec`. https://github.com/alecthomas/gometalinter/pull/505

hack/verify-lint.sh
```diff
@@ -29,6 +29,7 @@ for d in $(find . -type d -a \( -iwholename './pkg*' -o -iwholename './cmd*' \)
 		 --disable=aligncheck \
 		 --disable=gotype \
 		 --disable=gas \
+		 --disable=gosec \
```
mikebrow: ah that's why ci broke :-)

PR #865, opened 2018-07-26
Cache image reference

Random-Liu:
Fixes https://github.com/containerd/cri/issues/760.
Based on https://github.com/containerd/cri/pull/864.

This PR:
* Creates an in memory cache of containerd image references.
  * The cache is updated when:
    a) `cri` plugin creates or removes an image reference. This makes sure image reference cache is up-to-date after `cri` pulling or deleting an image.
    b) Event monitor receives an image event (create/update/delete). This makes sure image references created/deleted outside of `cri` can trigger a cache update, e.g. `ctr --namespace=k8s.io images pull`.
  * All other `cri` functions, which only read data image information, will treat the cache as the source of truth for images, so that `cri` doesn't need to frequently access containerd metadata/content store and easily index by image id.
  * Please note that all data in the cache is generated from containerd. a) and b) above only triggers a cache update. This makes sure there is no multi-writer race condition.
* Merge `RepoTags` and `RepoDigests` into a `References` field in the image store. For containerd, there is only image reference, only CRI has the `RepoTag` and `RepoDigest` concepts.
  * The `References` field reflects the references in containerd regardless of the reference format.
  * When `cri` generates CRI image status, it goes through all references, and pick out `RepoTags` and `RepoDigests`.
  * When `cri` pulls images, we'll still create repodigest and repotag as the image references, because of https://github.com/containerd/cri/issues/760#issuecomment-386752652. However, when users use `ctr --namespace=k8s.io images pull` to pull images, they can do whatever they want. They just need to understand that if their image doesn't have a reference in `repotag` or `repodigest` format, the corresponding CRI fields `RepoTags` and `RepoDigests` will be empty.

With this PR:
1) `cri` will be able to see images pulled/loaded with `ctr`, and we can remove `ctr cri` command once we have `ctr import` ready.
2) `repotag` which has pointed to a new image, won't show up in  `RepoTags` field of the old image. We have this bug before this PR.

/cc @crosbymichael @dmcgowan @stevvooe

### DIFF
--- a/pkg/server/events.go
+++ b/pkg/server/events.go
@@ -34,6 +34,7 @@ import (
 	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
 	"github.com/containerd/cri/pkg/store"
 	containerstore "github.com/containerd/cri/pkg/store/container"
+	imagestore "github.com/containerd/cri/pkg/store/image"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 )
 
@@ -49,6 +50,7 @@ const (
 type eventMonitor struct {
 	containerStore *containerstore.Store
 	sandboxStore   *sandboxstore.Store
+	imageStore     *imagestore.Store
 	ch             <-chan *events.Envelope
 	errCh          <-chan error
 	ctx            context.Context
@@ -76,12 +78,13 @@ type backOffQueue struct {
 
 // Create new event monitor. New event monitor will start subscribing containerd event. All events
 // happen after it should be monitored.
-func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store) *eventMonitor {
+func newEventMonitor(c *containerstore.Store, s *sandboxstore.Store, i *imagestore.Store) *eventMonitor {
 	// event subscribe doesn't need namespace.
 	ctx, cancel := context.WithCancel(context.Background())
 	return &eventMonitor{
 		containerStore: c,
 		sandboxStore:   s,
+		imageStore:     i,
 		ctx:            ctx,
 		cancel:         cancel,
 		backOff:        newBackOff(),
@@ -93,12 +96,13 @@ func (em *eventMonitor) subscribe(subscriber events.Subscriber) {
 	filters := []string{
 		`topic=="/tasks/exit"`,
 		`topic=="/tasks/oom"`,
+		`topic~="/images/"`,
 	}
 	em.ch, em.errCh = subscriber.Subscribe(em.ctx, filters...)
 }
 
 func convertEvent(e *gogotypes.Any) (string, interface{}, error) {
-	containerID := ""
+	id := ""
 	evt, err := typeurl.UnmarshalAny(e)
 	if err != nil {
 		return "", nil, errors.Wrap(err, "failed to unmarshalany")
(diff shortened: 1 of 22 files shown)


pkg/server/helpers.go
```diff
@@ -234,9 +229,9 @@ func getRepoDigestAndTag(namedRef reference.Named, digest imagedigest.Digest, sc
 	return repoDigest, repoTag
 }
 
-// localResolve resolves image reference locally and returns corresponding image metadata. It returns
-// nil without error if the reference doesn't exist.
-func (c *criService) localResolve(ctx context.Context, refOrID string) (*imagestore.Image, error) {
+// localResolve resolves image reference locally and returns corresponding image metadata. It
+// store.ErrNotExist if the reference doesn't exist.
```
mikebrow: /s/./or docker/distribution/digestset.ErrDigestInvalidFormat, or ErrDigestAmbiguous./

pkg/server/helpers.go
```diff
@@ -234,9 +229,9 @@ func getRepoDigestAndTag(namedRef reference.Named, digest imagedigest.Digest, sc
 	return repoDigest, repoTag
 }
 
-// localResolve resolves image reference locally and returns corresponding image metadata. It returns
-// nil without error if the reference doesn't exist.
-func (c *criService) localResolve(ctx context.Context, refOrID string) (*imagestore.Image, error) {
+// localResolve resolves image reference locally and returns corresponding image metadata. It
```
mikebrow: /s/It/It returns/

pkg/server/helpers.go
mikebrow: Add a todo, we should not be eating the err returned from util.NormalizeImageRef. We are currently assuming it must be a digest if it's not a valid image ref.. but it could just be a poorly formed image ref and not a digest. But right now if it's a poorly formed image ref we return the error from digest lookup.  Let's keep "both" errors and combine them here, so the user can see it's not a valid digest or image reference.

pkg/server/events.go
```diff
@@ -93,29 +96,36 @@ func (em *eventMonitor) subscribe(subscriber events.Subscriber) {
...
+		id = evt.(*eventtypes.ImageDelete).Name
 	default:
 		return "", nil, errors.New("unsupported event")
 	}
-	return containerID, evt, nil
+	return id, evt, nil
 }
 
-// start starts the event monitor which monitors and handles all container events. It returns
+// start starts the event monitor which monitors and handles all important events. It returns
```
mikebrow: /s/important/subscribed/

pkg/server/events.go
```diff
@@ -230,6 +240,18 @@ func (em *eventMonitor) handleEvent(any interface{}) error {
 		if err != nil {
 			return errors.Wrap(err, "failed to update container status for TaskOOM event")
 		}
+	case *eventtypes.ImageCreate:
```
mikebrow: nice!

dmcgowan:
This approach looks good

pkg/server/helpers.go
Random-Liu: Done

pkg/store/image/image.go
```diff
@@ -80,44 +212,29 @@ func (s *Store) Add(img Image) error {
...
+	// Or else, merge the references.
+	i.References = util.MergeStringSlices(i.References, img.References)
 	s.images[img.ID] = i
 	return nil
 }
 
-// Get returns the image with specified id. Returns store.ErrNotExist if the
-// image doesn't exist.
-func (s *Store) Get(id string) (Image, error) {
+func (s *store) get(id string) (Image, error) {
```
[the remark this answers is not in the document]
Random-Liu: Done.

Random-Liu:
@mikebrow Addressed comments.

Random-Liu:
Something with travis. All test passing, but the bullet doesn't turn green.

PR #866, opened 2018-07-26
sandbox: separate host accessing workload and privileged

Random-Liu:
Cherrypick #857 to catch the containerd 1.2 release train, or else we need to wait till end of September.


mikebrow:
just have to pick up the ci fix first then merge.. fun..

Random-Liu:
/test pull-cri-containerd-verify
--

PR #884, opened 2018-08-24
[WIP] Add timeout for container/sandbox recover.

Random-Liu:
Although we don't want, containerd-shim sometimes hangs, e.g. containerd/containerd#2438. We should definitely root cause and fix it.

However, on the other hand, we should make sure that the system still functional when some containerd-shims hang. An example is that `ctr task ls` will hang if one single containerd-shim hangs. --> We should probably fix it. @crosbymichael 

Luckily, the CRI plugin in most cases doesn't handle multiple containers at a time, this makes sure that a single container failure won't block other containers.

However, the only case that CRI plugin may handle multiple containers at a time is the recovery logic. This means that if a containerd-shim hangs, CRI plugin won't be able to restart. This is super bad.

This PR adds a timeout to per container/sandbox recovery logic, so that even one container hangs, we just don't load it, and continue dealing with other containers.

### DIFF
--- a/pkg/server/restart.go
+++ b/pkg/server/restart.go
@@ -146,8 +146,22 @@ func (c *criService) recover(ctx context.Context) error {
 	return nil
 }
 
+// loadContainerTimeout is the default timeout for loading a container/sandbox.
+// One container/sandbox hangs (e.g. containerd#2438) should not affect other
+// containers/sandboxes.
+// Most CRI container/sandbox related operations are per container, the ones
+// which handle multiple containers at a time are:
+// * ListPodSandboxes: Don't talk with containerd services.
+// * ListContainers: Don't talk with containerd services.
+// * ListContainerStats: Not in critical code path, a default timeout will
+// be applied at CRI level.
+// * Recovery logic: We should set a time for each container/sandbox recovery.
+const loadContainerTimeout = 10 * time.Second
+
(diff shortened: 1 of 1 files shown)


Random-Liu:
We should cherry-pick this into all supported branches.

yujuhong:
LGTM

One question: if the load request timed out, how is the container going to be represented in the CRI plugin -- does CRI show the existence of the container and what state will it be in?

Random-Liu:
@yujuhong Currently, won't show the container, but I think it would be better to show the container in Unknown state. Let me see whether showing a bad container will cause any issues.

BTW, how does Kubelet deal with unknown state? I remember it triggers a sync pod, and kubelet will try to start a new one.

yujuhong:
Yes, kubelet will try to kill the container in the unknown state and start a new one.

Random-Liu:
I found this PR doesn't actually help now, because:
1) There is no grpc between CRI plugin and containerd, so no one handles the timeout;
2) The context will be eventually passed to containerd-shim, and ideally the rpc client will handle the timeout. However, ttrpc doesn't support timeout yet https://github.com/containerd/ttrpc/issues/3.

Given so, I'll leave this PR here now... We need a better timeout solution to make the system more reliable. I'll file an issue.

PR #885, opened 2018-08-24
Fix an issue that container/sandbox can't be stopped.

Random-Liu:
Today, I found that some containers on my workstation can't be stopped with `crictl stop`.

It turns out that the container processes are moved to `/user.slice` from the original container cgroups. I guess it is because I simply run `./containerd` to test containerd, without setting `Delegate=true` on my systemd node.
```console
$ /usr/local/bin/ctr -n=k8s.io task ls | grep 206190
6f757ff0e1c2b544da02c3ab17488be73b843dcc780847ae32ca4ae756be936f    206190    RUNNING
$ cat /sys/fs/cgroup/cpu,cpuacct/user.slice/cgroup.procs | grep 206190
206190
$ cat /sys/fs/cgroup/cpu,cpuacct/kubepods/burstable/podcedaeae7-a4e4-11e8-a6f6-8cdcd43ac064/6f757ff0e1c2b544da02c3ab17488be73b843dcc780847ae32ca4ae756be936f/cgroup.procs
```

Although what I did was not best practice, I still think we should handle this case, not being able to stop container is very bad.

### DIFF
--- a/pkg/server/container_stop.go
+++ b/pkg/server/container_stop.go
@@ -33,7 +33,10 @@ import (
 
 // killContainerTimeout is the timeout that we wait for the container to
 // be SIGKILLed.
-const killContainerTimeout = 2 * time.Minute
+// The timeout is set to 1 min, because the default CRI operation timeout
+// for StopContainer is (2 min + stop timeout). Set to 1 min, so that we
+// have enough time for kill(all=true) and kill(all=false).
+const killContainerTimeout = 1 * time.Minute
 
 // StopContainer stops a running container with a grace period (i.e., timeout).
 func (c *criService) StopContainer(ctx context.Context, r *runtime.StopContainerRequest) (*runtime.StopContainerResponse, error) {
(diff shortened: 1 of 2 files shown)


Random-Liu:
Should cherrypick into all supported branches.

pkg/server/container_stop.go
```diff
@@ -81,52 +94,47 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
...
-	// Event handler will Delete the container from containerd after it handles the Exited event.
-	logrus.Infof("Kill container %q", id)
-	if task != nil {
-		if err = task.Kill(ctx, unix.SIGKILL, containerd.WithKillAll); err != nil {
-			if !errdefs.IsNotFound(err) {
-				return errors.Wrapf(err, "failed to kill container %q", id)
-			}
-			// Move on to make sure container status is updated.
-		}
+	logrus.WithError(err).Errorf("Kill container %q timed out", id)
```
mikebrow: waitContainerStop() has other error types.. should probably use "an error occurs during waiting for container %q to stop"

Random-Liu: Make sense.

Random-Liu: Done

Random-Liu:
@mikebrow Addressed comments. Please take another look. :)

PR #886, opened 2018-08-27
stream: can use user certificates

JulienBalestra:
This PR allows to configure containerd certificates used to stream.

This is needed if we want to use a safe `https` setup for `kubectl exec`.

```text
time="2018-08-27T18:30:06.366428577+02:00" level=info msg="Exec for "23db729b6d704da42d366ead46278dae498594ea4043ded7bb9af401bc589c50" with command [/bin/sh], tty true and stdin true"
time="2018-08-27T18:30:06.366497645+02:00" level=info msg="Exec for "23db729b6d704da42d366ead46278dae498594ea4043ded7bb9af401bc589c50" returns URL "https://192.168.128.173:10010/exec/_dPWQE1j""
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -28,7 +28,7 @@ The explanation and default value of each configuration item are as follows:
   # It generates a self-sign certificate unless the following x509_key_pair_streaming are both set.
   enable_tls_streaming = false
   
-  # "plugins.cri.x509_key_pair_streaming" constains a x509 valid key pair to stream with tls.
+  # "plugins.cri.x509_key_pair_streaming" contains a x509 valid key pair to stream with tls.
   [plugins.cri.x509_key_pair_streaming]
     # tls_cert_file is the filepath to the certificate paired with the "tls_key_file"
     tls_cert_file = ""
--- a/docs/config.md
+++ b/docs/config.md
@@ -24,8 +24,16 @@ The explanation and default value of each configuration item are as follows:
   # systemd_cgroup enables systemd cgroup support.
   systemd_cgroup = false
 
-  # enable_tls_streaming enables the TLS streaming support.
+  # enable_tls_streaming enables the TLS streaming support. 
+  # It generates a self-sign certificate unless the following x509_key_pair_streaming are both set.
   enable_tls_streaming = false
+  
+  # "plugins.cri.x509_key_pair_streaming" constains a x509 valid key pair to stream with tls.
+  [plugins.cri.x509_key_pair_streaming]
+    # tls_cert_file is the filepath to the certificate paired with the "tls_key_file"
+    tls_cert_file = ""
+    # tls_key_file is the filepath to the private key paired with the "tls_cert_file"
+    tls_key_file = ""
 
   # max_container_log_line_size is the maximum log line size in bytes for a container.
   # Log line longer than the limit will be split into multiple lines. -1 means no
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -114,14 +114,20 @@ type PluginConfig struct {
 	SystemdCgroup bool `toml:"systemd_cgroup" json:"systemdCgroup"`
 	// EnableTLSStreaming indicates to enable the TLS streaming support.
 	EnableTLSStreaming bool `toml:"enable_tls_streaming" json:"enableTLSStreaming"`
-	// TLSCertFileStreaming is the path to a certificate file
-	TLSCertFileStreaming string `toml:"tls_cert_file_streaming" json:"tlsCertFileStreaming"`
-	// TLSKeyFileStreaming is the path to a private key file
-	TLSKeyFileStreaming string `toml:"tls_key_file_streaming" json:"tlsKeyFileStreaming"`
+	// X509KeyPairStreaming is a x509 key pair used for TLS streaming
+	X509KeyPairStreaming `toml:"x509_key_pair_streaming" json:"x509KeyPairStreaming"`
 	// MaxContainerLogLineSize is the maximum log line size in bytes for a container.
 	// Log line longer than the limit will be split into multiple lines. Non-positive
 	// value means no limit.
 	MaxContainerLogLineSize int `toml:"max_container_log_line_size" json:"maxContainerLogSize"`
+}
+
+// X509KeyPairStreaming contains the x509 configuration for streaming
+type X509KeyPairStreaming struct {
+	// TLSCertFile is the path to a certificate file
+	TLSCertFile string `toml:"tls_cert_file" json:"tlsCertFile"`
+	// TLSKeyFile is the path to a private key file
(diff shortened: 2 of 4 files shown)


pkg/server/streaming.go
```diff
@@ -44,18 +44,30 @@ func newStreamServer(c *criService, addr, port string) (streaming.Server, error)
 	}
 	config := streaming.DefaultConfig
 	config.Addr = net.JoinHostPort(addr, port)
-	runtime := newStreamRuntime(c)
-	if c.config.EnableTLSStreaming {
-		tlsCert, err := newTLSCert()
+	run := newStreamRuntime(c)
```
JulienBalestra: I renamed `runtime` as `run` because it's a package name.

mikebrow:
Have to sign the commits pls.

pkg/config/config.go
```diff
@@ -114,6 +114,10 @@ type PluginConfig struct {
 	SystemdCgroup bool `toml:"systemd_cgroup" json:"systemdCgroup"`
 	// EnableTLSStreaming indicates to enable the TLS streaming support.
 	EnableTLSStreaming bool `toml:"enable_tls_streaming" json:"enableTLSStreaming"`
+	// TLSCertFileStreaming is the path to a certificate file
```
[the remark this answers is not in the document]
JulienBalestra: sure @mikebrow I added a struct, let me know if you prefer like that

pkg/server/streaming.go
```diff
@@ -44,18 +44,30 @@ func newStreamServer(c *criService, addr, port string) (streaming.Server, error)
...
-			return nil, errors.Wrap(err, "failed to generate tls certificate for stream server")
+			return nil, errors.Wrap(err, "failed to load x509 key pair for stream server")
 		}
 		config.TLSConfig = &tls.Config{
-			Certificates:       []tls.Certificate{tlsCert},
-			InsecureSkipVerify: true,
+			Certificates: []tls.Certificate{tlsCert},
 		}
+		return streaming.NewServer(config, run)
+	}
```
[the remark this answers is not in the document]
JulienBalestra: I can't use `errors.Wrap` because `err == nil`. I used `errors.Errorf` instead.

mikebrow:
@JulienBalestra looks good just need to reduce the length of the commit message to 90 characters :-)

mikebrow:
yeah the range is set by travis to the pull request.. don't know what broke there..

pkg/server/streaming.go
```diff
@@ -44,18 +44,35 @@ func newStreamServer(c *criService, addr, port string) (streaming.Server, error)
 	}
 	config := streaming.DefaultConfig
 	config.Addr = net.JoinHostPort(addr, port)
-	runtime := newStreamRuntime(c)
-	if c.config.EnableTLSStreaming {
-		tlsCert, err := newTLSCert()
+	run := newStreamRuntime(c)
+	if !c.config.EnableTLSStreaming {
+		if c.config.X509KeyPairStreaming.TLSCertFile != "" || c.config.X509KeyPairStreaming.TLSKeyFile != "" {
+			return nil, errors.Errorf("X509KeyPairStreaming.TLSCertFile and/or X509KeyPairStreaming.TLSKeyFile are set but EnableTLSStreaming is not set")
```
Random-Liu: nit: Should use `New` when there is no arguments.

pkg/server/streaming.go
```diff
@@ -44,18 +44,35 @@ func newStreamServer(c *criService, addr, port string) (streaming.Server, error)
...
+		return nil, errors.Errorf("must set both X509KeyPairStreaming.TLSCertFile and X509KeyPairStreaming.TLSKeyFile")
+	}
+	// generating self-sign certs
+	tlsCert, err := newTLSCert()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate tls certificate for stream server")
+	}
+	config.TLSConfig = &tls.Config{
+		Certificates:       []tls.Certificate{tlsCert},
+		InsecureSkipVerify: true,
```
Random-Liu: @mikebrow What does the `InsecureSkipVerify` mean here on the server side? Skip verifying client certificate?

Why don't we need it when X509KeyPairStreaming is set?

pkg/server/streaming.go
```diff
@@ -44,18 +44,35 @@ func newStreamServer(c *criService, addr, port string) (streaming.Server, error)
 	}
 	config := streaming.DefaultConfig
 	config.Addr = net.JoinHostPort(addr, port)
-	runtime := newStreamRuntime(c)
-	if c.config.EnableTLSStreaming {
-		tlsCert, err := newTLSCert()
+	run := newStreamRuntime(c)
+	if !c.config.EnableTLSStreaming {
```
Random-Liu: Can we move all the validation logic into a small function? A bit messy to split around this function.

JulienBalestra: Sure !
I choose to improve this by returning the type of listener instead of checking it again in `newStreamServer`.
I improved the error handling in this validation function and covered it by tests.

docs/config.md
```diff
@@ -24,8 +24,16 @@ The explanation and default value of each configuration item are as follows:
   # systemd_cgroup enables systemd cgroup support.
   systemd_cgroup = false
 
-  # enable_tls_streaming enables the TLS streaming support.
+  # enable_tls_streaming enables the TLS streaming support. 
+  # It generates a self-sign certificate unless the following x509_key_pair_streaming are both set.
   enable_tls_streaming = false
+  
+  # "plugins.cri.x509_key_pair_streaming" constains a x509 valid key pair to stream with tls.
```
lbernail: small typo: "constains"

pkg/server/streaming.go
Random-Liu: @JulienBalestra I guess in your case, apiserver also has a valid certificate? That is why you don't need `InsecureSkipVerify` on the containerd side?

Random-Liu:
@JulienBalestra LGTM with one last question.

Please not that:
* If you are using Kubernetes 1.10 and lower, you can continue doing what you are doing.
* **If you are using Kubernetes 1.11 or higher**, you can only enable the stream server TLS in containerd when you run kubelet with `--redirect-container-streaming`.

Actually for Kubernetes 1.11 or higher, the recommended way of making the stream connection secure is to:
1) Run kubelet **without** `--redirect-container-streaming`, so that kubelet behave as a proxy between containerd and apiserver;
2) Let containerd serves streaming on `localhost:0` without TLS.

In this way, you only need to make sure the connection between kubelet and apiserver is secure.

pkg/server/streaming.go
JulienBalestra: @Random-Liu I'm not sure we need to specify this configuration for a server listener.
I thought this option is for clients only but maybe I'm wrong.

Random-Liu: @JulienBalestra Yeah, that's why I'm a bit confused.

PR #891, opened 2018-09-04
Add RuntimeHandler support

tallclair:
Implement the changes proposed in https://github.com/containerd/cri/issues/879:

1. Deprecate `plugins.cri.containerd.untrusted_workload_runtime`. If it is provided, give it the implicit configuration name of `untrusted`.
2. Introduce a new `ContainerdConfig` field: `Runtimes map[string]Runtime`
3. Modify `criService.getSandboxRuntime` to take the runtime handler as a parameter, and map it to the correct runtime config

Integration tests to follow

/cc @Random-Liu


Random-Liu:
@tallclair Please use `git commit --amend -s` to sign your commits.

tallclair:
Done.

cri.go
```diff
@@ -104,6 +108,18 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	return s, nil
 }
 
+// validateConfig validates that the given configuration.
```
Random-Liu: nit: remove `that`?

pkg/config/config.go
```diff
@@ -183,3 +188,8 @@ func DefaultConfig() PluginConfig {
 		},
 	}
 }
+
+const (
+	// RuntimeUntrusted is the implicit runtime defined for ContainerdConfig.UntrustedWorkloadRuntime
+	RuntimeUntrusted = "unstrusted"
```
Random-Liu: s/unstrusted/untrusted

pkg/server/sandbox_run_test.go
```diff
@@ -595,6 +603,31 @@ func TestGetSandboxRuntime(t *testing.T) {
...
+			runtimes:                 map[string]criconfig.Runtime{"foo": fooRuntime},
+			expectedRuntime:          fooRuntime,
+		},
+		"should return error if runtime handler is required but not configured": {
+			sandboxConfig:  &runtime.PodSandboxConfig{},
+			runtimeHandler: "bar",
+			defaultRuntime: defaultRuntime,
+			runtimes:       map[string]criconfig.Runtime{"foo": fooRuntime},
+			expectErr:      true,
+		},
```
Random-Liu: missing test case: should return error if both annotation and runtime handler are specified.

pkg/config/config.go
```diff
@@ -33,10 +33,15 @@ type Runtime struct {
 type ContainerdConfig struct {
 	// Snapshotter is the snapshotter used by containerd.
 	Snapshotter string `toml:"snapshotter" json:"snapshotter"`
-	// DefaultRuntime is the runtime to use in containerd.
+	// DefaultRuntime is the default runtime to use in containerd.
```
mikebrow: The documentation for these fields has to be cut and pasted to here: https://github.com/containerd/cri/blame/master/docs/config.md#L52

pkg/server/sandbox_run.go
```diff
@@ -601,9 +601,13 @@ func hostAccessingSandbox(config *runtime.PodSandboxConfig) bool {
 // getSandboxRuntime returns the runtime configuration for sandbox.
 // If the sandbox contains untrusted workload, runtime for untrusted workload will be returned,
 // or else default runtime will be returned.
-func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig) (criconfig.Runtime, error) {
-	untrusted := false
+func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig, runtimeHandler string) (criconfig.Runtime, error) {
 	if untrustedWorkload(config) {
+		// If the untrusted annotation is provided, runtimeHandler MUST be empty.
+		if runtimeHandler != "" {
+			return criconfig.Runtime{}, errors.New("untrusted workload with explicit runtime handler is not allowed")
```
mikebrow: Why can't we match untrusted workloads with a runtime handler for untrusted workloads? I must be missing something.

pkg/config/config.go
Random-Liu: Thanks! I missed that.

cri.go
tallclair: Done.

pkg/config/config.go
tallclair: oh, good catch. Done.

pkg/server/sandbox_run.go
tallclair: Suppose I add the untrusted annotation, and ask for the 'native' runtime? Those are conflictiing. Originally, I was thinking RuntimeHandler should take precedence, but that would be "failing open" in this case, and potentially present a security risk.

However, to smooth roll out / upgrade, I decided we should allow setting the untrusted annotation and setting the 'untrusted' runtimeHandler.

pkg/config/config.go
```diff
@@ -33,10 +33,15 @@ type Runtime struct {
...
+	// This runtime is used when no runtime handler (or the empty string) is provided.
 	DefaultRuntime Runtime `toml:"default_runtime" json:"defaultRuntime"`
 	// UntrustedWorkloadRuntime is a runtime to run untrusted workloads on it.
+	// DEPRECATED: use Runtimes instead. If provided, this runtime is mapped to the runtime handler
+	//     named 'untrusted'. It is a configuration error to provide both.
 	UntrustedWorkloadRuntime Runtime `toml:"untrusted_workload_runtime" json:"untrustedWorkloadRuntime"`
+	// Runtimes maps a CRI RuntimeHandler string to a runtime configuration.
```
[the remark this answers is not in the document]
tallclair: It's not entirely accurate that the RuntimeHandler "specifies types of workloads". For example, one user might want their ngnix workload to be sandboxed, while another is comfortable running it with native isolation. It's not that those are different workloads, necessarily, just different threat models.

pkg/config/config.go
```diff
@@ -33,10 +33,15 @@ type Runtime struct {
 type ContainerdConfig struct {
 	// Snapshotter is the snapshotter used by containerd.
 	Snapshotter string `toml:"snapshotter" json:"snapshotter"`
-	// DefaultRuntime is the runtime to use in containerd.
+	// DefaultRuntime is the default runtime to use in containerd.
+	// This runtime is used when no runtime handler (or the empty string) is provided.
 	DefaultRuntime Runtime `toml:"default_runtime" json:"defaultRuntime"`
 	// UntrustedWorkloadRuntime is a runtime to run untrusted workloads on it.
+	// DEPRECATED: use Runtimes instead. If provided, this runtime is mapped to the runtime handler
+	//     named 'untrusted'. It is a configuration error to provide both.
```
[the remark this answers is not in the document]
tallclair: done.

pkg/config/config.go
tallclair: Done. Add a clarification about the example to the runtimes entry.

pkg/server/sandbox_run_test.go
tallclair: done

pkg/config/config.go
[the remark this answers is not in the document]
mikebrow: Could it be any type of classification. Development, Test, Production; or High, low, no threat model; or multi-tenant vs. uni-tenant; or ... So really it's any type system or class dreamed up by the user mapped to particular runtimes chosen by the user? Do I read that right?  At any rate the amount of threat secured by the runtime will be dependent on other factors outside just the version of the runtime selected. If we are saying these are threat model classes who's  to say which runtime is best for which type of threat model?

cri.go
```diff
@@ -104,6 +108,18 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	return s, nil
 }
 
+// validateConfig validates the given configuration.
+func validateConfig(c *criconfig.Config) error {
+	// It is an error to provide both an UntrustedWorkloadRuntime & define an 'untrusted' runtime.
+	if _, ok := c.ContainerdConfig.Runtimes[criconfig.RuntimeUntrusted]; ok {
+		if c.ContainerdConfig.UntrustedWorkloadRuntime.Type != "" {
+			return errors.New("conflicting untrusted runtimes defined")
```
mikebrow: maybe: "duplicate/conflicting definitions, configuration includes untrusted-workload-runtime and runtimes[`untrusted`]"

pkg/config/config.go
```diff
@@ -183,3 +191,8 @@ func DefaultConfig() PluginConfig {
 		},
 	}
 }
+
+const (
+	// RuntimeUntrusted is the implicit runtime defined for ContainerdConfig.UntrustedWorkloadRuntime
+	RuntimeUntrusted = "untrusted"
```
mikebrow: Would be nice if we could move this string over to CRI maybe a TODO?

cri.go
tallclair: done.

pkg/config/config.go
mikebrow: true.. yet it would be beneficial to have example patterns and reserved names that have more meaning than .. anything :-)

PR #892, opened 2018-09-05
Sort volume mount.

Random-Liu:
Fixes https://github.com/containerd/containerd/issues/2596.

We should cherry-pick this fix into all supported branches.

This PR:
1) Makes sure volume mounts are sorted based by depth, so that parent directory mount won't override the sub directory mount. (Corresponding docker code: https://github.com/moby/moby/blob/a9bd60d674f85c2161bb702404b23501d6b746c3/daemon/volumes_unix.go#L97)
2) Skip default mount or cri plugin controlled mount if the destination is mounted by CRI already.
3) Skip all default `/dev` mounts if `/dev` is mounted in CRI. (Corresponding docker code: https://github.com/moby/moby/blob/a9bd60d674f85c2161bb702404b23501d6b746c3/daemon/oci_linux.go#L555)

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -19,6 +19,7 @@ package server
 import (
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/892#pullrequestreview-152641023

PR #894, opened 2018-09-05
Support masked readonly paths

Random-Liu:
Add `MaskedPaths` and `ReadonlyPaths` support.


Random-Liu:
It's done here. https://github.com/containerd/cri/blob/master/pkg/server/container_create.go#L749

PR #895, opened 2018-09-05
Add HOSTNAME to env by default for pod containers

estesp:
To match expectations of users coming from Docker engine runtime, add
the HOSTNAME to the environment of new containers in a pod.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -341,6 +341,8 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 	for _, e := range config.GetEnvs() {
 		g.AddProcessEnv(e.GetKey(), e.GetValue())
 	}
+	// add the HOSTNAME variable to the environment to match Docker runtime default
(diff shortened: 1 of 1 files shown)


Random-Liu:
/lgtm

We may want to cherry-pick this to fix containerd 1.1.x.

estesp:
I'm happy to add it back on that branch as well

Random-Liu:
It looks like this PR broke the e2e test. The reason is:
1) cri-containerd e2e test starts failing after this PR merged; https://k8s-testgrid.appspot.com/sig-node-containerd#e2e-gci
2) containerd e2e test is fine, so kubernetes change can be mostly ruled out. https://k8s-testgrid.appspot.com/sig-node-containerd#containerd-e2e-gci

I'm going to revert this change for now, and see whether it brings back the test back to normal.
* If it does, we need to revert the changes in supported branches, and investigate why this breaks the test.
* If it doesn't, we can submit this PR again and find the actual root cause.

PR #898, opened 2018-09-06
Revert "Add HOSTNAME to env by default for pod containers"

Random-Liu:
This reverts commit 4c3e195db36e67d65bebd48b9987596799d9d543.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -342,8 +342,6 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 	for _, e := range config.GetEnvs() {
 		g.AddProcessEnv(e.GetKey(), e.GetValue())
 	}
-	// add the HOSTNAME variable to the environment to match Docker runtime default
-	g.AddProcessEnv("HOSTNAME", sandboxConfig.GetHostname())
 
 	securityContext := config.GetLinux().GetSecurityContext()
 	selinuxOpt := securityContext.GetSelinuxOptions()


yujuhong:
From the kube-addon-manager log:
```
INFO: Leader is bootstrap-e2e-master
INFO: Not elected leader, going back to sleep.
INFO: Leader is bootstrap-e2e-master
INFO: Not elected leader, going back to sleep.
INFO: Leader is bootstrap-e2e-master
INFO: Not elected leader, going back to sleep.
INFO: Leader is bootstrap-e2e-master
```
The addon manager did not think it's the leader `bootstrap-e2e-master` and refused to continue creating all the addons (including the cluster roles):
https://github.com/kubernetes/kubernetes/blob/d0439d417b0563d44c65b6e400e2070964dea7d1/cluster/addons/addon-manager/kube-addons.sh#L162-L164

yujuhong:
By the way, kube-addon-manager runs in the host network

yujuhong:
> By the way, kube-addon-manager runs in the host network

kubelet probably doesn't set hostname for pods running in the host network. Not sure how docker handles this.

estesp:
Interesting; it definitely looks like `HOSTNAME` is part of the problem here.

This is the code where Docker engine sets up `HOSTNAME` as a default env var in every container (https://github.com/moby/moby/blob/master/container/container.go#L694-L718), which is called from the OCI spec generator without any kind of checking if netns == host.

I also looked at dockershim and don't see any env replacement or changes happening; I can't even find the string "HOSTNAME" in any files in pkg/kubelet. So, something is definitely different, but I'm not sure what is occurring here.

estesp:
hmm, looks like the missing piece is that by the time that code runs in Docker, if `--net=host` it has set the metadata config's hostname to the actual host hostname; not sure if that is the case with the way this is happened here; Docker for Mac example:

```
docker run --rm --net=host alpine sh -c 'echo $HOSTNAME'
linuxkit-025000000001
```

The container's hostname is the LinuxKit VM hostname.

yujuhong:
> I also looked at dockershim and don't see any env replacement or changes happening; I can't even find the string "HOSTNAME" in any files in pkg/kubelet. So, something is definitely different, but I'm not sure what is occurring here.

Kubelet passes the hostname through the podsandbox config, but it doesn't add an env variable directly.


> hmm, looks like the missing piece is that by the time that code runs in Docker, if --net=host it has set the metadata config's hostname to the actual host hostname

That sounds reasonable. Docker probably derives this information before passing it through the config.

Random-Liu:
https://github.com/moby/moby/blob/master/daemon/container_operations.go#L936-L943

I'll make the change.

PR #901, opened 2018-09-10
Fix hostname env.

Random-Liu:
Ref https://github.com/containerd/cri/pull/898#issuecomment-419991978

Add `HOSTNAME` back. /cc @estesp 


estesp:
Nice; thanks @Random-Liu

Random-Liu:
We should cherrypick this into all supported branches.

PR #914, opened 2018-09-13
Fix addition gids

Random-Liu:
For ibm-messaging/mq-container#207.
Fixes https://github.com/containerd/cri/issues/910.
Depend on https://github.com/containerd/containerd/pull/2641.

/cc @estesp


Random-Liu:

estesp:
what's up with the temp mount `/etc/passwd` errors on the Travis CI test phase? Not sure what would have changed though to cause that

pkg/containerd/opts/spec.go
```diff
@@ -0,0 +1,51 @@
...
+	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// WithAdditionalGIDs adds any additional groups listed for a particular user in the
+// /etc/groups file of the image's root filesystem to the OCI spec's additionalGids array.
+func WithAdditionalGIDs(userstr string) oci.SpecOpts {
+	return func(ctx context.Context, client oci.Client, c *containers.Container, s *runtimespec.Spec) (err error) {
+		gids := s.Process.User.AdditionalGids
+		if err := oci.WithAdditionalGIDs(userstr)(ctx, client, c, s); err != nil {
+			return err
```
mikebrow: and if not should we eat the err?

Random-Liu:
@estesp It is because of this https://github.com/containerd/containerd/pull/2641#discussion_r217477155.

pkg/containerd/opts/spec.go
Random-Liu: We should eat the `os.IsNotExist(err)` in oci.WithAdditionalGIDs.

Random-Liu:
Updated to latest containerd, and get rid of the temporary local containerd change.

PR #924, opened 2018-09-21
Add timeout for recover and event monitor

Random-Liu:
This is another attempt of https://github.com/containerd/cri/pull/884.
1) Add timeout to both recovery logic and event monitor.
2) Generate Unknown state if fail to load container/sandbox status.

Update containerd to include ttrpc context cancellation support https://github.com/containerd/ttrpc/pull/29.

I've experimented, the context can now successfully stop pending calls to containerd-shim.
There is some wait group inside containerd, which could still block forever though, e.g. https://github.com/containerd/cri/blob/master/pkg/server/io/helpers.go#L63, which we should try to fix in the future.

This is still much better than before. We should cherry-pick this to v1.2.


pkg/server/restart.go
```diff
@@ -157,118 +172,128 @@ func (c *criService) loadContainer(ctx context.Context, cntr containerd.Containe
...
 
-	// Load up-to-date status from containerd.
-	var containerIO *cio.ContainerIO
-	t, err := cntr.Task(ctx, func(fifos *containerdio.FIFOSet) (_ containerdio.IO, err error) {
-		stdoutWC, stderrWC, err := c.createContainerLoggers(meta.LogPath, meta.Config.GetTty())
-		if err != nil {
-			return nil, err
-		}
-		defer func() {
+	// Generate container status based on checkedpointed status and actual state in containerd.
```
Random-Liu: I didn't change any logic, wrap the status generation logic in a function.

pkg/server/restart.go
```diff
@@ -324,43 +351,53 @@ func loadSandbox(ctx context.Context, cntr containerd.Container) (sandboxstore.S
...
 	var pid uint32
-	if notFound {
-		// Task does not exist, set sandbox state as NOTREADY.
-		state = sandboxstore.StateNotReady
-	} else {
-		if s.Status == containerd.Running {
-			// Task is running, set sandbox state as READY.
-			state = sandboxstore.StateReady
-			pid = t.Pid()
+	if err := func() error {
```
Random-Liu: I didn't change any logic, wrap the status generation logic in a function.

pkg/server/events.go
```diff
@@ -42,6 +42,13 @@ const (
 	backOffInitDuration        = 1 * time.Second
 	backOffMaxDuration         = 5 * time.Minute
 	backOffExpireCheckDuration = 1 * time.Second
+
+	// handleEventTimeout is the timeout for handling 1 event. Event monitor
+	// handles events in serial, if one event blocks the event monitor, no
+	// other events can be handled.
+	// Add a timeout for each event handling, timeout events will be requeued and
```
mikebrow: nit
/s/timeout events/events that timeout/

pkg/server/events.go
```diff
@@ -42,6 +42,13 @@ const (
...
+	// handled again in the future.
+	handleEventTimeout = 5 * time.Second
```
mikebrow: why 5 seconds? Is that a standard convention? This seems short, would not want to cause an endless loop.

pkg/server/restart.go
```diff
@@ -136,8 +136,23 @@ func (c *criService) recover(ctx context.Context) error {
...
+// containers/sandboxes.
+// Most CRI container/sandbox related operations are per container, the ones
+// which handle multiple containers at a time are:
+// * ListPodSandboxes: Don't talk with containerd services.
+// * ListContainers: Don't talk with containerd services.
+// * ListContainerStats: Not in critical code path, a default timeout will
+// be applied at CRI level.
+// * Recovery logic: We should set a time for each container/sandbox recovery.
+// * Event montior: We should set a timeout for each container/sandbox event handling.
+const loadContainerTimeout = 5 * time.Second
```
mikebrow: seems very short.. doesn't kube already do this? ok for sandbox but not for containers?

pkg/server/restart.go
mikebrow: why the edit?

pkg/server/restart.go
mikebrow: ok but why?

Random-Liu:
Based on https://github.com/kubernetes/kubernetes/issues/69060, kubelet will start new container although there is one in unknown state.

So it is not very useful to handle load failure containers as unknown. Given so, I prefer doing that in another PR, and not cherrypick it. In this PR I'll just simply add timeout.

pkg/server/events.go
Random-Liu: Done

pkg/server/events.go
Random-Liu: Not a convention. I just want to keep it as short as possible, because it will block the whole event monitor.
We don't do much when handling an event, I think 5 second should be good enough.

PR #926, opened 2018-09-26
Better external image handling

Random-Liu:
For https://github.com/containerd/cri/issues/909.
For https://github.com/containerd/containerd/issues/2678.

This PR:
1) Add `io.cri-containerd.image` label for `cri` plugin managed images.
2) For images pulled/imported by other clients, e.g. `ctr`, the `cri` plugin will generate the image id as the unique reference, and apply the `io.cri-containerd.image` label. This prevents the image from losing all references unexpectedly.
3) Add 2 integration tests: "image can be recovered during restart" and "image pulled with containerd client can be seen by CRI plugin".

/cc @dmcgowan

### DIFF
--- a/integration/containerd_image_test.go
+++ b/integration/containerd_image_test.go
@@ -1,0 +1,148 @@
+/*
+Copyright 2018 The containerd Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"golang.org/x/net/context"
+	"testing"
+	"time"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
+)
+
+// Test to test the CRI plugin should see image pulled into containerd directly.
+func TestContainerdImage(t *testing.T) {
+	testImage := "docker.io/library/busybox:latest"
+	ctx := context.Background()
+
+	t.Logf("make sure the test image doesn't exist in the cri plugin")
+	i, err := imageService.ImageStatus(&runtime.ImageSpec{Image: testImage})
+	require.NoError(t, err)
+	if i != nil {
+		require.NoError(t, imageService.RemoveImage(&runtime.ImageSpec{Image: testImage}))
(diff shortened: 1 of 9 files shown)


Random-Liu:
@mikebrow Can you take a look at this? This is a more serious issue, because if image loses reference silently, we `ContainerStatus` will always fail because it can't convert image id to repotag and repodigest. This line https://github.com/containerd/cri/blob/master/pkg/server/container_status.go#L45 will always fail.

pkg/server/image_load.go
```diff
@@ -44,8 +44,11 @@ func (c *criService) LoadImage(ctx context.Context, r *api.LoadImageRequest) (*a
 	}
 	for _, repoTag := range repoTags {
 		// Update image store to reflect the newest state in containerd.
-		if err := c.imageStore.Update(ctx, repoTag); err != nil {
-			return nil, errors.Wrapf(err, "failed to update image store %q", repoTag)
+		// Image imported by importer.Import is not treated as managed
+		// by the cri plugin, call `updateImage` to make it managed.
+		// TODO(random-liu): Replace this with the containerd library (issue #909).
+		if err := c.updateImage(ctx, repoTag); err != nil {
```
mikebrow: nit updateImage(ctx, repoTags...)

pkg/server/image_pull.go
```diff
@@ -174,20 +176,57 @@ func (c *criService) createImageReference(ctx context.Context, name string, desc
...
+		return nil
+	}
+	_, err = c.client.ImageService().Update(ctx, img, "target", "labels")
 	return err
 }
 
+// updateImage updates image store to reflect the newest state of an image reference
+// in containerd. If the reference is not managed by the cri plugin, the function also
+// generates necessary metadata for the image and make it managed.
+func (c *criService) updateImage(ctx context.Context, r string) error {
```
mikebrow: suggest: func (c *criService) updateImage(ctx context.Context, refs ...string) error

integration/containerd_image_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	t.Logf("pull the image into containerd")
+	_, err = containerdClient.Pull(ctx, testImage, containerd.WithPullUnpack)
+	assert.NoError(t, err)
+	defer func() {
+		t.Logf("image should be removed from the cri plugin if all references get deleted")
+		if err := containerdClient.ImageService().Delete(ctx, testImage); err != nil {
+			assert.True(t, errdefs.IsNotFound(err), err)
+		}
+		if err := containerdClient.ImageService().Delete(ctx, testImage); err != nil {
```
mikebrow: delete twice?

Random-Liu: My bad... The second time should be the image id... The test shouldn't pass then... Let me see.

Random-Liu: OK. The reason is that after the tag is deleted, we can't reference the image by tag, but we can still reference it by id. This is expected behavior.

I've updated the test.

pkg/server/image_load.go
Random-Liu: Let's do it one by one, we need the log below

pkg/server/image_pull.go
Random-Liu: ditto.

integration/containerd_image_test.go
Random-Liu: Done.

PR #936, opened 2018-10-03
Integration test dirty fix

mikebrow:
If the k8s.io namespace is not "free" from containers at the start of integration tests it will fail on the first container that does not have annotations set. This is not the intention of the test. 

So adding support for validating that we have stats for the right 3 containers.

### DIFF
--- a/integration/container_stats_test.go
+++ b/integration/container_stats_test.go
@@ -111,18 +111,28 @@ func TestContainerListStats(t *testing.T) {
 		if err != nil {
 			return false, err
 		}
+		var count int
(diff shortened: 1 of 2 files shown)


Random-Liu:
I prefer adding a random label on the test container/pod.

mikebrow:
@Random-Liu ok.. added random labels and annotations.

dmcgowan:
The master branch on this repository has been merged into [containerd/containerd](https://github.com/containerd/containerd). Please rebase your change onto containerd master and open a pull request there.

PR #937, opened 2018-10-03
cni result: add a debug message

JulienBalestra:
Add the Container Network Interface result output to the debug logs.

Output of this change:
```
time="2018-10-03T14:05:55.532915608+02:00" level=debug msg="cni result for sandbox "7db4d671ab60113bcec40ae9ca29dd059f6d22d44ca7b90bc9e379294b148edc"" cni-result="{"Interfaces":{"eth0":{"IPConfigs":[{"IP":"192.168.253.3","Gateway":"192.168.253.1"}],"Mac":"","Sandbox":""}},"DNS":[{},{}],"Routes":[{"dst":"0.0.0.0/0","gw":"192.168.253.1"}]}"
```

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -17,6 +17,7 @@ limitations under the License.
 package server
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -26,7 +27,7 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/runtime/linux/runctypes"
-	cni "github.com/containerd/go-cni"
+	"github.com/containerd/go-cni"
 	"github.com/containerd/typeurl"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
@@ -530,7 +531,7 @@ func (c *criService) unmountSandboxFiles(id string, config *runtime.PodSandboxCo
 // setupPod setups up the network for a pod
 func (c *criService) setupPod(id string, path string, config *runtime.PodSandboxConfig) (string, error) {
 	if c.netPlugin == nil {
-		return "", errors.New("cni config not intialized")
+		return "", errors.New("cni config not initialized")
 	}
 
 	labels := getPodCNILabels(id, config)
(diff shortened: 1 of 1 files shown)


pkg/server/sandbox_run.go
```diff
@@ -639,3 +641,15 @@ func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig, runtime
 	}
 	return handler, nil
 }
+
+func logDebugCNIResult(sandboxID string, result *cni.CNIResult) {
+	if logrus.GetLevel() != logrus.DebugLevel {
```
mikebrow: We also have a trace level that is DebugLevel + 1. See our use of "github.com/containerd/containerd/log" throughout. So let's go ahead and import that package here and do a switch or just make it >= DebugLevel. Then test it with trace on..

pkg/server/sandbox_run.go
```diff
@@ -639,3 +641,15 @@ func (c *criService) getSandboxRuntime(config *runtime.PodSandboxConfig, runtime
...
+		return
+	}
+	cniResult, err := json.Marshal(result)
+	if err != nil {
+		logrus.WithError(err).Errorf("Failed to marshal CNI result for sandbox %q: %v", sandboxID, err)
+		return
+	}
+	logrus.WithField("cni-result", string(cniResult)).Debugf("cni result for sandbox %q", sandboxID)
```
mikebrow: This would be the only use of field descriptive log output in the cri code. No need to say "cni-result" and "cni result for." So I'd pick one, with field if you think that is needed / have a reason for the field descriptive logging and remove the "cni result for" language, or just use logsrus.Debugf

mikebrow:
@JulienBalestra thanks for your contribution please sign your commit.

pkg/server/sandbox_run.go
JulienBalestra: I hope I addressed what described.
If the iota Level is lower than 5 (debug) we don't log (e.g: 4 == info).

pkg/server/sandbox_run.go
JulienBalestra: Sure this is quite redundant, I'll follow the codestyle of the project and don't use `WithField` 👌

pkg/server/sandbox_run.go
mikebrow: yes I meant continue if >= I forgot you were returning early so yeah return if < debug 
:-) heh

PR #949, opened 2018-10-10
Fix ip leakage

### LINKED ISSUE #948: IPAM IP leakage
Reported by @lbernail.

After a node reboot, all network namespaces in `/var/run/netns` will be wiped out.

In that case, we'll skip `teardown` the cni network, because the network namespace has already gone. https://github.com/containerd/cri/blob/master/pkg/server/sandbox_stop.go#L64

However, if `host-local` `ipam` is being used, it checkpoints allocated ip in `/var/lib/cni/containerd-net/`, and the checkpoint won't be wiped out, e.g.
```console
$ ls /var/lib/cni/networks/containerd-net/
10.88.0.2  last_reserved_ip.0  lock
```
**If we skip `teardown`, the ip will be leaked forever.**

To fix this problem, we need to:
1) Teardown the network even if network namespace doesn't exist.
2) Ignore "no such file or directory" error in go-cni.

Actually, we should be able to pass in empty network namespace if it doesn't exist (see [the PR](https://github.com/containernetworking/plugins/commit/72337159c1c28772f752e9abedf74f7b24fce11a)). However, the behavior is not consistent across different plugins right now, at least `loopback` and `host-device` still returns error. And we are using `loopback` by default.

I've filed https://github.com/containernetworking/plugins/issues/210 for this. Before that is fixed, let's just ignore the "no such file or directory" error, which the cni libraries in kubernetes are doing as well: [cni](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/network/cni/cni.go#L329), [kubenet](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go#L591).

We should fix this and cherrypick into all supported branches.

Random-Liu:
Fixes #948.
Based on https://github.com/containerd/go-cni/pull/31.

We should cherrypick this PR into all supported branches.

### DIFF
--- a/integration/test_utils.go
+++ b/integration/test_utils.go
@@ -24,11 +24,14 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/containerd/containerd"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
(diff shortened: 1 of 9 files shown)


integration/test_utils.go
```diff
@@ -295,6 +298,15 @@ func KillProcess(name string) error {
 	return nil
 }
 
+// KillPid kills the process by pid. kill is used.
+func KillPid(pid int) error {
+	output, err := exec.Command("kill", "-9", strconv.Itoa(pid)).CombinedOutput()
```
mikebrow: i think you wanted to send a SIGTERM here (-15) not a SIGKILL? or maybe the KillProcess is wrong as I see it's using default which I believe is SIGTERM.

Random-Liu: Use kill directly, which uses SIGTERM by default.

Random-Liu:
> need to update vendor I think

Done.

integration/test_utils.go
mikebrow: :-) good

PR #956, opened 2018-10-25
CreateContainerRequest should use PodSandboxConfig based on ID

jterry75:

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -76,12 +76,15 @@ func init() {
 // CreateContainer creates a new container in the given PodSandbox.
 func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateContainerRequest) (_ *runtime.CreateContainerResponse, retErr error) {
 	config := r.GetConfig()
-	sandboxConfig := r.GetSandboxConfig()
 	sandbox, err := c.sandboxStore.Get(r.GetPodSandboxId())
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to find sandbox id %q", r.GetPodSandboxId())
 	}
 	sandboxID := sandbox.ID
+	// PodSandboxConfig on `r` is optionally passed by convention. But it must
+	// be the same as the original request so just use the cached version in all
+	// cases.
+	sandboxConfig := sandbox.Config
 	s, err := sandbox.Container.Task(ctx, nil)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get sandbox container task")


jterry75:
@crosbymichael @Random-Liu - FYI. It turns out this causes an AV and crashes CRI if you send `nil` here. But the spec allows for `nil` and is only sent by convention.

jterry75:
No the goal is not to *harden* the API. As you point out a `nil` for `Config` would also AV CRI. However, we do not allow an empty `PodSandboxId` as we will fail the sandbox store check. Personally it doesnt make sense to me why you would be allowed to pass a `PodSandboxConfig` here given that it is **immutable in the lifetime of the pod** but that seems to be the API. Why would we ever take the passed in `SandboxConfig` over the one resolved via `PodSandboxId`?

jterry75:
Interestingly: [cri-o](https://github.com/kubernetes-sigs/cri-o/blob/71f84607110e1d9cba2f7012ba338528304c7fc7/server/container_create.go#L439) seems to follow this logic:
- `PodSandboxId` Required.
- `Config` Required.
- `SandboxConfig` Optional - Will not AV if not passed. BUT: [uses](https://github.com/kubernetes-sigs/cri-o/blob/8bbac5e061150c64f7434b742a907a0168e545d6/server/container_create_linux.go#L255) `sandboxConfig.GetLinux().GetSecurityContext().GetPrivileged()` and `sandboxConfig.GetLogDirectory()`.

So technically it wouldn't AV in the case we are but it would allow me to change the `SandboxConfig` options and have them used.

Random-Liu:
In what case the sandbox config will be nil? I don’t think that is allowed.

I think the reason we pass sandboxconfig again here is to avoid the
necessity to checkpoint it in the runtime. We do checkpoint it in
containerd, but dockershim doesn’t.

I think it should never be nil.

jterry75:
@Random-Liu - I don't follow. What do you mean *we need to checkpoint*. All this does is give CRI the sandbox config as it was originally passed via `RunPodSandbox`. If you are worried about a regression here I can make it only use the in-memory version if it is not `nil`.

Random-Liu:
I mean RunPodSandbox and CreateContainer are different request. To use the
sandbox config passed in RunPodSandbox in CreateContainer, we need to
checkpoint the sandbox config we get in RunPodSandbox. That is what sandbox
store does.

However, CRI doesn’t expect the runtime to do that, and dockershim doesn’t
do that, either. That is why we pass sandbox config again to
CreateContainer.

So in short, SandboxConfig passed to CreateContainer should not be nil, we
should be able to rely on it.

jterry75:
Ah. Ok no issues then. And we don't care about protecting against the AV in favor of a better error?

mikebrow:
> Ok no issues then. And we don't care about protecting against the AV in favor of a better error?

Thus my points about hardening and testing. One would need to modify the CRI docs to make the desired behavior explicit, develop the tests to ensure the desired behavior is correct, and add prs to the popular container runtimes as well. Otherwise we go with tyranny of the first implementor, docker shim :-)

jterry75:
Sounds good then closing.

PR #965, opened 2018-10-31
update the slack information

mikebrow:
see... https://github.com/containerd/containerd/pull/2754

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,8 @@ implementation.
 For sync communication we have a community slack with a #containerd channel that
 everyone is welcome to join and chat about development.
 
-**Slack:** https://dockr.ly/community
+**Slack:** Catch us in the #containerd and #containerd-dev channels on dockercommunity.slack.com.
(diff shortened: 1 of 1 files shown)


caniszczyk:
I'm hesitant to recommend the Docker Slack as it's corporate controlled and not neutral, can we just point people to a mailing list or other means for help?

estesp:
As usual, momentum is a barrier to shifting gears mid-stream--the community has been in a set of channels on the community Slack team for a long time, with 660 people in our main channel at this point. To be clear, dockercommunity.slack.com is a free community Slack team separate from Docker's corporate slack; but yes, is operated/managed by dev rel team at Docker.

estesp:
To be clear, we could put something else as a place for "help", but usually you want to point people to where discussion is actually happening and that is currently on the existing Slack channels that have been used for a few years now.

mikebrow:
> I'm hesitant to recommend the Docker Slack as it's corporate controlled and not neutral, can we just point people to a mailing list or other means for help?

the mailing list is included below.. this is where we chat actively about containerd. This replaced the irc channels. There is a related sig-node channel, but really the conversations about containerd happen at the channels listed.

PR #970, opened 2018-11-09
support DisableCgroup, DisableApparmor, RestrictOOMScoreAdj

AkihiroSuda:
Add following config for supporting "rootless" mode

* DisableCgroup: disable cgroup
* DisableApparmor: disable Apparmor
* RestrictOOMScoreAdj: restrict the lower bound of OOMScoreAdj

Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

---

A quick way to test this PR is to use Usernetes v20181109.0 ( https://github.com/rootless-containers/usernetes )

```console
$ ./run.sh default-containerd
$ ./kubectl.sh run -it --rm --image alpine foo
```

cc @giuseppe for consistency with CRI-O implementation (https://github.com/kubernetes-sigs/cri-o/blob/2accad9fab352c445bb4c414d9d3c8cdf351c524/server/rootless.go could be simplified as in this PR with the latest runc IIUC)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -142,6 +142,16 @@ type PluginConfig struct {
 	// Log line longer than the limit will be split into multiple lines. Non-positive
 	// value means no limit.
 	MaxContainerLogLineSize int `toml:"max_container_log_line_size" json:"maxContainerLogSize"`
+	// DisableCgroup indicates to disable the cgroup support.
+	// This is useful when the containerd does not have permission to access cgroup.
+	DisableCgroup bool `toml:"disable_cgroup" json:"disableCgroup"`
+	// DisableApparmor indicates to disable the apparmor support.
+	// This is useful when the containerd does not have permission to access Apparmor.
+	DisableApparmor bool `toml:"disable_apparmor" json:"disableApparmor"`
+	// RestrictOOMScoreAdj indicates to limit the lower bound of OOMScoreAdj to the containerd's
+	// current OOMScoreADj.
+	// This is useful when the containerd does not have permission to decrease OOMScoreAdj.
+	RestrictOOMScoreAdj bool `toml:"restrict_oom_score_adj" json:"restrictOOMScoreAdj"`
 }
 
 // X509KeyPairStreaming contains the x509 configuration for streaming
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -417,12 +417,18 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 
 	g.SetRootReadonly(securityContext.GetReadonlyRootfs())
 
-	setOCILinuxResource(&g, config.GetLinux().GetResources())
-
-	if sandboxConfig.GetLinux().GetCgroupParent() != "" {
-		cgroupsPath := getCgroupsPath(sandboxConfig.GetLinux().GetCgroupParent(), id,
-			c.config.SystemdCgroup)
-		g.SetLinuxCgroupsPath(cgroupsPath)
+	if c.config.DisableCgroup {
+		g.SetLinuxCgroupsPath("")
+	} else {
+		setOCILinuxResourceCgroup(&g, config.GetLinux().GetResources())
+		if sandboxConfig.GetLinux().GetCgroupParent() != "" {
+			cgroupsPath := getCgroupsPath(sandboxConfig.GetLinux().GetCgroupParent(), id,
+				c.config.SystemdCgroup)
+			g.SetLinuxCgroupsPath(cgroupsPath)
+		}
+	}
+	if err := setOCILinuxResourceOOMScoreAdj(&g, config.GetLinux().GetResources(), c.config.RestrictOOMScoreAdj); err != nil {
+		return nil, err
 	}
 
 	// Set namespaces, share namespace with sandbox container.
@@ -744,8 +750,8 @@ func setOCIBindMountsPrivileged(g *generate.Generator) {
 	spec.Linux.MaskedPaths = nil
 }
 
-// setOCILinuxResource set container resource limit.
-func setOCILinuxResource(g *generate.Generator, resources *runtime.LinuxContainerResources) {
+// setOCILinuxResourceCgroup set container cgroup resource limit.
+func setOCILinuxResourceCgroup(g *generate.Generator, resources *runtime.LinuxContainerResources) {
 	if resources == nil {
 		return
 	}
@@ -753,9 +759,26 @@ func setOCILinuxResource(g *generate.Generator, resources *runtime.LinuxContaine
 	g.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
 	g.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
 	g.SetLinuxResourcesMemoryLimit(resources.GetMemoryLimitInBytes())
-	g.SetProcessOOMScoreAdj(int(resources.GetOomScoreAdj()))
 	g.SetLinuxResourcesCPUCpus(resources.GetCpusetCpus())
 	g.SetLinuxResourcesCPUMems(resources.GetCpusetMems())
+}
+
+// setOCILinuxResourceOOMScoreAdj set container OOMScoreAdj resource limit.
+func setOCILinuxResourceOOMScoreAdj(g *generate.Generator, resources *runtime.LinuxContainerResources, restrictOOMScoreAdjFlag bool) error {
+	if resources == nil {
+		return nil
+	}
+	adj := int(resources.GetOomScoreAdj())
+	if restrictOOMScoreAdjFlag {
+		var err error
+		adj, err = restrictOOMScoreAdj(adj)
+		if err != nil {
+			return err
+		}
+	}
+	g.SetProcessOOMScoreAdj(adj)
+
+	return nil
 }
 
 // getOCICapabilitiesList returns a list of all available capabilities.
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -18,6 +18,7 @@ package server
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
@@ -495,4 +496,28 @@ func getRuntimeOptions(c containers.Container) (interface{}, error) {
 		return nil, err
 	}
 	return opts, nil
+}
+
+func getCurrentOOMScoreAdj() (int, error) {
+	b, err := ioutil.ReadFile("/proc/self/oom_score_adj")
+	if err != nil {
+		return 0, errors.Wrap(err, "could not get the daemon oom_score_adj")
+	}
+	s := strings.TrimSpace(string(b))
+	i, err := strconv.Atoi(s)
+	if err != nil {
+		return 0, errors.Wrap(err, "could not get the daemon oom_score_adj")
+	}
+	return i, nil
+}
+
+func restrictOOMScoreAdj(preferredOOMScoreAdj int) (int, error) {
+	currentOOMScoreAdj, err := getCurrentOOMScoreAdj()
+	if err != nil {
+		return preferredOOMScoreAdj, err
+	}
+	if preferredOOMScoreAdj < currentOOMScoreAdj {
+		return currentOOMScoreAdj, nil
+	}
+	return preferredOOMScoreAdj, nil
 }
--- a/docs/config.md
+++ b/docs/config.md
@@ -44,6 +44,19 @@ The explanation and default value of each configuration item are as follows:
   # limit.
   max_container_log_line_size = 16384
 
+  # disable_cgroup indicates to disable the cgroup support.
+  # This is useful when the daemon does not have permission to access cgroup.
+  disable_cgroup = false
+
+  # disable_apparmor indicates to disable the apparmor support.
+  # This is useful when the daemon does not have permission to access apparmor.
+  disable_apparmor = false
+
+  # restrict_oom_score_adj indicates to limit the lower bound of OOMScoreAdj to
+  # the containerd's current OOMScoreAdj.
+  # This is useful when the containerd does not have permission to decrease OOMScoreAdj.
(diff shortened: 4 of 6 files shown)


pkg/config/config.go
```diff
@@ -142,6 +142,16 @@ type PluginConfig struct {
...
+	// NoCgroup indicates to disable the cgroup support.
+	// This is useful when the daemon does not have permission to access cgroups.
+	NoCgroup bool `toml:"no_cgroup" json:"noCgroup"`
+	// NoApparmor indicates to disable the apparmor support.
+	// This is useful when the daemon does not have permission to access Apparmor.
+	NoApparmor bool `toml:"no_apparmor" json:"noApparmor"`
+	// RestrictOOMScoreAdj indicates to limit the lower bound of OOMScoreAdj to the daemon's
+	// current OOMScoreADj.
+	// This is useful when the daemon does not have permission to decrease OOMScoreAdj.
+	RestrictOOMScoreAdj bool `toml:"restrict_oom_score_adj" json:"restrictOOMScoreAdj"`
```
mikebrow: I believe these are runtime options and will have to start out as deprecated since they should be specified only for runc? runtimes?

mikebrow: See documentation location for updates to config options: https://github.com/containerd/cri/blob/master/docs/config.md

mikebrow: See: https://github.com/containerd/containerd/blob/130d07edd2e003b48709c0f4d8769be2ac164c7c/runtime/v2/runc/options/oci.pb.go#L57

AkihiroSuda: >  they should be specified only for runc?

Why? `OOMScoreAdj` is defined in OCI Runtime Spec: https://github.com/opencontainers/runtime-spec/blob/5806c35637336642129d03657419829569abc5aa/config.md

mikebrow: by runc I mean runc compatible runtimes.. as specified by the OCI Runtime Spec.

mikebrow: there's a new feature in kube now to support specifying which runtime(s) you want.. which may be one runc with different options.. say pod 1 with defaults and pod 2 rootless :)

AkihiroSuda: My changes are for how to generate OCI Runtime Spec `config.json`, not for OCI Runtime CLI flags. Is `runtime/v2/runc/options/oci.pb.go` designed to hold options for the OCI Runtime Spec `config.json` generator?

mikebrow: Yes.

sorry it took so long to get back to you was busy on another urgent project...

mikebrow: https://github.com/containerd/cri/pull/943

AkihiroSuda: Thanks for the info, I see the TOML section `[plugins.cri.containerd.default_runtime.options]` is designed to hold OCI Runtime Spec `config.json` generator options, however, I'm still not sure the v2 runtime itself is designed to modify `config.json` with options specified in `runtime/v2/runc/options/oci.pb.go`..

Maybe we should introduce a new field like `[plugins.cri.containerd.default_runtime.spec_options]` for `config.json` options?

mikebrow: number of ways to do it I suppose.. heading to china for kubecon shanghai. WDYT @Random-Liu?

Random-Liu:
In OCI spec, both `AppArmor` and `OOMScoreAdj` is a `Process` level configuration. I think we can match that, so I'm fine with a daemon level `NoAppArmor` and `RestrictOOMScoreAdj` option.

However, as for `NoCgroup`, although it is Linux specific, but it is still different from the existing runtime options, e.g. "NoPivot":
* Options like "NoPivot" or "SystemdCgroup" are actually runc specific (they are runc flags), and they are actually changing runc's behavior.
* However, `NoCgroup` is changing `cri` plugin's behavior - it changes how `cri` should config the OCI spec. Given so, I think it is a `cri` daemon level configuration. As for the fact it is Linux specific, we need to have different groups of daemon level configurations for Windows and Linux in the future anyway.

@AkihiroSuda Are these options only useful for `rootless`? If that is the case, can we add a `Rootless` section in the config? So that it is clear to the user that this is for rootless.

pkg/server/container_create.go
```diff
@@ -418,11 +418,20 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 
 	setOCILinuxResource(&g, config.GetLinux().GetResources())
 
-	if sandboxConfig.GetLinux().GetCgroupParent() != "" {
```
Random-Liu: We should:
1) Either move implementation of these options to the place the option was originally configured, e.g. `OOMScoreAdj` is configured in [`setOCILinuxResource`](https://github.com/containerd/cri/blob/70da14e4b37909431afb86508d9d83d49136a98c/pkg/server/container_create.go#L755);
2) Or add a separate function to handle Rootless, and move all the implementation of these options there. And we should make sure the `Rootless` function is called after the spec generation, so that it can cleanup unwanted options.

For example, the apparmor clearing up here won't take effect, because it will be overwrite by [this](https://github.com/containerd/cri/blob/70da14e4b37909431afb86508d9d83d49136a98c/pkg/server/container_create.go#L235).

mikebrow: I vote #1..

mikebrow:
> @AkihiroSuda Are these options only useful for rootless? If that is the case, can we add a Rootless section in the config? So that it is clear to the user that this is for rootless.

I was sort of thinking these are runtime config options required for filtering oci spec generation to make sure it will run rootless. But I don't see why we would want to filter/force all pods on the host to be rootless. Is that the goal here? Push rootless up the stack or to support running the shims/runc/containerd as rootless? Switching to rootless at the host/containerd/cri level seems like a pretty big switch. 

How will these changes effect kata support?

mikebrow:
ok. Even kubelet, interesting.

AkihiroSuda:
It might be complicating, if we want to support automatic detection, maybe we should change the variable type from `bool` to `string`, and reserve "auto" for the future implementation (Either way, probably we should allow setting `NoCgroups=false` even in rootless mode, because cgroups can be used when the fs is chowned)

AkihiroSuda:
@Random-Liu @mikebrow any thought?

mikebrow:
Suggest: using a string for rootless, "auto" should be default. False would make sure we are running as root otherwise throw error on init, true would make sure we are not running as root otherwise error on init.  Agree to implementing NoCgroup, NoAppArmor, and RestrictOOMScoreAdj as daemon level options. See @Random-Liu's comment above regarding:
> move implementation of these options to the place the option was originally configured, e.g. OOMScoreAdj is configured in setOCILinuxResource

pkg/server/helpers.go
```diff
@@ -496,3 +499,89 @@ func getRuntimeOptions(c containers.Container) (interface{}, error) {
 	}
 	return opts, nil
 }
+
+func getCurrentOOMScoreAdj() int {
+	b, err := ioutil.ReadFile("/proc/self/oom_score_adj")
+	if err != nil {
+		return 0
+	}
+	i, err := strconv.Atoi(string(b))
```
Random-Liu: Why ignore errors here and above?

pkg/config/config.go
```diff
@@ -142,6 +142,16 @@ type PluginConfig struct {
 	// Log line longer than the limit will be split into multiple lines. Non-positive
 	// value means no limit.
 	MaxContainerLogLineSize int `toml:"max_container_log_line_size" json:"maxContainerLogSize"`
+	// NoCgroup indicates to disable the cgroup support. (auto/true/false)
```
Random-Liu: 1) Update https://github.com/containerd/cri/blob/master/docs/config.md;
2) Add explanation for "auto".

Please do this for all 3 options.

pkg/config/config.go
```diff
@@ -142,6 +142,16 @@ type PluginConfig struct {
...
+	// This is useful when the daemon does not have permission to access cgroups.
+	NoCgroup string `toml:"no_cgroup" json:"noCgroup"`
```
Random-Liu: s/NoCgroup/DisableCgroup.

We use `enable_xxx` for the other options, let's follow the same pattern.

pkg/config/config.go
```diff
@@ -142,6 +142,16 @@ type PluginConfig struct {
...
+	// NoApparmor indicates to disable the apparmor support. (auto/true/false)
+	// This is useful when the daemon does not have permission to access Apparmor.
+	NoApparmor string `toml:"no_apparmor" json:"noApparmor"`
```
Random-Liu: DisableApparmor.

pkg/server/container_create.go
```diff
@@ -416,12 +416,14 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
...
-		cgroupsPath := getCgroupsPath(sandboxConfig.GetLinux().GetCgroupParent(), id,
-			c.config.SystemdCgroup)
-		g.SetLinuxCgroupsPath(cgroupsPath)
+	if !c.noCgroup {
+		if sandboxConfig.GetLinux().GetCgroupParent() != "" {
+			cgroupsPath := getCgroupsPath(sandboxConfig.GetLinux().GetCgroupParent(), id,
+				c.config.SystemdCgroup)
+			g.SetLinuxCgroupsPath(cgroupsPath)
+		}
 	}
```
Random-Liu: else, clear the cgroup path just in case there is default value.

pkg/server/container_create.go
```diff
@@ -416,12 +416,14 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
```
Random-Liu: Add unit test cases for this logic.

pkg/server/helpers.go
```diff
@@ -496,3 +499,89 @@ func getRuntimeOptions(c containers.Container) (interface{}, error) {
...
+	}
+}
+
+// parseAutoOrBool returns nil for "auto" and "", *bool for bool string.
+func parseAutoOrBool(v string) (*bool, error) {
+	b, err := strconv.ParseBool(v)
+	if err == nil {
+		return &b, nil
+	}
+	if v == "" || v == "auto" {
```
Random-Liu: > Suggest: using a string for a rootless option, "auto" should be default.

@mikebrow Let's not change the default behavior... It is dangerous to me.
You never know what people have been doing.

mikebrow: if true we might need to change runc.. I think they used auto there?

mikebrow: more thought is needed I think.

pkg/server/helpers.go
```diff
@@ -496,3 +499,89 @@ func getRuntimeOptions(c containers.Container) (interface{}, error) {
...
+// parseNoCgroup parses NoCgroup.
+func parseNoCgroup(v string) (bool, error) {
+	p, err := parseAutoOrBool(v)
+	if err != nil {
+		return false, err
+	}
+	if p != nil {
+		return *p, nil
+	}
+	b := rsystem.RunningInUserNS()
```
Random-Liu: Do something like https://github.com/containerd/containerd/blob/master/archive/tar_unix.go#L92

pkg/server/helpers.go
Random-Liu: Today's behavior is that we always enable cgroup, no restrict on OOMScore.

The default config should behave that way.

AkihiroSuda: Can we remove `auto` and revert the type to bool?

Random-Liu: > Can we remove auto and revert the type to bool?

I'm fine with either, as long as the default behavior is not affected.

Actually, I prefer only making the type bool, but error out with clear message if containerd is in user namespace, and unsupported options are enabled, e.g. cgroups, apparmor etc.

mikebrow: that works too! and covers the reason I figured we needed auto mode.

Random-Liu: > Actually, I prefer only making the type bool, but error out with clear message if containerd is in user namespace, and unsupported options are enabled, e.g. cgroups, apparmor etc. Instead of silently trying to do something smart.

Hm, I found that in this way, the default config will be kept the same, which is good and clear.

However, the default behavior is still changed:
* Before: Containerd in user namespace won't return error directly when apparmor or cgroup is enabled;
* Now: Containerd in user namespace returns error in that case.

However, if containerd in user namespace doesn't support those features, even in the previous behavior we'll eventually get an error. So it seems fine to me.

pkg/server/container_create.go
```diff
@@ -744,17 +746,25 @@ func setOCIBindMountsPrivileged(g *generate.Generator) {
...
+		g.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
+		g.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
+		g.SetLinuxResourcesMemoryLimit(resources.GetMemoryLimitInBytes())
+		g.SetLinuxResourcesCPUCpus(resources.GetCpusetCpus())
+		g.SetLinuxResourcesCPUMems(resources.GetCpusetMems())
+	}
 	g.SetProcessOOMScoreAdj(int(resources.GetOomScoreAdj()))
-	g.SetLinuxResourcesCPUCpus(resources.GetCpusetCpus())
-	g.SetLinuxResourcesCPUMems(resources.GetCpusetMems())
+	if restrictOOMScoreAdjFlag {
```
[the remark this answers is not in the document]
AkihiroSuda: restrictOOMScoreAdjFlag  depends on the result of SetProcessOOMScoreAdj

pkg/server/helpers.go
AkihiroSuda: Updated to print a warning, not an error (because even in userns cgroups might be available especially when running in LXD or with pam_cgfs.so)

pkg/server/container_create.go
AkihiroSuda: Can we cover this in an integrated test (in a follow-up PR)?

AkihiroSuda:
updated

AkihiroSuda:
thanks, updated

Issue #979, opened 2018-11-22
CNI config 


mikebrow:
We should add the CNI configuration details to the runtime.StatusResponse when verbose is enabled. 

Will help us .. and customers debug issues like this one:
https://github.com/containerd/containerd/issues/2804

mikebrow:
on second thought let's do this on inspectp / podsandboxstatus, since each pod may have it's own network configuration.

Random-Liu:
> We should add the CNI configuration details to the runtime.StatusResponse when verbose is enabled.

This is also useful. Let's also do it. At least, we should list which exact CNI config file containerd is using.

mikebrow:
i like that plan will push a commit if you don't beat me to it :)

Random-Liu:
@mikebrow Then do it. :)

abhi:
well go-cni config function will return the config at that moment. There is no guarantee that the same config was used for previously created  (and running) pods :)

mikebrow:
yeah was going to expose the one used to attach.. with a new api from the go-cni api before we garbage collect it .. then stow it away .. sigh

mikebrow:
Yeah .. what random said was already there cloning go-cni.. :)

mikebrow:
@Random-Liu interesting so you think paths to configs and conflist.. hmm

mikebrow:
Here's the new crictl inspectp output from a containerd/cri commit that I'll push when https://github.com/containerd/go-cni/pull/33 gets merged:

```
mike@mike-VirtualBox:~/go/src/crictl-test$ sudo crictl inspectp 46c
{
  "status": {
    "id": "46c14cc26e5f8c90de8b21d64b4f26bbe342f3d4ac0319cd2771dd6ff127abea",
    "metadata": {
      "attempt": 1,
      "name": "nginx-sandbox",
      "namespace": "default",
      "uid": "hdishd83djaidwnduwk28bcsb"
    },
    "state": "SANDBOX_READY",
    "createdAt": "2018-11-30T12:40:24.492124152-06:00",
    "network": {
      "ip": "10.88.10.229"
    },
    "linux": {
      "namespaces": {
        "options": {
          "ipc": "POD",
          "network": "POD",
          "pid": "POD"
        }
      }
    },
    "labels": {},
    "annotations": {}
  },
  "info": {
    "pid": 22045,
    "processStatus": "running",
    "netNamespaceClosed": false,
    "image": "gcr.io/google_containers/pause:3.1",
[...]

PR #980, opened 2018-11-26
adds cni results to verbose pod info

mikebrow:
Address issue #979 while we can ask them to re-run with debug logging turned on, it should be very useful to have the ability to inspect each pod's cni load results from users (clients) of containerd/cri. 

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -126,7 +126,7 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		// In this case however caching the IP will add a subtle performance enhancement by avoiding
 		// calls to network namespace of the pod to query the IP of the veth interface on every
 		// SandboxStatus request.
-		sandbox.IP, err = c.setupPod(id, sandbox.NetNSPath, config)
+		sandbox.IP, sandbox.CNIResult, err = c.setupPod(id, sandbox.NetNSPath, config)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to setup network for sandbox %q", id)
 		}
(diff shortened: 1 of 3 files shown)


mikebrow:
> I am assuming this will only be helpful in error cases right ? But on setupPod failure we will end up deleting the sandbox so pod info will not be there right ?
If not then 👍 LGTM

correct.. setupPod failures will continue to report back failure and the pod info will be lost. One would need to enable debug logging to get the failure results back. This pr will only be helpful in situations where there are networking failures after the network is successfully attached to the pod network namespace and the admin/developer is trying to debug his network config (check if the pod has the network configuration he expected.)

mikebrow:
It would be nice to have a network focused test validation suite that people could run to find configuration errors. crictl validate_network(for) pod.yaml

abhi:
Ok makes sense. But If things don't work after network setup ,I would expect to check the states in the CNI plugin. However I think this makes sense to keep as well. But may be short term. CNI is adding CHECK/GET for fetching the expected network states for an endpoint. So we could use that eventually.  And may be tie it to validation as well.

mikebrow:
> Ok makes sense. But If things don't work after network setup ,I would expect to check the states in the CNI plugin. However I think this makes sense to keep as well. But may be short term. CNI is adding CHECK/GET for fetching the expected network states for an endpoint. So we could use that eventually. And may be tie it to validation as well.

Agreed, on all points. I was surprised at the lack of a check/get in the current api.. I was looking for it when I saw that all I had access to was the setup results... Good to hear they are adding that.

Random-Liu:
This looks useful to me :+1: 
```
"cniResult": {
      "Interfaces": {
        "cnio0": {
          "IPConfigs": null,
          "Mac": "86:42:c2:29:96:14",
          "Sandbox": ""
        },
        "eth0": {
          "IPConfigs": [
            {
              "IP": "10.244.0.4",
              "Gateway": "10.244.0.1"
            }
          ],
          "Mac": "",
          "Sandbox": "/var/run/netns/cni-4a83253f-b80e-37e4-f205-5e64ac601a76"
        },
        "veth8906bfb0": {
          "IPConfigs": null,
          "Mac": "86:42:c2:29:96:14",
          "Sandbox": ""
        }
      },
      "DNS": [
        {},
        {}
      ],
      "Routes": [
        {
          "dst": "0.0.0.0/0"
        }
      ]
    }
```

/lgtm

PR #983, opened 2018-12-04
Fix kill when shared pid namespace.

Random-Liu:
For https://github.com/containerd/containerd/issues/2744.

Only set `KillAll` in `StopContainer` and `StopPodSandbox` is not enough, because:
1) The `Delete` in the event monitor will be first, if the init process exits with SIGTERM.
2) For containers which exit themselves, only `Delete` in the event monitor will be called.

This PR changes the event monitor to always delete task with `WithProcessKill`, this is equivalent to containerd behavior in 1.1 that a `runc kill --all` is always called after init process exits https://github.com/containerd/containerd/blob/release/1.1/linux/shim/service.go#L470.

Because the event monitor always delete task with `WithProcessKill` now, no need to do that in `StopContainer` and `StopPodSandbox` any more.


pkg/server/sandbox_stop.go
```diff
@@ -105,18 +104,8 @@ func (c *criService) stopSandboxContainer(ctx context.Context, sandbox sandboxst
...
-		return errors.Wrap(err, "failed to kill sandbox container")
-	}
-
-	if err = c.waitSandboxStop(ctx, sandbox, killContainerTimeout); err == nil {
-		return nil
-	}
-	logrus.WithError(err).Errorf("An error occurs during waiting for sandbox %q to be killed", sandbox.ID)
-
-	// Kill the sandbox container init process.
 	if err = task.Kill(ctx, unix.SIGKILL); err != nil && !errdefs.IsNotFound(err) {
```
mikebrow: did you mean to do a withkillall here?

mikebrow: I'm a bit nervous that we didn't check if we are stopped before sending the final kill (without an all option)

pkg/server/container_stop.go
```diff
@@ -105,36 +104,15 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
...
-	// cgroups, and kill them. However, sometimes the processes may be moved
-	// out from the container cgroup, e.g. users manually move them by mistake,
-	// or systemd.Delegate=true is not set.
-	// In these cases, we should try our best to do cleanup, kill the container
-	// without containerd.WithKillAll, so that runc can kill the container init
-	// process directly.
-	// NOTE(random-liu): If pid namespace is shared inside the pod, non-init processes
-	// of this container will be left running until the pause container is stopped.
-	logrus.Infof("Kill container %q init process", id)
 	if err = task.Kill(ctx, unix.SIGKILL); err != nil && !errdefs.IsNotFound(err) {
```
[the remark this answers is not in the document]
Random-Liu: `task.Delete(containerd.WithProcessKill)` in the event monitor will do the `KillAll` now.

pkg/server/sandbox_stop.go
Random-Liu: ditto.

PR #989, opened 2018-12-05
Test that CRI plugin should not see image in other namespaces.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/986.

### DIFF
--- a/integration/containerd_image_test.go
+++ b/integration/containerd_image_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/namespaces"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


integration/containerd_image_test.go
```diff
@@ -146,3 +147,38 @@ func TestContainerdImage(t *testing.T) {
...
+	t.Logf("cri plugin should not see the image")
+	checkImage := func() (bool, error) {
+		img, err := imageService.ImageStatus(&runtime.ImageSpec{Image: testImage})
+		if err != nil {
+			return false, err
+		}
+		return img == nil, nil
+	}
+	require.NoError(t, Consistently(checkImage, 100*time.Millisecond, time.Second))
+}
```
mikebrow: should probably add two tests .. 1) make sure we can also pull for our namespace and then see it .. 2) then delete one and make sure it's not visible again but still visible from the ns where it hasn't been deleted.

Random-Liu: 1) That is tested in the other test case `TestContainerdImage`;
2) We can add a case that image deleted in another namespace won't affect cri plugin. Will do.

Issue #990, opened 2018-12-06
Container life cycle should not rely on the existence of other image references.


Random-Liu:
Containerd manages images based on references. Image names (tag, id etc.), container references are all valid independent references.

They are independent, means that you can remove all image references, and the container can still run happily because the snapshots are still referenced. This also means that **we can't assume image references (and the content) exist across the whole life cycle of a container using the image**.

However, we are doing so in some container operations:
* ContainerStop: https://github.com/containerd/cri/blob/master/pkg/server/container_stop.go#L80
* ContainerStatus: https://github.com/containerd/cri/blob/master/pkg/server/container_status.go#L44

This causes issue like https://github.com/containerd/containerd/issues/2858.

There are 2 options to fix this:
* **Option 1**: Add logic in the cri plugin to avoid deleting image references when container is still using an image.
  * Problem 1: This is racy. It is always possible that after you check and make sure that an image is not in use, a new container is created referencing the image. Need to add synchronization for this case.
  * Problem 2: Users can always delete image references by talking with containerd directly. We don't have any control to that.
* **Option 2**: Always assume that image reference may be gone after container creation, and never rely on it in the container lifecycle.

Given the problem of option 1, I prefer option 2.

For option 2:
1) [For `StopContainer`, only `StopSignal` is needed](https://github.com/containerd/cri/blob/master/pkg/server/container_stop.go#L80), we should add it into container metadata during container creation; 
2) [For `ContainerStatus`, the image `id` to `tag` conversion is best effort anyway,](https://github.com/containerd/cri/blob/master/pkg/server/container_status.go#L46) just don't return error.

Note that adding `StopSignal` into metadata is adding a new field into our metadata checkpoint. We should handle in place upgrade properly.

mikebrow:
agree option 2

Random-Liu:
At a second thought, I think we need both option 1 and 2.

We need option 1, because that is Docker's behavior, I think people expect that in `crictl`. And even for Kubernetes garbage collection, it is an important piece to avoid race condition:
1) Kubelet compares container and image list, and find an unused image;
2) New container created using the image;
3) Kubelet garbage collects the image.

Currently, for Docker, 3) will fail; but for containerd 3) will remove the image.

mikebrow:
ok, thx for investigating! agree option 1 with tolerance for option 2.

mikebrow:
suggest adding an issue to enable the garbage collect of the in use image.. (breaking the bad employed exhibited by docker)

PR #991, opened 2018-12-06
Remove container lifecycle image dependency

Random-Liu:
Fixes https://github.com/containerd/cri/issues/990.

Test result without the fix:
```
--- FAIL: TestContainerLifecycleWithoutImageRef (1.66s)
    container_without_image_ref_test.go:29: Create a sandbox
    container_without_image_ref_test.go:42: Pull test image
    container_without_image_ref_test.go:49: Create test container
    container_without_image_ref_test.go:59: Remove test image
    container_without_image_ref_test.go:62: Container status should be running
        Error Trace:    container_without_image_ref_test.go:64
        Error:		Received unexpected error rpc error: code = Unknown desc = failed to get image "sha256:59788edf1f3e78cd0ebe6ce1446e9d10788225db3dedcfd1a59f764bad2b2690": does not exist
```

Test result with the fix:
```
=== RUN   TestContainerLifecycleWithoutImageRef
--- PASS: TestContainerLifecycleWithoutImageRef (1.74s)
    container_without_image_ref_test.go:29: Create a sandbox
    container_without_image_ref_test.go:42: Pull test image
    container_without_image_ref_test.go:49: Create test container
    container_without_image_ref_test.go:59: Remove test image
    container_without_image_ref_test.go:62: Container status should be running
    container_without_image_ref_test.go:67: Stop container
    container_without_image_ref_test.go:71: Container status should be exited
PASS
```

### DIFF
--- a/pkg/server/container_status.go
+++ b/pkg/server/container_status.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/net/context"
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 
+	"github.com/containerd/cri/pkg/store"
 	containerstore "github.com/containerd/cri/pkg/store/container"
 )
 
@@ -43,17 +44,20 @@ func (c *criService) ContainerStatus(ctx context.Context, r *runtime.ContainerSt
 	imageRef := container.ImageRef
 	image, err := c.imageStore.Get(imageRef)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get image %q", imageRef)
-	}
-	repoTags, repoDigests := parseImageReferences(image.References)
-	if len(repoTags) > 0 {
-		// Based on current behavior of dockershim, this field should be
-		// image tag.
-		spec = &runtime.ImageSpec{Image: repoTags[0]}
-	}
-	if len(repoDigests) > 0 {
-		// Based on the CRI definition, this field will be consumed by user.
-		imageRef = repoDigests[0]
+		if err != store.ErrNotExist {
+			return nil, errors.Wrapf(err, "failed to get image %q", imageRef)
+		}
+	} else {
+		repoTags, repoDigests := parseImageReferences(image.References)
+		if len(repoTags) > 0 {
+			// Based on current behavior of dockershim, this field should be
+			// image tag.
+			spec = &runtime.ImageSpec{Image: repoTags[0]}
+		}
+		if len(repoDigests) > 0 {
+			// Based on the CRI definition, this field will be consumed by user.
+			imageRef = repoDigests[0]
+		}
 	}
 	status := toCRIContainerStatus(container, spec, imageRef)
 	info, err := toCRIContainerInfo(ctx, container, r.GetVerbose())
--- a/pkg/store/container/container_test.go
+++ b/pkg/store/container/container_test.go
@@ -39,8 +39,9 @@ func TestContainerStore(t *testing.T) {
 					Attempt: 1,
 				},
 			},
-			ImageRef: "TestImage-1",
-			LogPath:  "/test/log/path/1",
+			ImageRef:   "TestImage-1",
+			StopSignal: "SIGTERM",
+			LogPath:    "/test/log/path/1",
 		},
 		"2abcd": {
 			ID:        "2abcd",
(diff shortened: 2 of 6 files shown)


pkg/store/container/metadata.go
```diff
@@ -56,6 +56,8 @@ type Metadata struct {
 	Config *runtime.ContainerConfig
 	// ImageRef is the reference of image used by the container.
 	ImageRef string
+	// StopSignal is the signal to send to stop the container.
+	StopSignal string
```
mikebrow: /s/StopSignal/Signal/

mikebrow: out of habit can we add new stuff to the end of the struct?

pkg/store/container/container_test.go
```diff
@@ -78,7 +81,8 @@ func TestContainerStore(t *testing.T) {
 					Attempt: 1,
 				},
 			},
-			ImageRef: "TestImage-4abcd",
```
mikebrow: Needs a test that does not have a StopSignal.

Needs tests or TODO + issue for tests that use different signals.

OCI spec:
```
StopSignal string, OPTIONAL

The field contains the system call signal that will be sent to the container to exit. 
The signal can be a signal name in the format SIGNAME, for instance SIGKILL or SIGRTMIN+3.
```

Random-Liu:
It is StopSignal. It matches the name in pic image config too

pkg/server/container_status.go
```diff
@@ -43,17 +44,20 @@ func (c *criService) ContainerStatus(ctx context.Context, r *runtime.ContainerSt
...
-		// Based on current behavior of dockershim, this field should be
-		// image tag.
-		spec = &runtime.ImageSpec{Image: repoTags[0]}
-	}
-	if len(repoDigests) > 0 {
-		// Based on the CRI definition, this field will be consumed by user.
-		imageRef = repoDigests[0]
+		if err != store.ErrNotExist {
+			return nil, errors.Wrapf(err, "failed to get image %q", imageRef)
+		}
```
mikebrow: Let's cache the image.References in meta, we don't want to rely on it sticking around. Don't like the container status results changing / omitting image data cause the image ref has been removed from the store or worse the image has been updated / removed, but not for the container. Or at least we can add a TODO and open an issue to do so.

mikebrow:
> It is StopSignal

yeah sorry hit the wrong button while reviewing I deleted that comment before finishing the review :-)

pkg/store/container/metadata.go
Random-Liu: Good point, will do.

pkg/store/container/container_test.go
Random-Liu: Will add an integration test for this.

pkg/server/container_status.go
Random-Liu: `image.References` is dynamically changed. I don't think we can cache it in container meta.

mikebrow: that's sort of my point, if the image that was used "changes" it's not what was used for the snapshot, thus the status information is not correct. I don't want to use the cache for anything but the status output.

image.References is just an array of strings.

Random-Liu: Are you ok with https://github.com/containerd/cri/issues/990#issuecomment-445323970?

pkg/store/container/metadata.go
```diff
@@ -56,6 +56,8 @@ type Metadata struct {
 	Config *runtime.ContainerConfig
 	// ImageRef is the reference of image used by the container.
 	ImageRef string
+	// StopSignal is the signal to send to stop the container.
```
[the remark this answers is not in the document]
Random-Liu: Done

pkg/store/container/metadata.go
Random-Liu: Done

pkg/store/container/container_test.go
Random-Liu: Added TODO. Will send a follow up PR.

pkg/server/container_status.go
mikebrow: yes, see responses on #990 

I still think we should cache the image.references at snapshot

mikebrow: nvm I see we do cache the original imageref.. that should be enough

Random-Liu: Are you OK with the change now? :)

PR #994, opened 2018-12-12
Do not remove image when it is being used by a container

Random-Liu:
For https://github.com/containerd/cri/issues/990.

This PR:
1) Added a check in image_remove.go to make sure image can only be removed when it is not used by a container;
2) Added a `leases` field in `Image` to avoid the race condition between the check and container creation.

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -36,7 +36,8 @@ test_setup ${REPORT_DIR}
 sudo PATH=${PATH} ${ROOT}/_output/integration.test --test.run="${FOCUS}" --test.v \
   --cri-endpoint=${CONTAINERD_SOCK} \
   --cri-root=${CRI_ROOT} \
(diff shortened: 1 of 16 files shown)


mikebrow:
@Random-Liu says no merge conflicts but maybe we should rebase and retest after the recent changes?

Random-Liu:
@mikebrow Rebased.

Note that this PR also changes the test containerd binary to be `containerd-test`, so that we can easily restart containerd in the test without breaking Docker 18.09+ on the node.

Random-Liu:
Hm, I'm really not sure whether we need this...

Let's wait for our customers' feedback. If removing image when containers exist cause issue in the future, we can always get back to this...

dmcgowan:
The master branch on this repository has been merged into [containerd/containerd](https://github.com/containerd/containerd). If we still need this, we can add it back there.

PR #995, opened 2018-12-14
Cherrypick #991 release 1.0

Random-Liu:


Random-Liu:
The failure is caused by a golang issue. https://github.com/golang/go/issues/29273

PR #1000, opened 2018-12-14
Teardown sandbox network after stop.

Random-Liu:
Fixes https://github.com/containerd/containerd/issues/2546.

I thought about this a bit.

I don't think reversing the order will cause any problem. Previously, I was worried that we remove the sandbox when the sandbox container is not running but network is not tore-down yet. So we teardown the network first then stop the sandbox container.

However, now we also checks network namespace before sandbox removal, it should be fine to reverse the order https://github.com/containerd/cri/blob/master/pkg/server/sandbox_remove.go#L57

### DIFF
--- a/pkg/server/sandbox_stop.go
+++ b/pkg/server/sandbox_stop.go
@@ -42,9 +42,8 @@ func (c *criService) StopPodSandbox(ctx context.Context, r *runtime.StopPodSandb
 	id := sandbox.ID
 
 	// Stop all containers inside the sandbox. This terminates the container forcibly,
-	// and container may still be so production should not rely on this behavior.
-	// TODO(random-liu): Delete the sandbox container before this after permanent network namespace
-	// is introduced, so that no container will be started after that.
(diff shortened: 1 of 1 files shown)


jterry75:
LGTM (not a maintainer)

AkihiroSuda:
What's current status?

mikebrow:
> What's current status?
Since this changes the timing .. I felt there was a risk if we didn't add a state flag for sandbox stop to halt new container creates, "TODO(random-liu): Introduce a state in sandbox to avoid future container creation."

After this:
https://github.com/containerd/cri/blob/master/pkg/server/container_create.go#L80

We'd check if sandbox.stopping... and return vs the possible error condition of creating a container after we've stopped all the containers in the sandbox.

Random-Liu:
Actually the risk is pretty low, because kubelet never does it. Kubelet
always stop all containers before tear down the sandbox. And it would never
create new containers at the same time.

Random-Liu:
My concern is for manual operation, which is relatively low priority.

PR #1004, opened 2018-12-19
Install libseccomp2 based on debian version

Random-Liu:
In Kubernetes test infra, we have both debian:jessie and debian:stretch test images.

We need to support both until the switch is completed.

/cc @bentheelder


Random-Liu:
s/libseccomp2-dev/libseccomp dev.

Reapply lgtm based on https://github.com/containerd/cri/pull/1004#pullrequestreview-186714761

Issue #1009, opened 2018-12-25
pod or container config file without metadata will crash containerd


teawater:
```
root@teawater-virtual-machine:~/vmshare/m# cat p.json
{
  "log_directory": "/tmp",
  "linux": {
  }
}
root@teawater-virtual-machine:~/vmshare/m# crictl runp p.json
DEBU[0000] RunPodSandboxRequest: &RunPodSandboxRequest{Config:&PodSandboxConfig{Metadata:nil,Hostname:,LogDirectory:/tmp,DnsConfig:nil,PortMappings:[],Labels:map[string]string{},Annotations:map[string]string{},Linux:&LinuxPodSandboxConfig{CgroupParent:,SecurityContext:nil,Sysctls:map[string]string{},},},}
2018/12/25 13:29:40 transport: http2Client.notifyError got notified that the client transport was broken EOF.
2018/12/25 13:29:40 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial unix /var/run/containerd/containerd.sock: connect: connection refused"; Reconnecting to {/var/run/containerd/containerd.sock <nil>}
DEBU[0000] RunPodSandboxResponse: nil
FATA[0000] run pod sandbox failed: rpc error: code = Internal desc = transport is closing

INFO[2018-12-25T13:29:40.769578744+08:00] RunPodSandbox with config &PodSandboxConfig{Metadata:nil,Hostname:,LogDirectory:/tmp,DnsConfig:nil,PortMappings:[],Labels:map[string]string{},Annotations:map[string]string{},Linux:&LinuxPodSandboxConfig{CgroupParent:,SecurityContext:nil,Sysctls:map[string]string{},},}
INFO[2018-12-25T13:29:40.769784271+08:00] RunPodSandbox for nil returns sandbox id ""
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x55e8c49f6c21]

goroutine 53 [running]:
github.com/containerd/containerd/vendor/github.com/containerd/cri/pkg/server.makeSandboxName(0x0, 0x40, 0x101000000000073)
	/home/teawater/gopath/src/github.com/containerd/containerd/vendor/github.com/containerd/cri/pkg/server/helpers.go:143 +0x41
[...]

PR #1010, opened 2018-12-25
Fix the issue that pod or container config file without metadata will…

teawater:
… crash containerd

Because RunPodSandbox and CreateContainer will access metadata
without check, pod or container config file without metadata will
crash containerd.

This patch add checks to handle the issue.

Fixes: #1009

Signed-off-by: Hui Zhu <teawater@hyper.sh>

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -92,7 +92,11 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	// Reserve the container name to avoid concurrent `CreateContainer` request creating
 	// the same container.
 	id := util.GenerateID()
-	name := makeContainerName(config.GetMetadata(), sandboxConfig.GetMetadata())
+	metadata := config.GetMetadata()
+	if metadata == nil {
+		return nil, errors.New("container config must include metadata")
+	}
+	name := makeContainerName(metadata, sandboxConfig.GetMetadata())
 	logrus.Debugf("Generated id %q for container %q", id, name)
 	if err = c.containerNameIndex.Reserve(name, id); err != nil {
 		return nil, errors.Wrapf(err, "failed to reserve container name %q", name)
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -58,7 +58,11 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 
 	// Generate unique id and name for the sandbox and reserve the name.
 	id := util.GenerateID()
-	name := makeSandboxName(config.GetMetadata())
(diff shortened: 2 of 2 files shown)


pkg/server/container_create.go
```diff
@@ -92,7 +92,11 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	// Reserve the container name to avoid concurrent `CreateContainer` request creating
 	// the same container.
 	id := util.GenerateID()
-	name := makeContainerName(config.GetMetadata(), sandboxConfig.GetMetadata())
+	metadata := config.GetMetadata()
+	if metadata == nil {
+		return nil, errors.New("GetMetadata failed")
```
AkihiroSuda: nit: Go error should start with a lower character, e.g. `errors.New("failed to get metadata")`

mikebrow: /s/GetMetadata failed/container config must include metadata/

add a TODO to make container create work using image name if metadata name is not set..

pkg/server/sandbox_run.go
```diff
@@ -58,7 +58,11 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 
 	// Generate unique id and name for the sandbox and reserve the name.
 	id := util.GenerateID()
-	name := makeSandboxName(config.GetMetadata())
+	metadata := config.GetMetadata()
+	if metadata == nil {
+		return nil, errors.New("GetMetadata failed")
```
mikebrow: /s/GetMetadata failed/sandbox config must include metadata/

pkg/server/container_create.go
```diff
@@ -91,8 +91,14 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	// Generate unique id and name for the container and reserve the name.
 	// Reserve the container name to avoid concurrent `CreateContainer` request creating
 	// the same container.
+	// TODO(teawater): make container create work using image name if metadata name
```
Random-Liu: 1) I'm not sure whether we'd like to do that.
2) Even if we'd like to do that, we should do it in crictl instead of the cri plugin.

Random-Liu:
Sorry for not mentioning this in the last pass.

Please add unit test cases for nil metadata. Just copy an existing unit test, change the metadata to nil, and assert an error instead of panic.

Random-Liu:
nvm, we don't have unit test for `CreateContainer` and `RunPodSandbox`. Can you add a simple integration test for it? An integration test example: https://github.com/containerd/cri/blob/master/integration/container_log_test.go

Random-Liu:
Let's get this in first. I'm fine with adding the integration test in another PR.

PR #1023, opened 2019-01-16
fix private docker registry configuration

guoruijing:
Signed-off-by: Guo Ruijing <ruijing.guo@intel.com>

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -7,11 +7,11 @@ With containerd, `docker.io` is the default image registry. You can also set up 
 To configure image registries create/modify the `/etc/containerd/config.toml` as follows:
 ```toml
 [plugins.cri.registry.mirrors]
-  [plugins.cri.registry.mirrors."docker.io"]
+  [plugins.cri.registry.mirrors."registry-1.docker.io"]
     endpoint = ["https://registry-1.docker.io"]
-  [plugins.cri.registry.mirrors."test.secure-registry.io"]
+  [plugins.cri.registry.mirrors."HostIP1:Port1"]
     endpoint = ["https://HostIP1:Port1"]
-  [plugins.cri.registry.mirrors."test.insecure-registry.io"]
+  [plugins.cri.registry.mirrors."HostIP2:Port2"]
     endpoint = ["http://HostIP2:Port2"]
 ```
 


mikebrow:
@Random-Liu we're getting hit by the log problem mentioned in the sig-node meeting.

Random-Liu:
@mikebrow Looking into this right now.

Random-Liu:
@guoruijing I don't quite understand your change. The original document seems correct to me.

Can you add some explanation?

guoruijing:
example with issue:
1.  config in /etc/containerd/config.toml
    [plugins.cri.registry]
      [plugins.cri.registry.mirrors]
        [plugins.cri.registry.mirrors."registry"]
          endpoint = ["http://registry:5000"]
2. restart containerd
3. pull images:
vagrant@worker1:~$ sudo crictl pull registry:5000/multus-cni
FATA[0000] pulling image failed: rpc error: code = Unknown desc = failed to resolve image "registry:5000/multus-cni:latest": no available registry endpoint: failed to do request: Head https://registry:5000/v2/multus-cni/manifests/latest: http: server gave HTTP response to HTTPS client

guoruijing:
I don't know how to build cri now and I remember the issue in pkg/server/image_pull.go

func (c *criService) getResolver(ctx context.Context, ref string, cred func(string) (string, string, error)) (remotes.Resolver, imagespec.Descriptor, error) {
        refspec, err := reference.Parse(ref)
        if err != nil {
                return nil, imagespec.Descriptor{}, errors.Wrap(err, "parse image reference")
        }
        // Try mirrors in order first, and then try default host name.
        for _, e := range c.config.Registry.Mirrors[refspec.Hostname()].Endpoints {

refspec.Hostname also include hostname & port. if configuration doesn't include port, then it mismatch and cannot find docker registry.

Random-Liu:
@guoruijing I feel like with your config, you should pull `registry/multus-cni` instead. And containerd will know it should access `http://registry:5000` based on your registry mirror config.

Random-Liu:
@guoruijing Do you still need anything?

Random-Liu:
I'll close this for now. Feel free to reopen if you still want it. :)

PR #1028, opened 2019-01-24
hack/Makefile: detect dirty vendor

fuweid:
use containerd/project script to detect dirty vendor.

close: #1012

### DIFF
--- a/hack/verify-vendor.sh
+++ b/hack/verify-vendor.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+#
+#   Copyright The containerd Authors.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


fuweid:
the k8s-ci workflow fails because of linter issue. wait for #1027.

Random-Liu:
@fuweid Can we add a `hack/verify-vendor.sh` and call it in `hack/verify.sh`.

I want this to be part of `pull-cri-containerd-verify` job, not travis job. :)

hack/verify-vendor.sh
```diff
@@ -0,0 +1,26 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+tmpdir="$(mktemp -d)"
+trap "rm -rf ${tmpdir}" EXIT
+
+git clone "https://github.com/containerd/project" "${tmpdir}"
+"${tmpdir}"/script/validate/vendor
+unset tmpdir
```
Random-Liu: If you unset `tmpdir` here, does trap still work?

Random-Liu: Hm, it still works, interesting.

We don't need to unset it actually, but it works anyway. :)

fuweid: :) I use `"` instead of `'`. I can remove the `unset` for the case, since it confuses.

PR #1029, opened 2019-01-24
Add a generic runtime options api.

Random-Liu:
This is an implementation based on discussion in https://github.com/containerd/containerd/pull/2916.

/cc @lifupan @jterry75 @crosbymichael


crosbymichael:
LGTM

jterry75:
A few questions:
1. I understand this needs to be staged but will we keep the `runcRuntime`? Seems like all `Runtime V2`'s should use this method.
2. Should CRI read `config_path` and forward all bytes as `proto.Any` to a `Runtime V2` or should a `Runtime V2` expect to receive this type and read `config_path` itself? If we read in CRI then each runtime still gets their specific opts and no changes are required in each runtime correct?

Random-Liu:
@jterry75 
1. The new runc option is already there, I think we should still support it. https://github.com/containerd/containerd/blob/master/runtime/v2/runc/options/oci.proto The generic option is for those who don't have in-tree option support.
2. The Value in `proto.Any` *must be a valid serialized protocol buffer of the above specified type* (see https://godoc.org/github.com/golang/protobuf/ptypes/any). This means that either CRI has to understand the content in the config file and serialize it with protobuf, or we'll need to have protobuf (binary format) inside the config file (not readable at all).

However, the runtime understands the config content, and knows how to use it.

jterry75:
@Random-Liu - I guess my point then is, a Runtime V2 shim (except `runc`) MUST accept `github.com/containerd/cri/pkg/runtimeoptions.Options` as its options type in order to work with CRI correct?

Random-Liu:
@jterry75 If the runtime v2 shim doesn't have in-tree options defined, they have to use `github.com/containerd/cri/pkg/runtimeoptions.Options`. If they have options pre-defined in tree, e.g. runc, runhcs etc., they can continue using the options they defined in-tree, and we'll support it in the `cri` plugin.

Random-Liu:
@jterry75 I'll wait for your approval. :)

crosbymichael:
Oh, hold on.  One quick idea!

What if we did define this in containerd where the options are part of this struct like so:

```proto
message Option {
string type = 1;
string config_path = 2;
protobuf.Any value = 3;
}
```

Then the runtime could accept this as a wrapper around how we currently provide options?  Just an quick idea after seeing some of the comments here.

Random-Liu:
> Then the runtime could accept this as a wrapper around how we currently provide options?

Sounds like it requires shim v2 api change?

crosbymichael:
@Random-Liu I don't think so, we just pass that `message` as the current Any field, we just have a consistent payload that can then be un-marshaled.  I don't know if it is better or worse.  @jterry75 thoughts?

jterry75:
This seems reasonable. For an orchestrator that knows the downstream runtimes (Docker) they can pass the values as `type, value` and for CRI where the runtimes are configurations via the `RuntimeHandler` and the `config.toml` they can pass it as `type, config_path` and the person setting up the CRI `config.toml` will also setup the appropriate `config.json` (or whatever) paths for each handler in `config.toml`. 

It does mean that to work with CRI though a RuntimeV2 SHOULD accept `CreateTaskRequest.Options` as `github.com/containerd/cri/pkg/runtimeoptions.Options` if they are "out of tree" like @Random-Liu  suggested and do the "one of" `config_path` or `value` like @crosbymichael  suggested.

All of this works for me! 👍

Random-Liu:
> @Random-Liu I don't think so, we just pass that message as the current Any field, we just have a consistent payload that can then be un-marshaled.

It doesn't seem too different with the approach in this PR, but requires all existing v2 shim to change? And the `type` field is duplicated with the `TypeUrl` in `value`. :)

jterry75:
The more I think about it just leave what you have @Random-Liu. I will just check the `TypeUrl` in `Create` and deserialize based on it and support a CRI based `config_path` and a Docker based `runhcsopts` for our shim. This works just fine.

Random-Liu:
@jterry75 Feel free to add `runhcsopts` support in the cri plugin, it is a "in-tree" option which we can support directly in tree.

jterry75:
Sounds good

crosbymichael:
LGTM

lifupan:
As this has been added in containerd/cri by https://github.com/containerd/cri/pull/1029, close this PR.

mikebrow:
> As this has been added in containerd/cri by #1029, close this PR.

@lifupan I think you meant to put this comment in another PR :)

lifupan:
@mikebrow  Aha, yeah, a mistake.

PR #1035, opened 2019-01-28
Add RUNTIME env in integration and cri test.

Random-Liu:
Add `RUNTIME` env in the integration and cri test.
This can be used to test different runtimes.


Random-Liu:
The test failure is caused by https://github.com/kubernetes/test-infra/issues/11001

PR #1042, opened 2019-02-08
Set /etc/hostname.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1041.

/cc @ibuildthecloud

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -233,7 +233,7 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		}
 	}()
 
-	// Setup sandbox /dev/shm, /etc/hosts and /etc/resolv.conf.
+	// Setup sandbox /dev/shm, /etc/hosts, /etc/resolv.conf and /etc/hostname.
 	if err = c.setupSandboxFiles(id, config); err != nil {
 		return nil, errors.Wrapf(err, "failed to setup sandbox files")
 	}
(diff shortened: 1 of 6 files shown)


Random-Liu:

pkg/server/sandbox_run.go
```diff
@@ -456,9 +456,21 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
...
 
-// setupSandboxFiles sets up necessary sandbox files including /dev/shm, /etc/hosts
-// and /etc/resolv.conf.
+// setupSandboxFiles sets up necessary sandbox files including /dev/shm, /etc/hosts,
+// /etc/resolv.conf and /etc/hostname.
 func (c *criService) setupSandboxFiles(id string, config *runtime.PodSandboxConfig) error {
+	sandboxEtcHostname := c.getSandboxHostname(id)
+	if config.GetHostname() == "" {
+		// copy host's /etc/hostname
+		if err := c.os.CopyFile(etcHostname, sandboxEtcHostname, 0644); err != nil {
```
mikebrow: I thought kubelet was ensuring /etc/hostname?? Shouldn't this be an error case?

pkg/server/sandbox_run.go
```diff
@@ -456,9 +456,21 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
 	return g.Config, nil
 }
 
-// setupSandboxFiles sets up necessary sandbox files including /dev/shm, /etc/hosts
-// and /etc/resolv.conf.
+// setupSandboxFiles sets up necessary sandbox files including /dev/shm, /etc/hosts,
+// /etc/resolv.conf and /etc/hostname.
```
mikebrow: we should probably start flagging *nix for content not applicable to other platforms..

pkg/server/sandbox_run.go
Random-Liu: Kubelet doesn't do that.

pkg/server/sandbox_run.go
Random-Liu: We'll do that when the windows support is in.

pkg/server/sandbox_run.go
mikebrow: sent you a txt message to chat over this for a min

mikebrow: yeah I was reading the ensure hosts code in kublet.. got off track.. thx

Random-Liu:
@mikebrow Updated the PR. Some OS doesn't have `/etc/hostname` (COS image used in GKE), so changed the PR to always `WriteFile` with correct hostname instead of copy `/etc/hostname` from the host.

PR #1045, opened 2019-02-12
Fix env performance issue

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1044.

This PR:
1) Remove config info log. It will be too spammy when the config is large, e.g. a lot of envs.
2) Add a map to cache envs in the spec generator.

We may eventually upstream the cache change, but I think we should make local change to avoid delaying release.

With the fix:
```console
$ time crictl create 9 container.json sandbox.json 
5c5bbe3d624bbdfb7030a9909e6f139c38c253ea436da2e8432d9ab02a2da213

real	0m0.298s
user	0m0.083s
sys	0m0.030s
```

Without the fix:
```console
$ time crictl create 9 container.json sandbox.json 
65ade40d8c4b77d6227c0fe36e11c5f3ec4b6cab4a8d76d4135bf4e273ca8544

real	0m41.998s
user	0m0.095s
sys	0m0.009s
```

Test env number:
```console
$ crictl exec 6 sh -c "env | wc -l"
26017
```

### DIFF
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -406,10 +406,17 @@ type generator struct {
 }
 
 func newCustomGenerator(g generate.Generator) generator {
-	return generator{
+	cg := generator{
 		Generator: g,
 		envCache:  make(map[string]int),
 	}
+	if g.Config != nil && g.Config.Process != nil {
+		for i, env := range g.Config.Process.Env {
+			kv := strings.SplitN(env, "=", 2)
+			cg.envCache[kv[0]] = i
+		}
+	}
+	return cg
 }
 
 // AddProcessEnv overrides the original AddProcessEnv. It uses
--- a/pkg/server/helpers.go
+++ b/pkg/server/helpers.go
@@ -391,10 +391,44 @@ func buildLabels(configLabels map[string]string, containerType string) map[strin
 }
 
 // newSpecGenerator creates a new spec generator for the runtime spec.
-func newSpecGenerator(spec *runtimespec.Spec) generate.Generator {
+func newSpecGenerator(spec *runtimespec.Spec) generator {
 	g := generate.NewFromSpec(spec)
 	g.HostSpecific = true
-	return g
+	return newCustomGenerator(g)
+}
+
+// generator is a custom generator with some functions overridden
+// used by the cri plugin.
+// TODO(random-liu): Upstream this fix.
+type generator struct {
+	generate.Generator
+	envCache map[string]int
+}
+
+func newCustomGenerator(g generate.Generator) generator {
+	return generator{
+		Generator: g,
+		envCache:  make(map[string]int),
+	}
+}
+
+// AddProcessEnv overrides the original AddProcessEnv. It uses
+// a map to cache and override envs.
+func (g *generator) AddProcessEnv(key, value string) {
+	if len(g.envCache) == 0 {
+		// Call AddProccessEnv once to initialize the spec.
+		g.Generator.AddProcessEnv(key, value)
+		g.envCache[key] = 0
+		return
+	}
+	spec := g.Config
+	env := fmt.Sprintf("%s=%s", key, value)
+	if idx, ok := g.envCache[key]; !ok {
+		spec.Process.Env = append(spec.Process.Env, env)
+		g.envCache[key] = len(spec.Process.Env) - 1
+	} else {
+		spec.Process.Env[idx] = env
+	}
 }
 
 func getPodCNILabels(id string, config *runtime.PodSandboxConfig) map[string]string {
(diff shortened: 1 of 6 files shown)


pkg/server/instrumented_service.go
```diff
@@ -142,8 +142,8 @@ func (in *instrumentedService) CreateContainer(ctx context.Context, r *runtime.C
 	if err := in.checkInitialized(); err != nil {
 		return nil, err
 	}
-	logrus.Infof("CreateContainer within sandbox %q with container config %+v and sandbox config %+v",
-		r.GetPodSandboxId(), r.GetConfig(), r.GetSandboxConfig())
+	logrus.Infof("CreateContainer within sandbox %q for container %+v",
```
mikebrow: same question

pkg/server/sandbox_run.go
```diff
@@ -149,7 +151,7 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to generate sandbox container spec")
 	}
-	logrus.Debugf("Sandbox container spec: %+v", spec)
+	logrus.Debugf("Sandbox container %q spec: %#+v", id, spew.NewFormatter(spec))
```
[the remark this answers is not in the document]
Random-Liu: Here is the spec output without spew:
```
&{Version:1.0.1-dev Process:0xc0005a6ea0 Root:0xc000b30f00 Hostname: Mounts:[{Destination:/proc Type:proc Source:proc Options:[nosuid noexec nodev]} {Destination:/dev Type:tmpfs Source:tmpfs Options:[nosuid strictatime mode=755 size=65536k]} {Destination:/dev/pts Type:devpts Source:devpts Options:[nosuid noexec newinstance ptmxmode=0666 mode=0620 gid=5]} {Destination:/dev/shm Type:tmpfs Source:shm Options:[nosuid noexec nodev mode=1777 size=65536k]} {Destination:/dev/mqueue Type:mqueue Source:mqueue Options:[nosuid noexec nodev]} {Destination:/sys Type:sysfs Source:sysfs Options:[nosuid noexec nodev ro]} {Destination:/dev/shm Type:bind Source:/run/containerd/io.containerd.grpc.v1.cri/sandboxes/766adb85c75b2665caebf96165d455141e2033cdf3dbb5475dcb32d70929b737/shm Options:[rbind ro]}] Hooks:<nil> Annotations:map[io.kubernetes.cri.container-type:sandbox io.kubernetes.cri.sandbox-id:766adb85c75b2665caebf96165d455141e2033cdf3dbb5475dcb32d70929b737] Linux:0xc000a3b5e0 Solaris:<nil> Windows:<nil> VM:<nil>}
```

Here is the config output without spew:
```
&PodSandboxConfig{Metadata:&PodSandboxMetadata{Name:nginx-sandbox1,Uid:hdishd83djaidwnduwk28bcsb11,Namespace:default1,Attempt:1,},Hostname:,LogDirectory:/tmp,DnsConfig:nil,PortMappings:[],Labels:map[string]string{},Annotations:map[string]string{k1: v1,k2: v2,},Linux:&LinuxPodSandboxConfig{CgroupParent:,SecurityContext:nil,Sysctls:map[string]string{},},}
```

The config shows pointers, but spec not. That's why we use spew for spec.

But it is weird that it works for one but not the other.

pkg/server/instrumented_service.go
```diff
@@ -52,7 +52,7 @@ func (in *instrumentedService) RunPodSandbox(ctx context.Context, r *runtime.Run
 	if err := in.checkInitialized(); err != nil {
 		return nil, err
 	}
-	logrus.Infof("RunPodSandbox with config %+v", r.GetConfig())
+	logrus.Infof("RunPodsandbox for %+v", r.GetConfig().GetMetadata())
```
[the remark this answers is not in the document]
Random-Liu: This is info level log, we still need it.

pkg/server/instrumented_service.go
Random-Liu: Same.

pkg/server/helpers.go
```diff
@@ -391,10 +391,44 @@ func buildLabels(configLabels map[string]string, containerType string) map[strin
...
+func newCustomGenerator(g generate.Generator) generator {
+	return generator{
+		Generator: g,
+		envCache:  make(map[string]int),
+	}
+}
+
+// AddProcessEnv overrides the original AddProcessEnv. It uses
+// a map to cache and override envs.
+func (g *generator) AddProcessEnv(key, value string) {
```
mikebrow: ToDo for RemoveProcessEnv? or is add key, "" good enough

Random-Liu: We don't need remove env for CRI. Always adding. :)

PR #1056, opened 2019-02-23
Add an OCI annotation for sandbox log directory.

Random-Liu:
Add an OCI annotation to pass down the pod log directory.

This is useful for sandbox runtime which has pod level logs, e.g. gvisor. Actually I chatted with @resouer about this, and he told me that kata used to have pod level logs as well.

Especially after https://github.com/kubernetes/kubernetes/pull/74441 lands, Kubelet will be responsible for monitoring the disk usage and cleanup these logs.

@mikebrow Are you ok with cherrypicking this to release/1.2? It doesn't break anyone. :)

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -32,6 +32,15 @@ const (
 	// SandboxID is the sandbox ID annotation
 	SandboxID = "io.kubernetes.cri.sandbox-id"
 
+	// SandboxLogDir is the pod log directory annotation.
+	// If the sandbox needs to generate any log, it will put it into this directory.
+	// Kubelet will be responsible for:
+	// 1) Monitoring the disk usage of the log, and including it as part of the pod
+	// ephemeral storage usage.
+	// 2) Cleaning up the logs when the pod is deleted.
+	// NOTE: Kubelet is not responsible for rotating the logs.
+	SandboxLogDir = "io.kubernetes.cri.sandbox-log-directory"
+
 	// UntrustedWorkload is the sandbox annotation for untrusted workload. Untrusted
 	// workload can only run on dedicated runtime for untrusted workload.
 	UntrustedWorkload = "io.kubernetes.cri.untrusted-workload"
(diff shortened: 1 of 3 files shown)


pkg/annotations/annotations.go
```diff
@@ -32,6 +32,15 @@ const (
 	// SandboxID is the sandbox ID annotation
 	SandboxID = "io.kubernetes.cri.sandbox-id"
 
+	// SandboxLogDir is the pod log directory annotation.
+	// If the sandbox needs to generate any log, it can put it into this directory.
```
mikebrow: /s/can/will/

pkg/annotations/annotations.go
```diff
@@ -32,6 +32,15 @@ const (
...
+	// Kubelet will be responsible to:
```
mikebrow: /s/to/for/

pkg/annotations/annotations.go
```diff
@@ -32,6 +32,15 @@ const (
...
+	// 1) Monitoring the disk usage of the log, and include it as part of the pod
```
mikebrow: /s/include/including/

pkg/annotations/annotations.go
```diff
@@ -32,6 +32,15 @@ const (
...
+	// ephemeral storage usage.
+	// 2) Cleanup the logs when the pod is deleted.
```
mikebrow: /s/Cleanup/Cleaning up/

pkg/annotations/annotations.go
```diff
@@ -32,6 +32,15 @@ const (
...
+	// NOTE that kubelet is not responsible for rotating the logs.
```
mikebrow: /s/Note that/Note: Kublet/

pkg/annotations/annotations.go
Random-Liu: Done

pkg/annotations/annotations.go
Random-Liu: Done

pkg/annotations/annotations.go
Random-Liu: Done

pkg/annotations/annotations.go
Random-Liu: Done

pkg/annotations/annotations.go
Random-Liu: Done

Random-Liu:
Addressed comments about the comment. Apply LGTM based on https://github.com/containerd/cri/pull/1056#pullrequestreview-207126001

PR #1070, opened 2019-03-01
Remove deprecated runtime options.

Random-Liu:
This PR:
1) Auto detect systemd parent cgroup, so that we don't need a daemon level `SystemdCgroup` option any more. However, `SystemdCgroup` option in the runc options is still available, should we also auto detect if it is not explicitly set? /cc @yujuhong @filbranden 
2) Remove all deprecated daemon level runtime options. Users should use runtime specific options since containerd 1.3. Note that we keep the `no_pivot` option right now, because there is no runtime specific option for it in shim v1, we can only remove it when shim v1 is deprecated.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -32,12 +32,6 @@ The explanation and default value of each configuration item are as follows:
   # stats_collect_period is the period (in seconds) of snapshots stats collection.
   stats_collect_period = 10
 
-  # systemd_cgroup enables systemd cgroup support. This only works for runtime
-  # type "io.containerd.runtime.v1.linux".
-  # DEPRECATED: use Runtime.Options for runtime specific config for shim v2 runtimes.
-  #   For runtime "io.containerd.runc.v1", use the option `SystemdCgroup`.
-  systemd_cgroup = false
-
   # enable_tls_streaming enables the TLS streaming support.
   # It generates a self-sign certificate unless the following x509_key_pair_streaming are both set.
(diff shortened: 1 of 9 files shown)


Random-Liu:
Need to fix the test. Will do.

lifupan:
> /cc @lifupan

Thanks for the reminder, it's LGTM from my side.

Random-Liu:
We can do both. Yeah, we should not silently ignore configs. Thanks for the suggestion!

We should:
1) Keep the config fields, and throw out warning if those deprecated fields are set;
2) Stop using those fields in the cri plugin;
3) Provide a upgrade guide.

docs/config.md
```diff
@@ -116,21 +95,6 @@ The explanation and default value of each configuration item are as follows:
...
-    [plugins.cri.containerd.untrusted_workload_runtime]
-      # runtime_type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
-      runtime_type = ""
-
-      # runtime_engine is the name of the runtime engine used by containerd.
-      runtime_engine = ""
-
-      # runtime_root is the directory used by containerd for runtime state.
-      runtime_root = ""
-
```
mikebrow: missing:
```[plugins.cri.containerd.runtimes]```

Random-Liu: We don't need that extra level in toml.

PR #1071, opened 2019-03-02
[WIP] Use format of the parent cgroup to determine whether to use cgroupfs or systemd cgroup driver

filbranden:
The parent cgroup will always be a systemd "slice" which can be detected by looking at a ".slice" suffix in the cgroup name. Use that information to drive whether to tell `runc` to use systemd or cgroupfs.

The advantage of this approach is that the full control of which cgroup driver to use is now in the CRI client. `containerd-cri` will simply match what the CRI client requested, and ask `runc` to do the same. This way it is possible to switch cgroup drivers by simply reconfiguring the CRI client (Kubelet.)

Note: There's a corner case for the root slice, which in theory would be "/" for both cases and auto-detection would not work. But the correct way to specify the root slice for the systemd case is "-.slice" (or "/-.slice", since we only consider the last component of the path), so that corner case actually needs to be handled in the CRI client and not here.

This still needs testing! But sending it along to get some early feedback on the idea itself.

There are a few follow up commits updating comments, documentation and other minor fixes.

It's unclear to me what can be done about the `systemd_cgroup` field of Runtime.Options, since that seems to be part of an external API, so not sure how easy it can be deprecated... Opinions?

/cc @Random-Liu @yujuhong


Random-Liu:
I've actually included this as part of https://github.com/containerd/cri/pull/1070. :p

PR #1077, opened 2019-03-06
Add an OCI annotation for sandbox VM params

harche:
OCI annotations can be used to set the sandbox related parameters. In case of VM based runtimes, such as kata, this can be used to dynamically select the `vmdisk` or `initrd` that can be custom built for that container image.

 e.g. With ppc64le ultravisors' upcoming capability to run secure VM which need to be encrypted with the public key of target system, user can specifically intruct the runtime, like kata, to use a specific initrd on the system for the given container image. 

Kata already supports setting those values via OCI annotations, it's just that until now CRI wasn't setting it. The corresponding k8s pod.yaml looks something like this,

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: foobar-kat3
  "annotations": {
    "com.github.containers.virtcontainers.InitrdPath" : "/usr/share/kata-containers/custom-secure-vm.img"
  }
spec:
  runtimeClassName: kataclass
  containers:
  - name: redis
    image: redis:latest 
```
Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -45,6 +45,7 @@ import (
 	"github.com/containerd/cri/pkg/netns"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 	"github.com/containerd/cri/pkg/util"
+	sandboxAnnotations "github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
 )
 
 func init() {
@@ -451,6 +452,27 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
 		}
 	}
 	g.SetProcessOOMScoreAdj(adj)
+
+	vcAnnotations := []string{
+		sandboxAnnotations.KernelPath,
+		sandboxAnnotations.ImagePath,
+		sandboxAnnotations.InitrdPath,
+		sandboxAnnotations.HypervisorPath,
+		sandboxAnnotations.FirmwarePath,
+		sandboxAnnotations.KernelHash,
+		sandboxAnnotations.ImageHash,
+		sandboxAnnotations.InitrdHash,
+		sandboxAnnotations.HypervisorHash,
+		sandboxAnnotations.FirmwareHash,
+	}
+
+	for _, a := range vcAnnotations {
+		value, ok := config.Annotations[a]
+		if !ok {
+			continue
+		}
+		g.AddAnnotation(a, value)
+	}
 
 	g.AddAnnotation(annotations.ContainerType, annotations.ContainerTypeSandbox)
 	g.AddAnnotation(annotations.SandboxID, id)
--- a/pkg/server/sandbox_run_test.go
+++ b/pkg/server/sandbox_run_test.go
@@ -34,9 +34,23 @@ import (
 	criconfig "github.com/containerd/cri/pkg/config"
 	ostesting "github.com/containerd/cri/pkg/os/testing"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
+	sandboxAnnotations "github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
 )
 
 func getRunPodSandboxTestData() (*runtime.PodSandboxConfig, *imagespec.ImageConfig, func(*testing.T, string, *runtimespec.Spec)) {
+
+	kernelPath := "/var/runtime/kernel"
+	initrdPath := "/var/runtime/initrd.img"
+	imagePath := "/var/runtime/image.img"
+	hypervisorPath := "/usr/bin/qemu-custom"
+	firmwarePath := "/var/run/custom-firmware/img"
+
+	kernelHash := "3971EBF11A18BCE250A282E6FF78739C9A0F7BCB9BC3552FE9CD8191E4A4DA573BB38983A47B2A86A63BFC4E41D3361D5D6AE1A1480BF3F27F730CC75FBC84DE"
+	initrdHash := "652DCF7057417D54C3236731A2CA47330B8809236B87DA89917403FEAAC34EA5DBE938EE60FDD606383B38DF9B51239D25968DDF1C0445C0E7AB448287197D64"
+	imageHash := "31D329DA77A2D0EF2943209EF432E3F0C7ED5B57124AFF1E37A5C7E6A01D780A473BD0FD594A3FA7CFBC2573FB150702BB11014C090D9222F9A0C048771F6F2A"
+	hypervisorHash := "0D4C6CCCCDCA141804EB236A07598FF118435450F58A181EED37E6E2D037720E1847473687558289F3972BDE6A9FD2E1F9EDC8C9B0550621D04198111FA955AD"
+	firmwareHash := "259870D2FF9DCB1C44E2D0C6FB467364422D9643E0577A55397E15ECEADA4668A9D1AD7FF776D1A93B9A70FE8C8C4EA77FE75E4BD21E7C24F1490FE09811CBD8"
+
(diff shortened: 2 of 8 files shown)


mikebrow:
Bad hash on travis.. odd.  Restarting the build.

pkg/annotations/annotations.go
```diff
@@ -35,4 +35,36 @@ const (
 	// UntrustedWorkload is the sandbox annotation for untrusted workload. Untrusted
 	// workload can only run on dedicated runtime for untrusted workload.
 	UntrustedWorkload = "io.kubernetes.cri.untrusted-workload"
+
+	vcAnnotationsPrefix = "com.github.containers.virtcontainers."
```
mikebrow: can we move these to a new vc.go either in the same package or vendor it in?

pkg/server/sandbox_run.go
```diff
@@ -452,6 +452,27 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
 	}
 	g.SetProcessOOMScoreAdj(adj)
 
+	assetAnnotations := []string{
```
mikebrow: maybe if the constants were already in an array of strings... specified in vc.go so you don't have to modify this block each time you add a new annotation.

mikebrow: vcAnnotations?

pkg/server/sandbox_run.go
```diff
@@ -452,6 +452,27 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
...
+		annotations.HypervisorPath,
+		annotations.FirmwarePath,
+		annotations.KernelHash,
+		annotations.ImageHash,
+		annotations.InitrdHash,
+		annotations.HypervisorHash,
+		annotations.FirmwareHash,
+	}
+
+	for _, a := range assetAnnotations {
```
mikebrow: maybe a helper utility that receives the arrays and adds them...

Random-Liu:
1) Can we use runtime options for this?

If not.

2) Can we make this configurable at daemon level, e.g. something like `PassThroughPodAnnotations`.
By default the list should be empty, but vendors can configure it for their experimental features.

I'm not a fan with arbitrary hard coded annotations for a specific use case in the code. :)

pkg/annotations/annotations.go
harche: OK, I will vendored it in.

pkg/server/sandbox_run.go
harche: Since I will vendor it, I will need that array of strings to confine the selection of annotations from all the available ones which may be relevant for this task.

harche: I will rename it.

AkihiroSuda:
+1 for PassThroughPodAnnotations

pkg/server/sandbox_run.go
harche: I thought over it. If I have to move this in a separate utility then I will have to send `Generator` pointer as well as `config.Annotations` to that utility. I am not so sure if it is any benefit to making the code more modular. What do you think?

harche:
@Random-Liu I took the inspiration for this PR from your PR to add a [gvisor specific annotation](https://github.com/containerd/cri/pull/1056). Also although at this point only Kata supports these annotations, nothing prevents other VM based runtimes to use the same annotations.

@Random-Liu @AkihiroSuda Can you elaborate on `PassThroughPodAnnotations`? Do you want the annotations be specified by vendors in a config file and allow the passthrough of only those annotations?

harche:
Like maybe adding an option in a config file described here? https://github.com/containerd/cri/blob/master/docs/config.md

```toml
[plugins.cri.passthroughAnnotations]
    com.github.containers.virtcontainers.InitrdPath = "/var/runtime/initrd.img"
    # tls_key_file is the filepath to the private key paired with the "tls_cert_file"
    com.github.containers.virtcontainers.InitrdHash = "3971EBF11A18BCE250A282E6FF78739C9A0F7BCB9BC3552FE9CD8191E4A4DA573BB38983A47B2A86A63BFC4E41D3361D5D6AE1A1480BF3F27F730CC75FBC84DE"
```

@Random-Liu @AkihiroSuda @mikebrow Let me know what do you think?

harche:
If we agree on that, I will close this PR and open another one for `PassThroughPodAnnotations`

pkg/server/sandbox_run.go
mikebrow: agree more busy work than benefit :-) if we get a 2nd customer for custom annotations we'll add the helper utility then.

mikebrow:
> Like maybe adding an option in a config file described here? https://github.com/containerd/cri/blob/master/docs/config.md
> 
> ```toml
> [plugins.cri]
>      # pod annotations can be passed down the runtime by using OCI annotations
>      passthrough_pod_annotations = [ "com.github.containers.virtcontainers.InitrdPath", "com.github.containers.virtcontainers.InitrdHash"]
> ```
> 
> The corresponding k8s pod yaml will look like,
> 
> ```yaml
> apiVersion: v1
> kind: Pod
> metadata:
>   name: foobar-kat3
>   "annotations": {
>     "com.github.containers.virtcontainers.InitrdPath" : "/usr/share/kata-containers/custom-secure-vm.img",
>     "com.github.containers.virtcontainers.InitrdHash" : "3971EBF11A18BCE250A282E6FF78739C9A0F7BCB9BC3552FE9CD8191E4A4DA573BB38983A47B2A86A63BFC4E41D3361D5D6AE1A1480BF3F27F730CC75FBC84DE"
>   }
> spec:
>   runtimeClassName: kataclass
>   containers:
>   - name: redis
>     image: redis:latest 
> ```
> 
> So the idea is, CRI will only pass annotations to the runtime which are mentioned against `passthrough_pod_annotations`
> 
> @Random-Liu @AkihiroSuda @mikebrow Let me know what do you think?

Or, trying to make it a bit easier to config, we could make it a flag on(off) for passing all pod annotations. Or an array of reg expressions: 
```
valid domain name:
`^(([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1})|([a-zA-Z]{1}[0-9]{1})|([0-9]{1}[a-zA-Z]{1})|([a-zA-Z0-9][a-zA-Z0-9-_]{1,61}[a-zA-Z0-9]))\.([a-zA-Z]{2,6}|[a-zA-Z0-9-]{2,30}\.[a-zA-Z
 ]{2,3})$`)``` 
or an array of domain prefixes with or without wildcards... "com.github.containers.virtcontainers.*"

Random-Liu:
@mikebrow Both are good suggestions.

Passing through all pod annotations is really powerful, and no config complexity at all.
However, it also makes the behavior very much opaque, unpredictable and unmanageable. When looking at a containerd setup, you don't get any hint about what annotation api is being used.

I like the second idea.
1) It makes the annotation api a bit more management at the containerd level. At least when you look at the containerd config, you get a basic idea about what annotations are supported, and those annotations are for which runtime;
2) It is a good supplement of our current runtime options. Today, we already support the [runtime specific options](https://github.com/containerd/cri/blob/master/pkg/config/config.go#L40). However, those options are "per-runtime-handler runtime specific options", there are no "per-pod runtime specific options". The new `passthrough_pod_annotations` can serve that purpose, we may want to just name it something like `runtime_option_annotation` or `runtime_option_pod_annotation` or something similar.

mikebrow:
@harche could you rebase pls?

harche:
Closing this PR in favor of https://github.com/containerd/cri/pull/1084

PR #1080, opened 2019-03-07
Use ParseSignal and AtomicWriteFile functions from containerd

zhsj:
Containerd has its own ParseSignal and AtomicWriteFile implementation.
So there's no need to use these function from github.com/docker/docker.


Random-Liu:
Thanks for cleaning this up!

PR #1084, opened 2019-03-11
Add support for passing sandbox annotations to runtime

harche:
OCI annotations can be used to set the sandbox related parameters. In the case of VM based runtimes, such as kata, this can be used to dynamically select the `vmdisk` or `initrd` that can be custom built for that container image.

 e.g. With ppc64le ultravisors' upcoming capability to run secure VM which need to be encrypted with the public key of target system, user can specifically intruct the runtime, like kata, to use a specific initrd on the system for the given container image. 

Kata already supports setting those values via OCI annotations, it's just that until now CRI wasn't setting it. The corresponding k8s pod.yaml looks something like this,

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: foobar-kat3
  "annotations": {
    "com.github.containers.virtcontainers.InitrdPath" : "/usr/share/kata-containers/custom-secure-vm.img",
    "com.github.containers.virtcontainers.InitrdHash" : "3971EBF11A18BCE250A282E6FF78739C9A0F7BCB9BC3552FE9CD8191E4A4DA573BB38983A47B2A86A63BFC4E41D3361D5D6AE1A1480BF3F27F730CC75FBC84DE" 
 }
spec:
  runtimeClassName: kataclass
  containers:
  - name: redis
    image: redis:latest 
```

```toml
[plugins.cri]
     # pod annotations can be passed down the runtime by using OCI annotations
     runtime_option_pod_annotations = [ "com.github.containers.virtcontainers.InitrdPath", 
                                        "com.github.containers.virtcontainers.InitrdHash"]
```

@Random-Liu I have renamed the `passthrough_pod_annotations` to `runtime_option_pod_annotations` as suggested in [this](https://github.com/containerd/cri/pull/1077#issuecomment-470877046) comment. 

@mikebrow configuration option is still under `[plugins.cri]`. I feel that this should not be tied to any specific runtime like kata. Instead, this feature can even be used by `runc` in the future to test experimental features. Also, I haven't added support for wildcard support that you mentioned in [this](https://github.com/containerd/cri/pull/1077#issuecomment-470626651) comment. I am still not sure if we should allow wildcards there for a couple of reasons. It leads to more code complexity (a simple key look up in a hash table vs full featured regex handling) and since pod annotations can contain any arbitary user defined annotations, I feel it would be good if we could precisey select and allow ony selected annotations to pass through. Let me know what do you think.  

@AkihiroSuda ping. 

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -138,6 +138,12 @@ The explanation and default value of each configuration item are as follows:
       # runtime_type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
       runtime_type = "io.containerd.runc.v1"
 
+      # pod_annotations is list of pod annotations passed to both pod sandbox as well as
+      # container OCI annotations. Pod_annotations also support golang supported
+      # regular expression - https://github.com/google/re2/wiki/Syntax.
+      # e.g. ["runc.com.github.containers.runc.*"]
+      pod_annotations = []
+
       # "plugins.cri.containerd.runtimes.runc.options" is options specific to
       # "io.containerd.runc.v1". Its corresponding options type is:
       #   https://github.com/containerd/containerd/blob/v1.2.0-rc.1/runtime/v2/runc/options/oci.pb.go#L39.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,6 +31,9 @@ type Runtime struct {
 	// This only works for runtime type "io.containerd.runtime.v1.linux".
 	// DEPRECATED: use Options instead. Remove when shim v1 is deprecated.
 	Engine string `toml:"runtime_engine" json:"runtimeEngine"`
+	// PodAnnotations is list of pod annotations passed to both pod sandbox as well as
+	// container OCI annotations.
+	PodAnnotations []string `toml:"pod_annotations" json:"PodAnnotations"`
 	// Root is the directory used by containerd for runtime state.
 	// DEPRECATED: use Options instead. Remove when shim v1 is deprecated.
 	// This only works for runtime type "io.containerd.runtime.v1.linux".
--- a/pkg/server/container_create_test.go
+++ b/pkg/server/container_create_test.go
@@ -36,6 +36,7 @@ import (
 	runtime "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
 
 	"github.com/containerd/cri/pkg/annotations"
+	criconfig "github.com/containerd/cri/pkg/config"
 	ostesting "github.com/containerd/cri/pkg/os/testing"
 	"github.com/containerd/cri/pkg/util"
 )
@@ -116,6 +117,7 @@ func getCreateContainerTestData() (*runtime.ContainerConfig, *runtime.PodSandbox
 			Namespace: "test-sandbox-ns",
 			Attempt:   2,
 		},
+		Annotations: map[string]string{"c": "d"},
 		Linux: &runtime.LinuxPodSandboxConfig{
 			CgroupParent: "/test/cgroup/parent",
 		},
@@ -193,7 +195,7 @@ func TestGeneralContainerSpec(t *testing.T) {
 	config, sandboxConfig, imageConfig, specCheck := getCreateContainerTestData()
 	c := newTestCRIService()
 	testSandboxID := "sandbox-id"
-	spec, err := c.generateContainerSpec(testID, testSandboxID, testPid, config, sandboxConfig, imageConfig, nil)
+	spec, err := c.generateContainerSpec(testID, testSandboxID, testPid, config, sandboxConfig, imageConfig, nil, []string{})
 	require.NoError(t, err)
 	specCheck(t, testID, testSandboxID, testPid, spec)
 }
@@ -248,7 +250,7 @@ func TestContainerCapabilities(t *testing.T) {
 	} {
 		t.Logf("TestCase %q", desc)
 		config.Linux.SecurityContext.Capabilities = test.capability
-		spec, err := c.generateContainerSpec(testID, testSandboxID, testPid, config, sandboxConfig, imageConfig, nil)
+		spec, err := c.generateContainerSpec(testID, testSandboxID, testPid, config, sandboxConfig, imageConfig, nil, []string{})
 		require.NoError(t, err)
 		specCheck(t, testID, testSandboxID, testPid, spec)
 		for _, include := range test.includes {
@@ -275,7 +277,7 @@ func TestContainerSpecTty(t *testing.T) {
 	c := newTestCRIService()
 	for _, tty := range []bool{true, false} {
 		config.Tty = tty
-		spec, err := c.generateContainerSpec(testID, testSandboxID, testPid, config, sandboxConfig, imageConfig, nil)
+		spec, err := c.generateContainerSpec(testID, testSandboxID, testPid, config, sandboxConfig, imageConfig, nil, []string{})
 		require.NoError(t, err)
 		specCheck(t, testID, testSandboxID, testPid, spec)
 		assert.Equal(t, tty, spec.Process.Terminal)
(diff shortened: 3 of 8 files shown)


mikebrow:
> @mikebrow configuration option is still under `[plugins.cri]`. I feel that this should not be tied to any specific runtime like kata. Instead, this feature can even be used by `runc` in the future to test experimental features. 

I think you may have miss-understood my point? I was saying per/runtime as in a new field in the runtime object which is specified for each runtime in runtimes[], not tied to any one particular runtime. Per @Random-Liu s comment this would show which runtime supports which pod annotations. 

> Also, I haven't added support for wildcard support that you mentioned in [this](https://github.com/containerd/cri/pull/1077#issuecomment-470626651) comment. I am still not sure if we should allow wildcards there for a couple of reasons. It leads to more code complexity (a simple key look up in a hash table vs full featured regex handling) and since pod annotations can contain any arbitary user defined annotations, I feel it would be good if we could precisey select and allow ony selected annotations to pass through. Let me know what do you think.

Yes, by wildcard I meant optional wildcard. IE. if you had a group of 1k optional annotations all scoped to one domain, I don't think you'd want to list each one. But yeah, we can add wildcards and / or regular expression support in a followup. 

> @Random-Liu I have renamed the `passthrough_pod_annotations` to `runtime_option_pod_annotations` as suggested in [this](https://github.com/containerd/cri/pull/1077#issuecomment-470877046) comment.

As for the naming.. if we do put it in per runtime in the runtime class, then the word runtime would be repetitive to use in a field for the runtime. How about just pod_annotations? Then explain in the comment that these are optional for passing a set of annotations through to the runtime for processing by the runtime.

docs/config.md
```diff
@@ -26,6 +26,9 @@ The explanation and default value of each configuration item are as follows:
   # enable_selinux indicates to enable the selinux support.
   enable_selinux = false
 
+  # runtime_option_pod_annotations is list of pod annotations passed to OCI annotations
+  runtime_option_pod_annotations = []
```
Random-Liu: Can we make this a per runtime option? As suggested in https://github.com/containerd/cri/pull/1077#issuecomment-470626651

pkg/server/sandbox_run.go
```diff
@@ -452,6 +452,14 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
 	}
 	g.SetProcessOOMScoreAdj(adj)
 
+	for _, r := range c.config.RuntimeOptionsPodAnnotations {
```
Random-Liu: Can we support regular expression match? As suggested in https://github.com/containerd/cri/pull/1077#issuecomment-470626651

Random-Liu:
> configuration option is still under [plugins.cri]. I feel that this should not be tied to any specific runtime like kata. Instead, this feature can even be used by runc in the future to test experimental features.

I don't think this should be tied to `kata`. I mean it should be tied to a runtime, the runtime could be any runtime, but for a specific runtime, a set of annotations are allowed.

You can add the field in the runtime config https://github.com/containerd/cri/blob/master/pkg/config/config.go#L27

harche:
@Random-Liu @mikebrow thanks for your feedback, I will update the PR with your suggestions.

Random-Liu:
> It leads to more code complexity (a simple key look up in a hash table vs full featured regex handling) and since pod annotations can contain any arbitary user defined annotations, I feel it would be good if we could precisey select and allow ony selected annotations to pass through.

We can use the golang https://golang.org/pkg/regexp/ library. In the document, we can just link to golang supported regular expression doc https://github.com/google/re2/wiki/Syntax.

I think in most cases, people will just need ".*"

harche:
> We can use the golang https://golang.org/pkg/regexp/ library. In the document, we can just link to golang supported regular expression doc https://github.com/google/re2/wiki/Syntax.
> 
> I think in most cases, people will just need ".*"

Yeah, that's what I have done. I will write some unit tests and send a PR by tomorrow.

pkg/server/sandbox_run.go
harche: done

docs/config.md
harche: done

harche:
@mikebrow @Random-Liu I moved `pod_annotations` in `Runtime` and also added wildcard support. I have added a unit test to cover possible scenarios. 

However, although `pod_annotations` are defined within [Runtime](https://github.com/containerd/cri/pull/1084/files#diff-091ace02b49a46f55f2586cfec89a160R35) when they get [injected in OCI Annotations](https://github.com/containerd/cri/pull/1084/files#diff-69e1275271049b9b0aae8014965cee3dR472) I wasn't able to find a way to specifically get a runtime for which the pod annotations are declared. 

e.g. in config.toml of containerd
```toml
     [plugins.cri.containerd.runtimes.runc]
         runtime_type = "io.containerd.runc.v1"
         pod_annotations = ["com.github.containers.virtcontainers.*", "abc.bbc.cdc"]
      [plugins.cri.containerd.runtimes.kata]
         runtime_type = "io.containerd.kata.v2"
```

Because of this, the allowed passthrough annotations get injected in any OCI spec regardless of the target runtime. I wasn't able to find a way to select the runtime [here](https://github.com/containerd/cri/pull/1084/files#diff-69e1275271049b9b0aae8014965cee3dR472) based on the info available in that scope.

pkg/server/sandbox_run.go
```diff
@@ -452,13 +453,42 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
...
 	g.AddAnnotation(annotations.SandboxLogDir, config.GetLogDirectory())
 
 	return g.Config, nil
 }
 
+func (c *criService) getPassthroughAnnotations(sandboxAnnotations map[string]string) (passthroughAnnotations map[string]string) {
+	passthroughAnnotations = make(map[string]string)
+
+	for podAnnoationKey, podAnnoationValue := range sandboxAnnotations {
+		for runtimeKey := range c.config.Runtimes {
```
mikebrow: within RunPodSandbox() call r.GetRuntimeHandler() to get the runtimeKey you need... pass that to this getPassthroughAnnotations() helper.. use the passed in key as the runtimeKey.. 

check if there is a runtime in c.config.Runtimes[runtimeKey] before you start the loop.. at the top of the function and return nil if there are no matching runtimes...

hope this helps

harche: Thanks @mikebrow . That really helped. Turns out `ociRuntime` is already [fetched](https://github.com/harche/cri/blob/passthrough_annotations/pkg/server/sandbox_run.go#L101) within `RunPodSandbox()` so I just have to [pass](https://github.com/harche/cri/blob/passthrough_annotations/pkg/server/sandbox_run.go#L151) the runtime pod annotations defined for that exact runtime all the way down till `getPassthroughAnnotations()`.

Thanks for the pointer. I have updated my PR.

docs/config.md
```diff
@@ -138,6 +138,11 @@ The explanation and default value of each configuration item are as follows:
       # runtime_type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
       runtime_type = "io.containerd.runc.v1"
 
+      # pod_annotations is list of pod annotations passed to OCI annotations
+      # pod_annotations support wildcard.
+      # e.g. ["com.github.containers.virtcontainers.*"]
```
Random-Liu: Let's use `runc`. `com.github.containers.runc.*`.

docs/config.md
```diff
@@ -138,6 +138,11 @@ The explanation and default value of each configuration item are as follows:
       # runtime_type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
       runtime_type = "io.containerd.runc.v1"
 
+      # pod_annotations is list of pod annotations passed to OCI annotations
```
Random-Liu: Are these annotations passed to the pod sandbox, or containers, or both the pod sandbox and all containers?

Let's clarify that.

pkg/config/config.go
```diff
@@ -31,6 +31,8 @@ type Runtime struct {
 	// This only works for runtime type "io.containerd.runtime.v1.linux".
 	// DEPRECATED: use Options instead. Remove when shim v1 is deprecated.
 	Engine string `toml:"runtime_engine" json:"runtimeEngine"`
+	// PodAnnotations is list of pod annotations passed to OCI annotations
```
Random-Liu: ditto.

pkg/server/sandbox_run.go
```diff
@@ -452,13 +453,40 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
...
+	runtimePodAnnotations []string) (passthroughAnnotations map[string]string) {
+	passthroughAnnotations = make(map[string]string)
+
+	for podAnnoationKey, podAnnoationValue := range sandboxAnnotations {
+		for _, r := range runtimePodAnnotations {
+			if r == podAnnoationKey {
+				passthroughAnnotations[podAnnoationKey] = podAnnoationValue
+				continue
+			}
+			if strings.HasSuffix(r, "*") {
```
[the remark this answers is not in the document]
Random-Liu: Why not just do regexp? Doing this by ourselves just makes this more error prone.

pkg/server/sandbox_run.go
```diff
@@ -452,13 +453,40 @@ func (c *criService) generateSandboxContainerSpec(id string, config *runtime.Pod
 	}
 	g.SetProcessOOMScoreAdj(adj)
 
+	for passthroughAnnotationKey, passthroughAnnotationvalue := range c.getPassthroughAnnotations(config.Annotations, runtimePodAnnotations) {
```
Random-Liu: nit: Line too long.

Random-Liu: So these annotations are not passed to containers?

I think we should probably pass them to both sandbox and containers. E.g. for a runc extended feature, passing annotations to the pause container doesn't make sense.

pkg/server/sandbox_run.go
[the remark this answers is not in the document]
harche: Done

docs/config.md
harche: Done

docs/config.md
harche: Done

pkg/config/config.go
harche: Done

pkg/server/sandbox_run.go
harche: Done

pkg/config/config.go
```diff
@@ -31,6 +31,9 @@ type Runtime struct {
 	// This only works for runtime type "io.containerd.runtime.v1.linux".
 	// DEPRECATED: use Options instead. Remove when shim v1 is deprecated.
 	Engine string `toml:"runtime_engine" json:"runtimeEngine"`
+	// pod_annotations is list of pod annotations passed to both pod sandbox as well as
```
Random-Liu: nit: PodAnnotations

pkg/util/annotations.go
```diff
@@ -0,0 +1,38 @@
...
+
+package util
+
+import (
+	"regexp"
+)
+
+// GetPassthroughAnnotations filters requested pod annotations by comparing
+// against permitted annotations for the given runtime.
+func GetPassthroughAnnotations(sandboxAnnotations map[string]string,
```
Random-Liu: nit: I think adding this to `pkg/server/helper.go` is good enough. I don't think we'll use this in other packages.

Random-Liu: s/sandboxAnnotations/podAnnotations

pkg/util/annotations.go
```diff
@@ -0,0 +1,38 @@
...
+	runtimePodAnnotations []string) (passthroughAnnotations map[string]string) {
+	passthroughAnnotations = make(map[string]string)
+
+	for podAnnoationKey, podAnnoationValue := range sandboxAnnotations {
```
Random-Liu: s/Annoation/Annotation

pkg/server/sandbox_run_test.go
```diff
@@ -82,6 +83,47 @@ func getRunPodSandboxTestData() (*runtime.PodSandboxConfig, *imagespec.ImageConf
 	return config, imageConfig, specCheck
 }
 
+func TestPodAnnotationPassthroughSandboxFilter(t *testing.T) {
```
Random-Liu: Can we test at `generateSandboxContainerSpec` level?

This test can't make sure that the annotations are actually applied in the OCI spec.

Random-Liu: And can you change the test to be table driven?

pkg/server/sandbox_run_test.go
```diff
@@ -82,6 +83,47 @@ func getRunPodSandboxTestData() (*runtime.PodSandboxConfig, *imagespec.ImageConf
...
+	c.config.Runtimes["kata"] = criconfig.Runtime{
+		// containerd config allows wildcard for following pod annotations passthrough
+		PodAnnotations: []string{"t*", "z.*", "y.c*"},
+	}
+
+	passthroughAnnotations = util.GetPassthroughAnnotations(config.Annotations, c.config.Runtimes["kata"].PodAnnotations)
+	assert.Equal(t, passthroughAnnotations["t.f"], "j")
+	assert.Equal(t, passthroughAnnotations["z.g"], "o")
+	assert.Equal(t, passthroughAnnotations["y.ca"], "b")
+}
```
Random-Liu: nit: empty line

pkg/server/container_create_test.go
```diff
@@ -287,6 +289,49 @@ func TestContainerSpecTty(t *testing.T) {
 	}
 }
 
+func TestPodAnnotationPassthroughContainerFilter(t *testing.T) {
```
Random-Liu: Test at `generateContainerSpec` level.

Random-Liu: And table driven please.

pkg/config/config.go
harche: done.

pkg/util/annotations.go
harche: Done.

harche:
@Random-Liu I have updated PR with table driven tests

pkg/server/sandbox_run_test.go
harche: done.

pkg/server/container_create_test.go
harche: Done.

Random-Liu:
There are still something I want to cleanup. I'll get this change in and send a cleanup PR.

pkg/server/sandbox_run.go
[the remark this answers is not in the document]
Random-Liu: Sorry, I thought more about this, and found that regular expression actually doesn't serve the domain name matching purpose very well, because we use `.` in domain name a lot, and in regular expression if we want to match `.`, we have to use `\.`.

I think your original solution is better. @harche /cc @mikebrow WDYT?
Sorry for the churns.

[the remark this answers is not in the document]
mikebrow: ^com\.github\.containers\.virtcontainers\.

PR #1085, opened 2019-03-12
Fix /etc/hostname backward compatibility issue for in-place upgrade.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1082.

If people do in-place upgrade, old sandboxes may not have the `hostname` file. In that case, we shouldn't try to mount it into containers.

Another option is to create the hostname file for the old sandbox if it doesn't have it. However, that also causes inconsistency, e.g. inside the same pod, old running containers don't have the hostname file, new restarted containers have it.

Neither of them is a perfect solution, skipping mounting is cleaner and keeps consistency inside the same pod, so I implemented that one.


Random-Liu:
We should cherry-pick this into supported branches. And let's include the fix in 1.2.5 https://github.com/containerd/containerd/pull/3086.

PR #1090, opened 2019-03-13
Add test for filtering container create masks when privileged

mikebrow:
Changed PR from add filtering to testing the filtering...

Add test case to confirm we are processing masks correctly when privileged. see https://github.com/kubernetes/kubernetes/issues/75238

Based on the code in dockershim... we should not be masking/adding read only paths for privileged containers (even if requested). Shrug. 

Added note: Since we already do mask, simply adding a test case and comments to verify and explain to future readers.


Random-Liu:
See https://github.com/containerd/cri/issues/1089#issuecomment-472731165

mikebrow:
@Random-Liu Thanks! I've updated the commit to just comment the code a bit and add a test cases.

PR #1093, opened 2019-03-15
Add advertise_stream_server_address option

### LINKED ISSUE #1011: Need advertising IP when using VIP or binding 0.0.0.0
CriService is using `StreamServerAddress` to serve function `GetExec`. There may be some cases that `GetExec` would return unusable url.

1. Configured StreamServerAddress of CriService  to "0.0.0.0", `GetExec` would return "https://0.0.0.0:$port/exec/$token", however client connected cri using real ip
2. client connected cri using VIP and the url `GetExec` returned is not connectible.

Proposal: Add a new config item of `advertise_stream_server_address`, `GetExec` would refer to it when composing Exec url. And when it's null, it fallback to use StreamServerAddress

RostyslavFridman:
Fixes #1011
We try to start `advertiseStreamServer` right before `GetExec` because it is possible that advertising address can appear later than containerd starts.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -121,6 +121,8 @@ type PluginConfig struct {
 	CniConfig `toml:"cni" json:"cni"`
 	// Registry contains config related to the registry
 	Registry Registry `toml:"registry" json:"registry"`
+	// AdvertiseStreamServerAddress is the ip address streaming server is advertising on.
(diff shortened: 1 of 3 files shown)


Random-Liu:
If we want to support this, we should add the support into k8s.io/kubernetes/pkg/kubelet/server/streaming.

dmcgowan:
The master branch on this repository has been merged into [containerd/containerd](https://github.com/containerd/containerd). Please rebase your change onto containerd master and open a pull request there.

PR #1096, opened 2019-03-22
Cleanup pod annotation test and only support wildcard

Random-Liu:
This is a follow up of #1084.

This PR:
1) Cleanup passthrough pod annotation test;
2) Change back to only support wildcard, reason https://github.com/containerd/cri/pull/1084#discussion_r268091851.

/cc @harche I think your original solution is better, because:
1) Regular expression is not convenient for domain name matching, especially for matching `.`.
2) A generic wildcard support requires dynamic planning, which makes the code complicated.
Let's just do tailing wildcard support.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -138,10 +138,14 @@ The explanation and default value of each configuration item are as follows:
       # runtime_type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
       runtime_type = "io.containerd.runc.v1"
 
-      # pod_annotations is list of pod annotations passed to both pod sandbox as well as
-      # container OCI annotations. Pod_annotations also support golang supported
-      # regular expression - https://github.com/google/re2/wiki/Syntax.
-      # e.g. ["runc.com.github.containers.runc.*"]
+      # pod_annotations is a list of pod annotations passed to both pod
+      # sandbox as well as container OCI annotations. Pod_annotations also
+      # supports golang path match pattern - https://golang.org/pkg/path/#Match.
+      # e.g. ["runc.com.*"], ["*.runc.com"], ["runc.com/*"].
+      #
+      # For the naming convention of annotation keys, please reference:
+      # * Kubernetes: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
+      # * OCI: https://github.com/opencontainers/image-spec/blob/master/annotations.md
       pod_annotations = []
 
       # "plugins.cri.containerd.runtimes.runc.options" is options specific to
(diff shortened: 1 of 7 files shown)


mikebrow:
```^com\.github\.containers\.virtcontainers\.```
not as pretty as I would like but ^ works

Random-Liu:
@mikebrow Initially I thought wildcard match is just a subset of regular expression, that's why I suggested that. But I'm wrong, sorry :P 

Just found that we do have a convention at https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

So basically the annotation key could only be:
1) name [required, 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.)];
2) DNS subdomain / name.

I think we can just use golang [`path.Match`](https://golang.org/pkg/path/#Match).
User can specify things like:
1) *.example.com
2) example.com.*
3) example.com/*
4) example.com.*/*

I didn't use `path.Match` initially, because I thought there might be multiple `/`, and it is hard for users to add patterns for all that: `a/*`, `a/*/*`, `a/*/*/*`.

However, since there can only be at most one slash, I think `path.Match` is good enough. :) @mikebrow WDYT?

mikebrow:
> @mikebrow Initially I thought wildcard match is just a subset of regular expression, that's why I suggested that. But I'm wrong, sorry :P
> 
> Just found that we do have a convention at https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
> 
> So basically the annotation key could only be:
> 
> 1. name [required, 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.)];
> 2. DNS subdomain / name.
> 
> I think we can just use golang [`path.Match`](https://golang.org/pkg/path/#Match). With that, users can specify things like:
> 
> 1. *.example.com
> 2. example.com.*
> 3. example.com/*
> 4. example.com.*/*
> 
> I didn't use `path.Match` initially, because I thought there might be multiple `/`, and it is hard for users to add patterns for all that: `a/*`, `a/*/*`, `a/*/*/*`.
> 
> However, since there can only be at most one slash, I think `path.Match` is good enough. :) @mikebrow WDYT?

Good find on that already existing convention in kube. Additionally, OCI defines: https://github.com/opencontainers/image-spec/blob/master/annotations.md

We should probably link to those in the doc for the field so admins will know how to remain OCI and kube compliant.

Yes, it looks like path.Match and restricting to these patterns is best!

Random-Liu:
@mikebrow Addressed comments.

harche:
@Random-Liu  lgtm

Random-Liu:
Just squashed the commits. Apply LGTM based on https://github.com/containerd/cri/pull/1096#pullrequestreview-218038021

PR #1099, opened 2019-03-27
Do not SIGKILL container if container stop is cancelled.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1098.


Random-Liu:

lbernail:
Thanks a lot!

PR #1101, opened 2019-03-27
bump opencontainers/selinux to v1.2

thaJeztah:
This repository now started doing tagged releases.

### DIFF
--- a/vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go
+++ b/vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go
@@ -4,6 +4,8 @@ package label
 
 import (
 	"fmt"
+	"os"
+	"os/user"
 	"strings"
 
 	"github.com/opencontainers/selinux/go-selinux"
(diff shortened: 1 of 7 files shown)


vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go
```diff
@@ -90,14 +104,36 @@ func SetProcessLabel(processLabel string) error {
 	return selinux.SetExecLabel(processLabel)
 }
 
+// SetSocketLabel takes a process label and tells the kernel to assign the
+// label to the next socket that gets created
+func SetSocketLabel(processLabel string) error {
```
thaJeztah: I'm not an SELinux expert (so definitely could use some eyes on this PR 😅 

I noticed this was added; not sure if this is needed / has to be used for anything in this repository?

vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go
```diff
@@ -130,13 +166,56 @@ func Relabel(path string, fileLabel string, shared bool) error {
 		return nil
 	}
 
-	exclude_paths := map[string]bool{"/": true, "/usr": true, "/etc": true, "/tmp": true, "/home": true, "/run": true, "/var": true, "/root": true}
+	exclude_paths := map[string]bool{
```
thaJeztah: pls double-check if the new list will cause any changes in behavior that we should be aware of 🤗 (left the same question on https://github.com/moby/moby/pull/38918#pullrequestreview-217130793)

vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go
Random-Liu: I'm not familiar with SELinux, either.

We don't use this function today.
Based on https://github.com/opencontainers/selinux/pull/38, this is a completely new function, so I don't think it will cause any regression.

Random-Liu:
Let's get this in after https://github.com/moby/moby/pull/38918

thaJeztah:
moby https://github.com/moby/moby/pull/38918 was merged. I see this one needs a rebase; gonna give that a go 👍

thaJeztah:
done; rebased 👍

PR #1102, opened 2019-03-28
Uts namespace and fix array

Random-Liu:
This PR:
1) Remove UTS namespace for hostnetwork. This is to keep the behavior consistent with dockershim https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/security_context.go#L204.
2) Stop assuming arrays don't contain duplicated elements.

1 is important, and we should cherry-pick to supported branches.
2 is not that critical.


Random-Liu:
@mikebrow Can you take a look at this as well? We need the UTS namespace fix. :)

PR #1116, opened 2019-04-04
Enable runc.v2 as the default runtime in test.

Random-Liu:
For https://github.com/containerd/cri/issues/1075.
/cc @crosbymichael @yujuhong 

Use per-pod shim in the node e2e and cluster e2e test.

We should consider switching the default runtime in `pkg/config/config.go`, once the test result looks good.


crosbymichael:
LGTM

PR #1118, opened 2019-04-05
Support comment in vendor.

Random-Liu:
The containerd build is broken, because we added a vendor comment for runc in https://github.com/containerd/containerd/pull/3184 and https://github.com/containerd/containerd/pull/3183.
* https://k8s-testgrid.appspot.com/sig-node-containerd#containerd-build-1.1
* https://k8s-testgrid.appspot.com/sig-node-containerd#containerd-build-1.2

This should not happen.

### DIFF
--- a/hack/utils.sh
+++ b/hack/utils.sh
@@ -81,7 +81,7 @@ from-vendor() {
                   BEGIN { rc=1 }                        # Assume we did not find what we were looking for.
                   // {
                      if ($1 == REPO) {
-                        if ($3 != "") { gsub(/http.*\/\//, "", $3); REPO = $3 };    # Override repo.
+                        if ($3 != "" && $3 !~ /#.*/ ) { gsub(/http.*\/\//, "", $3); REPO = $3 };    # Override repo.
                         printf("%s_VERSION=%s; %s_REPO=%s\n", WHAT, $2, WHAT, REPO);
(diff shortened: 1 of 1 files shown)


hack/utils.sh
```diff
@@ -81,7 +81,7 @@ from-vendor() {
                   BEGIN { rc=1 }                        # Assume we did not find what we were looking for.
                   // {
                      if ($1 == REPO) {
-                        if ($3 != "") { gsub(/http.*\/\//, "", $3); REPO = $3 };    # Override repo.
+                        if ($3 != "" && $3 !~ /#.*/ ) { gsub(/http.*\/\//, "", $3); REPO = $3 };    # Override repo.
```
yujuhong: What does `!~` do?

thaJeztah:
> The containerd build is broken, because we added a vendor comment for runc in containerd/containerd#3184 and containerd/containerd#3183.

oh, sorry! didn't know that would break things 😓

hack/utils.sh
Random-Liu: It is regular expression match.

Random-Liu:
@thaJeztah No worry. We should have fixed this long time ago. :)

I once wanted to add a comment in `cri/vendor.conf`, and broke things, but didn't get time to fix at that itme.

PR #1125, opened 2019-04-08
Add default runtime name

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1076.

This PR:
1) Add support for `DefaultRuntimeName`. And map `DefaultRuntime` to `DefaultRuntimeName=default` for legacy config support.
2) Change `DefaultRuntimeName` to `runc`, and start using `io.containerd.runc.v1` as the default runtime. (We may consider switching to v2 once https://github.com/containerd/cri/issues/1075 is fixed).
3) Add warning log for deprecated fields `DefaultRuntime` and `UntrustedWorkloadRuntime`.
4) Add deprecation policy for CRI config options.

/cc @tallclair Please take a look at the runtime config related change.

### DIFF
--- a/cri.go
+++ b/cri.go
@@ -119,7 +119,7 @@ func validateConfig(ctx context.Context, c *criconfig.Config) error {
 
 	// Validation for deprecated untrusted_workload_runtime.
 	if c.ContainerdConfig.UntrustedWorkloadRuntime.Type != "" {
-		log.G(ctx).Warning("untrusted_workload_runtime is deprecated, please use `untrusted` runtime in `runtimes` instead")
+		log.G(ctx).Warning("`untrusted_workload_runtime` is deprecated, please use `untrusted` runtime in `runtimes` instead")
 		if _, ok := c.ContainerdConfig.Runtimes[criconfig.RuntimeUntrusted]; ok {
 			return errors.Errorf("conflicting definitions: configuration includes both `untrusted_workload_runtime` and `runtimes[%q]`", criconfig.RuntimeUntrusted)
 		}
@@ -128,7 +128,7 @@ func validateConfig(ctx context.Context, c *criconfig.Config) error {
 
 	// Validation for deprecated default_runtime field.
 	if c.ContainerdConfig.DefaultRuntime.Type != "" {
-		log.G(ctx).Warning("default_runtime is deprecated, please use default_runtime_name instead")
+		log.G(ctx).Warning("`default_runtime` is deprecated, please use `default_runtime_name` to reference the default configuration you have defined in `runtimes`")
 		c.ContainerdConfig.DefaultRuntimeName = criconfig.RuntimeDefault
 		c.ContainerdConfig.Runtimes[criconfig.RuntimeDefault] = c.ContainerdConfig.DefaultRuntime
 	}
--- a/cri.go
+++ b/cri.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cri
 
 import (
+	"context"
 	"flag"
 	"path/filepath"
 	"time"
@@ -73,7 +74,7 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	}
 	log.G(ctx).Infof("Start cri plugin with config %+v", c)
 
-	if err := validateConfig(&c); err != nil {
+	if err := validateConfig(ctx, &c); err != nil {
 		return nil, errors.Wrap(err, "invalid config")
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -227,6 +230,8 @@ func DefaultConfig() PluginConfig {
 }
 
 const (
-	// RuntimeUntrusted is the implicit runtime defined for ContainerdConfig.UntrustedWorkloadRuntime
+	// RuntimeUntrusted is the implicit runtime defined for "io.kubernetes.cri.untrusted-workload" annotation.
 	RuntimeUntrusted = "untrusted"
+	// RuntimeDefault is the implicit runtime defined for the default runtime.
+	RuntimeDefault = "default"
```
tallclair: I'd actually prefer not to define this. I'd rather the `runc` runtime be named `runc` to make that explicit, and just use the DefaultRuntimeName field to make that the default. My fear is that defining `default` here will make having a `default` runtime a common practice.

Random-Liu: I can only map `DefaultRuntime` to this, and having the default config to use runc, are you OK with that?

tallclair: Sounds good.

Random-Liu:
@mikebrow Rebased. Can you take a look?

jterry75:
LGTM! This makes it easier on Windows. So far we have three runtimes:
"runhcs-wcow-process"
"runhcs-wcow-hypervisor"
"runhcs-lcow"
So this allows us to have the three runtime settings as well as define the default if not passed. Great work!

cri.go
```diff
@@ -111,14 +112,36 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
...
-		if c.ContainerdConfig.UntrustedWorkloadRuntime.Type != "" {
-			return errors.New("conflicting definitions: configuration includes untrusted_workload_runtime and runtimes['untrusted']")
+func validateConfig(ctx context.Context, c *criconfig.Config) error {
+	if c.ContainerdConfig.Runtimes == nil {
+		c.ContainerdConfig.Runtimes = make(map[string]criconfig.Runtime)
+	}
+
+	// Validation for deprecated untrusted_workload_runtime.
+	if c.ContainerdConfig.UntrustedWorkloadRuntime.Type != "" {
+		log.G(ctx).Warning("untrusted_workload_runtime is deprecated, please use `untrusted` runtime in `runtimes` instead")
```
mikebrow: the statement please use `untrusted` runtime has been confusing to users, because we don't actually do anything in our code with the `untrusted` runtimes[] name?, where before with the untrusted workload we did.  I think it's probably better to just say it's no longer necessary or supported and that they can manually configure and select any runtime they desire via the `runtimes` feature.

or... if we desire to continue promoting the specification of an `untrusted` runtime in runtimes... let's just add a field in the config that points to the untrusted runtime..  Maybe:  `default_untrusted_runtime_name` but with the Runtimes match feature.. don't really see a point to specifying untrusted workloads anymore, or maybe I've missed something?

cri.go
```diff
@@ -111,14 +112,36 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
...
+		if _, ok := c.ContainerdConfig.Runtimes[criconfig.RuntimeUntrusted]; ok {
+			return errors.Errorf("conflicting definitions: configuration includes both `untrusted_workload_runtime` and `runtimes[%q]`", criconfig.RuntimeUntrusted)
 		}
+		c.ContainerdConfig.Runtimes[criconfig.RuntimeUntrusted] = c.ContainerdConfig.UntrustedWorkloadRuntime
+	}
+
+	// Validation for deprecated default_runtime field.
+	if c.ContainerdConfig.DefaultRuntime.Type != "" {
+		log.G(ctx).Warning("default_runtime is deprecated, please use default_runtime_name instead")
```
mikebrow: instead... to reference the default configuration you have defined in the `runtimes` list/map.

docs/config.md
```diff
@@ -80,60 +80,22 @@ The explanation and default value of each configuration item are as follows:
...
-
-        # SystemdCgroup enables systemd cgroups.
-        SystemdCgroup = false
 
     # "plugins.cri.containerd.untrusted_workload_runtime" is a runtime to run untrusted workloads on it.
-    # DEPRECATED: use plugins.cri.runtimes instead. If provided, this runtime is mapped to the
-    #   runtime handler named 'untrusted'. It is a configuration error to provide both the (now
-    #   deprecated) UntrustedWorkloadRuntime and a handler in the Runtimes handler map (below) for
-    #   'untrusted' workloads at the same time. Please provide one or the other.
+    # DEPRECATED: use `untrusted` runtime in `plugins.cri.runtimes` instead.
```
mikebrow: see above comment

docs/config.md
```diff
@@ -205,3 +167,11 @@ The explanation and default value of each configuration item are as follows:
       [plugins.cri.registry.mirrors."docker.io"]
         endpoint = ["https://registry-1.docker.io", ]
 ```
+
+## Deprecation
+The config options of the CRI plugin follow the [Kubernetes deprecation
+policy of "admin-facing CLI components"](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli).
+
+Basically, when a config option is announced to be deprecated:
```
mikebrow: /s/Basically/In summary,

cri.go
Random-Liu: The untrusted workload pod annotation still needs to be supported. Since it is an api, it will take longer to deprecate it.

I prefer not adding a new field `default_untrusted_runtime_name`, because we want to deprecate the config, adding new config option just makes deprecating them even harder.

Random-Liu: The "untrusted" runtime was added in 1.2 https://github.com/containerd/cri/pull/891.

cri.go
Random-Liu: Done

docs/config.md
Random-Liu: Done

cri.go
mikebrow: ok.. then let's just explain it better in the doc to reduce confusion.

docs/config.md
```diff
@@ -205,3 +167,11 @@ The explanation and default value of each configuration item are as follows:
       [plugins.cri.registry.mirrors."docker.io"]
         endpoint = ["https://registry-1.docker.io", ]
 ```
+
```
mikebrow: If using an untrusted-workload pod annotation to request a pod be run using a runtime for untrusted workloads, the RuntimeHandler `plugins.cri.containerd.runtimes.untrusted` must be defined first. When the annotation `io.kubernetes.cri.untrusted-workload` is set to `true` the `untrusted` runtime will be used. For example, see [Create an untrusted pod using Kata Containers(https://github.com/kata-containers/documentation/blob/master/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md#create-an-untrusted-pod-using-kata-containers).

Random-Liu: Done

PR #1130, opened 2019-04-10
Update go-cni to 891c2a41e18144b2d7921f971d6c9789a68046b2.

### LINKED ISSUE #1078: Consider load cni conf in a goroutine to prevent stuckup in Status
When cni plugin is setting up a interface for a long time (>30s), a RLock is hold, which would prevent Load() to continue. Meanwhile cri Status() tried to load cni config and check it with Status(), see

https://github.com/containerd/cri/blob/95f564f95bbc0091ff229a8b707dd9f310aae089/pkg/server/status.go#L47

it will case hang until the setup is finished. So I guess we'd better load the config in an async goroutine.

Random-Liu:
Fixes #1078.


Random-Liu:
Just added a fake implementation for the new `GetConfig` function. Apply LGTM based on https://github.com/containerd/cri/pull/1130#pullrequestreview-224710363

PR #1133, opened 2019-04-12
Use wait instead of `TaskExit`.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1132.

This PR:
1) Uses `Wait` instead of `TaskExit` event. This should fix issues like https://github.com/containerd/containerd/issues/3177 and https://github.com/containerd/containerd/issues/3125;
2) Removes big sandbox and container start lock, because with `Wait` we can control the timing of exit event generation to avoid race condition. This is very important, because sandbox/container start lock is the last potential long running lock we have today. And in GKE, we do see slow container start holds container lock and blocks `docker ps` for Docker.

**Please note that for containers/sandboxes in unknown state, we still need to rely on containerd `TaskExit` event, because they were not successfully loaded, there are no corresponding exit monitors running.**

Since we'll support containerd 1.2 for a long time, I'd like to cherry-pick this into 1.2.
/cc @crosbymichael 

### DIFF
--- a/pkg/server/container_start.go
+++ b/pkg/server/container_start.go
@@ -38,64 +38,48 @@ import (
 
 // StartContainer starts the container.
 func (c *criService) StartContainer(ctx context.Context, r *runtime.StartContainerRequest) (retRes *runtime.StartContainerResponse, retErr error) {
-	container, err := c.containerStore.Get(r.GetContainerId())
+	cntr, err := c.containerStore.Get(r.GetContainerId())
 	if err != nil {
 		return nil, errors.Wrapf(err, "an error occurred when try to find container %q", r.GetContainerId())
 	}
 
-	var startErr error
-	// update container status in one transaction to avoid race with event monitor.
-	if err := container.Status.UpdateSync(func(status containerstore.Status) (containerstore.Status, error) {
-		// Always apply status change no matter startContainer fails or not. Because startContainer
-		// may change container state no matter it fails or succeeds.
-		startErr = c.startContainer(ctx, container, &status)
-		return status, nil
-	}); startErr != nil {
-		return nil, startErr
-	} else if err != nil {
-		return nil, errors.Wrapf(err, "failed to update container %q metadata", container.ID)
-	}
-	return &runtime.StartContainerResponse{}, nil
-}
-
-// startContainer actually starts the container. The function needs to be run in one transaction. Any updates
-// to the status passed in will be applied no matter the function returns error or not.
-func (c *criService) startContainer(ctx context.Context,
-	cntr containerstore.Container,
-	status *containerstore.Status) (retErr error) {
 	id := cntr.ID
 	meta := cntr.Metadata
 	container := cntr.Container
 	config := meta.Config
 
-	// Return error if container is not in created state.
-	if status.State() != runtime.ContainerState_CONTAINER_CREATED {
-		return errors.Errorf("container %q is in %s state", id, criContainerStateToString(status.State()))
+	// Set starting state to prevent other start/remove operations against this container
+	// while it's being started.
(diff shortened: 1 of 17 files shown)


jterry75:
LGTM! This is huge!

Random-Liu:
Actually I have a different idea about handling container/sandbox in unknown state (failed to be loaded at startup).

* Based on the state machine https://github.com/containerd/cri/blob/master/pkg/store/sandbox/status.go#L24-L53 and https://github.com/containerd/cri/blob/master/pkg/store/container/status.go#L31-L61, the only allowed operation to container/sandbox in unknown state is STOP;
* We've changed Kubernetes to always try stopping container/sandbox in unknown state before removing https://github.com/kubernetes/kubernetes/pull/73802

So actually we don't really care about whether a sandbox/container in unknown state exits itself or not, because Kubernetes will stop it right away. So we only need to know, when stopping it whether it is stopped or not.

In this case, we can start a synchronized `Wait` in stop for container/sandbox in unknown state, and we can **completely get rid of the dependency on `TaskExit`**.

I'll make the change.

mikebrow:
flying back today... will review in the morning!

pkg/store/sandbox/sandbox_test.go
```diff
@@ -125,21 +125,16 @@ func TestSandboxStore(t *testing.T) {
 		assert.Equal(sb, got)
 	}
 
-	t.Logf("should not be able to get unknown sandbox")
+	t.Logf("should be able to get sandbox with Get")
```
mikebrow: ... in unknown state

pkg/store/sandbox/status.go
```diff
@@ -53,23 +53,17 @@ import (
 // +-------------> DELETED
 
 // State is the sandbox state we use in containerd/cri.
-// It includes init and unknown, which are internal states not defined in CRI.
+// It includes unknown, which are internal states not defined in CRI.
```
mikebrow: /s/are/is/

which is an internal state not defined in CRI.. (INIT state has been removed)

pkg/server/container_start.go
```diff
@@ -38,64 +38,49 @@ import (
...
+				status.FinishedAt = time.Now().UnixNano()
+				status.ExitCode = errorStartExitCode
+				status.Reason = errorStartReason
+				status.Message = retErr.Error()
+				return status, nil
+			}); err != nil {
+				logrus.WithError(err).Errorf("failed to set start failure state for container %q", id)
+			}
+		}
+		// Reset starting if start failed.
```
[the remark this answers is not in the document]
Random-Liu: Yeah, forgot to remove the comment. Done.

pkg/server/container_start.go
```diff
@@ -133,15 +118,61 @@ func (c *criService) startContainer(ctx context.Context,
...
+			return status, errors.New("container is already in starting state")
+		}
+		status.Starting = true
+		return status, nil
+	})
+}
+
+// resetContainerStarting resets the container starting state on start failure. So
+// that we could remove the container later.
+func resetContainerStarting(container containerstore.Container) error {
```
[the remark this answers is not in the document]
Random-Liu: Should be idempotent

pkg/store/sandbox/sandbox_test.go
Random-Liu: Done

pkg/store/sandbox/status.go
Random-Liu: Done.

pkg/server/container_start.go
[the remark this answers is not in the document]
mikebrow: fair enough the set was not idempotent.. so the unset being so was unsettling ;-) But ok by me either way.

Issue #1132, opened 2019-04-10
Use `Wait` instead of `TaskExit` event.


Random-Liu:
https://github.com/containerd/containerd/issues/3177
https://github.com/containerd/containerd/issues/3125

It is super difficult to make event reliable, and event is never designed to be 100% reliable.
We should use `Wait` instead.

Random-Liu:
We may consider cherry-picking this into 1.2 if the fix is not too large.

PR #1136, opened 2019-04-18
Cherrypick #1133 release 1.2

Random-Liu:
Cherrypick #1133 release 1.2

This is not a small change, but it is an important bug fix. Since we are going to support containerd 1.2 for a long time, we should at least fix this in 1.2.

For #1132, https://github.com/containerd/containerd/issues/3125, https://github.com/containerd/containerd/issues/3177

### DIFF
--- a/pkg/server/container_remove.go
+++ b/pkg/server/container_remove.go
@@ -105,6 +105,9 @@ func setContainerRemoving(container containerstore.Container) error {
 		if status.State() == runtime.ContainerState_CONTAINER_UNKNOWN {
 			return status, errors.New("container state is unknown, to stop first")
 		}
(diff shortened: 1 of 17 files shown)


Random-Liu:
The verify failed because of the go 1.10/1.11 gofmt difference.

Random-Liu:
Mark hold based on https://github.com/containerd/cri/pull/1136#pullrequestreview-228440564.

Will get this in when master test looks good.

Random-Liu:
I think #1133 is stable enough.

The only issue I've found in node-e2e-benchmark job is https://github.com/containerd/containerd/pull/3244, which is not related to this change.

I'll unhold this change. :)

PR #1146, opened 2019-05-10
Forward sandbox config to PullImage request

jterry75:


Random-Liu:
Good catch.
/lgtm

PR #1147, opened 2019-05-10
Move from unix to syscall package for SIG* signals

jterry75:
To support cross compilation for SIG* signals perfer the syscall package over
the unix package.


Random-Liu:
/lgtm

Will merge after test passes.

PR #1148, opened 2019-05-10
Override container log path to empty if either of sandbox log directory or container log path is empty.

congliu01:
Override container log path to empty if either of sandbox log directory or container log path is empty.

If either of the log paths is empty, container won't start properly since the log path is not properly specified.

Signed-off-by: Cong Liu <conliu@google.com>

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -192,9 +192,14 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	meta.ImageRef = image.ID
 	meta.StopSignal = image.ImageSpec.Config.StopSignal
 
-	// Get container log path.
-	if config.GetLogPath() != "" {
+	// Validate log paths and compose full container log path.
+	if sandboxConfig.GetLogDirectory() != "" && config.GetLogPath() != "" {
 		meta.LogPath = filepath.Join(sandboxConfig.GetLogDirectory(), config.GetLogPath())
+		logrus.Debugf("Composed container full log path %q using sandbox log dir %q and container log path %q",
+			meta.LogPath, sandboxConfig.GetLogDirectory(), config.GetLogPath())
+	} else {
+		logrus.Infof("Logging will be disabled due to empty log paths for sandbox (%q) or container (%q)",
+			sandboxConfig.GetLogDirectory(), config.GetLogPath())
 	}
 
 	containerIO, err := cio.NewContainerIO(id,


Random-Liu:
@congliu01 You need to sign your commit with `git commit --amend -s`.

pkg/server/container_create.go
```diff
@@ -193,7 +193,7 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	meta.StopSignal = image.ImageSpec.Config.StopSignal
 
 	// Get container log path.
-	if config.GetLogPath() != "" {
+	if config.GetLogPath() != "" && sandboxConfig.GetLogDirectory() != "" {
```
mikebrow: can we update the comment line above and add a logrus.Debugf to aide in debugging the invalid config issue..

congliu01: Done. Let me know if you like the new comment and log message.

pkg/server/container_create.go
```diff
@@ -192,8 +192,9 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	meta.ImageRef = image.ID
 	meta.StopSignal = image.ImageSpec.Config.StopSignal
 
-	// Get container log path.
-	if config.GetLogPath() != "" {
+	// Validate log paths and compose full container log path.
+	logrus.Debugf("Composing container full log path using sandbox log dir %q and container log path %q", sandboxConfig.GetLogDirectory(), config.GetLogPath())
```
mikebrow: was more thinking that we needed log output for the else or ```if sandboxConfig.GetLogDirectory() == "" || config.GetLogPath() == ""``` case when we are not composing the full log path

mikebrow:
gitvalidation has a 90 character limit on the commit message :(

pkg/server/container_create.go
congliu01: Added more verbose logging in the `else` branch as well.

pkg/server/container_create.go
```diff
@@ -192,9 +192,12 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
...
 	meta.StopSignal = image.ImageSpec.Config.StopSignal
 
-	// Get container log path.
-	if config.GetLogPath() != "" {
+	// Validate log paths and compose full container log path.
+	if sandboxConfig.GetLogDirectory() != "" && config.GetLogPath() != "" {
 		meta.LogPath = filepath.Join(sandboxConfig.GetLogDirectory(), config.GetLogPath())
+		logrus.Debugf("Composed container full log path %q using sandbox log dir %q and container log path %q", meta.LogPath, sandboxConfig.GetLogDirectory(), config.GetLogPath())
+	} else {
+		logrus.Debugf("Empty log path in either sandbox or container config. Container full log path: %q, sandbox log dir: %q, container log path: %q", meta.LogPath, sandboxConfig.GetLogDirectory(), config.GetLogPath())
```
mikebrow: nit can we move this one to warning or error level...

pkg/server/container_create.go
```diff
@@ -192,9 +192,12 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	meta.ImageRef = image.ID
 	meta.StopSignal = image.ImageSpec.Config.StopSignal
 
-	// Get container log path.
-	if config.GetLogPath() != "" {
+	// Validate log paths and compose full container log path.
+	if sandboxConfig.GetLogDirectory() != "" && config.GetLogPath() != "" {
 		meta.LogPath = filepath.Join(sandboxConfig.GetLogDirectory(), config.GetLogPath())
+		logrus.Debugf("Composed container full log path %q using sandbox log dir %q and container log path %q", meta.LogPath, sandboxConfig.GetLogDirectory(), config.GetLogPath())
```
Random-Liu: The line is too long.

pkg/server/container_create.go
Random-Liu: 1) No need to print the paths, because they are empty.
2) I think a lot of crictl examples don't have log path specified, I think an info might be good enough.

congliu01: 1. Well there is at least one of them is empty but probably not both. So I will still print the paths for easy debugging.
2. Yeah I think info is OK.

pkg/server/container_create.go
congliu01: Split into 2 lines.

PR #1152, opened 2019-05-20
Return gRPC codes.DeadlineExceeded for all timeout operations

jterry75:
In order for a calling client to properly detect when a call failed due to a
timeout CRI should be returning codes.DeadlineExceeded from gRPC by contract.

### DIFF
--- a/pkg/server/container_stop.go
+++ b/pkg/server/container_stop.go
@@ -33,13 +33,6 @@ import (
 	containerstore "github.com/containerd/cri/pkg/store/container"
 )
 
-// killContainerTimeout is the timeout that we wait for the container to
-// be SIGKILLed.
-// The timeout is set to 1 min, because the default CRI operation timeout
-// for StopContainer is (2 min + stop timeout). Set to 1 min, so that we
-// have enough time for kill(all=true) and kill(all=false).
-const killContainerTimeout = 1 * time.Minute
-
 // StopContainer stops a running container with a grace period (i.e., timeout).
 func (c *criService) StopContainer(ctx context.Context, r *runtime.StopContainerRequest) (*runtime.StopContainerResponse, error) {
 	// Get container config from container store.
@@ -141,11 +134,21 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
 			return errors.Wrapf(err, "failed to stop container %q", id)
 		}
 
-		if err = c.waitContainerStop(ctx, container, timeout); err == nil || errors.Cause(err) == ctx.Err() {
-			// Do not SIGKILL container if the context is cancelled.
-			return err
+		sigTermCtx, sigTermCtxCancel := context.WithTimeout(ctx, timeout)
+		err = c.waitContainerStop(sigTermCtx, container)
+		if err == nil {
+			// Container stopped on first signal no need for SIGKILL
+			sigTermCtxCancel()
+			return nil
 		}
-		logrus.WithError(err).Errorf("An error occurs during waiting for container %q to be stopped", id)
+		// If the parent context was cancelled or exceeded return immediately
+		if ctx.Err() != nil {
+			sigTermCtxCancel()
+			return ctx.Err()
+		}
+		sigTermCtxCancel()
+		// sigTermCtx was exceeded. Send SIGKILL
+		logrus.Debugf("Stop container %q with signal %v timed out", id, sig)
 	}
 
 	logrus.Infof("Kill container %q", id)
@@ -154,21 +157,19 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
 	}
 
 	// Wait for a fixed timeout until container stop is observed by event monitor.
-	if err = c.waitContainerStop(ctx, container, killContainerTimeout); err == nil {
-		return nil
+	err = c.waitContainerStop(ctx, container)
+	if err != nil {
+		return errors.Wrapf(err, "an error occurs during waiting for container %q to be killed", id)
 	}
-	return errors.Wrapf(err, "an error occurs during waiting for container %q to be killed", id)
+	return nil
 }
 
-// waitContainerStop waits for container to be stopped until timeout exceeds or context is cancelled.
-func (c *criService) waitContainerStop(ctx context.Context, container containerstore.Container, timeout time.Duration) error {
-	timeoutTimer := time.NewTimer(timeout)
-	defer timeoutTimer.Stop()
+// waitContainerStop waits for container to be stopped until context is
+// cancelled or the context deadline is exceeded.
+func (c *criService) waitContainerStop(ctx context.Context, container containerstore.Container) error {
 	select {
 	case <-ctx.Done():
-		return errors.Wrapf(ctx.Err(), "wait container %q is cancelled", container.ID)
-	case <-timeoutTimer.C:
-		return errors.Errorf("wait container %q stop timeout", container.ID)
+		return errors.Wrapf(ctx.Err(), "wait container %q", container.ID)
 	case <-container.Stopped():
 		return nil
 	}
--- a/pkg/server/sandbox_stop.go
+++ b/pkg/server/sandbox_stop.go
@@ -142,18 +142,15 @@ func (c *criService) stopSandboxContainer(ctx context.Context, sandbox sandboxst
 		return errors.Wrap(err, "failed to kill sandbox container")
 	}
 
-	return c.waitSandboxStop(ctx, sandbox, killContainerTimeout)
+	return c.waitSandboxStop(ctx, sandbox)
 }
 
-// waitSandboxStop waits for sandbox to be stopped until timeout exceeds or context is cancelled.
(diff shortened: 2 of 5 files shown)


jterry75:
Actually, @mikebrow, @Random-Liu - I re-read the documentation. I don't believe that we should have a timeout on these calls at all.

1. RunPodSandbox has no user defied timeout. Thus should not return DeadlineExceeded (or any error for timeout). The call should be blocking.
2. StopContainer has a user defined timeout. BUT this timeout is the time before the container is SIGKILL'ed. So the workflow is timeout > 0, send SIGTERM, if not exited after timeout, send SIGKILL, but still the call should not return DeadlineExceeded (or any error for timeout). This call should be blocking.

If a client wants a deadline, they should do that client side to protect against a long running call. IE:
```go
ctx := context.WithDeadline(context.Background(), time.Minute * 1)
resp, err := client.StopPodSandbox(ctx, request)
if err != nil {
    if err == context.DeadlineExceeded {
        // retry or other client idea
    }
}
```

Random-Liu:
@jterry75 Client side deadline sounds more correct to me.

The client may even set context timeout for `RunPodSandbox`, and in that case only client side can reliably get that error. If we want to preserve that error and return on our side, we need to handle each error carefully, because all of them could be caused by context timeout excceeded.

jterry75:
The server can always return gRPC DeadlineExceeded on a gRPC timeout. But I don't think we need an additional timeout here is all I am saying. If your ok with it I will just remove the `killContainerTimeout` from `waitContainerStop` and `waitSandboxStop`

Random-Liu:
> The server can always return gRPC DeadlineExceeded on a gRPC timeout. But I don't think we need an additional timeout here is all I am saying. If your ok with it I will just remove the killContainerTimeout from waitContainerStop and waitSandboxStop. We should always honor the clients ctx though and return DeadlineExceeded on ctx.Done() if the user initiates the Cancel/Deadline. Does that make sense?

Sorry, I misunderstood what you meant.

Yeah, it makes sense, just remove the kill timeout and use context.

jterry75:
@Random-Liu - Broke into three commits. One for each since they are technically three separate issues. Let me know what you think.

-Justin

jterry75:
Looking at the failures. It seems the tests dont like the new error code I think

jterry75:
Ok lets try this again. It turns out that the ExecSync test in cri-test appears to require that after `Timeout` the code return an error. I'm not sure that is correct for what the API says but I just put that code back to how it was except honoring the parent context. Lets see if it passes tests now

jterry75:
@Random-Liu - any issues?

pkg/server/sandbox_stop.go
```diff
@@ -142,18 +144,18 @@ func (c *criService) stopSandboxContainer(ctx context.Context, sandbox sandboxst
...
+func (c *criService) waitSandboxStop(ctx context.Context, sandbox sandboxstore.Sandbox) error {
 	select {
 	case <-ctx.Done():
-		return errors.Wrapf(ctx.Err(), "wait sandbox container %q is cancelled", sandbox.ID)
-	case <-timeoutTimer.C:
-		return errors.Errorf("wait sandbox container %q stop timeout", sandbox.ID)
+		if ctx.Err() == context.Canceled {
+			return status.Errorf(codes.Canceled, "wait sandbox container %q is cancelled", sandbox.ID)
+		}
+		return status.Errorf(codes.DeadlineExceeded, "wait sandbox container %q stop timeout", sandbox.ID)
```
Random-Liu: If `ctx` has a client side timeout, what error do we get here for that timeout? `DeadlineExceeded` or `Canceled`?

pkg/server/container_stop.go
```diff
@@ -154,21 +156,22 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
...
-// waitContainerStop waits for container to be stopped until timeout exceeds or context is cancelled.
-func (c *criService) waitContainerStop(ctx context.Context, container containerstore.Container, timeout time.Duration) error {
-	timeoutTimer := time.NewTimer(timeout)
-	defer timeoutTimer.Stop()
+// waitContainerStop waits for container to be stopped until context is
+// cancelled or the context deadline is exceeded.
+func (c *criService) waitContainerStop(ctx context.Context, container containerstore.Container) error {
 	select {
 	case <-ctx.Done():
-		return errors.Wrapf(ctx.Err(), "wait container %q is cancelled", container.ID)
```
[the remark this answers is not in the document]
jterry75: Yea no problem. Was just keeping it the same as before in case a client was reading the text. I can make the update if you don't have concerns here?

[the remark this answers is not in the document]
Random-Liu: I see.

We can keep the message, but the error could still be ctx.Err(). :) Just in case new context errors added in the future.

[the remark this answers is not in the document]
jterry75: @Random-Liu - I think I misunderstood your question. Returning `ctx.Err()` will not suffice because gRPC does not automatically translate this to `codes.DeadlineExceeded`. So you still end up with: `rpc error: code = Unknown desc = context deadline exceeded` A client really needs `code` to be set appropriately. We could certainly add a handler in `instrumentedService` that tries to detect the `errors.Cause(err) == context.DeadlineExceeded` and translate it there but we cant just return `ctx.Err()` here.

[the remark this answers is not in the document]
Random-Liu: I see... My bad.

Random-Liu:
> I don't believe that we should have a timeout on these calls at all.

> The server can always return gRPC DeadlineExceeded on a gRPC timeout. But I don't think we need an additional timeout here is all I am saying. If your ok with it I will just remove the killContainerTimeout from waitContainerStop and waitSandboxStop. We should always honor the clients ctx though and return DeadlineExceeded on ctx.Done() if the user initiates the Cancel/Deadline. Does that make sense?

I'm a little bit confused. You said we don't need additional timeout, but you are still adding it, right?

Random-Liu:
> We could certainly add a handler in instrumentedService that tries to detect the errors.Cause(err) == context.DeadlineExceeded and translate it there but we cant just return ctx.Err() here.

I think that is the only reliable way to make sure `codes.DeadlineExceeded` is returned.

In the current code, when context is cancelled or timeout, it may return immediately before hitting the code you add, e.g. `process.Kill` could return error.

jterry75:
> > We could certainly add a handler in instrumentedService that tries to detect the errors.Cause(err) == context.DeadlineExceeded and translate it there but we cant just return ctx.Err() here.
> 
> I think that is the only reliable way to make sure `codes.DeadlineExceeded` or `codes.Cancel` is returned.
> 
> In the current code, when context is cancelled or timeout, it may return before hitting the code you add, e.g. context is canceled before `process.Kill` is called.

Np. I will do a wrap in instrumented service. Update soon.

jterry75:
@Random-Liu - Precursor for the last bit of work. Then I will rebase this PR. https://github.com/containerd/cri/pull/1177

jterry75:
@mikebrow - Any idea on these failures? Looks like containerd daemon stopped maybe? Not sure why we would get a connection refused that has nothing to do with this change

mikebrow:
ran fine on tip... so it was probably just beating on a back nic or whatnot.. Ill restart...

jterry75:
Lol. Thanks

jterry75:
@mikebrow / @yujuhong - Thoughts?

pkg/server/container_stop.go
```diff
@@ -141,11 +134,21 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
 			return errors.Wrapf(err, "failed to stop container %q", id)
 		}
 
-		if err = c.waitContainerStop(ctx, container, timeout); err == nil || errors.Cause(err) == ctx.Err() {
-			// Do not SIGKILL container if the context is cancelled.
-			return err
+		sigTermCtx, sigTermCtxCancel := context.WithTimeout(ctx, timeout)
```
Random-Liu: I think it is fine to keep the original timeout timer, making it purely context based doesn't make the code simpler.

Random-Liu:
LGTM with one nit.

pkg/server/container_stop.go
jterry75: I can do that for the container_stop code if you like it better. When the timer expires we just need to return `errors.Wrapf(context.DeadlineExceeded, "wait container %q stop timeout", container.ID)` and this should work too. Want me to do that?

jterry75: What we lose in that model is the ability for the client context to cause the cancellation

Random-Liu: I'm fine with using context. I'm just not a fan of having a `cancel` not in `defer`, which may cause trouble when we move code around in the future. :P

jterry75: The call is idempotent. I suppose we could defer it and cancel it early. Want me to send a subsequent update?

jterry75: Lol: https://github.com/containerd/cri/pull/1207

PR #1153, opened 2019-05-24
Update parseDNSOptions maxDNSSearches in error message

jterry75:
Minor correctness. We should use the value of the const in the error message
instead of hard coding it in the string so if maxDNSSearches ever changes so
does the error.

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -500,7 +500,7 @@ func parseDNSOptions(servers, searches, options []string) (string, error) {
 	resolvContent := ""
 
 	if len(searches) > maxDNSSearches {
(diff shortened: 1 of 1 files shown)


pkg/server/sandbox_run.go
```diff
@@ -500,7 +500,7 @@ func parseDNSOptions(servers, searches, options []string) (string, error) {
 	resolvContent := ""
 
 	if len(searches) > maxDNSSearches {
-		return "", errors.New("DNSOption.Searches has more than 6 domains")
+		return "", errors.Errorf("DNSOption.Searches has more than %d domains", maxDNSSearches)
```
Random-Liu: I think we also want to include the actual searches number.

Issue #1154, opened 2019-05-28
Make cni plugin config file num configurable


kuramal:
Make cni plugin config file num configurable， let containerd can make  mutil network card

Random-Liu:
CRI only reports one pod IP, how the other NICs of the pod will be used?

I'm fine with this change, just want to understand the background more. :)

Random-Liu:
@kuramal SG. Get the go-cni change in first, and then update the vendor on the CRI side.

kuramal:
@Random-Liu  hi, the pr of go-cni,  have already merged. Should I update the go-cni code of vendor, or just update the commit id in vendor.conf?

Random-Liu:
@kuramal Yeah! :D

Random-Liu:
#1155 is merged, this can be closed. @kuramal Thanks for adding this support! :)

PR #1155, opened 2019-05-28
add cni plugin config file max num config

kuramal:
Make cni plugin config file num configurable， let containerd can make mutil network card

#1154

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -149,6 +149,12 @@ The explanation and default value of each configuration item are as follows:
     # conf_dir is the directory in which the admin places a CNI conf.
     conf_dir = "/etc/cni/net.d"
 
+    # max_conf_num is the max number of CNI plugin config file will load
+    # By default, the containerd will load only 1 CRI plugin. If you want load multiple CRI plugins,
+    # set max_conf_num > 1. You should know exactly each CRI plugin do, whether they are compatible.
+    # max i<= 0 means no limit.
+    max_conf_num = 1
+
     # conf_template is the file path of golang template used to generate
     # cni config.
     # If this is set, containerd will generate a cni config file from the
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -71,6 +71,8 @@ type CniConfig struct {
 	NetworkPluginBinDir string `toml:"bin_dir" json:"binDir"`
 	// NetworkPluginConfDir is the directory in which the admin places a CNI conf.
 	NetworkPluginConfDir string `toml:"conf_dir" json:"confDir"`
+	// NetworkPluginMaxConfNum is the max number of plugin config file will load
+	NetworkPluginMaxConfNum int `toml:"max_conf_num" json:"maxConfNum"`
 	// NetworkPluginConfTemplate is the file path of golang template used to generate
 	// cni config.
 	// When it is set, containerd will get cidr from kubelet to replace {{.PodCIDR}} in
@@ -202,6 +204,7 @@ func DefaultConfig() PluginConfig {
 		CniConfig: CniConfig{
 			NetworkPluginBinDir:       "/opt/cni/bin",
 			NetworkPluginConfDir:      "/etc/cni/net.d",
+			NetworkPluginMaxConfNum:   1,
 			NetworkPluginConfTemplate: "",
 		},
 		ContainerdConfig: ContainerdConfig{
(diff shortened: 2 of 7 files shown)


pkg/config/config.go
```diff
@@ -71,6 +71,8 @@ type CniConfig struct {
 	NetworkPluginBinDir string `toml:"bin_dir" json:"binDir"`
 	// NetworkPluginConfDir is the directory in which the admin places a CNI conf.
 	NetworkPluginConfDir string `toml:"conf_dir" json:"confDir"`
+	// NetworkPluginMaxConfNum is the max num of plugin file will load
```
Random-Liu: s/max num of plugin file/max number of plugin config file

mikebrow: Can we explain what this number is and how it is used?

mikebrow: Can you link me to the value is used pls?

pkg/config/config.go
```diff
@@ -192,6 +194,7 @@ func DefaultConfig() PluginConfig {
 		CniConfig: CniConfig{
 			NetworkPluginBinDir:       "/opt/cni/bin",
 			NetworkPluginConfDir:      "/etc/cni/net.d",
+			NetworkPluginMaxConfNum:   1,
```
mikebrow: hmm.. should we be loading this from some vendor constant over in cni or go-cni? Sort of goes with the above question, where is this going to be used. Are there any test cases?

pkg/config/config.go
Random-Liu: Ha, yeah we should update docs/config.md.

kuramal: @Random-Liu Yeah, that's right. 
@mikebrow In our case, we add a second NIC of the pod, and set a LAN ip for it.The second NIC is Optional，and we control the ips for every deployment stable.

mikebrow: Ok, interesting use case... 

I don't see code in cni proper for using this value.. but I may have missed it? Is this for a custom plugin or a to be dropped pr for cni? Is this value just a pass through value? Is there a mockup of this pattern somewhere? Inquiring minds want to know :-)

Random-Liu: The value is used in `go-cni` library. :)

Random-Liu: @kuramal Can you update https://github.com/containerd/cri/blob/master/docs/config.md, and include some explanation of the option, like https://github.com/containerd/cri/pull/1155#discussion_r291023199?

Random-Liu:
Waiting for document change, otherwise LGTM

pkg/config/config.go
mikebrow: ok found it in go-cni/opts.go 

// loadFromConfDir detects network config files from the
// configured cni config directory and load them. max is
// the maximum network config to load (max i<= 0 means no limit).

pkg/config/config.go
mikebrow: this replaces the default .. which is by default 1 so no observable behavior change here.. just odd that we are not vendoring the default from go-cni

pkg/config/config.go
```diff
@@ -71,6 +71,8 @@ type CniConfig struct {
 	NetworkPluginBinDir string `toml:"bin_dir" json:"binDir"`
 	// NetworkPluginConfDir is the directory in which the admin places a CNI conf.
 	NetworkPluginConfDir string `toml:"conf_dir" json:"confDir"`
+	// NetworkPluginMaxConfNum is the max number of plugin config file will load
```
mikebrow: // NetworkPluginMaxConfNum (default = 1) is the max number of plugin config files that
// will be loaded from the cni config directory. Set the value to 0 to load all config files (no arbitrary limit).

mikebrow: And say something like that in the config.md.

pkg/config/config.go
kuramal: @Random-Liu  Done

docs/config.md
```diff
@@ -149,6 +149,12 @@ The explanation and default value of each configuration item are as follows:
     # conf_dir is the directory in which the admin places a CNI conf.
     conf_dir = "/etc/cni/net.d"
 
+    # max_conf_num is the max number of CNI plugin config file will load
+    # By default, the containerd will load only 1 CRI plugin. If you want load multiple CRI plugins,
+    # set max_conf_num > 1. You should know exactly each CRI plugin do, whether they are compatible.
+    # max i<= 0 means no limit.
```
mikebrow: typo nits:
    # max_conf_num specifies the maximum number of CNI plugin config files that will be loaded from the CNI config directory.    
    # By default, containerd will load only 1 CNI plugin config file. If you want to load multiple CNI plugin config files set max_conf_num to the number desired or 0, as 0 means no limit (load all CNI plugin config files from the CNI config directory).

mikebrow:
@Random-Liu I'm fine with merging this and doing a follow up for the text descriptions. The code is fine!

Random-Liu:
/lgtm

@mikebrow Can you send a doc update? Thanks!

mikebrow:
sure first thing in the morn...

PR #1156, opened 2019-05-28
Make sure exec process is killed when context is canceled.

Random-Liu:
When context is cancelled, the `process.Kill` won't work. We need to kill the process in the cleanup defer.


jterry75:
LGTM

PR #1163, opened 2019-06-11
Use the new v2 config

Random-Liu:
Ref https://github.com/containerd/containerd/pull/3322.
For https://github.com/containerd/cri/issues/1168.

Use the new v2 config in the test. Let's start testing this.


cluster/gce/configure.sh
```diff
@@ -158,29 +158,30 @@ fi
 log_level="${CONTAINERD_LOG_LEVEL:-"info"}"
 max_container_log_line="${CONTAINERD_MAX_CONTAINER_LOG_LINE:-16384}"
 cat > ${config_path} <<EOF
+version = 2
 # Kubernetes requires the cri plugin.
-required_plugins = ["cri"]
+required_plugins = ["io.containerd.grpc.v1.cri"]
```
mikebrow: ouch...  

is this prefix necessary now?

Random-Liu: Yes, it is now.

mikebrow: I see it now thx. Ouch but thx.

PR #1171, opened 2019-06-12
Fix http proxy ENV when TLS is enabled.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1169.

Support http proxy env.

The transport is the same with https://github.com/containerd/containerd/blob/master/cmd/ctr/commands/resolver.go#L82

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -21,9 +21,11 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/errdefs"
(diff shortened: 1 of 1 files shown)


pkg/server/image_pull.go
```diff
@@ -296,15 +299,14 @@ func (c *criService) getResolver(ctx context.Context, ref string, cred func(stri
 		}
 
 		if registryTLSConfig, ok := c.config.Registry.TLSConfigs[u.Host]; ok {
-			tlsTransport, err := c.getRegistryTLSTransport(ctx, registryTLSConfig)
+			transport.TLSClientConfig, err = c.getTLSConfig(registryTLSConfig)
```
mikebrow: that's better!

pkg/server/image_pull.go
```diff
@@ -340,3 +341,19 @@ func (c *criService) getResolver(ctx context.Context, ref string, cred func(stri
 	}
 	return resolver, desc, nil
 }
+
+// newTransport returns a new HTTP transport used to pull image.
+func newTransport() *http.Transport {
+	return &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
```
mikebrow: looks good..

Random-Liu:
> create some test cases unit and crictl or node that test TLS, Proxy, and other off the beaten path patterns for registry pull tests

It is hard to test this. :(

> create some common code in containerd that can be used here and in ctr.

Added TODO

Random-Liu:
Just added a TODO, apply lgtm based on https://github.com/containerd/cri/pull/1171#pullrequestreview-248761802

PR #1180, opened 2019-06-19
Fix runc and critools version in release.

Random-Liu:
Runc and critools are both versioned with containerd version, e.g. 1.2.6, without the fix.

### DIFF
--- a/hack/install/install-critools.sh
+++ b/hack/install/install-critools.sh
@@ -29,7 +29,7 @@ GOPATH=${TMPGOPATH}
 #Install crictl
 checkout_repo ${CRITOOL_PKG} ${CRITOOL_VERSION} ${CRITOOL_REPO}
 cd ${GOPATH}/src/${CRITOOL_PKG}
-make
+make VERSION=${CRITOOL_VERSION}
 ${SUDO} make install -e BINDIR=${CRITOOL_DIR} GOPATH=${GOPATH}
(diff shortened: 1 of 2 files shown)


thaJeztah:
Will it be confusing to give runc a new version, given that they also tag versions upstream?

mikebrow:
> Will it be confusing to give runc a new version, given that they also tag versions upstream?

this is only for containerd/cri dev/testing purposes... not prod build.

mikebrow:
@thaJeztah it is not a new version it is the version we have selected to pull and build from the git their repos eg: https://github.com/containerd/cri/blob/master/vendor.conf#L37

I believe what he meant above is we were previously failing to pass in the version we selected to the make builds and thus possibly using the wrong version tag in those builds. (The containerd version vs the one we have chosen in containerd/cri.) We have a who's on first cyclic versioning issue between containerd/containerd and containerd/cri that may affect workflow.

mikebrow:
at some point we probably need to stop building release binaries this way... I completely forgot we were still packaging up runc and crictl in a cri package.

thaJeztah:
Oh! Must have misunderstood, thought this would version all of them the same as the containerd version (thus "runc v1.2.7" if that's the version of containerd)

Random-Liu:
> Oh! Must have misunderstood, thought this would version all of them the same as the containerd version (thus "runc v1.2.7" if that's the version of containerd)

Without the fix, it would version all of them as the containerd version.

Issue #1181, opened 2019-06-21
Support a TCP endpoint


alculquicondor:
For an scenario where:

- processes running in the machine and remote clients are trusted
- an extra TCP/socket hop is undesired.

CRI could be served in a new TCP endpoint, additionally to being served in the main containerd socket. Exposing the whole containerd server in TCP might still be a security risk containerd/containerd#1324

TLS support could be provided as well.

mikebrow:
We can discuss this.  But IMO it really needs auth and/or tls with signed certs (not self signed) before the PR should be approved. Thus some config will be necessary to provide the certs, and/or some key repository integrated for the auth. Recognize that the typical config for container runtime integrations runs as root. Granted we'll be supporting non-root configs soon, but ... cart before the horse.

mikebrow:
See https://github.com/containerd/containerd/commit/b911ae3428717763377d038958597383c7ed0491

^ that commit does include tls with certs so... addressed my above issue.

alculquicondor:
I missed that feature too. I submitted a new PR :)

Random-Liu:
I think we should close this one.

Feel free to reopen if you need anything else.

PR #1182, opened 2019-06-21
Add extra TCP server for CRI service

alculquicondor:
If an endpoint is provided, CRI will be served over TCP, in addition to the service in the global containerd socket.

The feature is disabled by default.

#1181

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -138,6 +138,9 @@ type PluginConfig struct {
 	CniConfig `toml:"cni" json:"cni"`
 	// Registry contains config related to the registry
 	Registry Registry `toml:"registry" json:"registry"`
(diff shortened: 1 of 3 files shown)


Random-Liu:
Containerd supports tcp https://github.com/containerd/containerd/blob/2f69be5594cecacdbe42076ed12c7039a0638a2c/services/server/server.go#L108, we just need to register cri as an tcp service there.

mikebrow:
Cool:
```
mike@mike-VirtualBox:~$ containerd config default
version = 2
...
[grpc]
  address = "/run/containerd/containerd.sock"
  tcp_address = ""
  tcp_tls_cert = ""
  tcp_tls_key = ""
```

mikebrow:
> We can support both ttrpc and grpc

I think you meant grpc over both socket and tcp :-)

PR #1187, opened 2019-06-24
Add option to register on TCP server

alculquicondor:
If enabled, register CRI on containerd's TCP endpoint

#1181

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -15,6 +15,10 @@ version = 2
 # The 'plugins."io.containerd.grpc.v1.cri"' table contains all of the server options.
 [plugins."io.containerd.grpc.v1.cri"]
 
+  # disable_tcp_service disables serving CRI on the TCP server.
+  # Note that a TCP server is enabled for containerd if TCPAddress is set in section [grpc].
+  disable_tcp_service = true
+
   # stream_server_address is the ip address streaming server is listening on.
   stream_server_address = "127.0.0.1"
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -138,6 +138,8 @@ type PluginConfig struct {
 	CniConfig `toml:"cni" json:"cni"`
 	// Registry contains config related to the registry
 	Registry Registry `toml:"registry" json:"registry"`
+	// DisableTCPService disables serving CRI on the TCP server.
+	DisableTCPService bool `toml:"disable_tcp_service" json:"disableTCPService"`
 	// StreamServerAddress is the ip address streaming server is listening on.
 	StreamServerAddress string `toml:"stream_server_address" json:"streamServerAddress"`
 	// StreamServerPort is the port streaming server is listening on.
@@ -219,6 +221,7 @@ func DefaultConfig() PluginConfig {
 				},
 			},
 		},
+		DisableTCPService:   true,
 		StreamServerAddress: "127.0.0.1",
 		StreamServerPort:    "0",
 		StreamIdleTimeout:   streaming.DefaultConfig.StreamIdleTimeout.String(), // 4 hour
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -171,9 +171,15 @@ func NewCRIService(config criconfig.Config, client *containerd.Client) (CRIServi
 // Register registers all required services onto a specific grpc server.
 // This is used by containerd cri plugin.
 func (c *criService) Register(s *grpc.Server) error {
-	instrumented := newInstrumentedService(c)
-	runtime.RegisterRuntimeServiceServer(s, instrumented)
-	runtime.RegisterImageServiceServer(s, instrumented)
+	return c.register(s)
+}
+
+// RegisterTCP register all required services onto a GRPC server on TCP.
+// This is used by containerd CRI plugin.
+func (c *criService) RegisterTCP(s *grpc.Server) error {
+	if !c.config.DisableTCPService {
+		return c.register(s)
+	}
 	return nil
 }
 
@@ -264,6 +270,13 @@ func (c *criService) Close() error {
 	if err := c.streamServer.Stop(); err != nil {
 		return errors.Wrap(err, "failed to stop stream server")
 	}
+	return nil
+}
+
+func (c *criService) register(s *grpc.Server) error {
+	instrumented := newInstrumentedService(c)
+	runtime.RegisterRuntimeServiceServer(s, instrumented)
+	runtime.RegisterImageServiceServer(s, instrumented)
 	return nil
 }
 


alculquicondor:
@mikebrow what needs to be done to not be a WIP?

mikebrow:
when @Random-Liu gets back we should have a discussion regarding how this should work..

I'm thinking we don't need a new flag to enable it, we should build off the containerd config. 

It might be good to consider an integration test maybe use a modified crictl validate to hook up over tcpip instead of the socket.

alculquicondor:
Sounds good, thank you.
My rationale behind the flag is that some users might want to enable the TCP endpoint for containerd without exposing CRI.

The issue with crictl is that the validation for the protocol comes from kubelet code. But maybe we can craft a PR that maintains the kubelet expectations but we can still have crictl code work for TCP.

mikebrow:
There is already way to not enable/disable the cri plugin (cri services). Mixing the end point config just seems odd given that we have one socket ATM for containerd runtime (with an optional one for image). Now if we had cri as a different potential socket ala how runtime/image are done then I agree we could have different config options.  But if we do that then we would also want tcp config options ala how the config options were done in containerd... eg ip, cert, and key for cri.  I note that when they added tcp I don't think they split runtime and image? At least I didn't notice if they did.

IMO we should just ride on the containerd config vs have a different set of socket/tcp config options for cri. This to reduce complexity.

I just presumed when this is done it would be cri compatible, but over tcp instead of a unix socket.

alculquicondor:
I'm confused, don't we register both services in the same grpc server? https://github.com/containerd/cri/blob/322cd489653645e70f9bb2f17d2953c8fecbc66c/pkg/server/service.go#L175

mikebrow:
I see my msg was confusing.. edited. Yes we do and that's my point. If we are loaded we register period, there is no switch.

mikebrow:
to your edit .. yeah we don't "split" cri runtime and image services across two different containerd sockets.. even though that is an option at the cri layer.

docs/config.md
```diff
@@ -11,6 +11,10 @@ The explanation and default value of each configuration item are as follows:
 # The "plugins.cri" table contains all of the server options.
 [plugins.cri]
 
+  # enable_tcp_service indicates to enable serving CRI on TCP.
+  # TCPAddress is required in section [grcp].
```
Random-Liu: s/grcp/grpc

Random-Liu:
> I'm thinking we don't need a new flag to enable it, we should build off the containerd config.

I also had similar problem when reviewing the code.

Let me think about it more.

Random-Liu:
No containerd services have register TCP yet (no `RegisterTCP` implementation in containerd services), so I think at least for now this is a per service decision. Given so, I think an `enable_tcp_service` option makes sense.

docs/config.md
alculquicondor: Done

mikebrow:
> No containerd services have `RegisterTCP` implemented yet

That is my point, I think the intent was if it is defined, that is the "enable" switch that enables it for all services.. Or maybe I read it wrong? If we have a switch off it should be enable by default at any rate, IMO.

alculquicondor:
> That is my point, I think the intent was if it is defined, that is the "enable" switch that enables it for all services.. Or maybe I read it wrong? If we have a switch off it should be enable by default at any rate, IMO.

Each plugin can decide to implement `RegisterTCP`. If they do, they are added to the TCP server. However, the TCP server is disabled if there is no TCP address defined in the configuration.

The point of the flag is to give users the ability to disable cri on TCP even if they want TCP support for other plugins. Having it disabled by default is probably the best option because most users won't need it.

mikebrow:
> > That is my point, I think the intent was if it is defined, that is the "enable" switch that enables it for all services.. Or maybe I read it wrong? If we have a switch off it should be enable by default at any rate, IMO.
> 
> Each plugin can decide to implement `RegisterTCP`. If they do, they are added to the TCP server. However, the TCP server is disabled if there is no TCP address defined in the configuration.
> 
> The point of the flag is to give users the ability to disable cri on TCP even if they want TCP support for other plugins. Having it disabled by default is probably the best option because most users won't need it.

Let me put it another way. We do not have an enable flag for the socket.

alculquicondor:
> Let me put it another way. We do not have an enable flag for the socket.

True. Given that CRI is the first and only service to implement `RegisterTCP`, users have already all the means to have socket support only.

We can always re-evaluate the addition of a flag in the future. The approach of having the flag is simply more cautious. @Random-Liu wdyt?

mikebrow:
I'm not against a flag entirely... but I do think it should be a disable flag that defaults to false, giving customers a way to disable tcp for cri if they want to enable it for the rest of the containerd services by specifying the tcp options.

Random-Liu:
CRI endpoint is a "privileged" endpoint, because users can use it to mutate node state, e.g. pull image or create container. I do believe that users sometimes only want to enable tcp endpoint for readonly services, e.g. metrics, so **an option to disable CRI is required**.

> I'm not against a flag entirely... but I do think it should be a disable flag that defaults to false, giving customers a way to disable tcp for cri if they want to enable it for the rest of the containerd services by specifying the tcp options.

I agree.  My understanding about your comment is that: an `enable_tcp_service` can't really guarantee that CRI tcp service is enabled, because users still need to enable containerd tcp server, which may confuse users. However, a `disable_tcp_service` can guarantee that the CRI tcp service is always disabled, no matter tcp server is configured or not. Given so, **`disable_tcp_service` does sound better to me**. :)

mikebrow:
> CRI endpoint is a "privileged" endpoint, because users can use it to mutate node state, e.g. pull image or create container. I do believe that users sometimes only want to enable tcp endpoint for readonly services, e.g. metrics, so **an option to disable CRI is required**.
> 
> > I'm not against a flag entirely... but I do think it should be a disable flag that defaults to false, giving customers a way to disable tcp for cri if they want to enable it for the rest of the containerd services by specifying the tcp options.
> 
> I agree. My understanding about your comment is that: an `enable_tcp_service` can't really guarantee that CRI tcp service is enabled, because users still need to enable containerd tcp server. This may confuse users. However, a `disable_tcp_service` can guarantee that the CRI tcp service is always disabled, no matter tcp server is configured or not. In this case, **`disable_tcp_service` does sound better to me**. :)
> 
> As for the default value, I prefer `true`, because 1) it is not a common use case; 2) even though the default is `false`, it is not enabled by default, because tcp server is not enabled by default.

fair points. I strongly agree! 

We should make a note of pushing a doc PR to the containerd tcp config such that a user setting tcp config knows that, by default, the CRI tcp endpoints are disabled and thus have to be explicitly enabled by turning off the disable flag. 

At some point in the future it may make sense to have an array of white/black flags for tcp enabling of service(api) end points for CRI and possibly other services in containerd. But this will do for now :-)

alculquicondor:
Thanks for all the feedback. I think we arrived to a pretty good conclusion.

Regarding updating containerd docs, it might be a little early now, given that CRI is the only service using the endpoint. I expect users of CRI to first come to our docs and then learn that they have to add the endpoint in main containerd's config.

PR #1188, opened 2019-06-24
Update docs to v2 config

alculquicondor:
Config changed in #1163

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -8,8 +8,12 @@ for more information about containerd config.
 
 The explanation and default value of each configuration item are as follows:
 ```toml
-# The "plugins.cri" table contains all of the server options.
-[plugins.cri]
+# Use config version 2 to enable new configuration fields.
+# Config file is parsed as version 1 by default.
+version = 2
+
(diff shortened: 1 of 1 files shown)


alculquicondor:
anything else here?

mikebrow:
no just waiting for Lantau to get back as he had worked on this..

docs/config.md
```diff
@@ -8,8 +8,8 @@ for more information about containerd config.
 
 The explanation and default value of each configuration item are as follows:
 ```toml
```
Random-Liu: Add `version = 2`. Version 2 is not the default, we need to explicitly specify it.

See https://github.com/containerd/containerd/pull/3337

mikebrow: version 1 is the default if the field is not specified...

mikebrow: Here's my suggested text: 

The version field in the config file specifies the config's version. If no version number is specified inside
 the config file then it is assumed to be a version 1 config and parsed as such. Use version = 2 to enable/access to new/current config fields.

alculquicondor: Done.

PR #1198, opened 2019-07-13
feature: support ARM platform image

fuweid:
The image has multiple manifests for ARM platform. During unpacking
image for ARM platform, Manifest will read all the matched manifests,
sort and find best manifest. But Pull action will limit manifest number
as one. Therefore, the unpack will fail and say "content digest: xxx" is
not found.

Based on this case, use Fetch action which doesn't limit manifest
number andd call the unpack separately.

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -97,15 +97,18 @@ func (c *criService) PullImage(ctx context.Context, r *runtime.PullImageRequest)
 	// image has already been converted.
 	isSchema1 := desc.MediaType == containerdimages.MediaTypeDockerSchema1Manifest
 
-	image, err := c.client.Pull(ctx, ref,
+	img, err := c.client.Fetch(ctx, ref,
 		containerd.WithSchema1Conversion,
 		containerd.WithResolver(resolver),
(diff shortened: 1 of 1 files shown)


fuweid:
This is one option. Other option is to update image.Manifest.


https://github.com/containerd/containerd/blob/master/images/image.go#L191
```
case MediaTypeDockerSchema2ManifestList, ocispec.MediaTypeImageIndex:
			p, err := content.ReadBlob(ctx, provider, desc)
			if err != nil {
				return nil, err
			}

			var idx ocispec.Index
			if err := json.Unmarshal(p, &idx); err != nil {
				return nil, err
			}

			if platform == nil {
				return idx.Manifests, nil
			}

			var descs []ocispec.Descriptor
			for _, d := range idx.Manifests {
				if d.Platform == nil || platform.Match(*d.Platform) {
					descs = append(descs, d)
				}
			}

			wasIndex = true

			return descs, nil

		}
```

we can do the sort here and make sure that the Walk will not touch `not-found` manifest. 

cc @dmcgowan

fuweid:
close because merged containerd/containerd#3406

PR #1199, opened 2019-07-17
always update image in cri image store

sofat1989:
Fix: https://github.com/containerd/containerd/issues/3401

### DIFF
--- a/pkg/store/image/image.go
+++ b/pkg/store/image/image.go
@@ -106,9 +106,6 @@ func (s *Store) update(ref string, img *Image) error {
 		return nil
 	}
 	if oldExist {
-		if oldID == img.ID {
-			return nil
-		}
 		// Updated. Remove tag from old image.
 		s.store.delete(oldID, ref)
 	}


Random-Liu:
Hm, I feel like the problem is that we should not cache `containerd.Image` in our image store, because it can never be up to date given that fields like `Target`, `Labels` and `UpdatedAt` can all change.

mikebrow:
> Hm, I feel like the problem is that we should not cache `containerd.Image` in our image store, because it is hard to keep it up to date given that fields like `Target`, `Labels` and `UpdatedAt` can all change.

I don't disagree at this point in time.  It made more sense back when we were a remote daemon than it does now that we are a local plugin.

Random-Liu:
See https://github.com/containerd/cri/pull/1202

Random-Liu:
This should have been fxied by #1202.

Close for now.

Feel free to reopen if that doesn't solve your problem. :) @sofat1989

PR #1200, opened 2019-07-17
Assign ImageSpec User if SecurityContext is not set

jterry75:
By default the SecurityContext for Container activation can contain a Username
UID, GID. The order of precedences is username, UID, GID. If none of these
options are specified as a last resort attempt to set the ImageSpec username.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -227,10 +227,15 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	userstr, err := generateUserString(
 		securityContext.GetRunAsUsername(),
 		securityContext.GetRunAsUser(),
-		securityContext.GetRunAsGroup(),
-	)
+		securityContext.GetRunAsGroup())
+
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to generate user string")
+	}
+	if userstr == "" {
+		// Lastly, since no user override was passed via CRI try to set via OCI
+		// Image
+		userstr = image.ImageSpec.Config.User
 	}
 	if userstr != "" {
 		specOpts = append(specOpts, oci.WithUser(userstr))
@@ -589,7 +594,20 @@ func generateApparmorSpecOpts(apparmorProf string, privileged, apparmorEnabled b
 	}
 }
 
-// generateUserString generates valid user string based on OCI Image Spec v1.0.0.
+// generateUserString generates valid user string based on OCI Image Spec
+// v1.0.0.
+//
+// CRI defines that the following combinations are valid:
+//
+// (none) -> ""
+// username -> username
+// username, uid -> username
+// username, uid, gid -> username:gid
+// username, gid -> username:gid
+// uid -> uid
+// uid, gid -> uid:gid
+// gid -> error
+//
 // TODO(random-liu): Add group name support in CRI.
 func generateUserString(username string, uid, gid *runtime.Int64Value) (string, error) {
 	var userstr, groupstr string
--- a/pkg/server/container_create_test.go
+++ b/pkg/server/container_create_test.go
@@ -1230,3 +1230,74 @@ func TestDisableCgroup(t *testing.T) {
 	t.Log("cgroup path should be empty")
 	assert.Empty(t, spec.Linux.CgroupsPath)
 }
+
+func TestGenerateUserString(t *testing.T) {
+	type testcase struct {
+		// the name of the test case
+		name string
+
+		u        string
+		uid, gid *runtime.Int64Value
+
+		result        string
+		expectedError bool
+	}
+	testcases := []testcase{
+		{
+			name:   "Empty",
+			result: "",
+		},
+		{
+			name:   "Username Only",
+			u:      "testuser",
+			result: "testuser",
+		},
+		{
+			name:   "Username, UID",
+			u:      "testuser",
+			uid:    &runtime.Int64Value{Value: 1},
+			result: "testuser",
+		},
+		{
+			name:   "Username, UID, GID",
+			u:      "testuser",
+			uid:    &runtime.Int64Value{Value: 1},
+			gid:    &runtime.Int64Value{Value: 10},
+			result: "testuser:10",
+		},
+		{
+			name:   "Username, GID",
+			u:      "testuser",
+			gid:    &runtime.Int64Value{Value: 10},
+			result: "testuser:10",
+		},
+		{
+			name:   "UID only",
+			uid:    &runtime.Int64Value{Value: 1},
+			result: "1",
+		},
+		{
+			name:   "UID, GID",
+			uid:    &runtime.Int64Value{Value: 1},
+			gid:    &runtime.Int64Value{Value: 10},
+			result: "1:10",
+		},
+		{
+			name:          "GID only",
+			gid:           &runtime.Int64Value{Value: 10},
+			result:        "",
+			expectedError: true,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			r, err := generateUserString(tc.u, tc.uid, tc.gid)
+			if tc.expectedError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tc.result, r)
+		})
+	}
+}
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -169,6 +169,11 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to generate user string")
 	}
+	if userstr == "" {
+		// Lastly, since no user override was passed via CRI try to set via OCI
+		// Image
+		userstr = image.ImageSpec.Config.User
+	}
 	if userstr != "" {
 		specOpts = append(specOpts, oci.WithUser(userstr))
 	}


pkg/server/container_create.go
```diff
@@ -594,14 +595,17 @@ func generateApparmorSpecOpts(apparmorProf string, privileged, apparmorEnabled b
...
-func generateUserString(username string, uid, gid *runtime.Int64Value) (string, error) {
+func generateUserString(username, imageUser string, uid, gid *runtime.Int64Value) (string, error) {
 	var userstr, groupstr string
 	if uid != nil {
 		userstr = strconv.FormatInt(uid.GetValue(), 10)
 	}
 	if username != "" {
 		userstr = username
 	}
+	if userstr == "" {
```
mikebrow: if (userstr == "" && gid == nil)

pkg/server/container_create_test.go
```diff
@@ -1215,3 +1215,130 @@ func TestDisableCgroup(t *testing.T) {
...
+			i:      "testimage",
+			uid:    &runtime.Int64Value{Value: 1},
+			gid:    &runtime.Int64Value{Value: 10},
+			result: "1:10",
+		},
+		{
+			name:   "Image, GID",
+			i:      "testimage",
+			gid:    &runtime.Int64Value{Value: 10},
+			result: "testimage:10",
```
mikebrow: expectedError: true,

jterry75:
> See comments... I think we want it to fail if image name is set and gid is set but no uid.

Thanks Mike. I re-read the doc's for CRI and agree with what you said. We now only allow image if no username/uid/gid is passed via CRI. Otherwise we validate that the combination is acceptable to CRI and do not use the image.

mikebrow:
Looks good.  

Let's think this over a bit. Unless it already works this way with docker shim... we should consider making this an opt in feature at the node level, vs just adding it by default to existing. 

Down the road it might even make sense to have the pod spec enable/filter this option.

jterry75:
@mikebrow - I'm okay with it being opt in. It turns out this is the default behavior of Docker/CRI-O when no user credential is passed so I *assumed* this was a bug here. Should I make it a config option for now and we can remove the config opt if we agree it should be default?

mikebrow:
Let's go with the default docker shim behavior then.. and add an issue to track documenting the behavior somewhere and probably inviting some node folks to the discussion for further clarification, and probably Mr. Crosby for his opinion.

jterry75:
@crosbymichael / @yujuhong / @PatrickLang - Do any of you have input / opinions on this? My understanding from the code (Docker and CRI-O) is that its the default behavior that if no `LinuxContainerSecurityContext.RunAsUser`, `LinuxContainerSecurityContext.RunAsGroup`, `LinuxContainerSecurityContext.RunAsUsername` is passed that we would (as a last resort) attempt to use the OCI ImageSpec `v1.ImageConfig.User` which matches to the default user from the Dockerfile (if set).

Does this PR seem reasonable?

crosbymichael:
I think it seems reasonable, however, I believe that the user from the image can also come in the format of `1000:1000` specifying both the uid/gid or `root:bin` as well

jterry75:
> I think it seems reasonable, however, I believe that the user from the image can also come in the format of `1000:1000` specifying both the uid/gid or `root:bin` as well

Interesting your right. The struct documentation is not as verbose as the https://github.com/opencontainers/image-spec/blob/e562b04403929d582d449ae5386ff79dd7961a11/config.md

Will update.

jterry75:
> > I think it seems reasonable, however, I believe that the user from the image can also come in the format of `1000:1000` specifying both the uid/gid or `root:bin` as well
> 
> Interesting your right. The struct documentation is not as verbose as the https://github.com/opencontainers/image-spec/blob/e562b04403929d582d449ae5386ff79dd7961a11/config.md
> 
> Will update.

Wait. @crosbymichael - You are correct. But what is wrong with the current code for handling that case? We will return a `string` that is the `imageUser` value which can be any of the values in the docs. The `oci.WithUser` call will split it back apart in any format. Right?

crosbymichael:
This line: https://github.com/containerd/cri/pull/1200/files#diff-c656bacd6cb0fe9a7f64814b01256decR627 makes me thing that you could end up in a situation where it adds 3 `:` because userstr already contains `root:root` then gets a `root:root:root` at the end if gid was specified

mikebrow:
> This line: https://github.com/containerd/cri/pull/1200/files#diff-c656bacd6cb0fe9a7f64814b01256decR627 makes me thing that you could end up in a situation where it adds 3 `:` because userstr already contains `root:root` then gets a `root:root:root` at the end if gid was specified

my read of the code: userstr can not be imageUser if gid is not nil; groupstr is nil if gid is nil; so we don't get to the line of code with the concatenation while using imageUser from the container.

But the code is tricky. We should add testcases for the additional container default use cases mentioned by michael and the oci spec.

mikebrow:
that said... it looks like we "currently" let the caller pass in whatever on the username and optionally also add a gid without us generating an error ... thus we already have one code path that can generate root:root:root... Hmm. Needs more testcases and a decision if we should add a check in the code to fail early or let it pass through as is.

jterry75:
Ah I see the concern now. Based on the CRI docs lets say that if `RunAsUsername` contains `:` its invalid. The expected use from CRI is that `RunAsUsername` CAN be used with `RunAsGroup` (GID) but NOT with `RunAsUser` (UID). So my expectation is that we should allow `RunAsUsername:RunAsGroup`. Thus if `RunAsUsername` already had a `:Group/GID` it would fail to begin with. 

So basically:
RunAsUser -> Valid
RunAsUser, RunAsGroup -> Valid
RunAsUsername (no `:`) -> Valid
RunAsUsername, RunAsGroup -> Valid
RunAsUser, RunAsUsername -> Error
RunAsUsername (with `:`) -> Error
None of the above, default to "whatever" is in `ImageSpec.Config.User` which may be any combination by itself.

Random-Liu:
@jterry75 This logic is already handled by Kubelet.

Kubelet will get default user for a given image, consolidate it with `RunAsUser` field and then pass the correct user to us.

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_container.go#L197

mikebrow:
> @jterry75 This logic is already handled by Kubelet.
> 
> Kubelet gets default user for a given image, consolidate it with `RunAsUser` field and then pass the correct user to us. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_container.go#L197
> 
> Is there anything doesn't work for you so that you have to add this?

That works for the kublet case, ok. But just for our side of the CRI API sake, I think we should test.

jterry75:
@Random-Liu - Yes this works for Kublet as you pointed out. I noticed this writing some tests that use a default image user and was surprised to see them failing. When I figured it out it was because they were attempting to be run as root instead of the default image user.

I agree with @mikebrow we should handle this in CRI since things like crictl don't use Kublet.

Random-Liu:
@jterry75 SG. I just want to make sure you were not encountering any kubelet issue that broke the logic.

Having the CRI plugin default to image user SGTM

jterry75:
@Random-Liu - Great thanks!

jterry75:
@Random-Liu - Sorry maybe I misunderstood. Did I need to do anything else here?

pkg/server/container_create.go
```diff
@@ -222,6 +222,7 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	// the spec for us.
 	userstr, err := generateUserString(
 		securityContext.GetRunAsUsername(),
+		image.ImageSpec.Config.User,
```
Random-Liu: Can we use the `getUserFromImage` helper?
And only pass in uid and username when CRI hasn't specified them.

jterry75: I believe the latest commit is what you want. And we cant use `getUserFromImage` because we need the "full" name which could be `uid:gid` etc. The parsing will happen in `oci.WithUser` so we dont want to remove the context before passing.

jterry75:
@Random-Liu - PTAL. I think the updates are what you want.

pkg/server/container_create.go
```diff
@@ -589,7 +594,13 @@ func generateApparmorSpecOpts(apparmorProf string, privileged, apparmorEnabled b
 	}
 }
 
-// generateUserString generates valid user string based on OCI Image Spec v1.0.0.
+// generateUserString generates valid user string based on OCI Image Spec
+// v1.0.0.
+//
+// CRI defines that the following combinations are valid:
+//
+// uid, uid/gid, username, username/gid
```
Random-Liu: `/` -> `:`?

pkg/server/container_create_test.go
```diff
@@ -1230,3 +1230,78 @@ func TestDisableCgroup(t *testing.T) {
 	t.Log("cgroup path should be empty")
 	assert.Empty(t, spec.Linux.CgroupsPath)
 }
+
+func Test_generateUserString(t *testing.T) {
```
Random-Liu: nit: TestGenerateUserString to follow the convention in the file? :)

pkg/server/container_create_test.go
```diff
@@ -1230,3 +1230,78 @@ func TestDisableCgroup(t *testing.T) {
...
+		{
+			name:          "GID only",
+			gid:           &runtime.Int64Value{Value: 10},
+			result:        "",
+			expectedError: true,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			r, err := generateUserString(tc.u, tc.uid, tc.gid)
```
Random-Liu: nit: Use `assert` library to follow convention?

```golang
if tc.expectedErr {
  assert.Error(t, err)
} else {
  assert.NoError(t, err)
}
assert.Equal(t, tc.result, r)
```

pkg/server/container_create.go
Random-Liu: SG.

pkg/server/container_create_test.go
jterry75: The func is un-exported `generateUserString` and it turns out that go test doesnt find `TestgenerateUserString` because the first letter after `Test` is lowercase. So the only way to have it match is:
1. Name it as you did by changing the case of the func we are testing (which seems odd since its not exported).
2. Insert the `_` after the `Test` keyword. 

Not tied to either let me know what you prefer.

pkg/server/container_create.go
jterry75: I can see why this is confusing. I changed the comment.

pkg/server/container_create_test.go
jterry75: Cool!

jterry75:
@Random-Liu  - 17th times the charm?! :)

pkg/server/container_create_test.go
Random-Liu: I think Kubernetes's convention is `TestGernerateUserString`. :)

For example: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/eviction/helpers.go#L102
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/eviction/helpers_test.go#L43

pkg/server/container_create.go
Random-Liu: Ha, I was confused... Sorry. :P

pkg/server/container_create_test.go
Random-Liu: @jterry75 I also can't find `Test_` in containerd/containerd, let's follow the Kubernetes convention then.

Otherwise, ready to merge. :)

jterry75:
Got it this time for sure!

jterry75:
Thanks!

PR #1202, opened 2019-07-24
Do not cache image handler.

Random-Liu:
/cc @sofat1989 

Fixes https://github.com/containerd/containerd/issues/3401.


mikebrow:
the change looks good... need to figure out the test failure though.

Random-Liu:
There is some problem with this approach. Let me think about it a little bit.

Random-Liu:
@mikebrow Fixed. The problem is that when `ImageStatus` returns an image, the image ID reference in containerd may not be ready yet.

I changed the PR to get an actual reference from the reference list, which ensures that the reference should exist in containerd.

Random-Liu:
We should cherry-pick this to 1.2 to fix containerd/containerd#3401

PR #1207, opened 2019-07-25
Move context cancel into defer.

Random-Liu:
Just a small cleanup. :) Move `sigTermCtxCancel` into defer.

If `waitContainerStop` returns error, sigTermCtx should have already `Done`;
If `waitContainerStop` returns nil, sigTermCtxCancel will be called in defer immediately.

### DIFF
--- a/pkg/server/container_stop.go
+++ b/pkg/server/container_stop.go
@@ -135,18 +135,16 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
 		}
 
 		sigTermCtx, sigTermCtxCancel := context.WithTimeout(ctx, timeout)
+		defer sigTermCtxCancel()
 		err = c.waitContainerStop(sigTermCtx, container)
 		if err == nil {
 			// Container stopped on first signal no need for SIGKILL
-			sigTermCtxCancel()
 			return nil
 		}
 		// If the parent context was cancelled or exceeded return immediately
 		if ctx.Err() != nil {
-			sigTermCtxCancel()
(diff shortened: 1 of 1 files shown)


pkg/server/container_stop.go
```diff
@@ -135,18 +135,16 @@ func (c *criService) stopContainer(ctx context.Context, container containerstore
...
 			// Container stopped on first signal no need for SIGKILL
-			sigTermCtxCancel()
 			return nil
 		}
 		// If the parent context was cancelled or exceeded return immediately
 		if ctx.Err() != nil {
-			sigTermCtxCancel()
 			return ctx.Err()
 		}
-		sigTermCtxCancel()
```
jterry75: We should keep this one. The other two will run early at `return` but this one wont return until the SigKill call. This will be an early context to release `sigTermCtx` resources early.

mikebrow: what happens if you call cancel twice?  

how about just calling the cancel immediately after the wait?

Random-Liu: > We should keep this one. The other two will run early at return but this one wont return until the SigKill call. This will be an early context to release sigTermCtx resources early.

@mikebrow I just tried Cancel twice doesn't cause issue.

@jterry75 If the code reaches here, it means that the context has timeout. :) However, I'm fine with putting cancel here to be safe. :)

jterry75: Lol good point. Too tired. :)

PR #1225, opened 2019-08-07
Add flag to overload default privileged host device behaviour

### LINKED ISSUE #1213: runtime based privileged container translation
Right now when a container is specified as privileged, containerd/cri would expose all host devices to the container (https://github.com/containerd/cri/blob/master/pkg/server/container_create.go#L389). While the model works perfectly for runc, for vm based container such as kata, -- a lot of host devices doesn't make sense to be accessed in the guest.

Can we add runtime based policy here so that it is possible for a runtime to specify that no extra host devices to be added to the generated container spec (e.g. not appending `customopts.WithPrivilegedDevices`)?

Related kata issue: https://github.com/kata-containers/runtime/issues/1568

As mentioned in the kata issue [comment](https://github.com/kata-containers/runtime/issues/1568#issuecomment-516298599), it is possible for kata to have its own privileged container translation bypassing cri/containerd all together. But I still think the best way is to fix it in containerd/cri.

awprice:
This commit adds a flag to the runtime config that allows overloading of the default
privileged behaviour. When the flag is enabled on a runtime, host devices won't
be appended to the runtime spec if the container is run as privileged.

By default the flag is false to maintain the current behaviour of privileged.

Fixes #1213

Signed-off-by: Alex Price <aprice@atlassian.com>

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -28,6 +28,13 @@ import (
 	"github.com/containerd/containerd/contrib/seccomp"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/oci"
+	"github.com/containerd/cri/pkg/annotations"
+	"github.com/containerd/cri/pkg/config"
+	customopts "github.com/containerd/cri/pkg/containerd/opts"
+	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
+	cio "github.com/containerd/cri/pkg/server/io"
+	containerstore "github.com/containerd/cri/pkg/store/container"
+	"github.com/containerd/cri/pkg/util"
 	"github.com/containerd/typeurl"
 	"github.com/davecgh/go-spew/spew"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -35,13 +42,6 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
-
-	"github.com/containerd/cri/pkg/annotations"
-	customopts "github.com/containerd/cri/pkg/containerd/opts"
-	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
-	cio "github.com/containerd/cri/pkg/server/io"
-	containerstore "github.com/containerd/cri/pkg/store/container"
-	"github.com/containerd/cri/pkg/util"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


Random-Liu:
Update `docs/config.md`

pkg/server/container_create.go
```diff
@@ -168,7 +168,8 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	logrus.Debugf("Use OCI runtime %+v for sandbox %q and container %q", ociRuntime, sandboxID, id)
 
 	spec, err := c.generateContainerSpec(id, sandboxID, sandboxPid, config, sandboxConfig,
-		&image.ImageSpec.Config, append(mounts, volumeMounts...), ociRuntime.PodAnnotations)
+		&image.ImageSpec.Config, append(mounts, volumeMounts...), ociRuntime.PodAnnotations,
```
Random-Liu: nit: Let's pass in the `ociRuntime` instead to shorten the arg list

awprice: 👍 fixed

awprice:
👍 `docs/config.md` updated

AkihiroSuda:
Proposal for porting over this to Moby/Docker: https://github.com/moby/moby/issues/39697

jongwu:
hi @awprice, this feature sounds great. but how to use this feature. I open privileged_without_host_devices in /etc/containerd/config.toml and set this file as the default config for containerd. then I start a container use docker with --privileged. when I type the command "ls /dev" in container, I saw that the output is the same with host.
I think I may take the wrong step. can someone tell me how to test this feature? @Random-Liu @mikebrow @awprice

awprice:
@jongwu `privileged_without_host_devices` will only work when starting containers with the CRI plugin/interface in containerd, which is usually done through Kubernetes or with crictl.

There is a PR to add this functionality to docker, it might be worth trying out: https://github.com/moby/moby/pull/39702

jongwu:
thanks @awprice, I am not familiar with k8s, I think crictl is simple to test this feature, can you give me some info about how to test it using crictl. thanks.

awprice:
No worries @jongwu. Here is the documentation for crictl - https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md

jongwu:
thanks @awprice

jongwu:
hi @awprice, sorry to trouble you again, I have tried for a full day to start a container with --privileged, but failed.
my pod-config is 
{
    "metadata": {
        "name": "true222222222222",
        "namespace": "default",
        "attempt": 1,
        "uid": "hdishd83djaidwnduwk28bcsb"
    },
    "log_directory": "/tmp",
    "linux": {
        "securityContext": true
    }
}

my container-config.json is
{
  "metadata": {
      "name": "busybox1111111111111"
  },
  "image":{
      "image": "busybox"
  },
  "command": [
      "top"
  ],
  "log_path":"busybox/0.log",
  "linux": {
      "SecurityContext": {
              "privileged": true
      }
  }
}
the privileged_without_host_devices is false.
after I start a container, the picture under /dev is totally different with host.
I don't know what's wrong with it. can you help me? thanks!

PR #1227, opened 2019-08-07
New registry mirror

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1201.

This PR:
1) Updated containerd to 1.3.0-beta.0;
2) Use new containerd `RegistryHosts` library;
3) Add `registry.configs` field, move TLS config into `registry.configs` and deprecate `registry."registry.url".auths`;
4) Move config validation code into `pkg/config`.


jterry75:
@Random-Liu - Did you mean to PR this?

Random-Liu:
@jterry75 
https://github.com/containerd/containerd/pull/3503 is merged.

Updated containerd to include that.

jterry75:
Awesome thanks for doing that

pkg/server/image_pull.go
```diff
@@ -280,68 +262,63 @@ func (c *criService) getTLSConfig(registryTLSConfig config.TLSConfig) (*tls.Conf
...
-	// Try mirrors in order first, and then try default host name.
-	for _, e := range c.config.Registry.Mirrors[refspec.Hostname()].Endpoints {
-		u, err := url.Parse(e)
+		// Try mirrors in order, and then try the default registry if not tried.
+		defaultHost, err := docker.DefaultHost(host)
 		if err != nil {
-			return nil, imagespec.Descriptor{}, errors.Wrapf(err, "parse registry endpoint %q", e)
+			return nil, errors.Wrapf(err, "get default host")
+		}
+		defaultEndpoint := "https://" + defaultHost
```
mikebrow: docker registry can also communicate without ssl, right? ...do we need a todo for this? there may be an issue already open for non https defaults I think I saw one a while back.

Random-Liu: If users want insecure access, they can configure mirror `http://xxxx` to overwrite the default.

The default should be `https` :)

pkg/server/image_pull.go
```diff
@@ -95,7 +94,10 @@ func (c *criService) PullImage(ctx context.Context, r *runtime.PullImageRequest)
 	if ref != imageRef {
 		log.G(ctx).Debugf("PullImage using normalized image ref: %q", ref)
 	}
-	resolver, desc, err := c.getResolver(ctx, ref, c.credentials(r.GetAuth()))
+	resolver := docker.NewResolver(docker.ResolverOptions{
+		Hosts: c.registryHosts(r.GetAuth()),
+	})
+	_, desc, err := resolver.Resolve(ctx, ref)
```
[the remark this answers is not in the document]
Random-Liu: See https://github.com/containerd/containerd/blob/master/remotes/resolver.go#L35

The returned name can be different from the original reference.

We only care about whether the image is schema1 or not here.

Actually, I believe we can figure out whether an image is schema1 or not by checking labels. I can look into that and improve that later. :)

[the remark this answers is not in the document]
Random-Liu: Done https://github.com/containerd/cri/issues/1229

pkg/server/image_pull.go
```diff
@@ -280,68 +262,63 @@ func (c *criService) getTLSConfig(registryTLSConfig config.TLSConfig) (*tls.Conf
...
 
-// getResolver tries registry mirrors and the default registry, and returns the resolver and descriptor
-// from the first working registry.
-func (c *criService) getResolver(ctx context.Context, ref string, cred func(string) (string, string, error)) (remotes.Resolver, imagespec.Descriptor, error) {
-	refspec, err := reference.Parse(ref)
-	if err != nil {
-		return nil, imagespec.Descriptor{}, errors.Wrap(err, "parse image reference")
-	}
+// registryHosts is the registry hosts to be used by the resolver.
+func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHosts {
```
dmcgowan: Not sure how it was handled before but it is really better to handle this as a callback. It is best to assume that the credentials may not be accessible until they are requested by the server. Also, these auth configs should only be used when the host matches the server in the auth config. So for example if you are using a mirror of the Docker Hub and have hub credentials, those credentials should not be sent to the mirror, but the mirror can return 404 and have the client retry with the hub and send the credentials. Alternatively, in the configuration you should can explicitly authorize the mirror for another servers credentials if that is a use case you need to support.

pkg/config/config.go
```diff
@@ -257,3 +273,84 @@ const (
...
+		}
+		if r.Root != "" {
+			if r.Type != plugin.RuntimeLinuxV1 {
+				return errors.Errorf("`runtime_root` only works for runtime %s", plugin.RuntimeLinuxV1)
+			}
+			log.G(ctx).Warning("`runtime_root` is deprecated, please use runtime `options` instead")
+		}
+	}
+
+	// Validation for registry configurations.
```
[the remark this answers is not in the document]
Random-Liu: > validation and mapping deprecated Auths to Configs

Done.

> add todo to remove mapping in 1.4

Added TODO in the config object. :)

pkg/server/image_pull.go
mikebrow: Ok but if I remember right the issue opened by a customer was failing with an https error.. when setting a non http default.   hmm  maybe filter out https:/default if they included a mirror to http:/default...?

Random-Liu: @mikebrow You are right. I'll fix it.

pkg/server/image_pull.go
Random-Liu: > Not sure how it was handled before but it is really better to handle this as a callback.

Are you talking about the passed in `AuthConfig`?

> These auth configs should only be used when the host matches the server in the auth config

@dmcgowan The `ServerAddress` in the CRI `AuthConfig` is never properly set... https://github.com/kubernetes/enhancements/pull/1171#discussion_r310348605

It is something we should fix...

dmcgowan: > Are you talking about the passed in AuthConfig ?

Yes, ideally credentials should not be pre-fetched. Even if they are, the interface shouldn't require it.

> It is something we should fix...

Hmm, I guess the interface should pretend like it is doing the right thing at least, so it is easy to fix later. By "never properly set" do you mean it is empty or set to a completely useless value?

Random-Liu: @dmcgowan Even if users set `ServerAddress` in their docker file, once imported into Kubernetes, it is lost...

dmcgowan: So if it is empty, can we use it for all hosts. If it is set, then match the host. In this case, if Kubernetes fixes this issue, then CRI will already handle it properly.

If it is just set to the wrong value, that is harder to use.

Random-Liu: > So if it is empty, can we use it for all hosts. If it is set, then match the host. In this case, if Kubernetes fixes this issue, then CRI will already handle it properly.

SG. I was just thinking about that. :)

> Yes, ideally credentials should not be pre-fetched. Even if they are, the interface shouldn't require it.

What "credentials" are you referring to? The `tokens` in the `AuthConfig`? Or the `username`/`password`/`auth`?

BTW, can you briefly explain what `RegistryToken` and `IdentityToken` mean in the Docker `AuthConfig`? I believe those are copied from docker api in the first place, and we've never used them in Kubernetes.

pkg/server/image_pull.go

pkg/server/image_pull.go
Random-Liu: > So if it is empty, can we use it for all hosts. If it is set, then match the host. In this case, if Kubernetes fixes this issue, then CRI will already handle it properly.

Done

docs/registry.md
```diff
@@ -36,22 +36,25 @@ To configure the TLS settings for a specific registry, create/modify the `/etc/c
...
-  [plugins.cri.registry.configs."https://gcr.io".auth]
+  [plugins.cri.registry.configs."gcr.io".auth]
     username = ""
     password = ""
     auth = ""
     identitytoken = ""
 ```
 The meaning of each field is the same with the corresponding field in `.docker/config.json`.
 
+*NOTE: The registry host in the config MUST NOT contain url scheme or path.*
```
mikebrow: already mentioned... suggest picking one

pkg/config/config.go
```diff
@@ -129,7 +130,8 @@ type TLSConfig struct {
 type Registry struct {
 	// Mirrors are namespace to mirror mapping for all namespaces.
 	Mirrors map[string]Mirror `toml:"mirrors" json:"mirrors"`
-	// Configs are configs for each registry endpoint.
+	// Configs are configs for each registry.
+	// The key is the registry host address.
```
mikebrow: host address?  is that the same as fqdn or host name?  when I think address I usually think tcip address... Guess we are picking up the naming from the auth.ServerAddress field?

Random-Liu: Yeah, it is the hostname/ip

docs/registry.md
Random-Liu: Done. Moved into comment.

pkg/config/config.go
Random-Liu: Done.

Random-Liu:
Just squashed commits and addressed https://github.com/containerd/cri/pull/1227#pullrequestreview-273337150

Apply LGTM based on https://github.com/containerd/cri/pull/1227#pullrequestreview-273337150

PR #1233, opened 2019-08-10
allow non-mutual TLS

AkihiroSuda:
Previously, client keypair had needed to be specified even when unused.

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -37,6 +37,14 @@ To configure the TLS settings for a specific registry, create/modify the `/etc/c
 In the config example shown above, TLS mutual authentication will be used for communications with the registry endpoint located at https://my.custom.registry.
 `ca_file` is file name of the certificate authority (CA) certificate used to authenticate the x509 certificate/key pair specified by the files respectively pointed to by `cert_file` and `key_file`.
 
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Tested with Harbor 1.8

docs/registry.md
```diff
@@ -37,6 +37,8 @@ To configure the TLS settings for a specific registry, create/modify the `/etc/c
 In the config example shown above, TLS mutual authentication will be used for communications with the registry endpoint located at https://my.custom.registry.
 `ca_file` is file name of the certificate authority (CA) certificate used to authenticate the x509 certificate/key pair specified by the files respectively pointed to by `cert_file` and `key_file`.
 
+`cert_file` and `key_file` are not needed when TLS mutual authentication is unused.
```
mikebrow: Instead of saying it's not needed when TLS mutual auth is unused... can we expand on this by providing another example that only has the ca specified.. just to make it clear/easier to understand that is also a valid use for certain registries.

AkihiroSuda: done

PR #1235, opened 2019-08-13
Add wildcard mirror support.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1196.


pkg/server/image_pull.go
```diff
@@ -293,9 +293,7 @@ func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHost
 	return func(host string) ([]docker.RegistryHost, error) {
 		var registries []docker.RegistryHost
 
-		// Try mirrors in order, and then try the default registry if not tried.
-		endpoints, err := addDefaultEndpoint(
-			c.config.Registry.Mirrors[host].Endpoints, host)
+		endpoints, err := c.registryEndpoints(host)
 		if err != nil {
 			return nil, errors.Wrapf(err, "add default endpoint")
```
mikebrow: needs a new err message for registry endpoints call vs add default endpoint...

pkg/server/image_pull.go
```diff
@@ -347,9 +345,16 @@ func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHost
 	}
 }
 
-// addDefaultEndpoint add default registry endpoint if it does not
-// exist in the passed-in endpoint list.
-func addDefaultEndpoint(endpoints []string, host string) ([]string, error) {
+// registryEndpoints returns  endpoints for a given host. It adds default registry endpoint if it
```
mikebrow: s/  / /

mikebrow: should mention wildcarding.. here and in the md... to explain that host configured endpoints will be used first for matching host operations, then, if specified, wildcard specified host endpoints will be used, then if the default registry endpoint is not already specified via hostname or wildcard it will be added to the endpoints list for the given host.

pkg/server/image_pull_test.go
```diff
@@ -103,16 +105,50 @@ func TestParseAuth(t *testing.T) {
...
+			mirrors: map[string]criconfig.Mirror{
+				"*": {
+					Endpoints: []string{
+						"https://registry-1.io",
+						"https://registry-2.io",
+					},
+				},
+			},
+			host: "registry-3.io",
+			expected: []string{
```
mikebrow: needs a couples tests with host and wildcard specified...where host matches and doesn't to confirm ordering

Random-Liu: Good point.

pkg/server/image_pull.go

pkg/server/image_pull.go
Random-Liu: Done.

pkg/server/image_pull_test.go
Random-Liu: Done.

PR #1240, opened 2019-08-19
Fix apparmor for privileged.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1239.

@mikebrow Do you remember why we didn't clear apparmor when privileged = true before?
I don't remember we intentionally did that.


Random-Liu:
Updated the behavior based on Kubernetes documentation https://kubernetes.io/docs/tutorials/clusters/apparmor/#profile-reference

```
For Docker, this resolves to the docker-default profile for non-privileged containers, and unconfined (no profile) for privileged containers.
```

PR #1251, opened 2019-08-29
Add TaskMax=infinity in the systemd unit.

Random-Liu:
This PR:
1) added `TaskMax=infinity`, which we should have done long time ago;
2) ~~Removed `contrib/systemd-units`, because we already have https://github.com/containerd/containerd/blob/master/containerd.service~~ I found that `make release` still uses it, so keep it for now.

### DIFF
--- a/cluster/gce/cloud-init/master.yaml
+++ b/cluster/gce/cloud-init/master.yaml
@@ -45,6 +45,7 @@ write_files:
       # in the kernel. We recommend using cgroups to do container-local accounting.
       LimitNPROC=infinity
       LimitCORE=infinity
+      TasksMax=infinity
       ExecStartPre=/sbin/modprobe overlay
       ExecStart=/home/containerd/usr/local/bin/containerd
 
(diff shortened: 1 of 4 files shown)


cluster/gce/cloud-init/master.yaml
```diff
@@ -45,6 +45,9 @@ write_files:
       # in the kernel. We recommend using cgroups to do container-local accounting.
       LimitNPROC=infinity
       LimitCORE=infinity
+      # Comment TasksMax if your systemd version does not supports it.
+      # Only systemd 226 and above support this version.
```
thaJeztah: I think the comment may be redundant; we originally added the comment because the service should fail to start if it was set, but from testing, it only printed a _warning_, not an error; https://github.com/moby/moby/pull/24100#issuecomment-229669421

Current versions of our unit file set this option unconditionally, and (testing on CentOS 7.6, which has systemd 219), work just fine.

Random-Liu: Done

PR #1255, opened 2019-08-30
FDQN is a typo, and we don't support trailing dot in FQDN.

Random-Liu:
`FDQN` is a typo.

And actually we are not supporting `FQDN` which requires a trailing dot.


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1255#pullrequestreview-282241141 and https://github.com/containerd/cri/pull/1255#pullrequestreview-282259431

PR #1256, opened 2019-08-31
Clarify exec behavior in some edge cases

Random-Liu:
Dockershim leaves long running `kubectl exec -it` process running in the background after io is detached.

We have the different behavior in that case. The `execInContainer` function will wait until the exec process exits, thus `kubectl exec -it` will also hang until that even after io is detached.

The expected behavior is not defined. This PR added a comment to clarify that behavior. If this causes issue to anyone, we can address it in the future.

### DIFF
--- a/pkg/server/container_execsync.go
+++ b/pkg/server/container_execsync.go
@@ -72,6 +72,16 @@ type execOptions struct {
 
 // execInContainer executes a command inside the container synchronously, and
 // redirects stdio stream properly.
+// This function only returns when the exec process exits, this means that:
+// 1) As long as the exec process is running, the goroutine in the cri plugin
+// will be running and wait for the exit code;
+// 2) `kubectl exec -it` will hang until the exec process exits, even after io
+// is detached. This is different from dockershim, which leaves the exec process
+// running in background after io is detached.
+// https://github.com/kubernetes/kubernetes/blob/v1.15.0/pkg/kubelet/dockershim/exec.go#L127
+// For example, if the `kubectl exec -it` process is killed, IO will be closed. In
+// this case, the CRI plugin will still have a goroutine waiting for the exec process
+// to exit and log the exit code, but dockershim won't.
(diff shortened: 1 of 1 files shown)


Random-Liu:
Hm... After thinking about it more, I found that even if we don't wait for the exec exit, we'll still delete the exec process in the cleanup https://github.com/containerd/cri/blob/master/pkg/server/container_execsync.go#L137

Actually dockershim's behavior may not make sense. Let's describe the behavior difference in the comment, and see whether we got any issue report about the behavior difference in the future. :)

pkg/server/container_execsync.go
```diff
@@ -72,6 +72,13 @@ type execOptions struct {
 
 // execInContainer executes a command inside the container synchronously, and
 // redirects stdio stream properly.
+// This function only returns when the exec process exits, this means that:
+// 1) As long as the exec process is running, the goroutine in the cri plugin
+// will be running and wait for the exit code;
+// 2) `kubectl exec -it` will hang until the exec process exits, even after io
```
mikebrow: It would be nice to have -d detached option for kubectl exec.

Does it matter here "how" the io became detached wrt the way the container runtimes should behave, timeout.. interrupt... Maybe a todo is needed here.

Random-Liu: For example, if the `kubectl exec -it` process is deleted, the io will be detached, but the process will be left running.

mikebrow: right that's one example... and based on that example what is the behavior, expected, documented, test results... yada yada. By here I meant here in the comment, did you have some known manual tests that you wanted to add to the comment based on how you got here to push the PR  :-)

Random-Liu: Done.

mikebrow:
retest and test seem to be unavailable

Random-Liu:
@mikebrow Will get the windows change in first and rebase this PR.

The test should be fixed then.

PR #1258, opened 2019-09-02
Build CRI Plugin on Windows and Add Presubmit

Random-Liu:
For https://github.com/containerd/cri/issues/1257.

This PR moves linux specific logic into `_unix.go` files, and make this buildable on windows.

This PR also adds appveyor presubmit test. We don't use travis windows support because travis only supports windows 1803 (https://docs.travis-ci.com/user/reference/windows/#windows-version), but we need RS5 (17655+ build) for the containerd integration to work. appveyor supports it https://www.appveyor.com/docs/windows-images-software/#visual-studio-2019.

Note that this PR leaves all windows implementation empty. I have a local version which makes it work, I'll send out it one by one for easy reviewing and testing.

The first commit includes https://github.com/containerd/cri/pull/1259.


pkg/config/config_windows.go
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// DefaultConfig returns default configurations of cri plugin.
+func DefaultConfig() PluginConfig {
+	return PluginConfig{}
```
mikebrow: i was only expecting the paths to be different for the most part in things like configs... which would require a two different sets of paths approach based on windows/linux ... not two copies of config. For example, StreamServerAddress: "127.0.0.1", local host the same on windows? As another example, sandbox image, is it multi arch? If so should be the same image. Still another, docker registry should be the same right? 

Put another way... are we just forking entire sets of functions if there might be a difference of one path or one feature in a function, or are we forking because we know it will be very unique and no simple separation of paths and tools used could be determined. What's the plan?

.appveyor.yml
```diff
@@ -0,0 +1,28 @@
...
+image: Visual Studio 2019
+
+clone_folder: c:\gopath\src\github.com\containerd\cri
+
+environment:
+  GOPATH: C:\gopath
+  PATH: C:\go\bin;C:\tools\mingw64\bin;$(PATH)
+  CGO_ENABLED: 1
+  matrix:
+    - GO_VERSION: 1.12.9
```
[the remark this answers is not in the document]
mikebrow: other than the hard coding of the go version this presubmit test LGTM

cmd/containerd/builtins_windows.go
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	_ "github.com/containerd/containerd/diff/lcow"
```
jterry75: I'm fine with this but since you are only doing WCOW process for the first round we dont need any lcow/ stuff

pkg/config/config_windows.go
jterry75: It works in place :). Mostly haha. @Random-Liu - Should you fill this out with the defaults that matter from our fork for the checkin or do you want to do that as a followup?

pkg/netns/netns_windows.go
```diff
@@ -0,0 +1,50 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package netns
+
+// TODO(windows): Implement netns for windows.
+// NetNS holds network namespace.
+type NetNS struct {
```
jterry75: I keep wondering if we should just make this an interface rather than a hard type. What do you think?

pkg/os/os.go
```diff
@@ -23,27 +23,28 @@ import (
...
 	ResolveSymbolicLink(name string) (string, error)
 	FollowSymlinkInScope(path, scope string) (string, error)
 	CopyFile(src, dest string, perm os.FileMode) error
 	WriteFile(filename string, data []byte, perm os.FileMode) error
+	Hostname() (string, error)
+}
+
+// UNIX collects unix system level operations that need to be
+// mocked out during tests.
+type UNIX interface {
```
jterry75: Hmm. You didnt like the approach of defining `OS` in a `_unix.go` and in `_windows.go` so that both just use an `OS` interface but that has different methods on the different platforms?

pkg/server/container_create_windows.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/containerd/cri/pkg/config"
+)
+
+// No container mounts for windows.
+func (c *criService) containerMounts(sandboxID string, config *runtime.ContainerConfig) []*runtime.Mount {
+	return nil
+}
+
+// TODO(windows): Add windows container spec.
+func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint32, netNSPath string,
```
jterry75: I typically would return `nil, errors.New("not implemented")` so I dont accidentally skip an error handler.

pkg/server/container_stats.go
```diff
@@ -25,6 +25,8 @@ import (
 
 // ContainerStats returns stats of the container. If the container does not
 // exist, the call returns an error.
+// TODO(windows): hcsshim Stats is not implemented, add windows support after
```
jterry75: Ah. I have a PR for you. Will link you in

pkg/server/container_stats_list_windows.go
```diff
@@ -0,0 +1,35 @@
...
+	containerstore "github.com/containerd/cri/pkg/store/container"
+)
+
+// TODO(windows): Implement a dummy version of this, and actually support this
+// when stats is supported by the hcs containerd shim.
+func (c *criService) containerMetrics(
+	meta containerstore.Metadata,
+	stats *types.Metric,
+) (*runtime.ContainerStats, error) {
+	return nil, nil
```
jterry75: Again `nil, errors.New("not implemented")`

pkg/server/container_update_resources_windows.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/containerd/containerd/errdefs"
+	"golang.org/x/net/context"
+	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+)
+
+// UpdateContainerResources updates ContainerConfig of the container.
+// TODO(windows): Figure out whether windows support this.
+func (c *criService) UpdateContainerResources(ctx context.Context, r *runtime.UpdateContainerResourcesRequest) (*runtime.UpdateContainerResourcesResponse, error) {
+	return nil, errdefs.ErrNotImplemented
```
jterry75: There is only a `linux` section of the config in CRI-api so we have nothing to implement. This is correct.

pkg/server/helpers.go
```diff
@@ -51,46 +49,24 @@ const (
 	errorStartReason = "StartError"
 	// errorStartExitCode is the exit code when fails to start container.
 	// 128 is the same with Docker's behavior.
+	// TODO(windows): Figure out what should be used for windows.
 	errorStartExitCode = 128
```
jterry75: Is this for "command not exist"?

pkg/server/helpers.go
```diff
@@ -443,6 +340,7 @@ func getRuntimeOptions(c containers.Container) (interface{}, error) {
 
 const (
 	// unknownExitCode is the exit code when exit reason is unknown.
+	// TODO(windows): Figure out unknown exit status for windows.
 	unknownExitCode = 255
```
jterry75: Runtime V2 defined this as 255 so we return that for Windows as well

pkg/server/io/helpers_windows.go
```diff
@@ -0,0 +1,31 @@
...
+package io
+
+import (
+	"io"
+	"os"
+
+	"golang.org/x/net/context"
+)
+
+// TODO(windows): Add windows FIFO support.
```
jterry75: There is no such support. We have to do this with `npipe` unfortunately. Lets talk about this one before we pull the other code across.

Random-Liu:
> not sure how to review, it looks from this one (main) commit that we're forking most of the tree for things that might have some differences and setting the windows version of the fork to ... not implemented for now. Concerned that having two mostly different versions of the tree will make it harder to support the product. For example, you'll need to read both container_create_windows and container_create_unix before knowing how to modify either, and will have to test both independently, vs being able to just ignore simple path differences or sections of the code that say something like unix only or windows only

@mikebrow I share the same concern. Actually I've done a round of de-duplication. I think we've reduced to a reasonable amount of duplication. :)

I'll send out the actual windows implementation soon. If we still think that some part can be de-duplicated, we can cleanup after that is done.

pkg/server/io/helpers_windows.go
Random-Liu: I have an implementation locally. I just mean we need to have the same functionality implemented in windows. It is not a FIFO of course. :)

pkg/server/container_stats.go
Random-Liu: Thanks! I'll file an issue after my next PR, and will add the support once the hcsshim side is ready. :)

.appveyor.yml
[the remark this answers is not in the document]
Random-Liu: Hard coded because of this https://github.com/containerd/containerd/pull/3541

pkg/config/config_windows.go
Random-Liu: @mikebrow Here is the diff:
```diff
< // +build !windows
---
> // +build windows
30,32c30,32
< 			NetworkPluginBinDir:       "/opt/cni/bin",
< 			NetworkPluginConfDir:      "/etc/cni/net.d",
< 			NetworkPluginMaxConfNum:   1, // only one CNI plugin config file will be loaded
---
> 			NetworkPluginBinDir:       "C:\\Program Files\\containerd\\cni\\bin",
> 			NetworkPluginConfDir:      "C:\\Program Files\\containerd\\cni\\conf",
> 			NetworkPluginMaxConfNum:   1,
37c37
< 			DefaultRuntimeName: "runc",
---
> 			DefaultRuntimeName: "runhcs",
40,41c40,41
< 				"runc": {
< 					Type: "io.containerd.runc.v1",
---
> 				"runhcs": {
> 					Type: "io.containerd.runhcs.v1",
49d48
< 		EnableSelinux:       false,
55c54
< 		SandboxImage:            "k8s.gcr.io/pause:3.1",
---
> 		SandboxImage:            "mcr.microsoft.com/k8s/core/pause:1.0.0",
57d55
< 		SystemdCgroup:           false,
67d64
< 		DisableProcMount:       false,
```

Let's see whether we can further clean this up as part of https://github.com/containerd/cri/issues/1245.

Random-Liu: Added https://github.com/containerd/cri/issues/1245#issuecomment-527662192

cmd/containerd/builtins_windows.go
Random-Liu: The problem is that the `lcow` differ is required here. https://github.com/containerd/containerd/blob/4e6e61c44e83ba1f69aa37ea603a9f23383821fb/services/diff/service_windows.go#L22

pkg/netns/netns_windows.go
Random-Liu: It doesn't matter right now. :)

We can add it whenever needed.

pkg/os/os.go
Random-Liu: Done.

cmd/containerd/builtins_windows.go
jterry75: Oh funny. No problem just leave it. I dont think its worth the tag it wont work here anyways

pkg/netns/netns_windows.go
jterry75: Sounds good.

pkg/server/io/helpers_windows.go
jterry75: Ok cool!

pkg/server/container_create_windows.go
Random-Liu: DOne

pkg/server/container_stats_list_windows.go
Random-Liu: Done

pkg/server/container_update_resources_windows.go
Random-Liu: We can easily add a `windows` section if windows container supports vertical pod scaling.

That is some advanced feature we don't right now anyway.

pkg/config/config_windows.go
jterry75: @Random-Liu - Knowing there is a future here can we not add "runhcs" and use "runhcs-wcow-process". It makes more sense and is more descriptive once the other two: "runhcs-wcow-hypervisor" and "runhcs-lcow" are added

Random-Liu: Sure. Will do in the next PR when actually adding the default config.

pkg/server/helpers.go
Random-Liu: Nice. Sound good.

pkg/server/helpers.go
Random-Liu: Usually this is caused by command not found.

See this https://github.com/moby/moby/blob/master/daemon/start.go#L127

Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1258#pullrequestreview-283305729 and https://github.com/containerd/cri/pull/1258#pullrequestreview-283296520

Random-Liu:
Did a rebase. Apply LGTM based on https://github.com/containerd/cri/pull/1258#pullrequestreview-283305729 and https://github.com/containerd/cri/pull/1258#pullrequestreview-283296520

yujuhong:
Late to the party, but lgtm :-)

PR #1259, opened 2019-09-02
Fix panic for task in unknown state.

Random-Liu:
Similar with https://github.com/containerd/containerd/pull/3611.

### DIFF
--- a/pkg/server/io/container_io.go
+++ b/pkg/server/io/container_io.go
@@ -105,18 +105,20 @@ func (c *ContainerIO) Config() cio.Config {
 // to output stream.
 func (c *ContainerIO) Pipe() {
 	wg := c.closer.wg
-	wg.Add(1)
-	go func() {
-		if _, err := io.Copy(c.stdoutGroup, c.stdout); err != nil {
-			logrus.WithError(err).Errorf("Failed to pipe stdout of container %q", c.id)
-		}
-		c.stdout.Close()
-		c.stdoutGroup.Close()
-		wg.Done()
-		logrus.Infof("Finish piping stdout of container %q", c.id)
-	}()
+	if c.stdout != nil {
+		wg.Add(1)
+		go func() {
+			if _, err := io.Copy(c.stdoutGroup, c.stdout); err != nil {
+				logrus.WithError(err).Errorf("Failed to pipe stdout of container %q", c.id)
+			}
+			c.stdout.Close()
+			c.stdoutGroup.Close()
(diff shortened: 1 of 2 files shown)


pkg/server/io/container_io.go
```diff
@@ -105,18 +105,20 @@ func (c *ContainerIO) Config() cio.Config {
 // to output stream.
 func (c *ContainerIO) Pipe() {
 	wg := c.closer.wg
```
mikebrow: won't hurt to test if c.closer is nil

Random-Liu: The closer can never be nil.

If it could be nil, we need to test each of the `wg` is nil or not underlying. If so, I prefer just assume that closer won't be nil here. :)

mikebrow: never is a long time :) but yeah one would want to add a c.getcloserwg() func that logs a debug if c.closer test == nil and otherwise returns the wg, since this code isn't executed very often it's not a performance issue, vs doing the test 4 times.

pkg/server/io/container_io.go
```diff
@@ -105,18 +105,20 @@ func (c *ContainerIO) Config() cio.Config {
...
-		wg.Done()
-		logrus.Infof("Finish piping stdout of container %q", c.id)
-	}()
+	if c.stdout != nil {
+		wg.Add(1)
+		go func() {
+			if _, err := io.Copy(c.stdoutGroup, c.stdout); err != nil {
+				logrus.WithError(err).Errorf("Failed to pipe stdout of container %q", c.id)
+			}
+			c.stdout.Close()
```
mikebrow: are we only worried about null stdout/err at the start.. or could be be nulled out after the create?

Random-Liu: It could be nil if the task status is unknown. That's why we handle it here.

See https://github.com/containerd/containerd/pull/3611

mikebrow: maybe I'm reading the code wrong.. but it reads to me like if stdout isn't set then closer isn't set?

Random-Liu: Oops, I think I missed a fix.

Random-Liu: Added the missing fix... Confused by my PRs. :P

Random-Liu:
@mikebrow Thanks for reviewing! Can you also take a look at https://github.com/containerd/cri/pull/1252?

Tests are failing because of that.

PR #1260, opened 2019-09-04
Use container annotations when creating containers

bart0sh:
Container annotations should be used the same way as pod annotations when creating new containers. They should be also passed down to the OCI runtime.


jterry75:
@bart0sh - Was this a regression? I thought we already did this?

mikebrow:
> @bart0sh - Was this a regression? I thought we already did this?

So did I... but it looks like we only pass the pod annotations.. see my test case pr.

mikebrow:
You know what... it sort of depends how you read this:
      # pod_annotations is a list of pod annotations passed to both pod
      # sandbox as well as container OCI annotations. Pod_annotations also
      # supports golang path match pattern - https://golang.org/pkg/path/#Match.
      # e.g. ["runc.com.*"], ["*.runc.com"], ["runc.com/*"].
      #
      # For the naming convention of annotation keys, please reference:
      # * Kubernetes: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
      # * OCI: https://github.com/opencontainers/image-spec/blob/master/annotations.md

when reading the Kubernetes docs it's obvious or not so obvious that by objects they mean pod and container kinds...

When reading our config doc section above on pod_annotations one could read these are wildcards for pod annotations copied to the pod and container spec.. or one could read it as wild cards for container and pod kind annotations copied to pod and container specs respectively. In hind sight it should just have been passthrough_annotations.

mikebrow:
so... either we add an additional container_annotations for wildcarding container passthroughs... or we modify the config doc. Either way we should modify the config doc.

Random-Liu:
@mikebrow @bart0sh Why do we need to pass down "container annotations"?

There is no "container annotations" in the Kubernetes api, it is purely specified by kubelet. Why does the underlying runtime care about it? https://github.com/kubernetes/kubernetes/blob/4fec22b3e4da6034948fd684f60637171cb55561/pkg/kubelet/kuberuntime/labels.go#L108

Random-Liu:
Ah, I guess you want the device plugin annotations // Kubelet always overrides device plugin annotations if they are conflicting

jterry75:
@mikebrow / @Random-Liu - I vote we add a `container_annotations` map and clear document `pod_annotations` is for sandboxes, and `container_annotations` is for containers. There are many cases were annotations make sense as workarounds to the spec while we work on getting them upstream. For example, Kubernetes has no spec for many windows QoS settings but we can pass them down via annotations while we work on getting upstream KEP's in place.

And to be clear we only forward values: [ContainerConfig.Annotations](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1alpha2/api.proto#L738) that match `container_annotations` in the toml.

mikebrow:
https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

jterry75:
@mikebrow - I get the intended use of annotations and that our use of annotations is against that. But the truth is there is no other way to forward key/value pairs to the OCI spec. And since a sandbox and a container are both eventually an OCI spec for a container it seems reasonable to forward all annotations that match the prefixes defined the config. Kubernetes/Kublet can strictly adhere to what annotations "should" be by setting what is acceptable in the config but other authors can experiment on their own as needed as well without a hack. I think its a win win solution.

Random-Liu:
I prefer we define well known annotation domain for device plugin annotations in kubernetes like this https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/well_known_annotations_windows.go.

I'm fine with arbitrary annotation match for `pod_annotations`, because it is user facing, and different runtimes may want to have their own annotation domains for user facing API.

However, the device plugin annotation is just some implementation detail, I think it is possible to standardize it instead of defining arbitrary annotations.

For example, we can have something like `alpha.kubernetes.io/device-plugin`, and for different devices, they can do things like:
* `alpha.kubernetes.io/device-plugin/fpga/ocihook`
* `alpha.kubernetes.io/device-plugin/nvidia-gpu/ocihook`
* ...

mikebrow:
> I prefer we define well known annotation domain for device plugin annotations in kubernetes like this https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/well_known_annotations_windows.go.
> 
> I'm fine with arbitrary annotation match for `pod_annotations`, because it is user facing, and different runtimes may want to have their own annotation domains.
> 
> However, the device plugin annotation is just some underlying integration detail, I think it is possible to standardize it instead of defining arbitrary annotations.
> 
> For example, we can have something like `alpha.kubernetes.io/device-plugin`, and for different devices, they can do things like:
> 
> * `alpha.kubernetes.io/device-plugin/fpga/ocihook`
> * `alpha.kubernetes.io/device-plugin/nvidia-gpu/ocihook`
> * ...
> 
> @mikebrow @bart0sh does this work for you?

I really like the point of reserving Kubernetes keys, *kubernetes*/ *k8s*/ for annotations. 

And for device-plugin annotations it also makes very good sense to use *some-kube-domain*/device-plugin/some-predefined-names/* as a pattern.

mikebrow:
So yeah I vote we go for the win win.. reserve and support, by default, the passing of k8s & kubernetes annotation key ranges for the non-arbitrary use cases... and add the ability to specify container_annotations that can be passed through for the more arbitrary pod and container user facing annotation use cases.

bart0sh:
@Random-Liu 
> Ah, I guess you want the device plugin annotations.

Correct.

as for restricting annotations to some domain I have mixed feelings. Why would we want to do that? We can filter out any unwanted annotations using container_annotations.

I'll add container_annotations usage to this PR.

kad:
> I really like the point of reserving Kubernetes keys, _kubernetes_/ _k8s_/ for annotations.
> 
> And for device-plugin annotations it also makes very good sense to use _some-kube-domain_/device-plugin/some-predefined-names/* as a pattern.

The Kubernetes documentation mentions that kubernetes.io and k8s.io domains should be reserved to Kubernetes components. Device plugins are coming from device vendors and they are neither k8s components nor even CNCF projects. So, legally, they should not be using kubernetes.io/k8s.io domains for passing proprietary information.

pkg/config/config_unix.go
```diff
@@ -38,7 +38,8 @@ func DefaultConfig() PluginConfig {
 			NoPivot:            false,
 			Runtimes: map[string]Runtime{
 				"runc": {
-					Type: "io.containerd.runc.v1",
+					Type:                 "io.containerd.runc.v1",
+					ContainerAnnotations: []string{},
```
mikebrow: If we are going to show this here as an empty list, we'd also want to show PodAnnotations I think...

docs/config.md
```diff
@@ -117,6 +117,10 @@ version = 2
       # * OCI: https://github.com/opencontainers/image-spec/blob/master/annotations.md
       pod_annotations = []
 
+      # container_annotations is a list of container annotations passed to both pod
+      # sandbox as well as container OCI annotations.
```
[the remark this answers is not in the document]
bart0sh: thank you for the review! Should be fixed now.

pkg/config/config_unix.go
bart0sh: This is needed for 'containerd config default' to add this item to the generated content.
Yes, we'd want it for PodAnnotations as well, but not in this PR I believe.

pkg/config/config.go
```diff
@@ -38,6 +38,9 @@ type Runtime struct {
 	// PodAnnotations is a list of pod annotations passed to both pod sandbox as well as
 	// container OCI annotations.
 	PodAnnotations []string `toml:"pod_annotations" json:"PodAnnotations"`
+	// ContainerAnnotations is a list of container annotations passed to both pod sandbox as well as
+	// container OCI annotations.
```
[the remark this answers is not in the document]
bart0sh: should be fixed now.

pkg/config/config_unix.go
mikebrow: yeah we don't show all of the empty elements in the default.. it's on purpose

Random-Liu:
> The Kubernetes documentation mentions that kubernetes.io and k8s.io domains should be reserved to Kubernetes components. Device plugins are coming from device vendors and they are neither k8s components nor even CNCF projects. So, legally, they should not be using kubernetes.io/k8s.io domains for passing proprietary information.

If we go with the well-known device plugin annotation model, the annotation format will be defined as part of device plugin interface, which is a "property" of Kubernetes. Of course the device plugin itself belongs to the vendor.

kad:
Would it make sense to add filtering of container annotations in containerd only when it will be defined in Device Manager in Kubelet?

pkg/config/config_unix.go
bart0sh: ok, thanks for the explanation. Should I remove it then?

bart0sh: I set ContainerAnnotations to ["kubernetes.io/*", "k8s.io/*] by default. I think these are a reasonable default values. Please, review.

mikebrow: for container annotations it looks like the current keys are:
```
	// TODO: change those label names to follow kubernetes's format
	podDeletionGracePeriodLabel    = "io.kubernetes.pod.deletionGracePeriod"
	podTerminationGracePeriodLabel = "io.kubernetes.pod.terminationGracePeriod"

	containerHashLabel                     = "io.kubernetes.container.hash"
	containerRestartCountLabel             = "io.kubernetes.container.restartCount"
	containerTerminationMessagePathLabel   = "io.kubernetes.container.terminationMessagePath"
	containerTerminationMessagePolicyLabel = "io.kubernetes.container.terminationMessagePolicy"
	containerPreStopHandlerLabel           = "io.kubernetes.container.preStopHandler"
	containerPortsLabel                    = "io.kubernetes.container.ports"
```
but from the todo they may change soon, let's wait for @Random-Liu to chime in.  We might need to reserve two sets of patterns this io.kubernetes one currently in use and kubernetes.io suggested by the todo?

We'll also want to go ahead and set pod annotations default keys for k8s here I think so they will pass through as well,... I mean while we're at it, why not do the same for pod annotations as we are doing for container. It's that or we could put them all in the docs for explanation and to reserve them and users could config as they like, thus keeping the default as is, but allowing annotations to be passed through for 3rd party integrations. 

Again.. I'd like to hear from some other sig-node members regarding their opinion for defaulting the passthrough of pod and container annotations.

bart0sh: Changed ContainerAnnotations value to ["io.kubernetes.*", "io.k8s.*"]

In my opinion it's not that important. We can change this later on if we want to. The main purpose of this PR is to pass forgotten annotations to the runtime.

bart0sh: Would it work if we remove setting defaults for *Annotations settings from this PR and make another PR that sets them? Would it help to merge this PR faster?

mikebrow: > Would it work if we remove setting defaults for *Annotations settings from this PR and make another PR that sets them? Would it help to merge this PR faster?

probably yes. 

Just add in the config.md doc that *.kubernetes.* and *.k8s.* are reserved for both pod and container annotation keys.

Random-Liu: I don't think we need any default value here.

pkg/config/config.go
```diff
@@ -38,6 +38,8 @@ type Runtime struct {
 	// PodAnnotations is a list of pod annotations passed to both pod sandbox as well as
 	// container OCI annotations.
 	PodAnnotations []string `toml:"pod_annotations" json:"PodAnnotations"`
+	// ContainerAnnotations is a list of container annotations passed through to the OCI config of the containers.
```
Random-Liu: Add the following:
```
// Container annotations in CRI are usually generated by other Kubernetes node components (i.e.,
// not users). Currently, only device plugins populate the annotations.
```

pkg/config/config_unix.go
```diff
@@ -38,7 +38,8 @@ func DefaultConfig() PluginConfig {
 			NoPivot:            false,
 			Runtimes: map[string]Runtime{
 				"runc": {
-					Type: "io.containerd.runc.v1",
+					Type:                 "io.containerd.runc.v1",
+					ContainerAnnotations: []string{"io.kubernetes.*", "io.k8s.*"},
```
Random-Liu: Let's not put any default here.

If we really want to have a default value, I hope it is the device plugin domain we want to push, e.g. `io.kubernetes.device-plugin.*`.

bart0sh:
@Random-Liu thank you for the review! I've updated the PR as you've suggested. Please review and merge if you don't mind.

pkg/config/config_unix.go
bart0sh: > Let's not put any default here.

done

pkg/config/config.go
bart0sh: done

pkg/config/config_unix.go
bart0sh: I've removed default settings as @Random-Liu suggested

pkg/config/config.go
Random-Liu: Sorry, I mean adding this after or in front of you original comment. And please also update `docs/config.md` with the same text.

pkg/server/container_create_unix_test.go
```diff
@@ -367,7 +367,48 @@ func TestPodAnnotationPassthroughContainerSpec(t *testing.T) {
...
+		podAnnotations       []string
+		containerAnnotations []string
+		configChange         func(*runtime.PodSandboxConfig)
+		specCheck            func(*testing.T, *runtimespec.Spec)
+	}{
+		"passthrough annotations from pod and container should be passed as an OCI annotation": {
+			podAnnotations:       []string{"c"},
+			containerAnnotations: []string{"c*"}, // wildcard should pick up ca-c->ca-d pair in container
+			specCheck: func(t *testing.T, spec *runtimespec.Spec) {
+				assert.Equal(t, "d", spec.Annotations["c"])
```
Random-Liu: Can you add 2 test cases?
1) Annotations don't pass through if no pass through annotations are configured;
2) Unmatched annotations can't pass through even if pass through annotations are configured.

Random-Liu:
LGTM other than the comments.

pkg/server/container_create_unix_test.go
bart0sh: done

bart0sh:
@mikebrow @Random-Liu Thank you for the review! Updated the PR according to the review comments.

PR #1263, opened 2019-09-05
init cni in local cluster installation

gongguan:
When I tried containerd as kubernetes runtime in local cluster, node becomes notReady because of the `cni plugin not initialized` error in kubelet runtimeStatus api. It takes me some time to solve this problem.
I think cni initialization should be mentioned in the local cluster setup for beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ sudo containerd
 ```bash
 CONTAINER_RUNTIME=remote CONTAINER_RUNTIME_ENDPOINT='unix:///run/containerd/containerd.sock' ./hack/local-up-cluster.sh
 ```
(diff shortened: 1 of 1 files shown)


gongguan:
further more, can containerd/cri support no cni in local cluster such as k8s/no-op

Random-Liu:
@gongguan `make install.deps` should have setup cni for local test for you.

gongguan:
@Random-Liu cri release package doesn't have cni config and binaries, it may be a little confused to solve it.

gongguan:
> @gongguan `make install.deps` should have setup cni for local test for you.

@Random-Liu Can I close the pr now?

fuweid:
@gongguan please go ahead :)

PR #1268, opened 2019-09-08
add temp GOPATH as a bash env for install.deps

bg-chun:
add the GOPATH of temporally cloned packages to bash env to build dependency packages

Signed-off-by: byonggon chun <bg.chun@samsung.com>

PR #1271, opened 2019-09-09
need to move instr to more recent golang; remove integration/*.go from lint verify test

mikebrow:
Instructions point to golang 1.10.. which is a bit long in the tooth now. Time to ask them to use 1.12.9 see #1268


Random-Liu:
Hm, I think we want to skip `integration/*` in verify.

PR #1274, opened 2019-09-17
Add DualStack support

Random-Liu:
Carry on from https://github.com/containerd/cri/pull/1272.
For https://github.com/kubernetes/kubernetes/issues/82634

This PR adds dual stack support in `PodSandboxStatus` and `UpdateRuntimeConfig`.

Several things worth noting:
1) 2 new values `.PodCIDRRanges` and `.Routes` are supported in the cni config template. `.PodCIDRRanges` is a string array of all CIDRs assigned to the node; `.Routes` is a string array of all routes needed. They can be used to config CIDRs for dualstack in the ipam plugin, e.g.:
```
      "ipam": {
        "type": "host-local",
        "ranges": [{{range $i, $range := .PodCIDRRanges}}{{if $i}}, {{end}}[{"subnet": "{{$range}}"}]{{end}}],
        "routes": [{{range $i, $route := .Routes}}{{if $i}}, {{end}}{"dst": "{{$route}}"}{{end}}]
      }
```
2) IPv6 is enabled by default in the CNI config used in the CRI test and integration test. However, please note that pod startup will become slower when IPv6 is enabled, see https://github.com/containerd/cri/pull/563#issuecomment-359618326
3) The CNI config used in the cluster/node e2e test is also updated to use `.PodCIDRRanges`. However, IPv6 dual stack should only be enabled when `IPv6DualStack` feature gate is enabled in Kubernetes, so it shouldn't cause performance issue right now.

Anyway, please note that **Enabling IPv6 will slow down pod startup, we may want to warn user about that in the release note**.

/cc @aojea


cluster/gce/cni.template
```diff
@@ -7,11 +7,10 @@
       "mtu": 1460,
       "ipam": {
         "type": "host-local",
-        "subnet": "{{.PodCIDR}}",
+        "ranges": [{{range $i, $range := .PodCIDRRanges}}{{if $i}}, {{end}}[{"subnet": "{{$range}}"}]{{end}}],
         "routes": [
-          {
-            "dst": "0.0.0.0/0"
-          }
+          {"dst": "0.0.0.0/0"},
```
aojea: we had issues in kind using a config like this on hosts without IPv6 enabled, so we had to check if ipv6 was enabled to add the IPv6 default route, don't know if this can happen here

integration/pod_dualstack_test.go
```diff
@@ -0,0 +1,105 @@
...
+	ipv4Enabled, err := regexp.MatchString("inet .* scope global", string(content))
+	assert.NoError(t, err)
+	ipv6Enabled, err := regexp.MatchString("inet6 .* scope global", string(content))
+	assert.NoError(t, err)
+
+	if ipv4Enabled && ipv6Enabled {
+		t.Log("Dualstack should be enabled")
+		require.Len(t, additionalIps, 1)
+		assert.NotNil(t, net.ParseIP(ip).To4())
+		assert.Nil(t, net.ParseIP(additionalIps[0].GetIp()).To4())
```
aojea: do we need to use GetIP()? is not a slice of strings? `additionalIps[0].To4()`

Random-Liu: It is not.

cluster/gce/cni.template
Random-Liu: Good catch. Let me see.

Random-Liu: Kubenet always set it https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go#L286

pkg/server/sandbox_run.go
```diff
@@ -426,14 +427,29 @@ func toCNIDNS(dns *runtime.DNSConfig) *cni.DNS {
 	}
 }
 
-// selectPodIP select an ip from the ip list. It prefers ipv4 more than ipv6.
-func selectPodIP(ipConfigs []*cni.IPConfig) string {
+// selectPodIPs select an ip from the ip list. It prefers ipv4 more than ipv6
+// and returns the additional ips
+// TODO(random-liu): Revisit the ip order in the ipv6 beta stage.
+// https://github.com/kubernetes/kubernetes/blob/v1.16.0-rc.2/pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go#L611
```
aojea: nice catch

cluster/gce/cni.template
Random-Liu: Based on my experiment, it seems fine:
```console
$ sysctl net/ipv6/conf/all/disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 1
$ cat /etc/cni/net.d/10-containerd-net.conflist
{
  "cniVersion": "0.3.1",
  "name": "containerd-net",
  "plugins": [
    {
      "type": "bridge",
      "bridge": "cni0",
      "isGateway": true,
      "ipMasq": true,
      "promiscMode": true,
      "ipam": {
        "type": "host-local",
        "ranges": [
          [{
            "subnet": "10.88.0.0/16"
          }]
        ],
        "routes": [
          { "dst": "0.0.0.0/0" },
          { "dst": "::/0" }
        ]
      }
    },
    {
      "type": "portmap",
      "capabilities": {"portMappings": true}
    }
  ]
}
$ crictl runp sandbox.json
0103ae22345ebed28501b0388360b2f25cf0c112736f24c96883efd46733fbe1
```

aojea: this was the issue I was referring https://github.com/kubernetes-sigs/kind/issues/626, seems that only happen if ipv6 is disabled at boot time

Random-Liu: If that is the case, it is a bug in kubenet in Kubernetes 1.16 that we need to fix.

Let me look into it tomorrow.

Random-Liu: Thanks for the good catch! Filed https://github.com/kubernetes/kubernetes/pull/82800 for kubenet, and will fix this PR. @aojea

Random-Liu: Done

Random-Liu:
Switch golangci-lint to v1.18.0 because of the OOM issue (see https://github.com/containerd/containerd/pull/3657)

aojea:
SGTM
I'd love to test first it in kind though, but the build fails with the following error:
> containerd: symbol lookup error: containerd: undefined symbol: seccomp_api_set

does this version has new dependencies?

pkg/server/sandbox_run.go
```diff
@@ -426,14 +427,29 @@ func toCNIDNS(dns *runtime.DNSConfig) *cni.DNS {
 	}
 }
 
-// selectPodIP select an ip from the ip list. It prefers ipv4 more than ipv6.
-func selectPodIP(ipConfigs []*cni.IPConfig) string {
+// selectPodIPs select an ip from the ip list. It prefers ipv4 more than ipv6
```
mikebrow: TODO or ... Issue.. need a way to prefer ipv6 over ipv4. Or add preference to the ip order TODO below..

pkg/server/sandbox_run.go
```diff
@@ -137,7 +137,7 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		// In this case however caching the IP will add a subtle performance enhancement by avoiding
 		// calls to network namespace of the pod to query the IP of the veth interface on every
 		// SandboxStatus request.
-		sandbox.IP, sandbox.CNIResult, err = c.setupPod(ctx, id, sandbox.NetNSPath, config)
+		sandbox.IP, sandbox.AdditionalIPs, sandbox.CNIResult, err = c.setupPod(ctx, id, sandbox.NetNSPath, config)
```
mikebrow: starting to look a bit odd here might want to just pass a sandbox to setupPod.. and rename it to setupPodSandbox

Random-Liu:
> @Random-Liu the pods that were created before the containerd upgrade doesn't report the additional IPs after restarting containerd, is this the expected behavior?

@aojea Yeah, that is expected. We get the ips when creating the pod, and then checkpointed the IPs. For old pods, we don't have the checkpoint.

pkg/server/sandbox_status.go
```diff
@@ -67,39 +67,46 @@ func (c *criService) PodSandboxStatus(ctx context.Context, r *runtime.PodSandbox
...
 // toCRISandboxStatus converts sandbox metadata into CRI pod sandbox status.
-func toCRISandboxStatus(meta sandboxstore.Metadata, status sandboxstore.Status, ip string) *runtime.PodSandboxStatus {
+func toCRISandboxStatus(meta sandboxstore.Metadata, status sandboxstore.Status, ip string, additionalIPs []string) *runtime.PodSandboxStatus {
 	// Set sandbox state to NOTREADY by default.
 	state := runtime.PodSandboxState_SANDBOX_NOTREADY
 	if status.State == sandboxstore.StateReady {
 		state = runtime.PodSandboxState_SANDBOX_READY
 	}
 	nsOpts := meta.Config.GetLinux().GetSecurityContext().GetNamespaceOptions()
+	var ips []*runtime.PodIP
```
mikebrow: Hmm... seems a shame to have to convert string arrays to podip (string) arrays

pkg/store/sandbox/metadata.go
```diff
@@ -55,6 +55,8 @@ type Metadata struct {
 	NetNSPath string
 	// IP of Pod if it is attached to non host network
 	IP string
+	// AdditionalIPs of the Pod if it is attached to no host network
```
mikebrow: s/no/non/

nit: Maybe put this new field at the bottom?

hack/verify-lint.sh
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+GOLANGCI_LINT_VERSION="v1.18.0"
+
+if [ ! -x "$(command -v golangci-lint)" ]; then
```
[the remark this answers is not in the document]
Random-Liu: No version printed from `golangci-lint`. :(

Makefile
```diff
@@ -171,17 +171,13 @@ endif
...
 
-install.tools: .install.gitvalidation .install.golangci-lint .install.vndr ## install tools used by verify
+install.tools: .install.gitvalidation .install.vndr ## install tools used by verify
 	@echo "$(WHALE) $@"
 
 .install.gitvalidation:
 	@echo "$(WHALE) $@"
 	$(GO) get -u github.com/vbatts/git-validation
 
-.install.golangci-lint:
```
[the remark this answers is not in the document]
Random-Liu: Hm, linked to editor, never thought about it. Will do then.

Makefile
```diff
@@ -55,7 +55,7 @@ version: ## print current cri plugin release version
 
 lint:
 	@echo "$(WHALE) $@"
-	golangci-lint run --skip-files .*_test.go --skip-dirs='(integration)'
+	@./hack/verify-lint.sh
```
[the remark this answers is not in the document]
Random-Liu: What do you mean?

hack/verify-lint.sh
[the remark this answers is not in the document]
Random-Liu: Hm, it seems that lint works fine now https://github.com/containerd/cri/pull/1277

I'll probably remove the last commit for now and see whether it works. I prefer simple `go get -u` if it doesn't break things.

pkg/server/sandbox_run.go
Random-Liu: Filed https://github.com/containerd/cri/issues/1278

pkg/server/sandbox_run.go
Random-Liu: It only setup pod network.

Changed to `setupPodNetwork` and `teardownPodNetwork`.

pkg/server/sandbox_status.go
Random-Liu: But we have to. :( It is how it is defined in CRI.

pkg/store/sandbox/metadata.go
Random-Liu: I think `IP` and `AdditionalIPs` should better be together. :)

docs/config.md
```diff
@@ -183,6 +183,13 @@ version = 2
     # This is a temporary backward-compatible solution for kubenet users
     # who don't have a cni daemonset in production yet.
     # This will be deprecated when kubenet is deprecated.
+    # Currently supported template values are `.PodCIDR`, `.PodCIDRRanges`
```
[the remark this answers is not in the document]
Random-Liu: Added extra section in the doc for CNI Config Template.

Makefile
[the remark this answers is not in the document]
mikebrow: verify whether it's installed then if so use the installed (with version if that had been avail) vs verify does the install regardless. Didn't know there was no "version" odd, thought that was common for all go tools.

Makefile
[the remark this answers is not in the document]
mikebrow: Yeah this is the other part of why I was preferring check first before installing update. My local CI is fetch, merge, install.tools, checkout, edit... editor runs tests as I go, run make tests, push. Sometimes I forget make verify.. but the editor is running lint so I don't worry about anything but signing commit.

pkg/server/sandbox_run.go
mikebrow: by odd I also meant passing sandbox.NetNSPath... then lval assigning 3 sandbox values on the return, not just the name.

pkg/store/sandbox/metadata.go
mikebrow: yeah and json marshal is by field name not number.. so it's ok.. just normal convention to add new fields to bottom. Convention based on old tech where serialization was binary though.

pkg/server/sandbox_run.go
Random-Liu: Will do.

pkg/server/sandbox_run.go
```diff
@@ -302,31 +301,38 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 	return &runtime.RunPodSandboxResponse{PodSandboxId: id}, nil
 }
 
-// setupPod setups up the network for a pod
-func (c *criService) setupPod(ctx context.Context, id string, path string, config *runtime.PodSandboxConfig) (string, *cni.CNIResult, error) {
+// setupPodNetwork setups up the network for a pod
+func (c *criService) setupPodNetwork(ctx context.Context, sandbox *sandboxstore.Sandbox) error {
```
mikebrow: nice!

Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1274#pullrequestreview-290937107

Issue #1279, opened 2019-09-19
cri path issue


kfox1111:
Using containerd 1.2.8 or 1.2.9, containers fail to search the path if env PATH is not defined inside the container.

Simple reproducer
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - command:
    - sh
    - -cxe
    - |
      while true; do
        sleep 1000;
      done
    image: docker.io/gitlab/gitlab-runner-helper:x86_64-latest
    imagePullPolicy: Always
    name: main
```
Much more detail in https://github.com/containerd/containerd/issues/3597

Containers fail like so:
"ad62ed990987fba56502eb63d63b00a4804f2b0c77feb1a2f8b92cd3bb98b1be": OCI runtime exec failed: exec failed: container_linux.go:345: starting container process caused "exec: "sh": executable file not found in $PATH": unknown

estesp:
{ note I've moved this to the CRI as the PR to fix the issue is in this repository }
@kfox1111 disappointing that we missed the alternate path to creating an OCI spec in the CRI implementation that lost the default `$PATH` due to the changes in containerd. PR #1280 will fix this and then we can figure out a path to get the 1.2.x release line fixed.

Random-Liu:
This is bad...

PR #1280, opened 2019-09-19
Add back default UNIX env to container config

estesp:
Fixes: #1279 

Due to [changes to the defaults in containerd](https://github.com/containerd/containerd/pull/3551), the CRI path to creating a
container OCI config needs to add back in the default UNIX $PATH (and
any other defaults) as that is the expected behavior from other
runtimes.

This is a miss from our fix released in 1.2.9 after breaking in 1.2.8 with the standard path in containerd via the API and `ctr` client (reported in [containerd issue #3597](https://github.com/containerd/containerd/issues/3597) and fixed in [containerd PR #3599](https://github.com/containerd/containerd/pull/3599))

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -112,6 +112,7 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		customopts.WithoutDefaultSecuritySettings,
 		customopts.WithRelativeRoot(relativeRootfsPath),
 		customopts.WithProcessArgs(config, imageConfig),
+		oci.WithDefaultPathEnv,
 		// this will be set based on the security context below
 		oci.WithNewPrivileges,
 	}
--- a/pkg/server/container_create_unix_test.go
+++ b/pkg/server/container_create_unix_test.go
@@ -269,6 +269,27 @@ func TestContainerSpecTty(t *testing.T) {
 	}
 }
 
+func TestContainerSpecDefaultPath(t *testing.T) {
+	testID := "test-id"
+	testSandboxID := "sandbox-id"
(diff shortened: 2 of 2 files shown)


pkg/server/container_create_unix.go
```diff
@@ -124,6 +124,7 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 	if config.GetTty() {
 		specOpts = append(specOpts, oci.WithTTY)
 	}
+	specOpts = append(specOpts, oci.WithDefaultPathEnv)
```
Random-Liu: Move this after `WithoutRunMount`, that is where default should locate.

Random-Liu: And please add a unit test like https://github.com/containerd/cri/blob/master/pkg/server/container_create_unix_test.go#L735 to verify that PATH should not be empty by default. It can help us catch future issues.

pkg/server/container_create_unix_test.go
```diff
@@ -269,6 +269,27 @@ func TestContainerSpecTty(t *testing.T) {
 	}
 }
 
+func TestContainerSpecDefaultPath(t *testing.T) {
+	testID := "test-id"
+	testSandboxID := "sandbox-id"
+	testPid := uint32(1234)
+	expectedDefault := "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
Random-Liu: We should probably export Containerd default PATH, and use it here. We can do that in a following PR.

Random-Liu:
@estesp Can you please also send cherry-picks to 1.2 and 1.3? Thanks!

PR #1283, opened 2019-09-19
[release/1.2] Backport fix for default UNIX environment in OCI container config

estesp:
Backport test from #1280; harcoded `defaultUnixEnv` from containerd/containerd brought into the spec generator sequence, as there is no clear way to fix the bug using the `oci.With..` helper in this design.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -264,6 +264,7 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get runtime options")
 	}
+
 	opts = append(opts,
 		containerd.WithSpec(spec, specOpts...),
 		containerd.WithRuntime(sandboxInfo.Runtime.Name, runtimeOptions),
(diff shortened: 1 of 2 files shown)


pkg/server/container_create.go
```diff
@@ -321,6 +322,8 @@ func (c *criService) generateContainerSpec(id string, sandboxID string, sandboxP
 	// Set the relative path to the rootfs of the container from containerd's
 	// pre-defined directory.
 	g.SetRootPath(relativeRootfsPath)
+	// add default UNIX path; overwritten optionally by image config PATH env
+	g.AddProcessEnv("PATH", "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
```
Random-Liu: Do this in `defaultRuntimeSpec`.

estesp: Done; added to section where `spec.Linux` is non-nil as it only makes sense to have a default UNIX-like `PATH` for Linux specs.

Random-Liu: We only support linux in 1.2 actually. :)

But this is fine.

PR #1284, opened 2019-09-19
Add windows port forward support

liyanhui1228:
Fixes https://github.com/containerd/cri/issues/1273

### DIFF
--- a/pkg/server/container_execsync.go
+++ b/pkg/server/container_execsync.go
@@ -70,38 +70,13 @@ type execOptions struct {
 	timeout time.Duration
 }
 
-// execInContainer executes a command inside the container synchronously, and
-// redirects stdio stream properly.
-// This function only returns when the exec process exits, this means that:
-// 1) As long as the exec process is running, the goroutine in the cri plugin
-// will be running and wait for the exit code;
-// 2) `kubectl exec -it` will hang until the exec process exits, even after io
-// is detached. This is different from dockershim, which leaves the exec process
-// running in background after io is detached.
-// https://github.com/kubernetes/kubernetes/blob/v1.15.0/pkg/kubelet/dockershim/exec.go#L127
-// For example, if the `kubectl exec -it` process is killed, IO will be closed. In
-// this case, the CRI plugin will still have a goroutine waiting for the exec process
-// to exit and log the exit code, but dockershim won't.
-func (c *criService) execInContainer(ctx context.Context, id string, opts execOptions) (*uint32, error) {
+func (c *criService) execInternal(ctx context.Context, container containerd.Container, id string, opts execOptions) (*uint32, error) {
 	// Cancel the context before returning to ensure goroutines are stopped.
 	// This is important, because if `Start` returns error, `Wait` will hang
 	// forever unless we cancel the context.
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// Get container from our container store.
-	cntr, err := c.containerStore.Get(id)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to find container %q in store", id)
-	}
-	id = cntr.ID
-
-	state := cntr.Status.Get().State()
-	if state != runtime.ContainerState_CONTAINER_RUNNING {
-		return nil, errors.Errorf("container is in %s state", criContainerStateToString(state))
-	}
-
-	container := cntr.Container
 	spec, err := container.Spec(ctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get container spec")
@@ -204,4 +179,33 @@ func (c *criService) execInContainer(ctx context.Context, id string, opts execOp
 		log.G(ctx).Debugf("Stream pipe for exec process %q done", execID)
 		return &code, nil
 	}
+}
+
+// execInContainer executes a command inside the container synchronously, and
+// redirects stdio stream properly.
+// This function only returns when the exec process exits, this means that:
+// 1) As long as the exec process is running, the goroutine in the cri plugin
+// will be running and wait for the exit code;
+// 2) `kubectl exec -it` will hang until the exec process exits, even after io
+// is detached. This is different from dockershim, which leaves the exec process
+// running in background after io is detached.
+// https://github.com/kubernetes/kubernetes/blob/v1.15.0/pkg/kubelet/dockershim/exec.go#L127
+// For example, if the `kubectl exec -it` process is killed, IO will be closed. In
+// this case, the CRI plugin will still have a goroutine waiting for the exec process
+// to exit and log the exit code, but dockershim won't.
+func (c *criService) execInContainer(ctx context.Context, id string, opts execOptions) (*uint32, error) {
+	// Get container from our container store.
+	cntr, err := c.containerStore.Get(id)
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to find container %q in store", id)
+	}
+	id = cntr.ID
+
+	state := cntr.Status.Get().State()
+	if state != runtime.ContainerState_CONTAINER_RUNNING {
+		return nil, errors.Errorf("container is in %s state", criContainerStateToString(state))
+	}
+
+	return c.execInternal(ctx, cntr.Container, id, opts)
 }
--- a/pkg/server/sandbox_portforward_windows.go
+++ b/pkg/server/sandbox_portforward_windows.go
@@ -19,19 +19,62 @@ limitations under the License.
 package server
 
 import (
+	"bytes"
+	"fmt"
 	"io"
 
-	"github.com/containerd/containerd/errdefs"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
-	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	"k8s.io/utils/exec"
+
+	"github.com/containerd/cri/pkg/ioutil"
+	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 )
 
-// PortForward prepares a streaming endpoint to forward ports from a PodSandbox, and returns the address.
-// TODO(windows): Implement this for windows.
-func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequest) (*runtime.PortForwardResponse, error) {
-	return nil, errdefs.ErrNotImplemented
+func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
+	stdout := ioutil.NewNopWriteCloser(stream)
+	stderrBuffer := new(bytes.Buffer)
+	stderr := ioutil.NewNopWriteCloser(stderrBuffer)
+	// localhost is resolved to 127.0.0.1 in ipv4, and ::1 in ipv6.
+	// Explicitly using ipv4 IP address in here to avoid flakiness.
+	cmd := []string{"wincat.exe", "127.0.0.1", fmt.Sprint(port)}
+	err := c.execInSandbox(ctx, id, cmd, stream, stdout, stderr)
+	if err != nil {
+		return errors.Wrapf(err, "failed to execute port forward in sandbox: %s", stderrBuffer.String())
+	}
+	return nil
 }
 
-func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
-	return errdefs.ErrNotImplemented
+func (c *criService) execInSandbox(ctx context.Context, sandboxID string, cmd []string, stdin io.Reader, stdout, stderr io.WriteCloser) error {
+	// Get sandbox from our sandbox store.
+	sb, err := c.sandboxStore.Get(sandboxID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find sandbox %q in store", sandboxID)
+	}
+
+	// Check the sandbox state
+	state := sb.Status.Get().State
+	if state != sandboxstore.StateReady {
+		return errors.Errorf("sandbox is in %s state", fmt.Sprint(state))
+	}
+
+	opts := execOptions{
+		cmd:    cmd,
+		stdin:  stdin,
+		stdout: stdout,
+		stderr: stderr,
+		tty:    false,
+		resize: nil,
+	}
+	exitCode, err := c.execInternal(ctx, sb.Container, sandboxID, opts)
+	if err != nil {
+		return errors.Wrap(err, "failed to exec in sandbox")
+	}
+	if *exitCode == 0 {
+		return nil
+	}
+	return &exec.CodeExitError{
+		Err:  errors.Errorf("error executing command %v, exit code %d", cmd, *exitCode),
+		Code: int(*exitCode),
+	}
 }
--- a/test/windows/test.sh
+++ b/test/windows/test.sh
@@ -21,7 +21,7 @@ set -o pipefail
 export PATH="/c/Program Files/Containerd:$PATH"
 REPO_TAR="${REPO_TAR:-"/c/cri.tar.gz"}"
 FOCUS="${FOCUS:-"Conformance"}"
-SKIP="${SKIP:-"portforward"}"
+SKIP="${SKIP:-""}"
 REPORT_DIR="${REPORT_DIR:-"/c/_artifacts"}"
 
 repo="$GOPATH/src/github.com/containerd/cri"
--- a/hack/utils.sh
+++ b/hack/utils.sh
@@ -17,7 +17,7 @@
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 
 # Not from vendor.conf.
-CRITOOL_VERSION=84b8540c69e82671bfc244a0b0fe28a6a98381ce
+CRITOOL_VERSION=ad1e8a4f1e0e44b22c1c371b9dc4db92eba6f157
 CRITOOL_PKG=github.com/kubernetes-sigs/cri-tools
 CRITOOL_REPO=github.com/kubernetes-sigs/cri-tools
 
--- a/pkg/config/config_windows.go
+++ b/pkg/config/config_windows.go
@@ -54,7 +54,7 @@ func DefaultConfig() PluginConfig {
 			TLSKeyFile:  "",
 			TLSCertFile: "",
 		},
-		SandboxImage:            "mcr.microsoft.com/k8s/core/pause:1.0.0",
+		SandboxImage:            "mcr.microsoft.com/k8s/core/pause:1.2.0",
 		StatsCollectPeriod:      10,
 		MaxContainerLogLineSize: 16 * 1024,
 		Registry: Registry{
--- a/pkg/server/sandbox_portforward.go
+++ b/pkg/server/sandbox_portforward.go
@@ -1,0 +1,38 @@
+/*
+Copyright The containerd Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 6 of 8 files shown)


pkg/server/container_execsync.go
```diff
@@ -27,6 +28,8 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/oci"
+	"github.com/containerd/cri/pkg/store/container"
```
Random-Liu: Move cri vendors to the 3rd group.

pkg/server/container_execsync.go
```diff
@@ -70,6 +73,40 @@ type execOptions struct {
...
+}
+
+func checkSandboxState(sb *sandbox.Sandbox) error {
+	state := sb.Status.Get().State
+	if state != sandbox.StateReady {
+		return errors.Errorf("sandbox is in %s state", fmt.Sprint(state))
+	}
+	return nil
+}
+
```
Random-Liu: I prefer extract the common part (unchanged part in this PR) out to be a helper function and use it in `execInContainer` and `portforward`. :)

The function is called `execInContainer`, it feels weird to also handle sandbox here.

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,68 @@ limitations under the License.
...
+	"k8s.io/kubernetes/pkg/kubelet/util/ioutils"
+	"k8s.io/utils/exec"
+
+	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
+	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 )
 
 // PortForward prepares a streaming endpoint to forward ports from a PodSandbox, and returns the address.
 // TODO(windows): Implement this for windows.
 func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequest) (*runtime.PortForwardResponse, error) {
```
Random-Liu: This part can be shared between unix and windows, probably move to `sandbox_portforward.go`? We want as less duplication as possible.

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,68 @@ limitations under the License.
...
+	stderr := new(bytes.Buffer)
+	err = c.ExecInSandbox(id, []string{"wincat.exe", "localhost", fmt.Sprint(port)}, stream, ioutils.WriteCloserWrapper(stream), ioutils.WriteCloserWrapper(stderr), false, nil)
+	if err != nil {
+		return fmt.Errorf("%v: %s", err, stderr.String())
+	}
+
+	return nil
+}
+
+func (c *criService) ExecInSandbox(sandboxID string, cmd []string, stdin io.Reader, stdout, stderr io.WriteCloser,
```
Random-Liu: Keep the function private, we don't need to export it yet.

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,68 @@ limitations under the License.
...
 }
 
 func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
-	return errdefs.ErrNotImplemented
+	_, err := c.sandboxStore.Get(id)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find sandbox %q in store", id)
+	}
+	stderr := new(bytes.Buffer)
+	err = c.ExecInSandbox(id, []string{"wincat.exe", "localhost", fmt.Sprint(port)}, stream, ioutils.WriteCloserWrapper(stream), ioutils.WriteCloserWrapper(stderr), false, nil)
```
Random-Liu: The line is too long.

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,68 @@ limitations under the License.
...
+	if err != nil {
+		return fmt.Errorf("%v: %s", err, stderr.String())
```
Random-Liu: Use `errors.Wrap`

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,68 @@ limitations under the License.
...
+	if sandbox.Status.Get().State != sandboxstore.StateReady {
+		return nil, errors.New("sandbox container is not running")
+	}
+
+	return c.streamServer.GetPortForward(r)
 }
 
 func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
-	return errdefs.ErrNotImplemented
+	_, err := c.sandboxStore.Get(id)
```
Random-Liu: No need to try getting the sandbox if we don't need the result. :)

pkg/server/sandbox_portforward_windows.go
Random-Liu: We can remove `tty` and `resize` from the function signature if we don't need them any time soon.

Random-Liu: Pass in the context and use it instead of creating a new one.

liyanhui1228: Done. The context needs to have a namespace so I created it using the `ctrdutil.NamespacedContext()`.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Done.

pkg/server/container_execsync.go
liyanhui1228: Done.

pkg/server/container_execsync.go
liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Fixed.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Removed.

pkg/server/sandbox_portforward_windows.go
Random-Liu: The context passed to portforward should be namespaced, if not we should fix it. :) https://github.com/containerd/cri/blob/master/pkg/server/streaming.go#L159

liyanhui1228: Yeah I'm getting this error ` failed to get container spec: namespace is required: failed precondition` if directly passing the `ctx`. I'll looking into it.

Random-Liu:
```
pkg\server\sandbox_portforward_windows.go:28:2: cannot find package "k8s.io/kubernetes/pkg/kubelet/util/ioutils" in any of:
	C:\gopath\src\github.com\containerd\cri\vendor\k8s.io\kubernetes\pkg\kubelet\util\ioutils (vendor tree)
	c:\go\src\k8s.io\kubernetes\pkg\kubelet\util\ioutils (from $GOROOT)
	C:\gopath\src\k8s.io\kubernetes\pkg\kubelet\util\ioutils (from $GOPATH)
```

It seems that you need to update vendor:
```
go get -u github.com/LK4D4/vndr
vndr
```

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,66 @@ limitations under the License.
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to execute port forward in sandbox: %s", stderrBuffer.String())
+	}
+
+	return nil
 }
 
-func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
-	return errdefs.ErrNotImplemented
+func (c *criService) execInSandbox(sandboxID string, cmd []string, stdin io.Reader, stdout, stderr io.WriteCloser) error {
```
jterry75: Do you have a reference of how this works for dockershim? I dont see how this can work on Windows. This seems like it would of needed to be a cni policy to expose the ports in the namespace. I dont see how exec'ing in the sandbox namespace helps here?

liyanhui1228: https://github.com/kubernetes/kubernetes/pull/75479 is the implementation for dockershim. In here this is using the same approach that executing the wincat command inside pause image.

jterry75: @Random-Liu, @patricklang - Is "PortForward" supposed to be an internal port forward ie inside the namespace or is it an additional port to expose as an external interface?

Random-Liu: @jterry75 It is supposed to forward a port inside the namespace to a port on your desktop (the desktop you run `kubectl portforward`)

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,66 @@ limitations under the License.
...
+	ctrdutil "github.com/containerd/cri/pkg/containerd/util"
+	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 )
 
-// PortForward prepares a streaming endpoint to forward ports from a PodSandbox, and returns the address.
-// TODO(windows): Implement this for windows.
-func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequest) (*runtime.PortForwardResponse, error) {
-	return nil, errdefs.ErrNotImplemented
+func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
+	stdout := ioutils.WriteCloserWrapper(stream)
```
Random-Liu: We have our own `nopWriteCloser`. https://github.com/containerd/cri/blob/master/pkg/ioutil/write_closer.go#L54

Use that.

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,66 @@ limitations under the License.
...
+	stderrBuffer := new(bytes.Buffer)
+	stderr := ioutils.WriteCloserWrapper(stderrBuffer)
+	cmd := []string{"wincat.exe", "localhost", fmt.Sprint(port)}
+	err := c.execInSandbox(id, cmd, stream, stdout, stderr)
+	if err != nil {
+		return errors.Wrapf(err, "failed to execute port forward in sandbox: %s", stderrBuffer.String())
+	}
+
```
Random-Liu: Unnecessary empty line

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,66 @@ limitations under the License.
...
+	return nil
 }
 
-func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
-	return errdefs.ErrNotImplemented
+func (c *criService) execInSandbox(sandboxID string, cmd []string, stdin io.Reader, stdout, stderr io.WriteCloser) error {
+	ctx := ctrdutil.NamespacedContext()
```
Random-Liu: Pass in the context?

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,66 @@ limitations under the License.
...
+	opts := execOptions{
+		cmd:    cmd,
+		stdin:  stdin,
+		stdout: stdout,
+		stderr: stderr,
+		tty:    false,
+		resize: nil,
+	}
+	cntx, cancel := context.WithCancel(ctx)
```
Random-Liu: It is fine to reuse the name `ctx`

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,66 @@ limitations under the License.
...
+	}
+
+	return nil
 }
 
-func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
-	return errdefs.ErrNotImplemented
+func (c *criService) execInSandbox(sandboxID string, cmd []string, stdin io.Reader, stdout, stderr io.WriteCloser) error {
+	ctx := ctrdutil.NamespacedContext()
+	opts := execOptions{
```
Random-Liu: nit: Prefer moving the variable definition to where it will be used.

test/windows/test.sh
```diff
@@ -21,7 +21,6 @@ set -o pipefail
 export PATH="/c/Program Files/Containerd:$PATH"
 REPO_TAR="${REPO_TAR:-"/c/cri.tar.gz"}"
 FOCUS="${FOCUS:-"Conformance"}"
-SKIP="${SKIP:-"portforward"}"
```
Random-Liu: Keep the SKIP, just make it empty string.

pkg/server/container_execsync.go
```diff
@@ -70,38 +70,7 @@ type execOptions struct {
...
-	}
-	id = cntr.ID
-
-	state := cntr.Status.Get().State()
-	if state != runtime.ContainerState_CONTAINER_RUNNING {
-		return nil, errors.Errorf("container is in %s state", criContainerStateToString(state))
-	}
-
-	container := cntr.Container
+func (c *criService) execUsingContainerClient(ctx context.Context, container containerd.Container, id string, opts execOptions) (*uint32, error) {
```
Random-Liu: Can we come up with a better name? :p

liyanhui1228: Changed to `execInternal`.

test/windows/test.sh
liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Removed.

pkg/server/sandbox_portforward_windows.go
liyanhui1228: Done.

liyanhui1228:
The nginx server is occasionally returning 400 status code, looking into it.

pkg/server/container_execsync.go
```diff
@@ -205,3 +174,38 @@ func (c *criService) execInContainer(ctx context.Context, id string, opts execOp
...
+// running in background after io is detached.
+// https://github.com/kubernetes/kubernetes/blob/v1.15.0/pkg/kubelet/dockershim/exec.go#L127
+// For example, if the `kubectl exec -it` process is killed, IO will be closed. In
+// this case, the CRI plugin will still have a goroutine waiting for the exec process
+// to exit and log the exit code, but dockershim won't.
+func (c *criService) execInContainer(ctx context.Context, id string, opts execOptions) (*uint32, error) {
+	// Cancel the context before returning to ensure goroutines are stopped.
+	// This is important, because if `Start` returns error, `Wait` will hang
+	// forever unless we cancel the context.
+	ctx, cancel := context.WithCancel(ctx)
```
Random-Liu: I think we should do this in `execInternal`. It is always required.

liyanhui1228: Done.

pkg/server/sandbox_portforward_windows.go
jterry75: Who calls cni to add external acl to open the port then? My understanding of this code is that it will forward a port inside the namespace to another port inside the namespace. Am I missing something?

Random-Liu: @jterry75 It forwards the in/out of a port to stdin/out of the exec process, and eventually to the user who runs `kubectl portforward`. So there is only one port involved.

Random-Liu:
```

+ Failure [10.703 seconds]
[k8s.io] Streaming
C:/Users/prow/AppData/Local/Temp/cri-install-crictl.aIFq/src/github.com/kubernetes-sigs/cri-tools/pkg/framework/framework.go:72
  runtime should support streaming interfaces
  C:/Users/prow/AppData/Local/Temp/cri-install-crictl.aIFq/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/streaming.go:55
    runtime should support portforward [Conformance] [It]
    C:/Users/prow/AppData/Local/Temp/cri-install-crictl.aIFq/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/streaming.go:124
    The status code of response should be 200.
      
    Expected
        <int>: 400
    to equal
        <int>: 200
```

The flaky test you mentioned. @liyanhui1228 Can you collect some log from the failed pod, so that we can look into it more? :) Thanks!

pkg/server/sandbox_portforward_windows.go
jterry75: Oh ok cool. Thanks

liyanhui1228:
Update: I exec'ed into the container and there is no log files in `C:\inetpub\logs` directory when the status code is 400. But there will be log files if the status code is 200. Still investigating.

Random-Liu:
Let's re-run the test several times. :)

/test pull-cri-containerd-windows-cri

liyanhui1228:
I checked the host IP mapping file:
```
PS C:\windows\system32\drivers\etc> cat hosts
cat hosts
...some unrelated stuff
# localhost name resolution is handled within DNS itself.
#       127.0.0.1       localhost
#       ::1             localhost
```

And run command `ping localhost`
```
PS C:\windows\system32\drivers\etc> ping localhost
ping localhost

Pinging ed0bac81d3cf [::1] with 32 bytes of data:
Reply from ::1: time<1ms
Reply from ::1: time<1ms
Reply from ::1: time<1ms
Reply from ::1: time<1ms

Ping statistics for ::1:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 0ms, Maximum = 0ms, Average = 0ms
```

`localhost` is resolved to `::1` instead of `127.0.0.1` in here. That could be the reason.

Random-Liu:
@liyanhui1228 I think it makes sense. And actually we are also only supporting IPv4 on linux. https://github.com/containerd/cri/blob/master/pkg/server/sandbox_portforward_unix.go#L86
```
TCP4:<host>:<port>
Like TCP, but only supports IPv4 protocol (example).
Option groups: FD,SOCKET,IP4,TCP,RETRY
```

pkg/server/sandbox_portforward_windows.go
```diff
@@ -19,19 +19,60 @@ limitations under the License.
...
 
-// PortForward prepares a streaming endpoint to forward ports from a PodSandbox, and returns the address.
-// TODO(windows): Implement this for windows.
-func (c *criService) PortForward(ctx context.Context, r *runtime.PortForwardRequest) (*runtime.PortForwardResponse, error) {
-	return nil, errdefs.ErrNotImplemented
+func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
+	stdout := ioutil.NewNopWriteCloser(stream)
+	stderrBuffer := new(bytes.Buffer)
+	stderr := ioutil.NewNopWriteCloser(stderrBuffer)
+	cmd := []string{"wincat.exe", "127.0.0.1", fmt.Sprint(port)}
```
Random-Liu: Add an comment about the IPv4 thing.

liyanhui1228: Done.

PR #1287, opened 2019-09-20
Use type alias from containerd for cgroup metric types

crosbymichael:
This helps to decouple the direct dependency for cgroup metrics from the cgroup package.


Random-Liu:
```
I0920 14:13:09.061] These files were modified:
I0920 14:13:09.061] 
I0920 14:13:09.062]  D vendor/github.com/containerd/cgroups/go.mod

**Comment by Random-Liu** [comment]
Test windows test failure is the known ssh on windows flakiness:
```
Permission denied (publickey,keyboard-interactive)
```

/test pull-cri-containerd-windows-cri

File: vendor.conf [hunk 1]
```diff
@@ -44,7 +44,7 @@ github.com/containerd/ttrpc 92c8520ef9f86600c650dd540266a007bf03670f
 github.com/containerd/go-runc 9007c2405372fe28918845901a3276c0915689a1
 github.com/containerd/fifo bda0ff6ed73c67bfb5e62bc9c697f146b7fd7f13
 github.com/containerd/continuity f2a389ac0a02ce21c09edd7344677a601970f41c
-github.com/containerd/containerd 59a625defb21c958c25424fa5cc806167e22382e
+github.com/containerd/containerd ed16170c4c399c57f25d6aa1e97b345ed6ab96cb
 github.com/containerd/console 0650fd9eeb50bab4fc99dceb9f2e14cf58f36e7f
 github.com/containerd/cgroups c4b9ac5c7601384c965b9646fc515884e091ebb9
```
jterry75: You have to update this too right?

crosbymichael: not yet

crosbymichael: I'll update this reference from containerd then i'll update cri and cgroups in the containerd core and I think we are golden

jterry75: @Random-Liu - How does the ci/cd build work. Does it come from containerd/containerd or does it build from here? Seems like it builds from here in which case we need the updated vendor tree here for ci/cd builds but we need it updated in containerd/containerd for official builds. Or am I 100% crazy...?

Random-Liu:
The windows permission denied issue is so annoying.

I filed https://github.com/containerd/cri/issues/1288, and will try to fix it as soon as possible.

vendor.conf
Random-Liu: Yeah, all tests build from here, so we do need to update the dependencies here right now.

crosbymichael: Currently the cgroup commit in containerd and cri is in sync.  It should be fine for now.

PR #1291, opened 2019-09-23
Fix indent in cni.template.

Random-Liu:
Fix indent in cni.template.


mikebrow:
yeah not if the parser does not care.. and there is no format check...

PR #1292, opened 2019-09-24
Document simpler kube-up.sh

Random-Liu:
Now we just need a simple `KUBE_CONTAINER_RUNTIME=containerd` environment variable before kube-up.sh.


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1292#pullrequestreview-292126623

PR #1294, opened 2019-09-24
Forward SandboxConfig.Hostname to Workload container activation

jterry75:
For https://github.com/containerd/cri/issues/1257

1. For Windows the Hostname property is not inherited from the sandbox and must
be passed for the Workload container activations as well.


Random-Liu:
@jterry75 Hostname is not shared in the network namespace on windows?

jterry75:
Negative. We have an internal investigation going on but while that goes on here is a quick fix that doesn't have any risk of breaking anything.

Issue #1297, opened 2019-09-25
[WCOW] Revisit sandbox container cpu shares


Random-Liu:
Ideally, the sandbox container for windows should have different CPU shares with linux because:
1) windows sandbox container handles portforward;
2) windows cpu shares range is 5000-10000 (see [here](https://github.com/kubernetes/kubernetes/blob/v1.16.1-beta.0/pkg/kubelet/kuberuntime/helpers_windows.go)), which is different from 2-1024 on linux (see [here](https://github.com/kubernetes/kubernetes/blob/v1.16.1-beta.0/pkg/kubelet/kuberuntime/helpers_linux.go)).

However, we still use the same CPU shares for now, because:
1) The underlying job cpu rate control only supports 1-9 shares on windows, so it doesn't matter whether we use 2 or 20 here. (See [here](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/resource-controls#cpu-shares-without-hyper-v-isolation))
2) It is not clear how much cpu portforward will need. We may want to have a better default configured, or even make it configurable if it causes trouble for some users who rely on portforward heavily.

PR #1298, opened 2019-09-25
Set default sandbox container cpu shares on windows.

Random-Liu:
For https://github.com/containerd/cri/issues/1257

This PR sets the same default sandbox container cpu shares on windows.

For more details about this, see https://github.com/containerd/cri/issues/1297

@yujuhong @jterry75 

### DIFF
--- a/pkg/containerd/opts/spec.go
+++ b/pkg/containerd/opts/spec.go
@@ -31,6 +31,10 @@ import (
 	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
+// DefaultSandboxCPUshares is default cpu shares for sandbox container.
+// TODO(windows): Revisit cpu shares for windows (https://github.com/containerd/cri/issues/1297)
+const DefaultSandboxCPUshares = 2
+
 // WithRelativeRoot sets the root for the container
 func WithRelativeRoot(root string) oci.SpecOpts {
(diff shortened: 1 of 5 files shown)


yujuhong:
> For more details about this, see #1297

Read the issue. Have just one question. Does this result in the same effective CPU shares as what's used by dockershim today?

Random-Liu:
> Read the issue. Have just one question. Does this result in the same effective CPU shares as what's used by dockershim today?

Yeah, dockershim is also using `2`.

jterry75:
@Random-Liu - Is your goal here to simply lower the priority scheduling of the sandbox container process? I want to make sure we are setting the right thing here.

Random-Liu:
@jterry75 The sandbox container should be mostly paused, so we want to lower its cpu share.

We set the same thing in dockershim today. https://github.com/kubernetes/kubernetes/blob/v1.16.1-beta.0/pkg/kubelet/dockershim/docker_sandbox.go#L45

jterry75:
> @jterry75 The sandbox container should be mostly paused, so we want to lower its cpu share.
> 
> We set the same thing in dockershim today. https://github.com/kubernetes/kubernetes/blob/v1.16.1-beta.0/pkg/kubelet/dockershim/docker_sandbox.go#L45

Yep just making sure. Thanks!

LGTM

PR #1301, opened 2019-09-26
[release/1.0] Make sure exec process is killed when context is canceled.

Random-Liu:
Cherrypick https://github.com/containerd/cri/pull/1156 into release/1.0

I think this is the only pre-August thing we need to backport from the commits https://github.com/containerd/cri/commits/release/1.2


Random-Liu:
Apply lgtm label based on https://github.com/containerd/cri/pull/1301#pullrequestreview-293966311

PR #1302, opened 2019-09-26
[release/1.3] Update cri-tools to v1.16.1.

Random-Liu:
Let's update cri-tools to the newest release, which includes many useful bug fixes and features.


Random-Liu:
Verify test is not running because of https://github.com/kubernetes/test-infra/pull/14488

PR #1303, opened 2019-09-26
[release/1.0] Sync vendors with containerd before release.

Random-Liu:
See https://github.com/containerd/containerd/compare/v1.1.7...release/1.1#diff-4061fcef378a6d912e14e2ce162a1995R23

Prerequisite for https://github.com/containerd/containerd/pull/3693

The vendor.conf in the CRI directory isn't that important, but let's sync once anyway before release. :)


jterry75:
@Random-Liu - Just for my info why does this matter? The release is built out of cd so this vendor.conf doesnt matter anyways and is just for testing right?

Random-Liu:
@jterry75 For old release branch, we still need to build the cri release tarball manually https://github.com/containerd/cri/blob/master/docs/release.md. By default the dependency versions come from vendor.conf in this repo.

We can do something like https://github.com/containerd/containerd/pull/3692, but that needs to change the old release flow a little bit.

I'll add auto CRI release deploy into containerd/containerd release/1.2, hopefully we won't need to care about this anymore. And as you said, the vendor.conf will just be fore testing. :D

Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1303#pullrequestreview-293986870

PR #1309, opened 2019-10-04
Update containerd to a6a0c8b6e36415a151d93d096c1c0af9e0bd7977.

Random-Liu:
We've got more test coverage in the cri repo.

Let's update containerd to include https://github.com/containerd/containerd/pull/3711, so that we get more test coverage before cherrypicking it.


jterry75:
@Random-Liu - Is there a general strategy for updating this? Or said differently, is there a future where cri moves into containerd to avoid this need?

Random-Liu:
One test pod got killed with exit status 137, I guess it is OOM kill.

Triggered rerun.

> Is there a general strategy for updating this? Or said differently, is there a future where cri moves into containerd to avoid this need?

Moving the CRI plugin into `containerd/containerd` shouldn't be too hard. The annoying part is to update all the test/release scripts and infrastructure.

For example, do we want a 1hour windows CRI validation presubmit test in the containerd repo? I think not...

Random-Liu:
The PR includes 2 test changes:
1) Tolerant not found error in restart test. There is a race condition that containerd event monitor may delete the pod sandbox before the test after it is killed. The `state` call is faster now, which somehow exposed this race.
2) The unknown state restart test is removed. I really don't want to remove it. However, after the `runc state` fix, getting container state doesn't need `runc` anymore, the original way to move runc around to produce a container in unknown state doesn't work any more... And **I think very hard, but really can't find a way to create a container in unknown state in the test....**

integration/restart_test.go
```diff
@@ -195,150 +195,3 @@ func TestContainerdRestart(t *testing.T) {
 		assert.True(t, found, "should find image %+v", i1)
 	}
 }
-
-// Note: The test moves runc binary.
-// The test requires:
-// 1) The runtime is runc;
-// 2) runc is in PATH;
-func TestUnknownStateAfterContainerdRestart(t *testing.T) {
```
mikebrow: maybe a todo or issue to develop an integration test for unknownstate

Random-Liu: Done

Random-Liu: https://github.com/containerd/cri/issues/1310

Random-Liu:
Updated containerd again to include https://github.com/containerd/containerd/pull/3730 for the test.

Random-Liu:
Apply LGTM based on LGTMs above.

PR #1312, opened 2019-10-08
Explicitly specify trusty in travis.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1311.

This PR:
1) Fixes appveyor test. Command without `.exe` stopped working somehow.
2) Fixes golangci-lint version (same with containerd/containerd). We hit a lot of unexpected verification timeout failures.
3) Updates `.travis.yml` based on new default ubuntu distro `xenial`. https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment
4) Updates integration test to work with golang 1.13 https://github.com/golang/go/issues/31859#issuecomment-532415187

2 and 3 should be cherry-picked.

Signed-off-by: Lantao Liu <lantaol@google.com>

PR #1320, opened 2019-10-22
[release/1.0] Fix containerd build, use `libbtrfs-dev` when available.

Random-Liu:
Cherrypick #1319 to release/1.0.
Cherrypick #1312 to release/1.0.

We won't add new bug fix or feature into release/1.0, but this change is needed to fix the build. https://k8s-testgrid.appspot.com/sig-node-containerd#containerd-build-1.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,14 +13,12 @@ cache:
         - "${HOME}/google-cloud-sdk/"
 
 before_install:
-    # libseccomp in trusty is not new enough, need backports version.
-    - sudo sh -c "echo 'deb http://archive.ubuntu.com/ubuntu trusty-backports main restricted universe multiverse' > /etc/apt/sources.list.d/backports.list"
     - sudo apt-get update
 
 install:
     - sudo apt-get install btrfs-tools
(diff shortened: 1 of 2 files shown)


thaJeztah:
```
E: Essential packages were removed and -y was used without --allow-remove-essential.
The command "sudo apt-get install libseccomp2/trusty-backports" failed and exited with 100 during .
```

Random-Liu:
@thaJeztah Done

thaJeztah:
boo! looks like there's still failures 😢 (reading from my phone, so it's hard to see what the exact issue is - sorry!)

Random-Liu:
> boo! looks like there's still failures cry (reading from my phone, so it's hard to see what the exact issue is - sorry!)

The problem is that the ubuntu in travis is using Docker 18.06, which has containerd listens on 10010 https://github.com/coreos/bugs/issues/2524. And we have the same problem in the CRI plugin in the release/1.0 branch. :P

thaJeztah:
Oh the irony!
Is that the default that Travis itself has? (if so, they should definitely update, as 18.06 reached EOL quite some time ago)

PR #1322, opened 2019-10-22
[release/1.3] Fix containerd build, use `libbtrfs-dev` when available.

Random-Liu:
Cherrypick #1319 to release/1.3


Random-Liu:
Apply LGTM based on LGTMs above.

PR #1324, opened 2019-10-23
Avoid O_NONBLOCK on stdout in the windows test.

Random-Liu:
This is a trial to fix the windows test.

See https://github.com/kubernetes/test-infra/issues/14938 for more details.
/cc @BenTheElder


Random-Liu:
From the log, we can see that `O_NONBLOCK` is set by mistake:
```
+ python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); print(flags&os.O_NONBLOCK);'
2048
+ python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
```

So this fix makes sense.

PR #1328, opened 2019-10-28
Use `http` for localhost and 127.0.0.1 by default.

Random-Liu:
Fixes https://github.com/containerd/cri/issues/1327.

This changes the behavior to be consistent with Docker. We should make sure we mention this in the release note.

/cc @liyimeng 

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -310,6 +310,10 @@ func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHost
 				config    = c.config.Registry.Configs[u.Host]
 			)
 
+			if u.Scheme == "" {
+				u.Scheme = defaultScheme(u.Host)
+			}
+
 			if u.Scheme != "https" && config.TLS != nil {
 				return nil, errors.Errorf("tls provided for http endpoint %q", e)
 			}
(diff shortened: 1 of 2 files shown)


pkg/server/image_pull.go
```diff
@@ -346,6 +353,17 @@ func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHost
 	}
 }
 
+// defaultScheme returns the default scheme for a registry host.
+func defaultScheme(host string) (string, error) {
+	if h, _, err := net.SplitHostPort(host); err == nil {
+		host = h
+	}
+	if host == "localhost" || host == "127.0.0.1" {
```
mikebrow: || "::1" 

^ for ipv6

pkg/server/image_pull.go
```diff
@@ -346,6 +353,17 @@ func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHost
 	}
 }
 
+// defaultScheme returns the default scheme for a registry host.
+func defaultScheme(host string) (string, error) {
```
mikebrow: probably don't need to return error since it's always nil..

pkg/server/image_pull_test.go
```diff
@@ -236,3 +236,48 @@ func TestRegistryEndpoints(t *testing.T) {
...
+			expected: "https",
+		},
+		"should use https by default for remote ip": {
+			host:     "8.8.8.8",
+			expected: "https",
+		},
+		"should use https by default for remote ip with port": {
+			host:     "8.8.8.8:8080",
+			expected: "https",
+		},
```
mikebrow: see ipv6 case..

pkg/server/image_pull.go

pkg/server/image_pull.go
Random-Liu: Done. Left from old change.

pkg/server/image_pull_test.go
Random-Liu: Done

Random-Liu:
@mikebrow Can you take another look? Just want sure your comments are fully addressed. :)

PR #1329, opened 2019-10-29
Update support metrics.

Random-Liu:
Update support metrics to reflect that:
1) containerd 1.1 is end of life;
2) Kubernetes 1.13 is end of life.


Random-Liu:
Apply the label based on LGTM above.

PR #1330, opened 2019-10-29
missed the release version for HEAD in table-1

mikebrow:
minor update regarding HEAD support


Random-Liu:
It is reflected in the second table, right?

mikebrow:
> It is reflected in the second table, right?

Yes kube n-1 version number is reflected in the second table, not the first. 

Alternatively we could just remove the HEAD support row...

Random-Liu:
> Alternatively we could just remove the HEAD support rows...

SG. :)

Issue #1327, opened 2019-10-26
pulling localhost:5000 are forced to https only? 


liyimeng:
There is a private registry running http at localhost, crictl pull images there result failing as below:
```
Normal   Scheduled  <unknown>             default-scheduler  Successfully assigned default/vm-0 to iserver
  Warning  Failed     39m (x2 over 40m)     kubelet, iserver   Failed to pull image "localhost:5000/vm-runner:dev": rpc error: code = Unknown desc = failed to pull and unpack image "localhost:5000/vm-runner:dev": failed to resolve reference "localhost:5000/vm-runner:dev": failed to do request: Head https://localhost:5000/v2/vm-runner/manifests/dev: http: server gave HTTP response to HTTPS client

iserver [~]$ crictl -v
crictl github.com/rancher/containerd v1.16.1-k3s.1 
```
Is crictl behave differently from docker?, that is forcing to run https on localhost as well?

Random-Liu:
What containerd version are you using? This should have been fixed in containerd 1.3

liyimeng:
Thanks a lot @Random-Liu! The info is really helpful! could be nice to allow fallback into http on localhost. Since int many cases, especially in a development env. people want a quick and work solution.

thaJeztah:
> Let's see how Docker handle this.

Docker automatically adds `127.0.0.0/8` as insecure registries;

```
 Insecure Registries:
  127.0.0.0/8
```

Whether or not that was a good decision is another discussion (it had been going back-and-forth for a long time, and initially was rejected https://github.com/moby/moby/pull/2687#issuecomment-40853094, https://github.com/moby/moby/pull/5817#issuecomment-46952418, after which it finally got added in https://github.com/moby/moby/pull/9124).

For most uses, I guess that's ok, but there could be situations where you want to have that range to use a secure connection as well. I don't think that's currently possible (as it will always add that range to the list of insecure registries).

Random-Liu:
For the CRI plugin, we can make localhost insecure by default to keep consistency with Docker, but allow making it https.

PR #1331, opened 2019-10-31
Fix using `http` for localhost, 127.0.0.1 and ::1 by default

erikwilson:
It appears that PR #1328 for issue #1327 needs to be re-arranged a little to work, since c.registryEndpoints is pre-pending https the default scheme should be inserted there.

/cc @Random-Liu

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -310,10 +310,6 @@ func (c *criService) registryHosts(auth *runtime.AuthConfig) docker.RegistryHost
 				config    = c.config.Registry.Configs[u.Host]
 			)
 
-			if u.Scheme == "" {
-				u.Scheme = defaultScheme(u.Host)
-			}
-
 			if u.Scheme != "https" && config.TLS != nil {
 				return nil, errors.Errorf("tls provided for http endpoint %q", e)
 			}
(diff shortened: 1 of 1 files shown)


Random-Liu:
I think we `defaultScheme` at both places. If the user configured a non-default endpoint to localhost, the original `defaultScheme` call is still needed.

erikwilson:
I think the old `defaultScheme` call should be removed because we can say a non-default endpoint _must_ be a valid url (with scheme). If we consider a host with port `url.Parse("localhost:5000/foobar")` will have a scheme of `localhost` rather than the checked empty string, so the logic becomes messy.

erikwilson:
Also, godocs state: `Trying to parse a hostname and path without a scheme is invalid but may not necessarily return an error, due to parsing ambiguities.` 
`url.Parse("127.0.0.1:5000")` does produce an error, but should be considered valid with the concepts of the replaced code.

Not sure why the tests are currently failing @Random-Liu, or how added testing might help here.

erikwilson:
If it is necessary to support custom endpoints without a scheme, then it should probably happen in c.registryEndpoints before url.Parse, where we need to check for the substring `://` and prepend a scheme before parsing.

/cc @mikebrow

mikebrow:
@erikwilson pls sign the commit... 
```git commit -s --amend```
and push it back up so we can see the test results :-)

erikwilson:
Thanks @mikebrow, looks like everything is passing now. :)

Random-Liu:
> I think the old defaultScheme call should be removed because we can say a non-default endpoint must be a valid url (with scheme). If we consider a host with port url.Parse("localhost:5000/foobar") will have a scheme of localhost rather than the checked empty string, so the logic becomes messy.

SG. :)

PR #1335, opened 2019-11-12
Add support for / in Windows mounts

jterry75:
Kubelet sometimes passes Windows paths like c:/test. This change replaces all
'/' in HostPath and ContainerPath for Windows containers with '\'.

### DIFF
--- a/pkg/server/container_create_windows.go
+++ b/pkg/server/container_create_windows.go
@@ -19,6 +19,8 @@ limitations under the License.
 package server
 
 import (
+	"strings"
+
 	"github.com/containerd/containerd/oci"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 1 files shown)


jterry75:
@patricklang - FYI.

@adelina-t - Can you test with this change and make sure it works as expected?

jterry75:
@Random-Liu - I am surprised this passed the k8s e2e tests. Are we running those for Windows on CRI-Containerd?

adelina-t:
@jterry75 Unfortunately this doesn't work as expected. After being replaced, the mounts eventually get here: https://github.com/jterry75/cri/blob/314e8533f978e25dbaae4ba5b3c5d85791b57a63/pkg/containerd/opts/spec_windows.go#L65

There we have a problem:

criMounts - contains the mounts directly from config ( unaltered path ) [1]
extra - contains the altered mounts + volume mounts [2][3]

WithWindowsMounts, when trying to copy mounts from extramounts, if it finds the same path in both criMounts and extramounts, it will use criMounts[4]. What unfortunately happens when comparing the two paths, is that it uses filepath.Clean() which on Windows, has the undocumented side effect of replacing "/" with os.PathSeparator . So basically, when comparing the two mounts containerPaths here [5], even though one is the original path with "/" and the other the altered one with "\", they will be in fact equal because for the comparison they are transformed by filepath.Clean(). Since they are equal, the one from criMounts will be prefered, meaning the unaltered one.

[1] https://github.com/jterry75/cri/blob/314e8533f978e25dbaae4ba5b3c5d85791b57a63/pkg/containerd/opts/spec_windows.go#L58
[2]
https://github.com/jterry75/cri/blob/314e8533f978e25dbaae4ba5b3c5d85791b57a63/pkg/server/container_create.go#L151
[3] https://github.com/jterry75/cri/blob/314e8533f978e25dbaae4ba5b3c5d85791b57a63/pkg/server/container_create_windows.go#L79
[4]
https://github.com/jterry75/cri/blob/314e8533f978e25dbaae4ba5b3c5d85791b57a63/pkg/containerd/opts/spec_windows.go#L61
[5]
https://github.com/jterry75/cri/blob/314e8533f978e25dbaae4ba5b3c5d85791b57a63/pkg/containerd/opts/spec_windows.go#L65

Random-Liu:
Sent out https://github.com/containerd/cri/pull/1337 for a proper fix.

Random-Liu:
#1337 is merged. Close this for now.

Feel free to reopen if #1337 doesn't fix the issue.

PR #1341, opened 2019-11-22
Update docker/docker vendor to upstream latest

estesp:
See containerd/containerd#3839 for background.


estesp:
will need containerd/containerd#3842 first

Random-Liu:
Some other vendors also need to be updated. I can handle it in https://github.com/containerd/cri/pull/1338

Random-Liu:
LGTM

PR #1344, opened 2019-11-26
Provide resolvConf to sandbox container's mounts

darfux:
As https://github.com/kata-containers/runtime/issues/1603 discussed, kata relies on such mount spec to setup resolv.conf for pod VM properly.

Signed-off-by: Li Yuxuan <liyuxuan04@baidu.com>

### DIFF
--- a/pkg/server/sandbox_run_unix.go
+++ b/pkg/server/sandbox_run_unix.go
@@ -108,6 +108,13 @@ func (c *criService) sandboxContainerSpec(id string, config *runtime.PodSandboxC
 			Type:        "bind",
 			Options:     []string{"rbind", "ro"},
 		},
+		// Add resolv.conf for katacontainers to setup the DNS of pod VM properly.
+		{
+			Source:      c.getResolvPath(id),
+			Destination: resolvConfPath,
+			Type:        "bind",
+			Options:     []string{"rbind", "ro"},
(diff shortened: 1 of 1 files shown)


Random-Liu:
@darfux Is `/etc/hosts` and `/etc/hostname` needed?

darfux:
@Random-Liu  I think only resolv.conf is ok. `hostname` can be retrieved from ocispec [oci/utils.go#L774](https://github.com/kata-containers/runtime/blob/613fd0fb60a4a88da24012b7a422a30a3f30271d/virtcontainers/pkg/oci/utils.go#L774), while `/etc/hosts` is not considered and using the original VM rootfs version right now.

pkg/server/sandbox_run_unix.go
```diff
@@ -108,6 +108,12 @@ func (c *criService) sandboxContainerSpec(id string, config *runtime.PodSandboxC
 			Type:        "bind",
 			Options:     []string{"rbind", "ro"},
 		},
```
Random-Liu: Can you add a comment about why is this needed?

I already forgot why we mounted dev shim, so it would be better to put a comment here.

darfux: done, PTAL 😁

PR #1345, opened 2019-11-26
Add insecure_skip_verify option.

Random-Liu:
Fixes https://github.com/containerd/containerd/issues/3847.

We should cherrypick this into 1.3, because this is a bug.


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1345#pullrequestreview-323321936

PR #1349, opened 2019-12-03
Use containerD WithHostDevices

### LINKED ISSUE #1316: Use new host device options added in containerd
See https://github.com/containerd/containerd/pull/3718

This helps code deduplication.

Zyqsempai:
Fixes #1316 
Signed-off-by: bpopovschi <zyqsempai@mail.ru>

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -170,7 +170,7 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		}
 		specOpts = append(specOpts, oci.WithPrivileged)
 		if !ociRuntime.PrivilegedWithoutHostDevices {
-			specOpts = append(specOpts, customopts.WithPrivilegedDevices)
+			specOpts = append(specOpts, oci.WithHostDevices)
 		}
 	} else { // not privileged
 		specOpts = append(specOpts, customopts.WithDevices(c.os, config), customopts.WithCapabilities(securityContext))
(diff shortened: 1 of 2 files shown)


pkg/server/container_create_unix.go
```diff
@@ -168,10 +168,7 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		if !sandboxConfig.GetLinux().GetSecurityContext().GetPrivileged() {
 			return nil, errors.New("no privileged container allowed in sandbox")
 		}
-		specOpts = append(specOpts, oci.WithPrivileged)
-		if !ociRuntime.PrivilegedWithoutHostDevices {
```
mikebrow: could we add some comments here... I don't follow why are we removing/ignoring the privilegedwithouthostdevices config directive and replacing it with always withhostdevices when privileged.  

	// PrivilegedWithoutHostDevices overloads the default behaviour for adding host devices to the
	// runtime spec when the container is privileged. Defaults to false.

pkg/server/container_create_unix.go
```diff
@@ -168,10 +168,7 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
...
-			specOpts = append(specOpts, customopts.WithPrivilegedDevices)
```
mikebrow: I suspect you should just replace WithPrivilegedDevices with WithHostDevices. @Random-Liu thoughts?

Random-Liu: Yeah, and remove the code for `WithPrivilegedDevices`.

Random-Liu:
It seems that this change broke the kubernetes test.

The change was merged at 12/5 2:13 PM
The next build was at 12/5 2:15 PM
The first test failure started at 12/5 2:20 PM

https://k8s-testgrid.appspot.com/sig-node-containerd#e2e-gci

I'll revert this change to bring the test back to normal, and let's investigate this more.

PR #1350, opened 2019-12-04
sync create timestamp when loading the container

sofat1989:
sync createAt in the status if this container is loaded back
https://github.com/containerd/containerd/issues/3873


mikebrow:
/ok-to-test
Thanks for the PR @sofat1989

Random-Liu:
See https://github.com/containerd/cri/pull/1351

PR #1363, opened 2019-12-20
Validate and update the right config

Random-Liu:
`ValidatePluginConfig` may update the config for defaulting. It should be done before creating the service.

/cc @bentheelder


Random-Liu:
Apply LGTM based on https://github.com/containerd/cri/pull/1363#pullrequestreview-335487707

PR #1368, opened 2020-01-07
update for containerd vendor changes

mikebrow:
A number of vendor changes need to be done to start synching up with containerd/vendor.conf changes that went in last month up to commit 1fb1d93212af763bd481b476293040b1566ae00b  also addresses #1369 

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

PR #1369, opened 2020-01-08
Update vendored containerd/container.

adelina-t:
Update vendored containerd/containerd to use commit "v2: Fix missing ns when openShimLog on windows"
(1fb1d93212af763bd481b476293040b1566ae00b) in order to prevent the following failures when creating
containers on Windows:

"failed to create PodSandbox: rpc error: code = FailedPrecondition desc = failed
to create containerd task: open shim log pipe: namespace is required: failed precondition"

NOTE: Failures observer while running Kubernetes e2e tests on Windows with containerd. (https://testgrid.k8s.io/sig-windows#containerd-l2bridge-windows-master)


jterry75:
@mikebrow - Should we do this as part of #1368 at the same time?

mikebrow:
> @mikebrow - Should we do this as part of #1368 at the same time?

Yes, @jterry75 that makes sense... either way this PR would need a rebase.. so I'll merge her change and move up one commit on containerd... wrt my vendor merge efforts PR and pick up this change. Thanks @adelina-t!

PR #1371, opened 2020-01-09
cgroup2: unshare cgroup namespace for containers

AkihiroSuda:
In cgroup v1 container implementations, cgroupns is not used by default because
it was not available in the kernel until kernel 4.6 (May 2016), and the default
behavior will not change on cgroup v1 environments, because changing the
default will break compatibility and surprise users.

For cgroup v2, implementations are going to unshare cgroupns by default
so as to hide /sys/fs/cgroup from containers.

* Discussion: https://github.com/containers/libpod/issues/4363
* Podman PR (merged): https://github.com/containers/libpod/pull/4374
* Moby PR: https://github.com/moby/moby/pull/40174

This PR enables cgroupns for containers, but pod sandboxes are untouched
because probably there is no need to do.

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/containerd/cgroups"
 	"github.com/containerd/containerd/contrib/apparmor"
 	"github.com/containerd/containerd/contrib/seccomp"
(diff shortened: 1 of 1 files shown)


giuseppe:
> cc @giuseppe @mrunalp @saschagrunert for consistency with CRI-O

thanks, I've opened a PR for CRI-O as well: https://github.com/cri-o/cri-o/pull/3098

PR #1384, opened 2020-01-27
Fix store error serialization to gRPC status codes

jterry75:
The pkg/store errors are duplicated errors of NotFound and AlreadyExist from
containerd's errdefs package and thus do not properly serialize when running
errdefs.ToGRPC on them. CRI runs this function on every return from a CRI
method so the conversion fails if there is a cache miss from the store caches
for containers or sandboxes. This change verifies that the errors are properly
converted to their gRPC values.

### DIFF
--- a/pkg/store/errors.go
+++ b/pkg/store/errors.go
@@ -16,12 +16,18 @@ limitations under the License.
 
 package store
 
-import "errors"
+import "github.com/containerd/containerd/errdefs"
 
 var (
 	// ErrAlreadyExist is the error returned when data added in the store
 	// already exists.
-	ErrAlreadyExist = errors.New("already exists")
+	//
+	// This error has been DEPRECATED and will be removed in 1.5. Please switch
+	// usage directly to `errdefs.ErrAlreadyExists`.
+	ErrAlreadyExist = errdefs.ErrAlreadyExists
 	// ErrNotExist is the error returned when data is not in the store.
-	ErrNotExist = errors.New("does not exist")
+	//
+	// This error has been DEPRECATED and will be removed in 1.5. Please switch
+	// usage directly to `errdefs.ErrNotFound`.
+	ErrNotExist = errdefs.ErrNotFound
 )
--- a/pkg/store/errors_test.go
+++ b/pkg/store/errors_test.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


jterry75:
@Random-Liu - What is the issue here with the Windows pull cri? Anything I can help with to make it more reliable?

pkg/store/errors_test.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
mikebrow: 2020

jterry75: Oh funny. Want me to update other headers in a follow up?

mikebrow: :) no need to update... might be best to remove the year(s) per the way containerd/containerd is doing it at some point in time.

jterry75:
@Random-Liu - Any chance you can take a look here?

thaJeztah:
@jterry75 does this need backporting to the 1.2, 1.3 branches?

pkg/store/errors.go
```diff
@@ -16,12 +16,12 @@ limitations under the License.
...
+import "github.com/containerd/containerd/errdefs"
 
 var (
 	// ErrAlreadyExist is the error returned when data added in the store
 	// already exists.
-	ErrAlreadyExist = errors.New("already exists")
+	ErrAlreadyExist = errdefs.ErrAlreadyExists
 	// ErrNotExist is the error returned when data is not in the store.
-	ErrNotExist = errors.New("does not exist")
+	ErrNotExist = errdefs.ErrNotFound
```
thaJeztah: Thinking out loud: wondering if the abstraction should be kept, or if these should be marked deprecated and locations that use these errors to use the `containerd/errdefs` package directly

jterry75: I'm ok removing the abstraction. I was hoping @Random-Liu would chime in on an approach as well. So I did this as the least invasive to the overall project references.

thaJeztah: Definitely not urgent, and I agree on keeping the diff for this PR small; 

if it's decided to deprecate (for a follow-up), these aliases can be kept (to not break potential consumers of this package), but marked "deprecated" (motivating them to switch to the containerd/errdefs package).

jterry75: Yep I can do that as a followup to this PR

mikebrow: SGTM to just mark them as deprecated in the comments along with a TODO: to remove in 1.5

jterry75: Sounds good. Are we ok with this PR or should we wait for @Random-Liu ?

mikebrow: can you add the comment deprecating it with the TODO?

jterry75: My bad. I misunderstood this to be a followup PR. Done now.

mikebrow: :-)

PR #1387, opened 2020-01-28
vendor: bump gopkg.in/yaml.v2 v2.2.8

thaJeztah:
full diff: https://github.com/go-yaml/yaml/compare/v2.2.4...v2.2.8

includes:

- go-yaml/yaml@f90ceb4f409096b60e2e9076b38b304b8246e5fa Fix check for non-map alias merging in v2
    - fix for "yaml.Unmarshal crashes on "assignment to entry in nil map""
- go-yaml/yaml 543 Port stale simple_keys fix to v2
- go-yaml/yaml@1f64d6156d11335c3f22d9330b0ad14fc1e789ce Fix issue in simple_keys improvements
    - fixes "Invalid simple_keys now cause panics later in decode"
- go-yaml/yaml 555 Optimize cases with long potential simple_keys

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -67,7 +67,7 @@ k8s.io/client-go kubernetes-1.17.1
 k8s.io/api kubernetes-1.17.1
 k8s.io/apiserver kubernetes-1.17.1
 k8s.io/apimachinery kubernetes-1.17.1
-gopkg.in/yaml.v2 v2.2.4
+gopkg.in/yaml.v2 53403b58ad1b561927d19068c655246f2db79d48 # v2.2.8
 gopkg.in/inf.v0 v0.9.1
 golang.org/x/time 9d24e82272b4f38b78bc8cff74fa936d31ccd8ef
 golang.org/x/oauth2 0f29369cfe4552d0e4bcddc57cc75f4d7e672a33
--- a/vendor/gopkg.in/yaml.v2/decode.go
+++ b/vendor/gopkg.in/yaml.v2/decode.go
@@ -319,10 +319,14 @@ func (d *decoder) prepare(n *node, out reflect.Value) (newout reflect.Value, unm
 }
 
 const (
-	// 400,000 decode operations is ~500kb of dense object declarations, or ~5kb of dense object declarations with 10000% alias expansion
+	// 400,000 decode operations is ~500kb of dense object declarations, or
+	// ~5kb of dense object declarations with 10000% alias expansion
 	alias_ratio_range_low = 400000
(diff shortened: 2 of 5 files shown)


thaJeztah:
opening for consideration; note that Kubernetes 1.17.1 does not yet use this version, so I wasn't sure if this project wants to stick to the exact version used in Kubernetes, or if being "ahead" is ok

mikebrow:
> opening for consideration; note that Kubernetes 1.17.1 does not yet use this version, so I wasn't sure if this project wants to stick to the exact version used in Kubernetes, or if being "ahead" is ok

Being slightly ahead of k8s is ok for short stretches so long as it's for a sound reason and we are not breaking compatibility with the supported k8s releases. I don't see us using yaml parsing for any of the CRI interfaces. We do use yaml for some install / ansible code, but that should be ok and we would not yet be using the new fields offered by this update. So.. let's test this out.

dims:
LGTM

mikebrow:
@thaJeztah Thanks a ton.. I think just need this yaml bump cherry picked to 1.3?  need a spread sheet :)

thaJeztah:
Yes, we should cherry-pick this one, and possibly update kubernetes to the latest 1.16.x version (see https://github.com/containerd/cri/pull/1386#issuecomment-579677073)

mikebrow:
"possibly" Yes, as you properly surmised this would be a big change for what is now a fairly old (in k8s terms) service branch. Notably Kube release support is only n-2 (plus CVE special cases) and they bump every quarter, making the release march with new features and deprecations only 9months long. If there are breaking changes in k8s due to deprecations, api additions, or just accidents we may be forcing 1.2 users of containerd to bump up their k8s. Gulp. You'll also note we have differences in CI testing as well between 1.2 and 1.3 and master, meaning some breaks might not get tested unless that is also merge where intertwined with the 1.3 release. Further, if cloud users don't bump k8s or do but don't go from 1.12 to 1.16 the mix of 1.13,14,15 releases on 1.2 code base... It might make more sense to ask k8s for 1.12 cve prs, than to bump our maintenance release (1.2) up. I'll be pinging sig-node folks on this.

thaJeztah:
oh, sorry meant the 1.3 branch, which is already on kubernetes 1.16, but currently on 1.16.3 (which is not the latest patch release of 1.16); https://github.com/containerd/cri/blob/b350e25534aeb5a3ba23954752ae839c96c58727/vendor.conf#L57-L61

mikebrow:
whew thx

PR #1392, opened 2020-01-30
[release/1.2] Sync vendors with containerd 1.2.11

dims:
Look through the `containerd dependencies` in vendor.conf, for each
dependency / SHA, check if the SHA is newer in containerd's vendor.conf
in 1.2.11 tag. Update if necessary.

Drop unused packages. Add package that was not present.


thaJeztah:
note that the containerd 1.2 _branch_ (so what will be in 1.2.12) also has https://github.com/containerd/containerd/pull/3711 backported (through https://github.com/containerd/containerd/pull/3755), so https://github.com/containerd/cri/pull/1309 might be needed as well in this branch

(it looks to be test-changes only though, so perhaps can wait until after 1.2.12 has been released - it's a bit of a chicken-and-egg situation)

thaJeztah:
On a side-note: @mikebrow is there a need to update the kubernetes dependencies in this branch to the latest 1.12.x patch release? I notice this branch has kubernetes 1.12.0, but latest patch release for kubernetes 1.12 is 1.12.10 https://github.com/kubernetes/kubernetes/releases/tag/v1.12.10

mikebrow:
> On a side-note: @mikebrow is there a need to update the kubernetes dependencies in this branch to the latest 1.12.x patch release? I notice this branch has kubernetes 1.12.0, but latest patch release for kubernetes 1.12 is 1.12.10 https://github.com/kubernetes/kubernetes/releases/tag/v1.12.10

I think that would be a very good idea. After a cursory check through the patches nothing stands out as a possible break to any of our interaction points with k8s, other than a couple obvious possibilities that we are already covering like moving up to go 1.10.x, and updating crictl. But there are too many commits there, to go over them with any great detail and for assurance.  In theory it should just be a validation test with containerd against the k8s patches, a good thing to do given that kubernetes patch testing (to my knowledge) does not test against the various runtimes outside the default.

dims:
@mikebrow  @thaJeztah nothing much to gain by moving to 1.12.10 :) https://paste.centos.org/view/raw/24c3e7f8

mikebrow:
In hind sight we could move up to 1.14 k8s .. esp since we are already testing containerd 1.2 against 1.14.  https://github.com/kubernetes/test-infra/blob/master/config/jobs/containerd/cri/containerd-cri-presubmit-jobs.yaml#L120

thaJeztah:
> @mikebrow @thaJeztah nothing much to gain by moving to 1.12.10 :)

In that case, looks like there's no risk; it would be just for "sanity" (and taking away any doubt that we didn't include possible fixes); but for sure doesn't need to be in this PR

(my 0.02c - I'm not a maintainer for this repository 🤗)

mikebrow:
> In hind sight we could move up to 1.14 k8s .. esp since we are already testing containerd 1.2 against 1.14. https://github.com/kubernetes/test-infra/blob/master/config/jobs/containerd/cri/containerd-cri-presubmit-jobs.yaml#L120

@dims @thaJeztah which reminds me .. we have not yet moved up the version of kubernetes that we are testing against for presubmit jobs.  @Random-Liu @yujuhong When should we move up the version in infra that we are testing against from 1.15 to 1.16..

mikebrow:
I see AWS moved to containerd on their infra tests so that's good.

thaJeztah:
oh, @dims could you add `[release/1.2]` prefix to the PR title (easier to find it back)


> we have not yet moved up the version of kubernetes that we are testing against for presubmit jobs.

I'll leave the kubernetes bumps up to the maintainers here 😅 - perhaps could be something for containerd 1.2.13 (to not block the 1.12.12 release?)

dims:
whoooops! done.

+1 to leave it alone for now (avoid blocking 1.12.12)

PR #1405, opened 2020-02-27
reload cni network config if has fs change events

fuweid:
With go RWMutex design, no goroutine should expect to be able to
acquire a read lock until the read lock has been released, if one
goroutine call lock.

The original design is to reload cni network config on every single
Status CRI gRPC call. If one RunPodSandbox request holds read lock
to allocate IP for too long, all other RunPodSandbox/StopPodSandbox
requests will wait for the RunPodSandbox request to release read lock.
And the Status CRI call will fail and kubelet becomes NOTReady.

Reload cni network config at every single Status CRI call is not
necessary and also brings NOTReady situation. To lower the possibility
of NOTReady, CRI will reload cni network config if there is any valid fs
change events from the cni network config dir.

### DIFF
--- a/pkg/server/cni_conf_syncer.go
+++ b/pkg/server/cni_conf_syncer.go
@@ -1,0 +1,121 @@
+/*
+Copyright The Containerd Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"os"
+	"sync"
+
+	cni "github.com/containerd/go-cni"
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// cniNetConfSyncer is used to reload cni network conf triggered by fs change
+// events.
+type cniNetConfSyncer struct {
+	// only used for lastSyncStatus
+	sync.RWMutex
+	lastSyncStatus error
+
+	watcher   *fsnotify.Watcher
+	confDir   string
+	netPlugin cni.CNI
+	loadOpts  []cni.CNIOpt
+}
+
+// newCNINetConfSyncer creates cni network conf syncer.
+func newCNINetConfSyncer(confDir string, netPlugin cni.CNI, loadOpts []cni.CNIOpt) (*cniNetConfSyncer, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create fsnotify watcher")
+	}
+
+	if err := os.MkdirAll(confDir, 0700); err != nil {
+		return nil, errors.Wrapf(err, "failed to create cni conf dir=%s for watch", confDir)
+	}
+
+	if err := watcher.Add(confDir); err != nil {
+		return nil, errors.Wrapf(err, "failed to watch cni conf dir %s", confDir)
+	}
+
+	syncer := &cniNetConfSyncer{
+		watcher:   watcher,
+		confDir:   confDir,
+		netPlugin: netPlugin,
+		loadOpts:  loadOpts,
+	}
+
+	if err := syncer.netPlugin.Load(syncer.loadOpts...); err != nil {
+		logrus.WithError(err).Error("failed to load cni during init, please check CRI plugin status before setting up network for pods")
+		syncer.updateLastStatus(err)
+	}
+	return syncer, nil
+}
+
+// syncLoop monitors any fs change events from cni conf dir and tries to reload
+// cni configuration.
+func (syncer *cniNetConfSyncer) syncLoop() error {
+	for {
+		select {
+		case event := <-syncer.watcher.Events:
+			// Only reload config when receiving write/rename/remove
+			// events
+			//
+			// TODO(fuweid): Might only reload target cni config
+			// files to prevent no-ops.
+			if event.Op&(fsnotify.Chmod|fsnotify.Create) > 0 {
+				logrus.Debugf("ignore event from cni conf dir: %s", event)
+				continue
+			}
+			logrus.Debugf("receiving change event from cni conf dir: %s", event)
+
+			lerr := syncer.netPlugin.Load(syncer.loadOpts...)
+			if lerr != nil {
+				logrus.WithError(lerr).
+					Errorf("failed to reload cni configuration after receiving fs change event(%s)", event)
+			}
+			syncer.updateLastStatus(lerr)
+
+		case err := <-syncer.watcher.Errors:
+			if err != nil {
+				logrus.WithError(err).Error("failed to continue sync cni conf change")
+				return err
+			}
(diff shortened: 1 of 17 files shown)


fuweid:
```
ERROR: gcloud crashed (MetadataServerException): HTTP Error 400: Bad Request
If you would like to report this issue, please run the following command:
  gcloud feedback
```

no idea what it is. 😢

fuweid:
Forgot one thing that rlock will wait when the wlock is pending. Looping to reload CNI network conf still can cause that NotReady issue. Need inotify to load conf on demand.

mikebrow:
> ```
> ERROR: gcloud crashed (MetadataServerException): HTTP Error 400: Bad Request
> If you would like to report this issue, please run the following command:
>   gcloud feedback
> ```
> 
> 
> no idea what it is. 😢

@yujuhong @Random-Liu still having problems running infra on gcloud is this a known issue? thoughts?

pkg/server/service.go
```diff
@@ -243,6 +254,22 @@ func (c *criService) register(s *grpc.Server) error {
...
 
+// cniNetworkConfSyncer loads latest CNI network conf.
+func (c *criService) cniNetworkConfSyncer(du time.Duration) {
+	var tick = time.NewTicker(du)
+
+	go func() {
+		defer tick.Stop()
+
+		for {
+			if err := c.netPlugin.Load(c.cniLoadOptions()...); err != nil {
```
mikebrow: suggest using a switch here and have a case for exiting the loop cleanly...

mikebrow: agree we need to load once before waiting on for the duration... so when moving to a switch need that to be taken into consideration :)

pkg/server/status.go
```diff
@@ -41,10 +41,6 @@ func (c *criService) Status(ctx context.Context, r *runtime.StatusRequest) (*run
 		Type:   runtime.NetworkReady,
 		Status: true,
 	}
-	// Load the latest cni configuration to be in sync with the latest network configuration
-	if err := c.netPlugin.Load(c.cniLoadOptions()...); err != nil {
-		log.G(ctx).WithError(err).Errorf("Failed to load cni configuration")
```
mikebrow: suggest we get the status of the last async reload request and if there was an error, output it's failure here as well..

pkg/server/service.go
```diff
@@ -243,6 +254,22 @@ func (c *criService) register(s *grpc.Server) error {
...
+				logrus.WithError(err).Errorf("Failed to load cni configuration")
```
mikebrow: suggest mentioning that this is the async reload

pkg/server/status.go
mikebrow: further since it is async the time since the last successful reload and/or time till the next reload may actually be interesting from a status information perspective. For example, someone sets the timer to a week.

fuweid:
@mikebrow updated. PTAL. Thanks

----
It seems that `cri-containerd-node-e2e` fails because containerd doesn't start. Any idea how to check the log?

mikebrow:
> @mikebrow updated. PTAL. Thanks
> 
> Any idea how to check the log?

Sorry I don't.  @Random-Liu how would one get a containerd log dump, would we have to insert a print of the log at failure?

vendor.conf
```diff
@@ -89,6 +89,7 @@ sigs.k8s.io/yaml                                    9fc95527decd95bb9d28cc2eab08
 github.com/containerd/go-cni                        0d360c50b10b350b6bb23863fd4dfb1c232b01c9
 github.com/containernetworking/cni                  4cfb7b568922a3c79a23e438dc52fe537fc9687e # v0.7.1
 github.com/containernetworking/plugins              9f96827c7cabb03f21d86326000c00f61e181f6a # v0.7.6
+github.com/fsnotify/fsnotify                        4bf2d1fec78374803a39307bfb8d340688f4f28e
```
mikebrow: can we use a versioned fsnotify? I see it's not had a new version tag since 18.. hmm Trying to get to a good place for go mod

pkg/server/service.go
```diff
@@ -233,7 +253,7 @@ func (c *criService) Close() error {
 	if err := c.streamServer.Stop(); err != nil {
 		return errors.Wrap(err, "failed to stop stream server")
 	}
-	return nil
+	return c.cniNetConfMonitor.stop()
```
mikebrow: let's wrap the error with something like  "failed to stop cniNetConfMonitor"

mikebrow: it feels weird to possibly be doing a cni config update after the event monitor has been stopped... wondering if we should be closing the conf monitor first in this Close() func. last in first out :)

fuweid: Sure. Updated. I call the `c.cniNetConfMonitor.stop()` first and log error instead of return error. It is not blocked issue during shutdown period.

pkg/server/cni_conf_syncer.go
```diff
@@ -0,0 +1,116 @@
...
+// syncLoop monitors any fs change events from cni conf dir and tries to reload
+// cni configuration.
+func (syncer *cniNetConfSyncer) syncLoop() error {
+	for {
+		select {
+		case event := <-syncer.watcher.Events:
+			// Only reload config when receiving write/rename/remove
+			// events
+			//
+			// FIXME(fuweid): Might only reload target cni config
```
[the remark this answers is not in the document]
fuweid: Updated

vendor.conf
fuweid: Yeah. The current master `4bf2d1fec78374803a39307bfb8d340688f4f28e` contains the pr https://github.com/fsnotify/fsnotify/pull/219, which used to prevent fd leak from cni call or other exec commands. It is good patch, but the latest tag `v1.4.7` doesn't include this one. 

So I still use `4bf2d1fec78374803a39307bfb8d340688f4f28e` commit here. Is it good choice?

fuweid: ping @cpuguy83

cpuguy83: Yes this is a good commit. The project does not have very much changes.
Drafting a new release for fsnotify now, just waiting on some other maintainers to OK it.

cpuguy83: And 4bf2d1fec78374803a39307bfb8d340688f4f28e will be the tagged commit.

cpuguy83: https://github.com/fsnotify/fsnotify/releases/tag/v1.4.8

fuweid: @cpuguy83 Thank you! will update it

fuweid: updated. but still no idea about node e2e testing

fuweid:
ping @Random-Liu and how to check containerd log in e2e 😂

fuweid:
@mikebrow I check kubernetes with extra-logs and upload logs into artifacts :) finally get it

pkg/server/cni_conf_syncer.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// newCNINetConfSyncer creates cni network conf syncer.
+func newCNINetConfSyncer(confDir string, netPlugin cni.CNI, loadOpts []cni.CNIOpt) (*cniNetConfSyncer, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create fsnotify watcher")
+	}
+
+	if err := os.MkdirAll(confDir, 0700); err != nil {
```
fuweid, line 49: the previous test failed because the conf dir doesn't exist.

fuweid:
ping @mikebrow is it ok to move on?

mikebrow:
@fuweid yes we have a fix for the windows cri test ... it has been moved to github actions!

fuweid:
@mikebrow great. Thanks!

PR #1416, opened 2020-03-09
Use local "ensureRemoveAll" instead of docker/pkg/system

thaJeztah:
Trying to reduce (or remove) the dependency on github.com/docker/docker

depends on:

- [x] https://github.com/moby/moby/pull/40648 pkg/symlink: don't depend on pkg/system and pkg/longpath
- [x] ~https://github.com/containerd/containerd/pull/4096, which implements UnmountRecursive and EnsureRemoveAll, which should provide the same functionality as [`github.com/docker/docker. RecursiveUnmount()`](https://github.com/moby/moby/blob/19008faf03fbf36d9e3e4b7c5769201d3516f613/pkg/mount/mount.go#L131)
and [`github.com/docker/docker/pkg/system.EnsureRemoveAll()`](https://github.com/moby/moby/blob/0211de67ab92cd62ab4484bac28c33f9cbd92b7f/pkg/system/rm.go#L24).~

With this, containerd/cri no longer has to import those packages.

Remaining packages that are used from docker/docker:

- github.com/docker/docker/pkg/symlink
    - perhaps this package should be copied to containerd, or copied here
- github.com/docker/docker/pkg/truncindex
    - perhaps this package should be copied to this repository


vendor.conf
```diff
@@ -1,6 +1,6 @@
 # cri dependencies
 github.com/docker/distribution                      0d3efadf0154c2b8a4e7b6621fff9809655cc580
-github.com/docker/docker                            d1d5f6476656c6aad457e2a91d3436e66b6f2251
+github.com/docker/docker                            standalone_symlink https://github.com/thaJeztah/docker.git
```
thaJeztah: temporarily using my fork until https://github.com/moby/moby/pull/40648 is merged

vendor.conf
```diff
@@ -10,7 +10,7 @@ github.com/BurntSushi/toml                          3012a1dbe2e4bd1391d42b32f057
 github.com/cespare/xxhash/v2                        d7df74196a9e781ede915320c11c378c1b2f3a1f # v2.1.1
 github.com/containerd/cgroups                       7347743e5d1e8500d9f27c8e748e689ed991d92b
 github.com/containerd/console                       8375c3424e4d7b114e8a90a4a40c8e1b40d1d4e6
-github.com/containerd/containerd                    01310155947cb6eec37dcae29742a165e56acb4a
+github.com/containerd/containerd                    ensure_remove_all https://github.com/thaJeztah/containerd.git
```
thaJeztah: temporarily using my fork until https://github.com/containerd/containerd/pull/4096 is merged

thaJeztah:
Hm..

```
	github.com/containerd/cri/cmd/containerd
import cycle not allowed
package github.com/containerd/cri/cmd/containerd
	imports github.com/containerd/containerd/cmd/containerd/command
	imports github.com/containerd/containerd/mount
	imports github.com/containerd/containerd/sys
	imports github.com/containerd/containerd/mount
vendor/github.com/containerd/containerd/cmd/containerd/command/notify_linux.go:25:2: cannot find package "github.com/coreos/go-systemd/v22/daemon" in any of:
	/home/prow/go/src/github.com/containerd/cri/vendor/github.com/coreos/go-systemd/v22/daemon (vendor tree)
	/usr/local/go/src/github.com/coreos/go-systemd/v22/daemon (from $GOROOT)
	/home/prow/go/src/github.com/coreos/go-systemd/v22/daemon (from $GOPATH)
make: *** [Makefile:88: _output/containerd] Error 1
+ EXIT_VALUE=2
+ set +o xtrace
```

pkg/server/helpers_windows.go
```diff
@@ -156,3 +158,62 @@ func fixLongPath(path string) string {
...
+// a directory.
+//
+// Because of the way `os.Remove` (and by extension `os.RemoveAll`) works, there
+// can be a race between reading directory entries and then actually attempting
+// to remove everything in the directory.
+// These types of errors do not need to be returned since it's ok for the dir to
+// be gone we can just retry the remove operation.
+//
+// This should not return a `os.ErrNotExist` kind of error under any circumstances
+func ensureRemoveAll(_ context.Context, dir string) error {
```
thaJeztah: This is the same as the linux variant, with exception of the unmounting. Alternatively we could make windows-specific implementations of unmountRecursive and unmount, and make them a no-op.

pkg/server/helpers_unix.go
```diff
@@ -141,3 +148,97 @@ func (c *criService) seccompEnabled() bool {
...
+// the deepest mount first.
+func unmountRecursive(ctx context.Context, target string) error {
+	mounts, err := mount.Self()
+	if err != nil {
+		return err
+	}
+
+	var toUnmount []string
+	for _, m := range mounts {
+		if strings.HasPrefix(m.Mountpoint, target) {
```
dmcgowan: Does `target` always end in a `/` here? Since this is string comparison I always feel more comfortable doing path cleanup and explicitly checking for equality or cleaned prefix + `/` to ensure it couldn't catch something like `/something` and `/something-variant`.

pkg/server/helpers_unix.go
```diff
@@ -141,3 +148,97 @@ func (c *criService) seccompEnabled() bool {
...
+func ensureRemoveAll(ctx context.Context, dir string) error {
+	notExistErr := make(map[string]bool)
+
+	// track retries
+	exitOnErr := make(map[string]int)
+	maxRetry := 50
+
+	// Attempt to unmount anything beneath this dir first. Ignore errors,
+	// as they will show up below
+	_ = unmountRecursive(ctx, dir)
```
dmcgowan: Debug log the errors? The error that shows up below might not indicate why the unmount failed or that it is related

thaJeztah:
@dmcgowan updated; ptal

pkg/server/helpers_unix.go
```diff
@@ -141,3 +148,102 @@ func (c *criService) seccompEnabled() bool {
...
+		p, err := filepath.Rel(target, m.Mountpoint)
+		if err != nil {
+			return err
+		}
+		if !strings.HasPrefix(p, "..") {
+			toUnmount = append(toUnmount, m.Mountpoint)
+		}
+	}
+
+	sort.Sort(sort.Reverse(sort.StringSlice(toUnmount)))
```
mikebrow: does reverse sort order always guarantee depth first? would be worried about certain allowable special chars having a sort value less than `/`

thaJeztah: Updated to explicitly sort by length

PR #1422, opened 2020-03-25
Bump golang to v1.13.9 and v1.14.1

chenrui333:


chenrui333:
oh yes, doing that now.

PR #1424, opened 2020-03-27
Start docker unconditionally

fvoznika:
`systemctl is-enabled` is not a reliable indicator that the docker service should be started. Docker service may be running even if marked as disabled.


Random-Liu:
You need to sign your commit

PR #1425, opened 2020-03-27
use the same dependency version for x/sys from containerd/containerd

dims:
The versions of x/sys drifted between the repos:
- https://github.com/containerd/cri/blob/master/vendor.conf#L53
- https://github.com/containerd/containerd/blob/master/vendor.conf#L49

Picking the one that is newer:
https://github.com/golang/sys/compare/52ab431487773bc9dd1b0766228b1cf3944126bf...5c8b2ff67527cb88b770f693cebf3799036d8bc0


dims:
Ack @mikebrow i'll try to remember that for next time.

PR #1429, opened 2020-04-06
Create etcd user in cloud init.

Random-Liu:
Copied the change from https://github.com/kubernetes/kubernetes/pull/88856.

All Kubernetes containerd e2e tests are failing because of this.

See more details at https://github.com/kubernetes/kubernetes/issues/89847

@cheftako

### DIFF
--- a/cluster/gce/cloud-init/master.yaml
+++ b/cluster/gce/cloud-init/master.yaml
@@ -1,5 +1,11 @@
 #cloud-config
 
+users:
+- name: etcd
+  homedir: /var/etcd
+  lock_passwd: true
+  ssh_redirect_user: true
+
 write_files:
 # Setup containerd.
   - path: /etc/systemd/system/containerd-installation.service


mikebrow:
> /lgtm
> 
> lgtm to unblock. Not sure we should have k8s/k8s config in containerd/cri.

agree to unblock please feel free to open issue... I agree we should review given the recent positive changes wrt k8s testing of container runtimes other than docker.

Random-Liu:
> lgtm to unblock. Not sure we should have k8s/k8s config in containerd/cri.

Since there is already containerd service in the new COS. We should be able to rely on that and only replace the binaries.

mikebrow:
@Random-Liu compute instance looks like it's down "(gcloud.compute.ssh) Could not SSH into the instance."

Random-Liu:
I believe the windows test is also affected by https://github.com/kubernetes/kubernetes/issues/89847#issuecomment-610109930

Sigh.

dims:
LGTM

mikebrow:
@Random-Liu pls rebase.. pull-cri-containerd-windows-cri test has been moved to git hub actions .

mikebrow:
had to close and re-open to get the appveyor test started again

PR #1440, opened 2020-04-14
Containerd-Cri with supports multi-architectures

jr-santos98:
While the containerd build a file getting information about the architecture and the operating system it is running on. The cri assumes that you are using a Linux computer with x86. This commit makes the CRI adapt to the system and architecture of the computer on which it is being used.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -42,6 +42,8 @@ OFFICIAL_RELEASE=${OFFICIAL_RELEASE:-false}
 # LOCAL_RELEASE indicates that containerd has been built and released
 # locally.
 LOCAL_RELEASE=${LOCAL_RELEASE:-false}
+GOOS=$(go env GOOS)
+GOARCH=$(go env GOARCH)
 
 
 destdir=${BUILD_DIR}/release-stage
@@ -66,7 +68,7 @@ download_containerd() {
 
 # copy_local_containerd copies local containerd release.
 copy_local_containerd() {
-  local -r tarball="${GOPATH}/src/github.com/containerd/containerd/releases/containerd-${VERSION}.linux-amd64.tar.gz"
+  local -r tarball="${GOPATH}/src/github.com/containerd/containerd/releases/containerd-${VERSION}.${GOOS}-${GOARCH}.tar.gz"
   if [[ ! -e "${tarball}" ]]; then
     echo "Containerd release is not built"
     exit 1
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -42,8 +42,14 @@ OFFICIAL_RELEASE=${OFFICIAL_RELEASE:-false}
 # LOCAL_RELEASE indicates that containerd has been built and released
 # locally.
 LOCAL_RELEASE=${LOCAL_RELEASE:-false}
-GOOS=$(go env GOOS)
-GOARCH=$(go env GOARCH)
+if [ -z "${GOOS:-}" ]
+then
+    GOOS=$(go env GOOS)
+fi
+if [ -z "${GOARCH:-}" ]
+then
+    GOARCH=$(go env GOARCH)
+fi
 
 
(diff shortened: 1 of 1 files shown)


mikebrow:
please squash and sign your commits

jr-santos98:
@mikebrow How do I do this? Failure in the file would be for commit validation.

mikebrow:
normally when modifying a prior commit to make more changes.. I would:
git add filename
git commit -s --amend

To squash if you didn't amend the second commit you can:
git rebase -i HEAD~2
you want the top commit to say pick and just change the second one to squash...
save and exit
git push -f branchname

Cheers, Mike

mikebrow:
to configure your git for signing with the -s option:
$ git config --global user.name "full name here"
$ git config --global user.email "email address here"

AkihiroSuda:
ping @jr-santos98

jr-santos98:
@mikebrow What's the problem now ?? Me running the command locally. It is informed that everything is ok.
The [link](http://www.students.ic.unicamp.br/~ra195727/log.txt) shows the output for the `make .gitvalidation` command that I obtained locally.

mikebrow:
> @mikebrow What's the problem now ?? Me running the command locally. It is informed that everything is ok.
> The [link](http://www.students.ic.unicamp.br/~ra195727/log.txt) shows the output for the `make .gitvalidation` command that I obtained locally.

nothing.. the github action host encountered a problem pulling the files in the pull request threw a 128 network error  then github "checks" code got in a bad state .. 

so I closed/reopened the PR to force the state reset (as github refused to restart the failed action due to an internal error)

sigh.

hack/release.sh
```diff
@@ -42,6 +42,8 @@ OFFICIAL_RELEASE=${OFFICIAL_RELEASE:-false}
 # LOCAL_RELEASE indicates that containerd has been built and released
 # locally.
 LOCAL_RELEASE=${LOCAL_RELEASE:-false}
+GOOS=$(go env GOOS)
```
mikebrow, line 45: We may get these from elsewhere so let's only set these if not already set..
for example: `GOOS=$(GOOS:-go env GOOS)`

jr-santos98: I think it's a good idea

mikebrow: bash is so much fun sometimes
```
if [ -z "${GOOS}" ]
then
    GOOS=$(go env GOOS)
fi
```

jr-santos98:
@mikebrow Interesting idea, what do you think of the change I made?

jr-santos98:
@mikebrow I was wrong ... But, I tested it, the condition `if [ -z "${GOOS:-}" ] then  GOOS=$(go env GOOS) fi`.
It will now work and pass the tests.

mikebrow:
```
cat run.sh

#!/bin/bash
if [ -z "${GOOS}" ]
then
    GOOS=$(go env GOOS)
fi
echo $GOOS

$ GOOS=foo ./run.sh
foo

$ ./run.sh
linux
```
what failed?

mikebrow:
the kubernetes e2e stuff is just flakey ATM ... nothing to do with your script..

jr-santos98:
The error "unbound variable" is displayed in some of the CI points.
As can be seen in this [code](https://repl.it/@Yukine_/condition#main.sh).
This error happens while running my build as well. Even without that dependency.
In this format:"${GOOS:-}" ... or: "${GOOS}", there is no error in the tests, and I was able to run locally as well.

mikebrow:
> The error "unbound variable" is displayed in some of the CI points.
> As can be seen in this [code](https://repl.it/@Yukine_/condition#main.sh).
> This error happens while running my build as well. Even without that dependency.
> In this format:"${GOOS:-}" ... or: "${GOOS}", there is no error in the tests, and I was able to run locally as well.

Interesting.. works fine on ubuntu :-) Thx!

Issue #1421, opened 2020-03-20
can not pull image from http or insecure https in same registry domain


yylt:
Hi, there is a question

docker pull will try from http and https protocol, but cri plugin will not, and i write configure file will not work

```
    [plugins."io.containerd.grpc.v1.cri".registry]
      [plugins."io.containerd.grpc.v1.cri".registry.configs]
        [plugins."io.containerd.grpc.v1.cri".registry.configs."hub.xx.io"]
          [plugins."io.containerd.grpc.v1.cri".registry.configs."hub.xx.io".tls]
            insecure_skip_verify = true
      [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
        [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
          endpoint = ["https://registry-1.docker.io"]
        [plugins."io.containerd.grpc.v1.cri".registry.mirrors."hub.xx.io"]
          endpoint = ["https://hub.xx.io","http://hub.xx.io"]
    [plugins."io.containerd.grpc.v1.cri".x509_key_pair_streaming]
```

Could you  support http and https from same registry, and pull from registry by http protocol will not check tls  even if tls config had set.

fuweid:
Hi, does option `insecure_skip_verify` help?

```
[plugins.cri.registry.configs."my.custom.registry".tls]
  insecure_skip_verify = true
```

https://github.com/containerd/cri/pull/1346/commits

yylt:
`insecure_skip_verify` can only help skip tls check, and cri will ont use http.

In my case, `docker pull` work good, but when migrate to containerd(v1.3.3) runtime, `crictl pull` can not work

https://github.com/containerd/cri/blob/5fb103d8e66f0c6292971b00f05b2aed4c99f902/pkg/server/image_pull.go#L322

In this line, it will failed if there are tls config when use http protocol

fuweid:
Sorry @yylt for the late reply. If you are using http, why add tls for toml?

mikebrow:
note: moving to what will "hopefully" be more docker like config for containerd & containerd/cri config ->  https://github.com/containerd/containerd/pull/4138

yylt:
I just want it  like docker to pull image, and I don't understand why to judge it is incorrect to use http scheme and certificate configure at one time

@mikebrow the PR more like want to use self-signed certificate

mikebrow:
[...]

PR #1449, opened 2020-04-16
removes the error when tls is configured for https but http is tried first

mikebrow:
Address issue #1421

@Random-Liu


dims:
LGTM

PR #1455, opened 2020-04-21
fix incomplete host device for PrivilegedWithoutHostDevices

thbtcllt:
For a privilege pods with PrivilegedWithoutHostDevices set to true
host device specified in the config are not provided (whereas it is done for
non privilege pods or privilege pods with PrivilegedWithoutHostDevices set
to false as all devices are included).

Add them in this case.

Fixes: 3353ab76d922 ("Add flag to overload default privileged host device behaviour")
Signed-off-by: Thibaut Collet <thibaut.collet@6wind.com>

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -172,6 +172,9 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		specOpts = append(specOpts, oci.WithPrivileged)
 		if !ociRuntime.PrivilegedWithoutHostDevices {
 			specOpts = append(specOpts, oci.WithHostDevices, oci.WithAllDevicesAllowed)
+		} else {
+			// add requested devices by the config as host devices are not automatically added
+			specOpts = append(specOpts, customopts.WithDevices(c.os, config), customopts.WithCapabilities(securityContext))
 		}
 	} else { // not privileged
 		specOpts = append(specOpts, customopts.WithDevices(c.os, config), customopts.WithCapabilities(securityContext))


thbtcllt:
This change is linked to https://github.com/kata-containers/documentation/issues/637

pkg/server/container_create_unix.go
```diff
@@ -172,6 +172,9 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		specOpts = append(specOpts, oci.WithPrivileged)
 		if !ociRuntime.PrivilegedWithoutHostDevices {
 			specOpts = append(specOpts, oci.WithHostDevices, oci.WithAllDevicesAllowed)
+		} else {
+			// add requested devices by the config as host devices are not automatically added
+			specOpts = append(specOpts, customopts.WithDevices(c.os, config), customopts.WithCapabilities(securityContext))
```
mikebrow, line 177: I believe we are doing what is expected here by allowing a user to configure CRI to not pass host devices to privileged containers. The default behavior is to pass them through. Thoughts?
```
      # privileged_without_host_devices allows overloading the default behaviour of passing host
      # devices through to privileged containers. This is useful when using a runtime where it does
      # not make sense to pass host devices to the container when privileged. Defaults to false -
      # i.e pass host devices through to privileged containers.
```
will run the bucket to show the test result hopefully we catch the change as an error :-)

mikebrow: Without means not.. IOW the default for the flag is false... which negates the without.. causing it to pass the host devices through.  Set it to true and it withholds (without) passing the host devices through.

mikebrow:
my bad on the first review... can you rebase please? I removed greetings from the test checklist.

PR #1456, opened 2020-04-21
remove broken greetings github action

mikebrow:
Resolving greetings test failure here:
https://github.com/containerd/cri/pull/1455

See https://github.com/actions/first-interaction/issues/10
removing for now might bring it back in at some point but isn't useable as is.


dmcgowan:
LGTM

PR #1462, opened 2020-04-25
introduce managed namespaces

dweomer:
CRI clients are containerd clients. containerd supports multi-tenancy
via containerd namespaces. Clients interact with resources in non-default
namespaces by specifying a namespace name in the client invocation that
is curried to the server via gRPC. This means that the basic building
blocks for a client-directed (aka honor-system) multi-tenancy are ready
to be leveraged.

Introducing "managed namespaces" for containerd/cri. This change-set is
largely concerned with allowing for multi-tenant CRI to coexist and
enabling interactions with CRI resources in non-default (aka not
"k8s.io") namespaces by NOT overriding the client-provided namespace.
This means that the default CRI namespace, "k8s.io", is only enforced
for clients that have not specified a namespace. A CRI namespace is one
that is labeled with "io.cri-containerd=managed". Invoking the CRI with
a namespace that either does not exist or isn't properly labeled will
return an "unmanaged namespace" error.

How was this implemented?
    
- the plugin initialization aspect of `criService` and the check/logging of `instrumentedService` have been merged into `criServiceManager`
- the `criServiceManager` maintains an event exchange that is subscribed to containerd events (filtering on task and image events, as before)
- the `criServiceManager` manages `criService` instances in a sync.Map keyed by namespace name
- all `criService` instances are subscribed to the `criServiceManager` event exchange filtering on their namespace name
- all `criService` instances are currently created and started on NamespaceCreate/NamespaceUpdate having the label "io.cri-containerd == managed"
  - managed `criService` instances are stopped on NamespaceDelete or NamespaceUpdate when "io.cri-containerd != managed"
- the `DeferContext` and `NamespacedContext` in ./pkg/containerd/util have been changed to accept a namespace name
- the default CRI namespace, k8s.io, is configured, as before, by the CRI plugin TOML
- all other namespaces are configured by TOML that is the standard, aka "version 1", emitted by serializing a `config.PluginConfig`
  - located at `${containerd.root_dir}/io.containerd.grpc.v1.cri/namespaces/${namespace.name}/config.toml`
  - this config is serialized for all namespaces, including k8s.io, at the same (parameterized) location
  - non-default namespaces lacking a config file will be started with a "default" having:
    - `cni.conf_dir = /opt/cri/${namespace.name}/etc/cni/net.d` instead of the usual default `/etc/cni/net.d`

/cc @mikebrow @Random-Liu @ibuildthecloud

Signed-off-by: Jacob Blain Christen <jacob@rancher.com>

### DIFF
--- a/cri.go
+++ b/cri.go
@@ -30,6 +30,7 @@ import (
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/leases"
 	"github.com/containerd/containerd/log"
+	ctdnamespaces "github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/containerd/plugin"
 	"github.com/containerd/containerd/services"
@@ -63,7 +64,7 @@ func init() {
 func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	ic.Meta.Platforms = []imagespec.Platform{platforms.DefaultSpec()}
 	ic.Meta.Exports = map[string]string{"CRIVersion": constants.CRIVersion}
-	ctx := ic.Context
+	ctx := ctdnamespaces.WithNamespace(ic.Context, "cri.containerd.io")
 	pluginConfig := ic.Config.(*criconfig.PluginConfig)
 	if err := criconfig.ValidatePluginConfig(ctx, pluginConfig); err != nil {
 		return nil, errors.Wrap(err, "invalid plugin config")
@@ -104,7 +105,7 @@ func initCRIService(ic *plugin.InitContext) (interface{}, error) {
 	}
 
 	go func() {
-		if err := s.Run(); err != nil {
+		if err := s.Run(ctx); err != nil {
 			log.G(ctx).WithError(err).Fatal("Failed to run CRI service")
 		}
 		// TODO(random-liu): Whether and how we can stop containerd.
--- a/integration/namespace_lifecycle_test.go
+++ b/integration/namespace_lifecycle_test.go
@@ -1,0 +1,86 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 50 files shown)


AkihiroSuda:
Is this for non-kubelet clients? Especially k3c?

dweomer:
> Is this for non-kubelet clients? Especially k3c?

@AkihiroSuda yes, I need this specifically for k3c as it would enable running the `k3c daemon` (repackaged containerd) underneath k3s. As things stand currently, `k3c` and `k3s` sharing the same container runtime namespace results in each daemon reaping sandboxes that the other owns. My choices are to fork the kubelet or enhance the underlying container runtime to allow for a permeable multi-tenancy (which opens up a number of features we want for k3c).

dweomer:
@mikebrow I think I have the configuration changes worked out locally but I am also thinking of changing up this PR a bit to limit the number of files changes. It will mean that the new `criNamespace` type reverts back to `criService` and that the old `criService` type will be replaced with a `criPlugin` (or `criDispatch`) type. This will limit the number of files touched and this make it easier to merge without conflict. It will also re-instate the CRIAPI interface contracts being implemented by the containerd/cri service returned via the introspection service. Thoughts?

dweomer:
> @mikebrow I think I have the configuration changes worked out locally but I am also thinking of changing up this PR a bit to limit the number of files changes. It will mean that the new `criNamespace` type reverts back to `criService` and that the old `criService` type will be replaced with a `criPlugin` (or `criDispatch`) type. This will limit the number of files touched and thus make it easier to merge without conflict re: other outstanding changes. It will also re-instate the CRIAPI interface contracts being implemented by the containerd/cri service returned via the introspection service. Thoughts?

I went ahead and refactored. Down to 48 changed files from the original 69, nice.

AkihiroSuda:
needs rebase

dweomer:
This ended up being a bigger can of worms than I wanted to open. While this change-set worked and I wrote it to be as maintainable as possible (assuming I had to maintain a fork) it simply touched on too many areas in this codebase. I've moved on to an easier-to-maintain gRPC header trick to hide CRI containers from the kubelet by default but expose them to savvy clients such as k3c.

mikebrow:
ok.. will keep the proposal in mind then on future changes.. and maybe take a peek at your header trick. I like to keep abreast of the components sitting between containerd/cri and kubelet.. crictl for example.

PR #1466, opened 2020-04-27
Pass chained layer digests to snapshotter for parallel snapshot preparation

ktock:
Currently, CRI plugin passes each layer digest to remote snapshotters sequentially and snapshotters prepares snapshots sequentially. But it costs extra time especially for remote snapshotters which need to connect to the remote backend store (e.g. registries) for checking the snapshot existence on each preparation.

This commit solves this problem by introducing new label `containerd.io/snapshot/cri.chain` for passing all layer digests in an image to snapshotters and by allowing them to prepare these snapshots in parallel, which leads to speed up the preparation.

This label is supported by stargz snapshotter: https://github.com/containerd/stargz-snapshotter/pull/83

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -455,6 +455,10 @@ const (
 	// targetDigestLabel is a label which contains layer digest and will be passed
 	// to snapshotters.
 	targetDigestLabel = "containerd.io/snapshot/cri.layer-digest"
+	// targetImageLayersLabel is a label which contains layer digests contained in
+	// the target image and will be passed to snapshotters for preparing layers in
+	// parallel.
+	targetImageLayersLabel = "containerd.io/snapshot/cri.image-layers"
 )
 
 // appendInfoHandlerWrapper makes a handler which appends some basic information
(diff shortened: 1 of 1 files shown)


pkg/server/image_pull.go
```diff
@@ -455,6 +455,9 @@ const (
 	// targetDigestLabel is a label which contains layer digest and will be passed
 	// to snapshotters.
 	targetDigestLabel = "containerd.io/snapshot/cri.layer-digest"
+	// targetChainLabel is a label which contains layer digests contained in the target
+	// image and will be passed to snapshotters for preparing layers in parallel.
+	targetChainLabel = "containerd.io/snapshot/cri.chain"
```
AkihiroSuda, line 460: I'd expect this terminology to be reserved for something related to OCI ChainID: https://github.com/opencontainers/image-spec/blob/e562b04403929d582d449ae5386ff79dd7961a11/config.md#layer-chainid

ktock: Thanks for the feedback. Should we use another word here like `Stack` or `relatedLayers`?

ktock: Fixed it in the new commit.

AkihiroSuda: Curious, can we just make it compatible with OCI ChainID?

AkihiroSuda: If not the reason and the differences should be noted as code comment lines.

ktock: If my understanding is correct...,
Snapshotters requires the list of layer digests in the target image but ChainID is a recursively digested list of layers, with which we cannot restore the original list of layer digest.
So just passing ChainID to snapshotter isn't enough?

ktock: Or using " "(white space) as a delimiter instead of "," makes the layer digest list compatible to OCI-defined `chain`?

ktock:
@AkihiroSuda made chain compatible to OCI on 7585176060452f086912a631102030bbd8b2b5fc, could you review it? Or should we revert to ca661c8 (your approved commit)?

AkihiroSuda:
revert

PR #1467, opened 2020-04-29
Update to latest hcsshim - 0.8.9

dims:
Switch to ad tagged version of hcsshim. Context here:
https://github.com/kubernetes/kubernetes/issues/87420#issuecomment-620837610


thaJeztah:
does that mean tags should be used again? (there was some confusion in https://github.com/microsoft/hcsshim/pull/700, also what stability/compatibility guarantees the tags provided)

dims:
@thaJeztah yep, i believe so. the main problem was the recursive dependency which we worked through in https://github.com/microsoft/hcsshim/pull/789 to split the cri-api dependencies into a separate go.mod file in the test/ directory

thaJeztah:
Ah, thanks!

oh, wow, it's vendoring itself (we were actually discussing that recently what would happen if you use vendoring in combination with a local path replace)

thaJeztah:
> This dependency is one of the containerd/containerd dependencies. 

Good catch; yes, so it should be updated in containerd/containerd first, and then bumped here together with containerd/containerd (and other dependencies in containerd that were updated)

dims:
closing this. will fix containerd/containerd first

PR #1468, opened 2020-05-01
RunPodSandbox: destroy network if fails or invalid

fuweid:
Should destroy the pod network if fails to setup or return invalid
net interface, especially multiple CNI configurations.


mikebrow:
@fuweid good catch

PR #1469, opened 2020-05-04
Remove dependency on libcontainer/system

thaJeztah:
Follow-up to / depends on;

- [x] https://github.com/containerd/containerd/pull/4227 sys/proc: remove GetSystemCPUUsage as it is unused
- [x] https://github.com/containerd/containerd/pull/4230 Remove remaining uses of libcontainer/system package
- [x] https://github.com/containerd/cri/pull/1473 vendor: update containerd and dependencies to ed261720c86d1e700cd5d39175128322baac6dd


pkg/server/service_unix.go
```diff
@@ -72,3 +72,21 @@ func (c *criService) initPlatform() error {
 func (c *criService) cniLoadOptions() []cni.CNIOpt {
 	return []cni.CNIOpt{cni.WithLoNetwork, cni.WithDefaultConf}
 }
+
+// runningInUserNS detects whether we are currently running in a user namespace.
+// Originally copied from github.com/lxc/lxd/shared/util.go
+func runningInUserNS() bool {
+	uidmap, err := user.CurrentProcessUIDMap()
```
thaJeztah, line 79: Actually, given that (within scope of CRI itself) this is the only use of libcontainer/user; I'm wondering if we should simplify this to the original implementation in https://github.com/lxc/lxd/blob/master/shared/util.go#L690-L710

@AkihiroSuda wdyt?

pkg/server/service_unix.go
```diff
@@ -34,7 +37,7 @@ const networkAttachCount = 2
 func (c *criService) initPlatform() error {
 	var err error
 
-	if runcsystem.RunningInUserNS() {
```
AkihiroSuda, line 37: I think we can safely keep this

AkihiroSuda: Ah, I saw this is being moved into https://github.com/containerd/containerd/pull/4230 . That's fine to me as well.

mikebrow:
nice... though I had to look up that 0 0 4294967295 means :-)

mikebrow:
nice... though I had to look up 0 0 4294967295 in http://man7.org/linux/man-pages/man7/user_namespaces.7.html :-)

thaJeztah:
@mikebrow do you prefer a `const` for it, or a comment to be added?

mikebrow:
> @mikebrow do you prefer a `const` for it, or a comment to be added?

Since it's a copy.. I'd leave it as is. Maybe cite to the linx man page ^ 
```
The initial user namespace has no parent namespace, but, for
       consistency, the kernel provides dummy user and group ID mapping
       files for this namespace.  Looking at the uid_map file (gid_map is
       the same) from a shell in the initial namespace shows:

           $ cat /proc/$$/uid_map
                    0          0 4294967295

       This mapping tells us that the range starting at user ID 0 in this
       namespace maps to a range starting at 0 in the (nonexistent) parent
       namespace, and the length of the range is the largest 32-bit unsigned
       integer.  This leaves 4294967295 (the 32-bit signed -1 value)
       unmapped.  This is deliberate: (uid_t) -1 is used in several inter‐
       faces (e.g., setreuid(2)) as a way to specify "no user ID".  Leaving
       (uid_t) -1 unmapped and unusable guarantees that there will be no
       confusion when using these interfaces.
```

thaJeztah:
Rebased, and removed it out of draft now that the other changes are in; updated the comment to include the description from http://man7.org/linux/man-pages/man7/user_namespaces.7.html (as I think you suggested)


@AkihiroSuda @mikebrow PTAL

thaJeztah:
I wasn't sure, and thought: let's save the reader from having to open a browser 🤷‍♂️

pkg/server/service_unix.go
```diff
@@ -67,3 +70,51 @@ func (c *criService) initPlatform() error {
 func (c *criService) cniLoadOptions() []cni.CNIOpt {
 	return []cni.CNIOpt{cni.WithLoNetwork, cni.WithDefaultConf}
 }
+
+// runningInUserNS detects whether we are currently running in a user namespace.
+// Originally copied from github.com/lxc/lxd/shared/util.go
+func runningInUserNS() bool {
```
AkihiroSuda, line 76: containerd/sys.RunningInUserNS

thaJeztah: There was some discussion in an earlier PR about CRI not depending too much on containerd (see https://github.com/containerd/containerd/pull/4096#issuecomment-596778362), which is why I added it here; let me push a commit to see what it'd look like if we did that (so that we can discuss)

thaJeztah:
@AkihiroSuda ptal if that's what you meant, and if it's ok to depend on that package (I'll squash commits it if looks good)

pkg/server/service_unix.go
AkihiroSuda: archive and diff use that function

thaJeztah:
Let me squash commits to clean up

thaJeztah:
Squashed; should be ok now

thaJeztah:
Thanks!

PR #1470, opened 2020-05-06
Get rid of socat for port forwarding

aojea:
use goroutines to copy the data from the stream to the TCP
connection, and viceversa, removing the socat dependency.

Quoting Lantao Liu, the logic is as follow:

When one side (either pod side or user side) of portforward
is closed, we should stop port forwarding.

When one side is closed, the io.Copy use that side as source will close,
but the io.Copy use that side as dest won't.

Fixes: https://github.com/containerd/cri/issues/730

### DIFF
--- a/pkg/server/sandbox_portforward_unix.go
+++ b/pkg/server/sandbox_portforward_unix.go
@@ -19,28 +19,27 @@
 package server
 
 import (
-	"bytes"
 	"fmt"
 	"io"
-	"os/exec"
-	"strings"
+	"net"
+	"time"
 
 	"github.com/containerd/containerd/log"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
+
 	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
-// portForward requires `socat` on the node. It uses netns to enter the sandbox namespace,
-// and run `socat` inside the namespace to forward stream for a specific port. The `socat`
-// command keeps running until it exits or client disconnect.
-func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriter) error {
+// portForward uses netns to enter the sandbox namespace, and forwards a stream inside the
+// the namespace to a specific port. It keeps forwarding until it exits or client disconnect.
+func (c *criService) portForward(ctx context.Context, id string, port int32, stream io.ReadWriteCloser) error {
 	s, err := c.sandboxStore.Get(id)
 	if err != nil {
 		return errors.Wrapf(err, "failed to find sandbox %q in store", id)
 	}
+
 	var netNSDo func(func(ns.NetNS) error) error
 	// netNSPath is the network namespace path for logging.
 	var netNSPath string
@@ -62,48 +61,64 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
 		netNSPath = "host"
 	}
 
-	socat, err := exec.LookPath("socat")
-	if err != nil {
-		return errors.Wrap(err, "failed to find socat")
-	}
-
-	// Check https://linux.die.net/man/1/socat for meaning of the options.
-	args := []string{socat, "-", fmt.Sprintf("TCP4:localhost:%d", port)}
-
-	log.G(ctx).Infof("Executing port forwarding command %q in network namespace %q", strings.Join(args, " "), netNSPath)
+	log.G(ctx).Infof("Executing port forwarding in network namespace %q", netNSPath)
 	err = netNSDo(func(_ ns.NetNS) error {
-		cmd := exec.Command(args[0], args[1:]...)
-		cmd.Stdout = stream
-
-		stderr := new(bytes.Buffer)
-		cmd.Stderr = stderr
-
-		// If we use Stdin, command.Run() won't return until the goroutine that's copying
-		// from stream finishes. Unfortunately, if you have a client like telnet connected
-		// via port forwarding, as long as the user's telnet client is connected to the user's
-		// local listener that port forwarding sets up, the telnet session never exits. This
-		// means that even if socat has finished running, command.Run() won't ever return
-		// (because the client still has the connection and stream open).
-		//
-		// The work around is to use StdinPipe(), as Wait() (called by Run()) closes the pipe
-		// when the command (socat) exits.
-		in, err := cmd.StdinPipe()
+		defer stream.Close()
+		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default if the system has IPv6 enabled.
+		// Theoretically happy eyeballs will try IPv6 first and fallback to IPv4
+		// but resolving localhost doesn't seem to return and IPv4 address, thus failing the connection.
+		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
 		if err != nil {
-			return errors.Wrap(err, "failed to create stdin pipe")
+			return errors.Wrapf(err, "failed to dial %d", port)
 		}
+		defer conn.Close()
+
+		errCh := make(chan error, 2)
+		// Copy from the the namespace port connection to the client stream
 		go func() {
-			if _, err := io.Copy(in, stream); err != nil {
-				logrus.WithError(err).Errorf("Failed to copy port forward input for %q port %d", id, port)
-			}
-			in.Close()
-			logrus.Debugf("Finish copying port forward input for %q port %d", id, port)
+			log.G(ctx).Debugf("PortForward copying data from namespace %q port %d to the client stream", id, port)
+			_, err := io.Copy(stream, conn)
+			errCh <- err
 		}()
 
-		if err := cmd.Run(); err != nil {
-			return errors.Errorf("socat command returns error: %v, stderr: %q", err, stderr.String())
+		// Copy from the client stream to the namespace port connection
+		go func() {
+			log.G(ctx).Debugf("PortForward copying data from client stream to namespace %q port %d", id, port)
+			_, err := io.Copy(conn, stream)
+			errCh <- err
+		}()
+
+		// Wait until the first error is returned by one of the connections
+		// we use errFwd to store the result of the port forwarding operation
+		// if the context is cancelled close everything and return
+		var errFwd error
+		select {
+		case errFwd = <-errCh:
+			log.G(ctx).Debugf("PortForward stop forwarding in one direction in network namespace %q port %d: %v", id, port, errFwd)
+		case <-ctx.Done():
+			log.G(ctx).Debugf("PortForward cancelled in network namespace %q port %d: %v", id, port, ctx.Err())
+			return ctx.Err()
 		}
-		return nil
+		// give a chance to terminate gracefully or timeout
+		// 0.5s is the default timeout used in socat
+		// https://linux.die.net/man/1/socat
+		timeout := time.Duration(500) * time.Millisecond
+		select {
+		case e := <-errCh:
+			if errFwd == nil {
+				errFwd = e
+			}
+			log.G(ctx).Debugf("PortForward stopped forwarding in both directions in network namespace %q port %d: %v", id, port, e)
+		case <-time.After(timeout):
+			log.G(ctx).Debugf("PortForward timed out waiting to close the connection in network namespace %q port %d", id, port)
+		case <-ctx.Done():
+			log.G(ctx).Debugf("PortForward cancelled in network namespace %q port %d: %v", id, port, ctx.Err())
+			errFwd = ctx.Err()
+		}
+
+		return errFwd
 	})
+
 	if err != nil {
 		return errors.Wrapf(err, "failed to execute portforward in network namespace %q", netNSPath)
 	}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,8 +107,7 @@ jobs:
           sudo apt-get install -y \
             btrfs-tools \
             libseccomp2 \
-            libseccomp-dev \
-            socat
+            libseccomp-dev
           make install.deps
         working-directory: ${{github.workspace}}/src/github.com/containerd/cri
 
--- a/README.md
+++ b/README.md
@@ -77,11 +77,10 @@ specifications as appropriate.
 (Fedora, CentOS, RHEL). On releases of Ubuntu <=Trusty and Debian <=jessie a
 backport version of `libseccomp-dev` is required. See [travis.yml](.travis.yml) for an example on trusty.
 * **btrfs development library.** Required by containerd btrfs support. `btrfs-tools`(Ubuntu, Debian) / `btrfs-progs-devel`(Fedora, CentOS, RHEL)
-2. Install **`socat`** (required by portforward).
-3. Install **`pkg-config`** (required for linking with `libseccomp`).
-4. Install and setup a Go 1.13.10 development environment.
-5. Make a local clone of this repository.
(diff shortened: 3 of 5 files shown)


aojea:
xref https://github.com/kubernetes-sigs/kind/issues/1560

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,52 +61,60 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
-		// The work around is to use StdinPipe(), as Wait() (called by Run()) closes the pipe
-		// when the command (socat) exits.
-		in, err := cmd.StdinPipe()
+		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default if the system has IPv6 enabled.
+		// Theoretically happy eyeballs will try IPv6 first and fallback to IPv4
+		// but resolving localhost doesn't seem to return and IPv4 address, thus failing the connection.
+		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
 		if err != nil {
-			return errors.Wrap(err, "failed to create stdin pipe")
+			return errors.Wrapf(err, "failed to dial %q", port)
```
saschagrunert, line 71:
```suggestion
			return errors.Wrapf(err, "failed to dial :%d", port)
```

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,52 +61,60 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+			_, err := io.Copy(stream, conn)
+			errCh <- err
+		}()
+
+		// Wait until one direction exits and store the result
+		// with the reason why the connection is being closed
+		errFwd := <-errCh
+
+		// give a chance to terminate gracefully or timeout
+		timeout := time.Duration(1) * time.Second
```
saschagrunert, line 98: I'm wondering why we need this. :thinking:

aojea: it mimics socat https://linux.die.net/man/1/socat

> When one of the streams effectively reaches EOF, the closing phase begins. Socat transfers the EOF condition to the other stream, i.e. tries to shutdown only its write stream, giving it a chance to terminate gracefully. For a defined time socat continues to transfer data in the other direction, but then closes all remaining channels and terminates.

I make up the timeout number though, 1 sec sounds a good compromise, it should be enough time so the application that is using the channel can close gracefully

BenTheElder:
cc @dims @mikebrow

dims:
in general looks good to me, i'd like to see if we can raise @Random-Liu to take a look.

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,56 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+
+		errCh := make(chan error, 2)
+		// Copy from the client stream to the namespace port connection
 		go func() {
-			if _, err := io.Copy(in, stream); err != nil {
-				logrus.WithError(err).Errorf("Failed to copy port forward input for %q port %d", id, port)
-			}
-			in.Close()
-			logrus.Debugf("Finish copying port forward input for %q port %d", id, port)
+			// tries to shutdown only its write stream, giving it a chance to terminate gracefully
```
mikebrow, line 77: s/write/destination/

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,56 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+			defer conn.Close()
+			log.G(ctx).Debugf("PortForward copying data from client stream to namespace %q port %d", id, port)
+			_, err := io.Copy(conn, stream)
+			errCh <- err
 		}()
 
-		if err := cmd.Run(); err != nil {
-			return errors.Errorf("socat command returns error: %v, stderr: %q", err, stderr.String())
+		// Copy from the the namespace port connection to the stream
```
mikebrow, line 84: s/stream/client stream/

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,56 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+		go func() {
```
mikebrow, line 85: nit changes opening sequence from outbound first to inbound first..

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,56 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+
+		// give a chance to terminate gracefully or timeout
+		timeout := time.Duration(1) * time.Second
+		select {
+		case e := <-errCh:
+			if e != nil {
+				// This error may be caused because we are shutting down the connection, only log it
+				log.G(ctx).Debugf("PortForward closed the connection with error %v in network namespace %q", e, id)
+			} else {
+				log.G(ctx).Debugf("PortForward closed the connection gracefully in network namespace %q", id)
```
mikebrow, line 105: pls add port to these as well..

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,56 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+		errCh := make(chan error, 2)
+		// Copy from the client stream to the namespace port connection
 		go func() {
-			if _, err := io.Copy(in, stream); err != nil {
-				logrus.WithError(err).Errorf("Failed to copy port forward input for %q port %d", id, port)
-			}
-			in.Close()
-			logrus.Debugf("Finish copying port forward input for %q port %d", id, port)
+			// tries to shutdown only its write stream, giving it a chance to terminate gracefully
+			defer conn.Close()
```
mikebrow, line 78: did you want to defer the close till after the timeout?

pkg/server/sandbox_portforward_unix.go
mikebrow: 1sec seems like a short timeout...

mikebrow: however https://manpages.debian.org/stretch/socat/socat.1.en.html says the default for socat is .5sec .. so

pkg/server/sandbox_portforward_unix.go
aojea: One of the problems I've observed is that despite it forwards the port, once the client disconnect it was leaking connections inside the namespace. I think this guarantees that we always close the connection.
Is it possible to not leak connections without deferring?

dims:
changed my mind. Let's wrap up Mike's comments and get this in to iterate. (no need to wait for Lantao)

pkg/server/sandbox_portforward_unix.go
mikebrow: my question is to the desired scope of the deferred close.. Did you want to defer the close to the end of each go func?  If just for connection leaks I'd think about doing these defers at the `func(_ ns.NetNS) error { ` scope.  We probably need to be using cancellable context and check for case <-ctx.Done(): to make sure we get a chance to close. Hate connection leaks :-)

aojea: ahh ok, it has to happen inside the go func() because it "closes" the other direction and "gives a chance to terminate gracefully", otherwise we'll always end with the timeout.
I think that this way we are mimicking socat completely

mikebrow: I thought it was leaking before, not sure we should mimic that :)  Will this way not leak on context close? Closing the other direction abruptly ends the other direction with no bytes and EOF.. which is not as graceful as a select with wait for context close or timeout then close forcefully.  If my read of the code is right.. it can't timeout unless the close is locked out for some reason?  Did socat always time out?  hmm

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,58 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+		timeout := time.Duration(500) * time.Millisecond
+		select {
+		case e := <-errCh:
+			if e != nil {
+				// This error may be caused because we are shutting down the connection, only log it
+				log.G(ctx).Debugf("PortForward closed the connection with error %v in network namespace %q port %d", e, id, port)
+			} else {
+				log.G(ctx).Debugf("PortForward closed the connection gracefully in network namespace %q port %d", id, port)
+			}
+		case <-time.After(timeout):
```
fuweid, line 109: how about using var `ctx` for timeout?

mikebrow: or both.. in case ctx isn't timing out for some odd reason waiting on the streams to close :)

aojea: then I think that the context cancellation should be done when we are blocked waiting for the first error, right?

aojea: 5e7b3e4e35e8c4a94dd967681b06e2a2b61ff4c3

mikebrow:
sry each commit has to be signed :)

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,63 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+		}()
+
+		// Wait until the first error is returned or the context cancels the operation
+		// we use errFwd to store the result of the port forwarding operation
+		var errFwd error
+		select {
+		case errFwd = <-errCh:
+		case <-ctx.Done():
+			log.G(ctx).Debugf("PortForward cancelled by %v in network namespace %q %d", ctx.Err(), id, port)
+			errFwd = ctx.Err()
```
mikebrow, line 100: if ctx is done I think it might be preferable to close your connections so they don't leak and return.

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,63 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
 		}
-		return nil
+		// give a chance to terminate gracefully or timeout
+		// 0.5s is the default timeout used in socat
+		// https://linux.die.net/man/1/socat
+		timeout := time.Duration(500) * time.Millisecond
+		select {
```
mikebrow, line 106: ctx may move to done while waiting on the other channel to close.

pkg/server/sandbox_portforward_unix.go
aojea: > mimicking socat completely

I mean socat itself, not to  the previous implementation with socat :sweat_smile: 

I think that this comment in the code we are replacing is significant too, using the pipe it was forcing to close everything
 
https://github.com/containerd/cri/pull/1470/files#diff-a60e53338184ff4aee2ff77caed7cabaL81-L89

> 		// If we use Stdin, command.Run() won't return until the goroutine that's copying	
> 		// from stream finishes. Unfortunately, if you have a client like telnet connected	
> 		// via port forwarding, as long as the user's telnet client is connected to the user's	
> 		// local listener that port forwarding sets up, the telnet session never exits. This	
> 		// means that even if socat has finished running, command.Run() won't ever return	
> 		// (because the client still has the connection and stream open).	
> 		//	
> 		// The work around is to use StdinPipe(), as Wait() (called by Run()) closes the pipe	
> 		// when the command (socat) exits



> Closing the other direction abruptly ends the other direction with no bytes and EOF..

we are closing the writer, so `io.Copy` will only exit if it tries to write, or the consumer of that stream close the connection (gracefully), otherwise, it will keep waiting forever ... that's the problem described by Lantao here https://github.com/containerd/cri/pull/729#issue-180773387

The previous attempt to replace socat was using a waiting group, hence the deadlock, currently we only need any of the streams to close and then we try to close everything, giving an opportunity to the applications that are using the stream to close it by themselves.

pkg/server/sandbox_portforward_unix.go
aojea: I understand now your suggestion to move the close() out of the goroutine, sorry, I wasn't getting it before

mikebrow: :-) all good, it's Friday! just having some fun with OSS. Thanks for doing this for us.

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,65 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+			_, err := io.Copy(conn, stream)
+			errCh <- err
+		}()
+
+		// Wait until the first error is returned by one of the connections
+		// we use errFwd to store the result of the port forwarding operation
+		// if the context is cancelled close everything and return
+		var errFwd error
+		select {
+		case errFwd = <-errCh:
```
mikebrow, line 96: debugf here for the first stream is closed

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,65 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
+		// https://linux.die.net/man/1/socat
+		timeout := time.Duration(500) * time.Millisecond
+		select {
+		case e := <-errCh:
+			if e != nil {
+				// This error may be caused because we are shutting down the connection, only log it
+				log.G(ctx).Debugf("PortForward closed the connection with error %v in network namespace %q port %d", e, id, port)
+			} else {
+				log.G(ctx).Debugf("PortForward closed the connection gracefully in network namespace %q port %d", id, port)
+			}
```
mikebrow, line 112: this if `e!=nil` logic made sense when we had code closing the other stream.. not so much now.

maybe `if errFwd == nil {errFwd = e}` and a sing line of output indicating the other stream is closed

thaJeztah:
Love the branch name 😁 `gocat` 😁

pkg/server/sandbox_portforward_unix.go
```diff
@@ -62,48 +61,64 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
-		// (because the client still has the connection and stream open).
-		//
-		// The work around is to use StdinPipe(), as Wait() (called by Run()) closes the pipe
-		// when the command (socat) exits.
-		in, err := cmd.StdinPipe()
+		defer stream.Close()
+		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default if the system has IPv6 enabled.
+		// Theoretically happy eyeballs will try IPv6 first and fallback to IPv4
+		// but resolving localhost doesn't seem to return and IPv4 address, thus failing the connection.
+		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
```
saschagrunert, line 70: @aojea could we do something like this?
```suggestion
		conn, err := net.Dial("tcp4", fmt.Sprintf("%s:%d", s.IP, port))
```

aojea: I was just keeping the old behaviour because of my ignorance on how the ports are exposed inside the namespace.

I can explain at the network level the problem and why I think is forcing IPv4 here.
`localhost`, depending on some factors, may resolve to `[::1]` or `127.0.0.1`, Long history short, commonly it resolves to [::1] if you are more interested in the whole history I explain it in this link https://gist.github.com/aojea/94f6f483173641647c731f582e52f0b0

But typically, the applications use to listen only 127.0.0.1 inside the container :shrug: , that's the part I'm missing. 

If we are able to identify correctly the address that the app is listening, then is better using ti, so this works for both IPv4 and IPv6, but I prefer having @mikebrow guidance here.

mikebrow: it's hardcoded localhost(or 127.0.0.1):random port for sec. reasons... or at least that's what the decision was way back when.

Any change here would likely need a kep that sets up streaming ip(s) and port(s) at the CRI request and/or response level.

PR #1472, opened 2020-05-08
Add config flag to default empty seccomp profile

mxpv:
This change adds `default_seccomp_profile` config switch to apply default seccomp profile when not provided by k8s.

Signed-off-by: Maksym Pavlenko <pavlenko.maksym@gmail.com>

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -286,7 +286,7 @@ func (c *criService) containerSpecOpts(config *runtime.ContainerConfig, imageCon
 		specOpts = append(specOpts, apparmorSpecOpts)
 	}
 
-	seccompSpecOpts, err := generateSeccompSpecOpts(
+	seccompSpecOpts, err := c.generateSeccompSpecOpts(
 		securityContext.GetSeccompProfilePath(),
 		securityContext.GetPrivileged(),
 		c.seccompEnabled())
@@ -300,10 +300,13 @@ func (c *criService) containerSpecOpts(config *runtime.ContainerConfig, imageCon
 }
 
 // generateSeccompSpecOpts generates containerd SpecOpts for seccomp.
-func generateSeccompSpecOpts(seccompProf string, privileged, seccompEnabled bool) (oci.SpecOpts, error) {
+func (c *criService) generateSeccompSpecOpts(seccompProf string, privileged, seccompEnabled bool) (oci.SpecOpts, error) {
 	if privileged {
 		// Do not set seccomp profile when container is privileged
 		return nil, nil
+	}
+	if seccompProf == "" {
+		seccompProf = c.config.DefaultSeccompProfile
 	}
 	// Set seccomp profile
 	if seccompProf == runtimeDefault || seccompProf == dockerDefault {
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -306,7 +306,7 @@ func (c *criService) generateSeccompSpecOpts(seccompProf string, privileged, sec
 		return nil, nil
 	}
 	if seccompProf == "" {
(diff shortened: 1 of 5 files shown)


pkg/server/container_create_unix.go
```diff
@@ -300,11 +300,14 @@ func (c *criService) containerSpecOpts(config *runtime.ContainerConfig, imageCon
 }
 
 // generateSeccompSpecOpts generates containerd SpecOpts for seccomp.
-func generateSeccompSpecOpts(seccompProf string, privileged, seccompEnabled bool) (oci.SpecOpts, error) {
+func (c *criService) generateSeccompSpecOpts(seccompProf string, privileged, seccompEnabled bool) (oci.SpecOpts, error) {
 	if privileged {
 		// Do not set seccomp profile when container is privileged
 		return nil, nil
 	}
+	if seccompProf == "" {
```
mikebrow, line 308: I think we also want to check `|| seccompProf == runtimeDefault`... since we'll now be supporting the configuring of the container runtime default to be something other than dockerDefault..

FYI from the cri api spec:
```
	// Seccomp profile for the sandbox, candidate values are:
	// * runtime/default: the default profile for the container runtime
	// * unconfined: unconfined profile, ie, no seccomp sandboxing
	// * localhost/<full-path-to-profile>: the profile installed on the node.
	//   <full-path-to-profile> is the full path of the profile.
	// Default: "", which is identical with unconfined.
```

mikebrow: not sure why there's no description of `docker/default` at the cri spec hmm..

crosbymichael: i'm not sure, the runtime/default my mean something, like the docker or containerd provided default seccomp profile

mikebrow: right feels like runtime/default is supposed to mean custom configured profile for the runtime and docker/default the original ..

mikebrow: we're changing "" from unconfined to some other default...

mikebrow: checking kublet code.. there was a deprecation for docker/default.. that's why it's missing from the docs on CRI

mikebrow: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/helpers_linux.go#L51

crosbymichael: I don't think changing "" from unconfined to something else is a big deal because you have to explicitly set this in the config as an admin.  You know you are making the change for your workloads.

mikebrow: fair.. Would submit then if we go that way we should push a change to the CRI to say something to the effect unless otherwise restricted by the container runtime.. Then on this pr name the field defaultUnsetSeccompProfile or maybe defaultUnconfinedSeccompProfile and indicate as such in the config default output and config.md. Thoughts?

mxpv: @mikebrow should we drop either `default` or `unset`? `seccompProfileIfUnset`? (naming is the most challenging part)

mikebrow: all good

mikebrow: so long as it's explained in the docs where they see the option

mikebrow: @mxpv We've always talked about eventually having a way to configure the runtime/default seccomp profile, so when I saw your defaultSeccompProfile I just presumed it would be for that. Didn't even read the "empty" part in the title of the PR. Now that I see it's just for the "" request... I like ```unsetSeccompProfile (default `unconfined`) If the seccompProfile is unset (`""`) for a container, containerd/cri will use the unsetSeccompProfile. ```

PR #1475, opened 2020-05-10
newTransport(): remove deprecated DualStack option

thaJeztah:
The `DualStack` option was deprecated in Go 1.12, and is now enabled by default
(through commit github.com/golang/go@efc185029bf770894defe63cec2c72a4c84b2ee9).

> The Dialer.DualStack field is now meaningless and documented as deprecated.
>
> To disable fallback, set FallbackDelay to a negative value.

The default `FallbackDelay` is 300ms; to make this more explicit, this patch
sets `FallbackDelay` to the default value.

Note that Docker Hub currently does not support IPv6 (DNS for registry-1.docker.io
has no AAAA records, so we should not hit the 300ms delay).

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -425,9 +425,9 @@ func newTransport() *http.Transport {
 	return &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
 		DialContext: (&net.Dialer{
-			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
-			DualStack: true,
+			Timeout:       30 * time.Second,
+			KeepAlive:     30 * time.Second,
+			FallbackDelay: 300 * time.Millisecond,
 		}).DialContext,
 		MaxIdleConns:          10,
 		IdleConnTimeout:       30 * time.Second,


pkg/server/image_pull.go
```diff
@@ -425,9 +425,9 @@ func newTransport() *http.Transport {
 	return &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
 		DialContext: (&net.Dialer{
-			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
-			DualStack: true,
+			Timeout:       30 * time.Second,
+			KeepAlive:     30 * time.Second,
+			FallbackDelay: 300 * time.Millisecond,
```
mikebrow, line 430: fair to remove the dual stack setting since it's now true by default, but why introduce an explicit constant for the 0 maps to 300mil fallback delay?  Would personally prefer `// TODO () need a config option for fallbackdelay default is zero which is mapped to 300mil by fallbackDelay() in golang net/dial.go`

thaJeztah: I wasn't sure if we _did_ want to have an option for this, hence kept it in line with the other options which set a default here (and make sure that possible updates in Golang's defaults wouldn't introduce a change in behaviour). Happy to remove it though if you prefer that.

As to adding a "TODO"; I see there's an existing TODO on the function itself;

```
TODO(random-liu): Create a library and share this code with `ctr`.
```

I _think_ it's worth looking at that (as a follow-up); there's some work done in containerd to provide a package for pulling images (https://github.com/containerd/containerd/pull/3728); possibly look at what's missing and add those parts, then use it in this repository

PR #1477, opened 2020-05-12
Increase port-forward timeout to 1s to fix e2e test

saschagrunert:
We encountered two failing end-to-end tests after the adoption of https://github.com/containerd/cri/pull/1470 in https://github.com/cri-o/cri-o/pull/3749:

```
Summarizing 2 Failures:
[Fail] [sig-cli] Kubectl Port forwarding With a server listening on 0.0.0.0 that expects a client request [It] should support a client that connects,
sends DATA, and disconnects
test/e2e/kubectl/portforward.go:343

[Fail] [sig-cli] Kubectl Port forwarding With a server listening on localhost that expects a client request [It] should support a client that connects
, sends DATA, and disconnects
test/e2e/kubectl/portforward.go:343
```

Increasing the timeout to 1s fixes the issue.

cc @aojea

### DIFF
--- a/pkg/server/sandbox_portforward_unix.go
+++ b/pkg/server/sandbox_portforward_unix.go
@@ -100,9 +100,9 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
 			return ctx.Err()
 		}
 		// give a chance to terminate gracefully or timeout
-		// 0.5s is the default timeout used in socat
+		// after 1s
 		// https://linux.die.net/man/1/socat
-		timeout := time.Duration(500) * time.Millisecond
+		const timeout = time.Second
 		select {
 		case e := <-errCh:
 			if errFwd == nil {


pkg/server/sandbox_portforward_unix.go
```diff
@@ -100,12 +100,12 @@ func (c *criService) portForward(ctx context.Context, id string, port int32, str
...
 		// give a chance to terminate gracefully or timeout
-		// 0.5s is the default timeout used in socat
+		// after 1s
 		// https://linux.die.net/man/1/socat
-		timeout := time.Duration(500) * time.Millisecond
+		const timeout = time.Second
 		select {
 		case e := <-errCh:
-			if errFwd == nil {
+			if errFwd == nil && e != nil {
```
saschagrunert, line 108: I think this can result in a nil assignment, which would make the check in L122 fail.

aojea:
I didn't know there were kubernetes tests covering this, otherwise, I should've tested the patch, sorry.

Anyway, I can confirm that the timeout it's too short and make the e2e Kubernetes fail, using 1 second the tests pass.

pkg/server/sandbox_portforward_unix.go
aojea: it only will be a nil assignment if errFwd was nil. Otherwise, errFwd keeps the error obtained from the first io.Copy() independently of the value of `e`

dims:
LGTM

/ok-to-test

dims:
LGTM

yes please!

pkg/server/sandbox_portforward_unix.go
saschagrunert: Yes, removed that additional check

aojea:
It seems that socat approach had these failures too.
Definitively this timeout seems related https://github.com/kubernetes/kubernetes/issues/86355 and maybe something we should tune in the future?

saschagrunert:
> It seems that socat approach had these failures too.
> Definitively this timeout seems related [kubernetes/kubernetes#86355](https://github.com/kubernetes/kubernetes/issues/86355) and maybe something we should tune in the future?

Is the actual issue that the client (container runtime) does not know when there is no more content to be expected from the streaming server? If yes, then I'm wondering if we can optimize there:
https://github.com/kubernetes/kubernetes/blob/70948498bcc696e66780edb4e4652366d4862e65/pkg/kubelet/server/portforward/httpstream.go#L36-L68

Or there: https://github.com/kubernetes/kubernetes/blob/70948498bcc696e66780edb4e4652366d4862e65/pkg/kubelet/server/portforward/httpstream.go#L208-L234

mikebrow:
> > It seems that socat approach had these failures too.
> > Definitively this timeout seems related [kubernetes/kubernetes#86355](https://github.com/kubernetes/kubernetes/issues/86355) and maybe something we should tune in the future?
> 
> Is the actual issue that the client (container runtime) does not know when there is no more content to be expected from the streaming server? If yes, then I'm wondering if we can optimize there:
> https://github.com/kubernetes/kubernetes/blob/70948498bcc696e66780edb4e4652366d4862e65/pkg/kubelet/server/portforward/httpstream.go#L36-L68
> 
> Or there: https://github.com/kubernetes/kubernetes/blob/70948498bcc696e66780edb4e4652366d4862e65/pkg/kubelet/server/portforward/httpstream.go#L208-L234

IMO It's a test case error.
https://github.com/kubernetes/kubernetes/blob/89ba90573f163ee3452b526f30348a035d54e870/test/e2e/kubectl/portforward.go#L322-L344

The test sends  'abc' to the stream for it's writer on the port, then immediately closes their writer. Then the code starts reading. At some point it is expected that the code in the container is getting ready to get run read the abc and do something like send data back over the port, but we have received the close already because they closed half their writer. Now it's a race condition.  If the test had waited to close the writer there would be no race.

saschagrunert:
Thank you for the insights @mikebrow, I can propose a PR to fix that and we can reduce the timeout here again afterwards. :)

mikebrow:
> Thank you for the insights @mikebrow, I can propose a PR to fix that and we can reduce the timeout here again afterwards. :)

Sure, thx. I like your idea for using an http idle wait after one or the other side has been closed, I had considered that as well but figured it would need to be in this short time range anyway so.. Maybe a config value for the timeout..

saschagrunert:
Good, a fix the e2e test is incoming here: https://github.com/kubernetes/kubernetes/pull/91045

PR #1485, opened 2020-05-18
vendor.conf: back to using tags

thaJeztah:
When I changed the vendor.conf format to use tags, many of the
dependencies didn't use tagged versions, and the column format
made the file slightly more consistent / easier to read.

With many dependencies moving to go modules, we see more deps
tagging releases, and we're now more actively trying to use
tagged releases for our dependencies.

With containerd/containerd changing the format to use tags as
default, it makes sense to do the same here as well (to allow
for easier comparing the vendor.conf files between repositories)


dims:
LGTM

PR #1487, opened 2020-05-19
Add SELinux Support for CRI

crosbymichael:
This is a carry of #1246.  

We are splitting up the work and I will be carrying this PR to merge.

### DIFF
--- a/hack/utils.sh
+++ b/hack/utils.sh
@@ -17,7 +17,7 @@
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 
 # Not from vendor.conf.
-CRITOOL_VERSION=v1.18.0
+CRITOOL_VERSION=89384cc13a27bb9128553c9fe75a7cc07c6a95bb
 CRITOOL_PKG=github.com/kubernetes-sigs/cri-tools
 CRITOOL_REPO=github.com/kubernetes-sigs/cri-tools
 
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -29,7 +29,6 @@ import (
 	cni "github.com/containerd/go-cni"
 	"github.com/containerd/typeurl"
 	"github.com/davecgh/go-spew/spew"
-	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
@@ -43,6 +42,7 @@ import (
 	"github.com/containerd/cri/pkg/netns"
 	sandboxstore "github.com/containerd/cri/pkg/store/sandbox"
 	"github.com/containerd/cri/pkg/util"
+	selinux "github.com/opencontainers/selinux/go-selinux"
 )
 
 func init() {
@@ -161,7 +161,7 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 	sandbox.ProcessLabel = spec.Process.SelinuxLabel
 	defer func() {
 		if retErr != nil {
-			_ = label.ReleaseLabel(sandbox.ProcessLabel)
+			selinux.ReleaseLabel(sandbox.ProcessLabel)
 		}
 	}()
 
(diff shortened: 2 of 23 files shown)


crosbymichael:
I went ahead and squashed this as it had many WIP commits and unsigned commits from the previous PR causing CI to fail.

hack/utils.sh
```diff
@@ -17,9 +17,9 @@
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/..
 
 # Not from vendor.conf.
-CRITOOL_VERSION=v1.18.0
+CRITOOL_VERSION=selinux
 CRITOOL_PKG=github.com/kubernetes-sigs/cri-tools
-CRITOOL_REPO=github.com/kubernetes-sigs/cri-tools
+CRITOOL_REPO=github.com/erikwilson/cri-tools
```
crosbymichael, line 22: Just making a comment as a note, we will have to update this back to upstream once all the upstream PRs are merged into cri-tools

mikebrow:
e2e bucket timed out for no obvious reason restarting

crosbymichael:
I'm not getting good info on why this is timing out....

hack/utils.sh
erikwilson: The initial PR for MCS testing has been merged and is now available at https://github.com/kubernetes-sigs/cri-tools/commit/89384cc13a27bb9128553c9fe75a7cc07c6a95bb

crosbymichael: Thanks, i'll update back to upstream

crosbymichael:

it looks like a normal timeout now for a single test, going to rerun

crosbymichael:
victory!

crosbymichael:
Tested labeling for volumes:

```bash
system_u:system_r:container_t:s0:c192,c450 root 9226 0.0  0.1 10636 5168 ?       Ss   11:57   0:00 nginx: master process nginx -g daemon off;
system_u:system_r:container_t:s0:c192,c450 101 9240 0.0  0.0 11100 2552 ?        S    11:57   0:00 nginx: worker process
```

process ^^ 

```bash
drwxr-xr-x. 2 root root unconfined_u:object_r:mnt_t:s0                  6 May 26 11:57 mnt
drwxr-xr-x. 2 root root system_u:object_r:container_file_t:s0:c192,c450 6 May 26 11:57 test
```

mount ^^

crosbymichael:
I think we could merge this and then follow up on adding reference counting to the pod level and off of the container.

pkg/server/sandbox_run.go
```diff
@@ -157,6 +158,18 @@ func (c *criService) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		return nil, errors.Wrap(err, "failed to generate sandbox container spec")
 	}
 	log.G(ctx).Debugf("Sandbox container %q spec: %#+v", id, spew.NewFormatter(spec))
+	sandbox.ProcessLabel = spec.Process.SelinuxLabel
+	defer func() {
+		if retErr != nil {
+			_ = label.ReleaseLabel(sandbox.ProcessLabel)
```
mxpv, line 164: nit: this one is deprecated. We can use `selinux.ReleaseLabel` which returns nothing, so we don't have to write `_ = X` everywhere.

crosbymichael: I'll add this for the next around of updates.

crosbymichael: i went ahead and updated this

crosbymichael:
Updated, please take another look

pkg/server/service.go
```diff
@@ -99,12 +100,13 @@ type criService struct {
...
 func NewCRIService(config criconfig.Config, client *containerd.Client) (CRIService, error) {
 	var err error
+	labels := label.NewStore()
 	c := &criService{
 		config:             config,
 		client:             client,
 		os:                 osinterface.RealOS{},
-		sandboxStore:       sandboxstore.NewStore(),
-		containerStore:     containerstore.NewStore(),
+		sandboxStore:       sandboxstore.NewStore(labels),
```
dmcgowan, line 108: We should try to decouple this a little bit, maybe having the label store could be registered through the plugin system in the future.

dmcgowan: In a follow up or maybe post merge back into containerd

crosbymichael:
Ya, the store is what I’m currently working on and will be putting a PR up shortly.

Michael Crosby

PR #1497, opened 2020-05-27
would like to have the commit sha in the log file names

mikebrow:
adding the pull request sha to the log filenames

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
       - name: Upload Integration Log File
         uses: actions/upload-artifact@v1
         with:
-          name: integration-test.log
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -128,7 +128,7 @@ jobs:
       - name: Upload Integration Log File
         uses: actions/upload-artifact@v1
         with:
-          name: integration-test.log
+          name: integration-test-${{ github.event.pull_request.head.sha }}.log
```
mxpv, line 131: This CI script can be triggered with `push` event as well, so this might not work for all cases. 
It's better to rely on `GITHUB_SHA` env: https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables

PR #1498, opened 2020-05-28
Specify base OCI runtime spec

mxpv:
This PR adds `base_runtime_spec` to runtime config to use a file with base OCI spec.

Details: https://github.com/containerd/cri/issues/1488

Signed-off-by: Maksym Pavlenko <pavlenko.maksym@gmail.com>

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -135,6 +135,11 @@ version = 2
       # i.e pass host devices through to privileged containers.
       privileged_without_host_devices = false
 
+      # base_runtime_spec is a file path to a JSON file with the OCI spec that will be used as the base spec
+      # that all container's are created from.
+      # Use containerd's `ctr oci default-spec > /etc/containerd/cri-base.json` to output initial spec file.
+      base_runtime_spec = ""
+
       # 'plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options' is options specific to
       # "io.containerd.runc.v1" and "io.containerd.runc.v2". Its corresponding options type is:
       #   https://github.com/containerd/containerd/blob/v1.3.2/runtime/v2/runc/options/oci.pb.go#L26 .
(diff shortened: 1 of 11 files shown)


docs/config.md
```diff
@@ -135,6 +135,11 @@ version = 2
       # i.e pass host devices through to privileged containers.
       privileged_without_host_devices = false
 
+      # base_runtime_spec is a file path to a JSON file with the OCI spec that will be used as the base spec
+      # that all container's are created from.
+      # Use containerd's `ctr oci default-spec > /etc/containerd/cri-base.json` to output initial spec file.
+      base_runtime_spec = ""
```
mikebrow, line 141: nit.. may want to mention a restart of containerd is required to load changes to the one default spec...
still running containers and restarted containers will still be using the original spec from which that container was created.

mxpv: Done

pkg/server/container_create.go
```diff
@@ -297,12 +297,33 @@ func (c *criService) volumeMounts(containerRootDir string, criMounts []*runtime.
...
+		if !ok {
+			return nil, errors.Errorf("can't find base OCI spec %q", baseSpecFile)
+		}
+
+		spec := oci.Spec{}
+		if err := util.DeepCopy(&spec, &baseSpec); err != nil {
+			return nil, errors.Wrap(err, "failed to clone OCI spec")
+		}
+
+		if err := oci.ApplyOpts(ctx, nil, container, &spec, opts...); err != nil {
```
crosbymichael, line 316: After doing this you need to make sure that you setup the cgrouppath based on the namespace of context.

```go
				var s oci.Spec
		if err := json.NewDecoder(f).Decode(&s); err != nil {
			return errors.Wrap(err, "decode spec")
		}
		s.Linux.CgroupsPath = filepath.Join("/", ns, c.ID)

		if err := oci.ApplyOpts(ctx, client, c, &s, opts...); err != nil {
			return errors.Wrap(err, "apply opts")
		}
```

mxpv: Good catch, thank you.

crosbymichael:
LGTM

PR #1501, opened 2020-06-01
Tolerate missing HugeTLB cgroups controller

dims:
Kubernetes has been sending hugepage limits for a few releases now. Things used to work because containerd-cri failed to honor them. 

However starting in the following [diff](https://github.com/containerd/containerd/compare/97ca1be0..7a5fcf61#diff-85b069277b32de3389720de87247f54eR451-R456) we noticed that k8s failed hard with the following error:

```
 ay 31 02:47:12 kind-control-plane containerd[133]: time="2020-05-31T02:47:12.295625667Z" level=error msg="StartContainer for \"6426531767995db591ea89b05e52ebfa8887d2cd23c095ddf8e1727d1d6ec8cb\" failed" error="failed to create containerd task: OCI runtime create failed: container_linux.go:346: starting container process caused \"process_linux.go:449: container init caused \\\"process_linux.go:415: setting cgroup config for procHooks process caused \\\\\\\"cannot set hugetlb limit: container could not join or create cgroup\\\\\\\"\\\"\": unknown"
```
We tracked it down to the following commit:
https://github.com/containerd/cri/commit/c02c24847f191ea5d38bb3917ec5c54222ba5501

Essentially, containerd was trying to set hugepages limits when the cgroup controller (`hugeTLB`) was absent. So this failed in a variety of legacy environments. So in this PR we detect this situation and have a toggle to keep the old behavior until we can patch/fix k8s in the wild somehow. Since CRI-O is ignoring the hugepage limit in the same exact situation and this was the previous behavior, let's please do the same.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -230,6 +230,10 @@ type PluginConfig struct {
 	// UnsetSeccompProfile is the profile containerd/cri will use If the provided seccomp profile is
 	// unset (`""`) for a container (default is `unconfined`)
 	UnsetSeccompProfile string `toml:"unset_seccomp_profile" json:"unsetSeccompProfile"`
+	// TolerateMissingHugePagesCgroupController if set to false will error out on create/update
+	// container requests with huge page limits if the cgroup controller for hugepages is not present.
+	// This helps with supporting Kubernetes <=1.18 out of the box. (default is `true`)
+	TolerateMissingHugePagesCgroupController bool `toml:"tolerate_missing_hugepages_controller" json:"tolerateMissingHugePagesCgroupController"`
 }
 
 // X509KeyPairStreaming contains the x509 configuration for streaming
--- a/pkg/containerd/opts/spec_unix.go
+++ b/pkg/containerd/opts/spec_unix.go
@@ -27,6 +27,8 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
+	"syscall"
 
 	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/log"
@@ -36,6 +38,7 @@ import (
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 
@@ -405,7 +408,7 @@ func WithSelinuxLabels(process, mount string) oci.SpecOpts {
 }
 
 // WithResources sets the provided resource restrictions
-func WithResources(resources *runtime.LinuxContainerResources) oci.SpecOpts {
+func WithResources(resources *runtime.LinuxContainerResources, tolerateMissingHugePagesCgroupController bool) oci.SpecOpts {
 	return func(ctx context.Context, client oci.Client, c *containers.Container, s *runtimespec.Spec) (err error) {
 		if resources == nil {
 			return nil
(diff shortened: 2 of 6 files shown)


pkg/containerd/opts/spec_unix.go
```diff
@@ -448,16 +450,47 @@ func WithResources(resources *runtime.LinuxContainerResources) oci.SpecOpts {
...
+			controllers, err := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
+			if err == nil {
+				if strings.Contains(string(controllers), "hugetlb") {
+					_cgroupHasHugetlb = true
+				}
+			}
+		}
+	})
+	return _cgroupHasHugetlb
+}
```
mikebrow, line 492: needs this logic:
```
		supportsHugetlb := false
		if IsCgroup2UnifiedMode() {
			supportsHugetlb, _ = cgroupv2HasHugetlb()
		} else {
			supportsHugetlb, err = cgroupv1HasHugetlb()
		}
```

uses:
```
var (
	_cgroupv1HasHugetlbOnce sync.Once
	_cgroupv1HasHugetlb     bool
	_cgroupv1HasHugetlbErr  error
	_cgroupv2HasHugetlbOnce sync.Once
	_cgroupv2HasHugetlb     bool
	_cgroupv2HasHugetlbErr  error
)

// cgroupv1HasHugetlb returns whether the hugetlb controller is present on
// cgroup v1.
func cgroupv1HasHugetlb() (bool, error) {
	_cgroupv1HasHugetlbOnce.Do(func() {
		if _, err := ioutil.ReadDir("/sys/fs/cgroup/hugetlb"); err != nil {
			_cgroupv1HasHugetlbErr = errors.Wrap(err, "readdir /sys/fs/cgroup/hugetlb")
			_cgroupv1HasHugetlb = false
		} else {
			_cgroupv1HasHugetlbErr = nil
			_cgroupv1HasHugetlb = true
		}
	})
	return _cgroupv1HasHugetlb, _cgroupv1HasHugetlbErr
}

// cgroupv2HasHugetlb returns whether the hugetlb controller is present on
// cgroup v2.
func cgroupv2HasHugetlb() (bool, error) {
	_cgroupv2HasHugetlbOnce.Do(func() {
		controllers, err := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
		if err != nil {
			_cgroupv2HasHugetlbErr = errors.Wrap(err, "read /sys/fs/cgroup/cgroup.controllers")
			return
		}
		_cgroupv2HasHugetlb = strings.Contains(string(controllers), "hugetlb")
	})
	return _cgroupv2HasHugetlb, _cgroupv2HasHugetlbErr
}

// IsCgroup2UnifiedMode returns whether we are running in cgroup v2 unified mode.
func IsCgroup2UnifiedMode() bool {
	isUnifiedOnce.Do(func() {
		var st syscall.Statfs_t
		if err := syscall.Statfs("/sys/fs/cgroup", &st); err != nil {
			panic("cannot statfs cgroup root")
		}
		isUnified = st.Type == unix.CGROUP2_SUPER_MAGIC
	})
	return isUnified
}
```

dims: Done! thanks @mikebrow

mikebrow:
restarted windows appveyor and... success !

pkg/config/config.go
```diff
@@ -230,6 +230,10 @@ type PluginConfig struct {
 	// UnsetSeccompProfile is the profile containerd/cri will use If the provided seccomp profile is
 	// unset (`""`) for a container (default is `unconfined`)
 	UnsetSeccompProfile string `toml:"unset_seccomp_profile" json:"unsetSeccompProfile"`
+	// TolerateMissingHugePagesCgroupController if set to false will error out on create/update
+	// container requests with huge page limits if the cgroup controller for hugepages is not present.
+	// This helps with supporting Kubernetes <=1.18 out of the box. (default is `true`)
```
BenTheElder, line 235: should this be 1.19?

dims: @BenTheElder, we haven't released 1.19 yet right? So we could add similar code in 1.19 to detect and stop telling containerd/runc to do HugePages, so we can remove this flag in containerd at some point in time. Hence the `Kubernetes <=1.18` as we won't go back to fix older k8s already in the field.

dims:
@BenTheElder confirmed that this patch works!

pkg/config/config.go
mikebrow: I could see changing the default to false at some point thus moving from the warning about skipping the limits to an error and at the same time we'd deprecate the flag. 

I think saying 1.18 is ok for the context of when it was pushed.

mikebrow:
@dims I think we need to cherry pick this one (and the commit that copied the huge page limits) My reasoning is we are currently silent to the issue of not supporting huge page limits. It's that or we should post up a notice in the service release notes, highlighting that we ignore huge page limits with that release, and request those who need said limits to move up to 1.4.

dims:
@mikebrow here's where we picked up the containerd/cri commit into containerd/containerd - https://github.com/containerd/containerd/commit/12c7d69769b43b8f432fe4dd779335f02c52c036

So looks like we don't need to touch 1.3.x releases at all ( https://github.com/containerd/containerd/tree/release/1.3/vendor/github.com/containerd/cri/pkg/containerd/opts).

Since we have not cut a branch for release-1.4 in containerd/containerd, we are good to go here. Once this lands, it goes into next beta or rc for 1.4.

Please take a look to see if i am missing something

pkg/containerd/opts/spec_unix.go
```diff
@@ -448,16 +451,93 @@ func WithResources(resources *runtime.LinuxContainerResources) oci.SpecOpts {
...
+	_cgroupv2HasHugetlbOnce sync.Once
+	_cgroupv2HasHugetlb     bool
+	_cgroupv2HasHugetlbErr  error
+	isUnifiedOnce           sync.Once
+	isUnified               bool
+)
+
+// cgroupv1HasHugetlb returns whether the hugetlb controller is present on
+// cgroup v1.
+func cgroupv1HasHugetlb() (bool, error) {
```
fuweid, line 502: Is it good to move it cgroup repo?

mikebrow: agree this would be good in containerd/cgroups

pkg/config/config.go
BenTheElder: Quick question: will containerd / CRI ignore unknown fields? Or will we have to make sure we only set this starting at some minimum version?

We'll attempt a kind upgrade as soon as this is vendored into the main repo.

BenTheElder:
Yes -- 1.3.X has been unaffected, we rolled back to roughly 1.3.4

Thanks all

mikebrow:
Removing cherry pick needed flag until there is a request for it. The lack of support should be documented.

PR #1505, opened 2020-06-08
Add GMSA credential spec passing

dcantah:
Signed-off-by: Daniel Canter <dcanter@microsoft.com>

### DIFF
--- a/pkg/server/container_create_windows.go
+++ b/pkg/server/container_create_windows.go
@@ -68,13 +68,30 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 
 	specOpts = append(specOpts, customopts.WithWindowsMounts(c.os, config, extraMounts))
 
-	specOpts = append(specOpts, customopts.WithWindowsResources(config.GetWindows().GetResources()))
+	// Start with the image config user and override below if RunAsUsername is not "".
+	username := imageConfig.User
 
-	username := config.GetWindows().GetSecurityContext().GetRunAsUsername()
-	if username != "" {
-		specOpts = append(specOpts, oci.WithUser(username))
+	windowsConfig := config.GetWindows()
+	if windowsConfig != nil {
+		specOpts = append(specOpts, customopts.WithWindowsResources(windowsConfig.GetResources()))
+		securityCtx := windowsConfig.GetSecurityContext()
+		if securityCtx != nil {
+			runAsUser := securityCtx.GetRunAsUsername()
+			if runAsUser != "" {
+				username = runAsUser
+			}
+			cs := securityCtx.GetCredentialSpec()
+			if cs != "" {
+				specOpts = append(specOpts, customopts.WithWindowsCredentialSpec(cs))
+			}
+		}
 	}
-	// TODO(windows): Add CredentialSpec support.
+
+	// There really isn't a good Windows way to verify that the username is available in the
+	// image as early as here like there is for Linux. Later on in the stack hcsshim
+	// will handle the behavior of erroring out if the user isn't available in the image
+	// when trying to run the init process.
+	specOpts = append(specOpts, oci.WithUser(username))
 
 	for pKey, pValue := range getPassthroughAnnotations(sandboxConfig.Annotations,
 		ociRuntime.PodAnnotations) {
--- a/pkg/containerd/opts/spec_windows.go
+++ b/pkg/containerd/opts/spec_windows.go
@@ -188,3 +188,15 @@ func WithWindowsDefaultSandboxShares(ctx context.Context, client oci.Client, c *
 	s.Windows.Resources.CPU.Shares = &i
 	return nil
 }
+
+// WithWindowsCredentialSpec assigns `credentialSpec` to the
+// `runtime.Spec.Windows.CredentialSpec` field.
+func WithWindowsCredentialSpec(credentialSpec string) oci.SpecOpts {
+	return func(ctx context.Context, client oci.Client, c *containers.Container, s *runtimespec.Spec) error {
+		if s.Windows == nil {
+			s.Windows = &runtimespec.Windows{}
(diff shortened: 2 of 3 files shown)


pkg/server/container_create_windows.go
```diff
@@ -70,11 +70,17 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 
 	specOpts = append(specOpts, customopts.WithWindowsResources(config.GetWindows().GetResources()))
 
-	username := config.GetWindows().GetSecurityContext().GetRunAsUsername()
-	if username != "" {
-		specOpts = append(specOpts, oci.WithUser(username))
+	securityContext := config.GetWindows().GetSecurityContext()
```
mikebrow, line 73: Suggest cache the reference returned from config.GetWindows up on lne 71 and check it for nil...

pkg/server/container_create_windows.go
```diff
@@ -70,11 +70,17 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
...
+	if securityContext != nil {
+		username := securityContext.GetRunAsUsername()
```
mikebrow, line 75:
```
// WindowsContainerSecurityContext holds windows security configuration that will be applied to a container.
type WindowsContainerSecurityContext struct {
	// User name to run the container process as. If specified, the user MUST
	// exist in the container image and be resolved there by the runtime;
	// otherwise, the runtime MUST return error.
	RunAsUsername string `protobuf:"bytes,1,opt,name=run_as_username,json=runAsUsername,proto3" json:"run_as_username,omitempty"`
	// The contents of the GMSA credential spec to use to run this container.
```
I read that as: 
```
if username != "" && username != imageConfig.User { return nil, errors.Wrap(err, "failed to match username with image username")}
```

mikebrow: if so.. we should probably add that code ^ and test over here:
https://github.com/containerd/cri/blob/master/pkg/server/container_create_windows_test.go

to make sure all is well with the CRI rule above.

dcantah:
@mikebrow Thanks, will fix up and add the test :)

pkg/server/container_create_windows.go
dcantah: @mikebrow Sorry for not getting back to this yesterday. So there's a couple things regarding this. 

1. We can't just check if the user name isn't equal to the imageConfigs user as that would inhibit running as a local system account/another default container user (ContainerUser and ContainerAdministrator are the two default virtual accounts) if this wasn't what's present in the imageConfig. For example, on nanoserver:1809 the image config user is specified as ContainerUser, but ContainerAdministrator and the local accounts (NT AUTHORITY\NETWORK SERVICE etc) are all valid options as well. On our end in hcsshim we actually do error out when trying to create the init process with a user that doesn't exist in the container so this case is handled, just later down in the stack. 

2. What can be changed however is that it looks like we don't currently uphold the image config user if no override is passed, it seems we just default to ContainerAdministrator. Going through cri and running nanoserver, the container user ends up being ContainerAdministrator when it truthfully should be ContainerUser if we want parity with dockers behavior. So we can check if there is a user in the image config --> if not then leave as is and default to ContainerAdministrator, otherwise set this as spec.Process.User.Username unless RunAsUsername is specified. I can then add a test for this behavior. 

Let me know if this sounds good!

pkg/server/container_create_windows.go
```diff
@@ -68,13 +68,30 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
...
-	if username != "" {
-		specOpts = append(specOpts, oci.WithUser(username))
+	// Start with the image config user and override below if RunAsUsername is not "".
+	username := imageConfig.User
+
+	windowsConfig := config.GetWindows()
+	if windowsConfig != nil {
+		specOpts = append(specOpts, customopts.WithWindowsResources(windowsConfig.GetResources()))
+		securityCtx := windowsConfig.GetSecurityContext()
+		if securityCtx != nil {
```
kevpar, line 78: I don't think we need all these nil checks. The `Get*` functions in protobuf should propagate nils forward, so if you do `config.GetWindows().GetSecurityContext().GetRunAsUsername()` it should return "" even if `GetWindows()` returns nil.

dcantah: Can update 👍

dcantah:
@mikebrow What's the process of getting cri revendored into containerd?

pkg/server/container_create_windows.go
mikebrow: which nil check is not needed they all look valid to me..

mikebrow: IMO it's not a matter of will it panic.. it won't .. It's a matter of should the code proceed as if the get was successful.

dcantah: The caching of securityCtx and if securityCtx != nil could be removed in favor of just checking if `windowsConfig.GetSecurityContext.GetRunAsUsername()` and `windowsConfig.GetSecurityContext.GetCredentialSpec()` aren't empty strings. Either way works. I'm going to leave unless you feel strongly about it @kevpar

kevpar: I don't feel strongly. It seemed cleaner the other way, but agree it's a bit of an optimization to avoid doing some of the other work if an earlier `Get` call fails.

mikebrow:
that's weird I hit quote reply and got edit instead :-) ^

mikebrow:
> @mikebrow What's the process of getting cri revendored into containerd? 

summary: we'll update all our dependencies here to cover containerd dependencies considering if we need other dependencies updated as well, then when that tests out ok we'll do the same over on containerd/containerd merging containerd/cri dependency updates into containerd and the merge commit for containerd/cri that we decide to merge from...

PR #1508, opened 2020-06-14
Remove the protocol filter from the HostPort management

janosi:
Reason: originally it was introduced to prevent the loading of the SCTP kernel module on the nodes. But iptables chain creation alone does not load the kernel module. The module would be loaded if an SCTP socket was created, but neither cri nor the portmap CNI plugin starts managing SCTP sockets if hostPort / portmappings are defined.

Trigger:
Failing SCTP HostPort e2e test in K8s: https://github.com/kubernetes/kubernetes/issues/92041

Containerd issue (if this one is merged containerd/vendor shall be updated):
https://github.com/containerd/containerd/issues/4321

### DIFF
--- a/pkg/server/sandbox_run.go
+++ b/pkg/server/sandbox_run.go
@@ -414,9 +414,6 @@ func toCNIPortMappings(criPortMappings []*runtime.PortMapping) []cni.PortMapping
 		if mapping.HostPort <= 0 {
 			continue
 		}
-		if mapping.Protocol != runtime.Protocol_TCP && mapping.Protocol != runtime.Protocol_UDP {
-			continue
-		}
 		portMappings = append(portMappings, cni.PortMapping{
 			HostPort:      mapping.HostPort,
 			ContainerPort: mapping.ContainerPort,
--- a/pkg/server/sandbox_run_test.go
+++ b/pkg/server/sandbox_run_test.go
@@ -175,6 +175,12 @@ func TestToCNIPortMappings(t *testing.T) {
 					HostPort:      8765,
 					HostIp:        "126.125.124.123",
 				},
+				{
+					Protocol:      runtime.Protocol_SCTP,
+					ContainerPort: 1234,
+					HostPort:      5678,
+					HostIp:        "123.124.125.126",
+				},
 			},
 			cniPortMappings: []cni.PortMapping{
 				{
@@ -188,6 +194,12 @@ func TestToCNIPortMappings(t *testing.T) {
 					ContainerPort: 4321,
 					Protocol:      "tcp",
 					HostIP:        "126.125.124.123",
+				},
+				{
+					HostPort:      5678,
+					ContainerPort: 1234,
+					Protocol:      "sctp",
+					HostIP:        "123.124.125.126",
 				},
 			},
 		},
@@ -216,12 +228,6 @@ func TestToCNIPortMappings(t *testing.T) {
 		},
 		"CRI port mapping with unsupported protocol should be skipped": {
 			criPortMappings: []*runtime.PortMapping{
-				{
-					Protocol:      runtime.Protocol_SCTP,
-					ContainerPort: 1234,
-					HostPort:      5678,
-					HostIp:        "123.124.125.126",
-				},
 				{
 					Protocol:      runtime.Protocol_TCP,
 					ContainerPort: 4321,


aojea:
it LGTM, @janosi  I think you have to sign the commit

/assign @mikebrow 
/cc @bentheelder

janosi:
@aojea Thanks! Signed. 
Though, still failing. According to GitHub the commit is "Verified". So, I am not sure why.

aojea:
> Though, still failing. According to GitHub the commit is "Verified". So, I am not sure why.

It uses this https://github.com/containerd/project/blob/master/script/validate/dco

I think that you should format better the subject and the message, and use `git commit -s` to sign in the message

janosi:
@aojea Thanks! I re-signed it (now with github -s not with github -S) and the signature is OK now.

mikebrow:
@Random-Liu FYI
https://github.com/containerd/cri/pull/911

@janosi is there "never" a case where this should be filtered on any node?  If there is we could introduce a switch to enable the filter with a default of off.

janosi:
@mikebrow Please check my following logic, just to be sure. But if the setup of this port mapping on a node depends on the scheduling of the pod/container that specifies this port mapping to that node, then such a "I do not want to have this mapping happening on this node" should be configured with K8s node selectors and/or taints. I.e. I think the user has the configuration options to manage a heterogenous cluster from this perspective.
If we provided a configuration option on the container runtime level the user should label the affected nodes on K8s level, too, so pods/containers that have such hostPort definition are not scheduled to the affected nodes. But we get to the same point as without the configuration option on containerd level, I think. I mean, the user has to use K8s scheduling parameters on the pod to avoid the nodes where such mapping cannot be setup due to the containerd config.  But if the user does so, it does not matter anymore whether the node can be configured to setup such mappings or not.

mikebrow:
Your logic makes sense "if" said selectors and/or taints are setup.  If they are not setup and instead the user / cloud admin is relying on the container runtimes to filter "always" for all pods/containers... then we need a switch to resume original behavior or customer has to add new selectors and/or taints.

janosi:
Yes, indeed. On the other hand I think there is another aspect: the current situation is something like "you can configure SCTP host ports and your pods are scheduled fine but your desired functionality does not work". Which is not the best user experience, I think. If it is an intended and desired restriction the cluster operator shall rather deploy e.g. K8s validation webhooks to prevent the creation of pods with SCTP host ports. Or the cluster operator shall employ the aforementioned labels to provide a better UX.
The other thing to consider is, whether the current situation is a desired one from the cluster admin's perspective, or rather a "it just happened to be like this, and so far I have not really cared about it". I mean:
- it is recommended to avoid the usage of host port usage according to the K8s documentation the K8s documentation, in general
- SCTP support in K8s has been an alpha feature up til now
- Docker (dockershim in the kubelet) has set up these mappings for SCTP without problems
- the sum of the previous points (host port usage + alpha + works with Docker) means a lower chance that someone expects the current functionality as an intended logic
- the current change is not about preventing the load of an extra kernel module (with the possible additional attack vectors of that), because what we enable here will not load those kernel modules either
- what would be the reason to silently ignore the setup of host port mappings for SCTP and not informing the user about such a restriction (see my comment above with the validating webhook)?

mikebrow:
Good questions.. 
> what would be the reason to silently ignore the setup of host port mappings for SCTP and not informing the user about such a restriction (see my comment above with the validating webhook)?

silent.. I agree there should be a warning if filtered. Recognize this is two years after the filter was put in place. many things have changed. The request for the filter, came from the k8s CRI team, I thought. If they agree to take it out.. of course I'm good with that so long as there is documentation of the change in release notes. 

> the current change is not about preventing the load of an extra kernel module (with the possible additional attack vectors of that), because what we enable here will not load those kernel modules either

Sometimes we have multiple layers of protection. What happens if the SCTP module is in fact loaded? Removing this filter would then add an additional vector correct? .. 

If SCTP is not loaded why not filter.. if it is loaded do we want to enable?

What is the scenario we are trying to enable.. and should that scenario be default, configurable, or default with optional filter by admin?

mikebrow:
suggest
```
if !sctpEnabled() || config.filterSCTP {
  log.G(ctx).Debugf("filtering %v see ___ to enable",... 
  continue
}
```
or maybe a warning instead of debug out

janosi:
> If SCTP is not loaded why not filter.. if it is loaded do we want to enable?

The filtering prevents the creation of the iptables rules for the host port mapping. The filtering does not prevent the SCTP module loading, because that will not happen without the filtering either. I.e. currently the filtering prevents the host port feature from working with SCTP, but does not prevent any (known) security problem.

> What is the scenario we are trying to enable

To enable the creation of iptables rules for SCTP HostPorts.

Yep, I am aware of that CVE. And I know SCTP is not a usual protocol in IT environments ("esoteric"). On the other hand:
- SCTP is the basis of the M3UA and IUA protocol stacks which are replacements of all traditional SS7 protocol stacks in telecommunications. Where SS7 is the major protocol stack in both mobile and fixed telco networks. 
- SCTP is also a used protocol in the voice over IP telco networks, i.e. not only a basis of a "migrate the old to a newer protocol stack, and let it fade away as new things come", but also the basis of the new networks  
- I am sure there are CVEs for the other, less ("esoteric") protocols as well

mikebrow:
ok found this in the kubernetes thread: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0015-20180614-SCTP-support.md#the-solution-in-the-kubernetes-sctp-support-implementation

gives me some understanding of what this is about.

danwinship:
@mikebrow copying [from the kube bug](https://github.com/kubernetes/kubernetes/issues/92041#issuecomment-644092139):

> For TCP and UDP HostPorts, the runtime is expected to do two things:
> 1. open a listening socket on that port so the kernel will think it's in use and not allocate it to anyone else
> 2. create iptables rules (or whatever) to route traffic from the host port to the container port
> 
> For _SCTP_ HostPorts, the runtime _should_ do the second part (iptables rules), but _should not_ do the first part (listening socket), for two reasons:
> - Some users want to use "userspace SCTP", in which case they will be referring to SCTP at the Kubernetes level but need to keep the `sctp.ko` kernel module from ever being loaded. (This is discussed more [in the KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0015-20180614-SCTP-support.md#the-solution-in-the-kubernetes-sctp-support-implementation).)
> - If `sctp.ko` is not loaded when Kubernetes starts, we want to avoid having Kubernetes itself cause it to be auto-loaded, for security reasons. (More discussion of this [on the sig mailing list](https://groups.google.com/forum/#!msg/kubernetes-sig-network/KSi0DI-Gw80/fNpWFnAdCAAJ).)
> 
> So, runtimes should be filtering out SCTP for the opening-listening-sockets part, but not for the making-iptables-rules part. (Some implementations, notably CNI `portmap`, don't actually do the opening-listening-sockets part, so they wouldn't be filtering at all.)

mikebrow:
if anyone needs us to optionally add the filter back I'll do it on a subsequent PR

BenTheElder:
thanks everyone! :pray: 
what's left to merge this?

mikebrow:
> thanks everyone! 🙏
> what's left to merge this?

I asked for another set of eyes on this from the containerd maintainers.. expecting the extra review to happen today.

crosbymichael:
LGTM

PR #1509, opened 2020-06-15
[WIP] Close streams when process exits

dims:


dims:
possible fix for https://github.com/kubernetes/kubernetes/issues/92057

PR #1513, opened 2020-06-19
Change "failed to stop sandbox" error message to use state name instead of numeric value

brianpursley:
This PR changes the "failed to stop sandbox container" error message to show the state name instead of a numeric value, so it will look like this instead:

Old message example:
```
failed to stop sandbox container "91fc4d82d15624121520720bdf0c733bea8506bc129ea856bd6b2f0bb9d1c980" in '\x00' state
```

New message example:
```
failed to stop sandbox container "91fc4d82d15624121520720bdf0c733bea8506bc129ea856bd6b2f0bb9d1c980" in "StateReady" state
```

### DIFF
--- a/pkg/store/sandbox/status.go
+++ b/pkg/store/sandbox/status.go
@@ -17,8 +17,11 @@
 package sandbox
 
 import (
+	"strconv"
 	"sync"
 	"time"
+
+	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
 // The sandbox state machine in the CRI plugin:
@@ -63,7 +66,7 @@ type State uint32
 const (
 	// StateReady is ready state, it means sandbox container
 	// is running.
-	StateReady = iota
+	StateReady State = iota
 	// StateNotReady is notready state, it ONLY means sandbox
 	// container is not running.
 	// StopPodSandbox should still be called for NOTREADY sandbox to
@@ -74,6 +77,21 @@ const (
 	// into unknown state when its status fails to be loaded.
 	StateUnknown
 )
+
+// String returns the string representation of the state
+func (s State) String() string {
+	switch s {
+	case StateReady:
+		return runtime.PodSandboxState_SANDBOX_READY.String()
+	case StateNotReady:
+		return runtime.PodSandboxState_SANDBOX_NOTREADY.String()
+	case StateUnknown:
+		// PodSandboxState doesn't have an unknown state, but State does, so return a string using the same convention
+		return "SANDBOX_UNKNOWN"
+	default:
+		return "invalid sandbox state value: " + strconv.Itoa(int(s))
+	}
+}
 
 // Status is the status of a sandbox.
 type Status struct {
--- a/pkg/store/sandbox/status_test.go
+++ b/pkg/store/sandbox/status_test.go
@@ -18,6 +18,7 @@ package sandbox
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


mikebrow:
please rebase off master to pick up fix to the pull-cri-containerd-verify test bucket

fuweid:
ping @brianpursley the CI error is not related to your change. please rebase upstream and push it again~

brianpursley:
Thanks, I will rebase tomorrow...

brianpursley:
PR has been rebased

pkg/server/sandbox_stop.go
```diff
@@ -65,7 +65,11 @@ func (c *criService) StopPodSandbox(ctx context.Context, r *runtime.StopPodSandb
 	state := sandbox.Status.Get().State
 	if state == sandboxstore.StateReady || state == sandboxstore.StateUnknown {
 		if err := c.stopSandboxContainer(ctx, sandbox); err != nil {
-			return nil, errors.Wrapf(err, "failed to stop sandbox container %q in %q state", id, state)
+			stateName := "Ready"
```
mikebrow, line 68: suggest adding a String() function for sandbox.Status.State
See cri api for example:
```
type PodSandboxState int32

const (
	PodSandboxState_SANDBOX_READY    PodSandboxState = 0
	PodSandboxState_SANDBOX_NOTREADY PodSandboxState = 1
)

var PodSandboxState_name = map[int32]string{
	0: "SANDBOX_READY",
	1: "SANDBOX_NOTREADY",
}

var PodSandboxState_value = map[string]int32{
	"SANDBOX_READY":    0,
	"SANDBOX_NOTREADY": 1,
}

func (x PodSandboxState) String() string {
	return proto.EnumName(PodSandboxState_name, int32(x))
}
```
If cri wasn't missing the unknown pod state I'd say use that.. but since it's missing we should just create our own I suppose. Or you could do a switch .. either way maybe we should use the name format ^ cri is using.

brianpursley: Great idea! I updated the PR. See if this is what you had in mind.

Just to be clear, what names do you prefer for the states?  I just used the enum names exactly: StateReady, StateNotReady, and StateUnknown, but can use all caps if you prefer. In the example you posted it looks like it left off the prefix part, but the namig convention is a bit odd.

brianpursley: Forgot to gofmt, pushed again.

brianpursley: I reread your comment... it sounds like you like the all upper case names SANDBOX_READY and SANDBOX_NOTREADY, and add one called SANDBOX_UNKNOWN?

mikebrow: yes.. let's use their existing naming convention to reduce confusion and give them something to google :-)

pkg/store/sandbox/status.go
```diff
@@ -75,6 +76,20 @@ const (
...
+func (s State) String() string {
+	switch s {
+	case StateReady:
+		return "StateReady"
+	case StateNotReady:
+		return "StateNotReady"
+	case StateUnknown:
+		return "StateUnknown"
+	default:
+		return strconv.Itoa(int(s))
```
mikebrow, line 89: +"invalid state"

mikebrow:
Doesn't like the ordering of the imports..

```
I0627 20:14:41.410] pkg/store/sandbox/status.go:20: File is not `goimports`-ed (goimports)
151
I0627 20:14:41.410] 	"k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
152
W0627 20:14:49.936] make: *** [Makefile:60: lint] Error 1
```

brianpursley:
Whoops, should have done the verify before pushing... will fix it

brianpursley:
fixed imports

mikebrow:
Thank you!

brianpursley:
Thanks for your help and patience @mikebrow :+1:

PR #1516, opened 2020-06-24
ignore image_pull.go SA1019 tlsConfig.BuildNameToCertificate is deprecated

mikebrow:
Resolves:

> Got an error that seems unrelated to my changes:
> 
> ```
> I0624 07:55:20.688] pkg/server/image_pull.go:287:3: SA1019: tlsConfig.BuildNameToCertificate is deprecated: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (staticcheck)
> I0624 07:55:20.688] 		tlsConfig.BuildNameToCertificate()
> I0624 07:55:20.688] 		^
> ```

let's see... BuildNameToCertificate seems to have been deprecated in go 1.14.. and must've been cherry picked in the linter we are using so..

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -284,7 +284,7 @@ func (c *criService) getTLSConfig(registryTLSConfig criconfig.TLSConfig) (*tls.C
 		if len(cert.Certificate) != 0 {
 			tlsConfig.Certificates = []tls.Certificate{cert}
 		}
-		tlsConfig.BuildNameToCertificate()
+		tlsConfig.BuildNameToCertificate() // nolint:staticcheck
 	}
 
 	if registryTLSConfig.CAFile != "" {


thaJeztah:
Curious; Is there an alternative that would work without ignoring?

mikebrow:
> Curious; Is there an alternative that would work without ignoring?

Nothing that I found.  Field NameToCertificate set by BuildNameToCertificate() was deprecated without replacement.

```
// NameToCertificate maps from a certificate name to an element of
// Certificates. Note that a certificate name can be of the form
// '*.example.com' and so doesn't have to be a domain name as such.
//
// Deprecated: NameToCertificate only allows associating a single
// certificate with a given name. Leave this field nil to let the library
// select the first compatible chain from Certificates.
NameToCertificate map[string]*Certificate
```
IOW it only remembers one name to certificate in the NameToCertificate map.. if you have two.. you get the last one..

thaJeztah:
ugh. I hate deprecated without clear replacement.

not at my keyboard, but will have a look if the commit history provides more context

mikebrow:
> ugh. I hate deprecated without clear replacement. not at my keyboard, but will have a look if the commit history provides more context
> […](#)

yeah have looked through the code.. now.  If you don't set NameToCertificate... the library will just loop through all certificates looking for a match, vs doing the match based on the one you set in BuildNameToCertificate() 

Not sure that is better or worse and would be a behavior change.

mikebrow:
Not sure what broke the k8s e2e bucket.. sigh :-)

mikebrow:
ok fix for node e2e bucket here: 
https://github.com/containerd/cri/pull/1517

mikebrow:
will just merge this over to the e2e bucket...

PR #1519, opened 2020-07-01
config: TolerateMissingHugePagesCgroupController -> TolerateMissingHugetlbController

AkihiroSuda:
ref: https://github.com/containerd/cri/pull/1519#issuecomment-652331640

Also, the TOML tag didn't correspond with the Go struct name and the JSON tag.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -235,7 +235,7 @@ type PluginConfig struct {
 	// TolerateMissingHugePagesCgroupController if set to false will error out on create/update
 	// container requests with huge page limits if the cgroup controller for hugepages is not present.
 	// This helps with supporting Kubernetes <=1.18 out of the box. (default is `true`)
-	TolerateMissingHugePagesCgroupController bool `toml:"tolerate_missing_hugepages_controller" json:"tolerateMissingHugePagesCgroupController"`
+	TolerateMissingHugePagesCgroupController bool `toml:"tolerate_missing_hugepages_cgroup_controller" json:"tolerateMissingHugePagesCgroupController"`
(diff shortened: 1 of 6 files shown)


AkihiroSuda:
@dims PTAL

pkg/config/config.go
```diff
@@ -235,7 +235,7 @@ type PluginConfig struct {
 	// TolerateMissingHugePagesCgroupController if set to false will error out on create/update
 	// container requests with huge page limits if the cgroup controller for hugepages is not present.
 	// This helps with supporting Kubernetes <=1.18 out of the box. (default is `true`)
-	TolerateMissingHugePagesCgroupController bool `toml:"tolerate_missing_hugepages_controller" json:"tolerateMissingHugePagesCgroupController"`
+	TolerateMissingHugePagesCgroupController bool `toml:"tolerate_missing_hugepages_cgroup_controller" json:"tolerateMissingHugePagesCgroupController"`
```
AkihiroSuda, line 238: Another options is to match the go struct field name and the JSON tag to the previous TOML tag.

dims:
@AkihiroSuda looks like it's officially called "HugeTLB Controller" in Kernel documentation?

I am fine either way @AkihiroSuda :) if i remember right, i was just trying to reduce the length of the key.

[1] https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/hugetlb.html

pkg/config/config.go
mikebrow, line 238: missing from the docs... here:
https://github.com/containerd/cri/blame/master/docs/config.md#L89

AkihiroSuda:
Updated to use the short name and also added the doc

PR #1520, opened 2020-07-01
CI: run apt-get update befor apt-get install

AkihiroSuda:
`apt-get update` is always required before `apt-get install` to avoid 404 errors


dims:
LGTM

PR #1523, opened 2020-07-02
Update to runc v1.0.0 rc91

dims:
Typically, the runc changes happen in containerd/containerd first and then rolled into containerd/cri. Unfortunately we have chicken-egg scenario first identified here: https://github.com/containerd/containerd/pull/4314#issue-431547031

So we need to update containerd/cri first.


AkihiroSuda:
Duplicate https://github.com/containerd/cri/pull/1521

PR #1527, opened 2020-07-07
[1.3 - DO NOT MERGE] User namespace support

mauriciovasquezbernal:
This is the containerd/cri implementation for the [Kubernetes Node-Level User Namespaces Design Proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-usernamespace-remapping.md). The patches are based on the `release/1.3` branch. It is tested on Kubernetes 1.17 with patches adapted from PR 64005 (https://github.com/kinvolk/kubernetes/pull/4).

The purpose of this PR is to gather some early feedback from the community on this feature to start the discussion again. This PR is based on the `release/1.3` branch, we don't intend to merge it as it is so it should not be a problem. We are planning to create / update a KEP to have a proper discussion about the design of this feature.

The main changes are:
- Extend the configuration with uid/guid mappings.
- Import the new CRI API from Kubernetes PR 64005.
- Implement `GetRuntimeConfigInfo` returning the configured mappings.
- Use the `WithRemappedSnapshot` snapshotter.
- Fix sysfs mount with correct ownership of netns (see commitmsg for details).
- Additional mount restrictions on /dev/shm (nosuid, noexec, nodev).
- Chown /dev/shm appropriately.
- Fix etc-hosts mounts with supplementary groups (see commitmsg for details).
- Use custom options `WithoutNamespace`, `WithUserNamespace`, `WithLinuxNamespace` at the right places.

At the OCI level (config.json), we have the following changes:
- sandbox container with **new** "user" namespace and **UidMappings**
- normal containers with "user" namespace from a **path** and **UidMappings**

Demo:

Minimal example of the containerd configuration file:
```
version = 2
[plugins]
  [plugins."io.containerd.grpc.v1.cri"]
    [plugins."io.containerd.grpc.v1.cri".node_wide_uid_mapping]
      container_id = 0
      host_id = 100000
      size = 65536
    [plugins."io.containerd.grpc.v1.cri".node_wide_gid_mapping]
      container_id = 0
      host_id = 100000
      size = 65536
```

```
$ kubectl apply -f userns-tests/node-standard.yaml
pod/pod-simple created
$ kubectl apply -f userns-tests/pod-standard.yaml
pod/pod-userns created
$ kubectl exec -ti node-standard -- /bin/sh -c 'cat /proc/self/uid_map'
         0          0 4294967295
$ kubectl exec -ti pod-standard -- /bin/sh -c 'cat /proc/self/uid_map'
         0     100000      65535
```

<details>

* userns-tests/pod-standard.yaml

```
# Pod user with userns set to "node" mode.
apiVersion: v1
kind: Pod
metadata:
  name: node-standard
  namespace: default
  annotations:
    alpha.kinvolk.io/userns: "node"
spec:
  restartPolicy: Never
  containers:
  - name: container1
    image: busybox
    command: ["sh"]
    args: ["-c", "sleep infinity"]
```

* userns-tests/pod-standard.yaml:

```
# Pod user with userns set to "pod" mode. Pod creation would fail if userns
# not supported by runtime.
apiVersion: v1
kind: Pod
metadata:
  name: pod-standard
  namespace: default
  annotations:
    alpha.kinvolk.io/userns: "pod"
spec:
  restartPolicy: Never
  containers:
  - name: container1
    image: busybox
    command: ["sh"]
    args: ["-c", "sleep infinity"]

```
</details>

/cc @alban @rata


mikebrow:
>We don't intend to merge it as it is (that's the reason we used release/1.3 as the base) 

? we would not merge a PR that is in [wip] or hold against the contributor's wishes...

mauriciovasquezbernal:
> ? we would not merge a PR that is in [wip]/hold against the contributor's wishes...

I don't mean that we based it on release/1.3 to avoid merging. What I wanted to say is that it's not a problem to be based on release/1.3 because we don't want to merge it...

zhsj:
It is same as https://github.com/rootless-containers/usernetes?

And maybe https://github.com/kubernetes/enhancements/pull/1371 should replace [Kubernetes Node-Level User Namespaces Design Proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-usernamespace-remapping.md)

alban:
@zhsj It is not the same: while both use user namespaces as underlying technology, they achieve different things:
- usernetes uses unprivileged user namespaces to be able to run container runtimes without being root on the host. All containers run as the same user.
- this user namespace support PR allows containers to run in a user namespace (and possibly soon with different id mappings), hardening the isolation between containers. This does not allow by itself to run kubelet/containerd/runc as non-root.

Both are useful, depending on the use cases.

AkihiroSuda:
What's current status?

mauriciovasquezbernal:
> What's current status?

We're waiting reviews on the [Kubernetes KEP](https://github.com/kubernetes/enhancements/pull/2101) in order to define the CRI changes.

AkihiroSuda:
Can we implement  [the "io.kubernetes.cri-o.userns-mode" convention](https://github.com/cri-o/cri-o/pull/3944) until the KEP settles, or do we want to wait until KEP settles?

rptaylor:
Will this allow a user to start an unprivileged kubernetes pod, and then use e.g. Singularity inside their pod to create a nested container?

mauriciovasquezbernal:
> Can we implement [the "io.kubernetes.cri-o.userns-mode" convention](https://github.com/cri-o/cri-o/pull/3944) until the KEP settles, or do we want to wait until KEP settles?

I'd like to get the KEP merged first and then implement this support in containerd. I fear that we would have a lot of heterogeneous implementations if we start implementing this support in the different runtimes without having a proper specification to do it. I also agree that people needing to have this support quickly could implement such mechanism to avoid waiting on Kubernetes on the meanwhile.

mauriciovasquezbernal:
> Will this allow a user to start an unprivileged kubernetes pod, and then use e.g. Singularity inside their pod to create a nested container?

Yes, that should work. User namespaces used in this context allow to run pods that require elevated privileges in a safer way. If you give the pod the correct capabilities to create containers and set correct k8s settings like https://kubernetes.io/docs/concepts/policy/pod-security-policy/#allowedprocmounttypes that should be fine. However it's something we haven't tried and there could be unknown issues.

PR #1528, opened 2020-07-08
Update to go 1.13 - experiment with go 1.15beta1

mikebrow:
- first commit modifies to test on latest version of go1.13 as it moves along
  - also making a note in the readme that we will not be supporting go 1.14
- second commit adds experimental testing for go1.15beta1 (skipping critest for now for go1.15beta1 will push a PR later to fix critest on 1.15)


mikebrow:
odd it reported pass in my repo.. reports fail here..

PR #1529, opened 2020-07-09
allow disabling hugepages

AkihiroSuda:
This helps with running rootless mode + cgroup v2 + systemd without hugetlb delegation.
Systemd does not (and will not, perhaps) support hugetlb delegation as of systemd v245. https://github.com/systemd/systemd/issues/14662

From https://github.com/rootless-containers/usernetes/blob/502bc5427e57236382db58a0af7996a145381803/src/patches/containerd/0001-DIRTY-VENDOR-cri-allow-disabling-hugepages.patch

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -236,6 +236,10 @@ type PluginConfig struct {
 	// container requests with huge page limits if the cgroup controller for hugepages is not present.
 	// This helps with supporting Kubernetes <=1.18 out of the box. (default is `true`)
 	TolerateMissingHugetlbController bool `toml:"tolerate_missing_hugetlb_controller" json:"tolerateMissingHugetlbController"`
+	// DisableHugetlbController indicates to silently disable the hugetlb controller, even when it is
+	// present in /sys/fs/cgroup/cgroup.controllers.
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -236,6 +236,10 @@ type PluginConfig struct {
 	// container requests with huge page limits if the cgroup controller for hugepages is not present.
 	// This helps with supporting Kubernetes <=1.18 out of the box. (default is `true`)
 	TolerateMissingHugetlbController bool `toml:"tolerate_missing_hugetlb_controller" json:"tolerateMissingHugetlbController"`
+	// DisableHugetlbController indicates to disable the hugetlb controller even when it is
+	// present in /sys/fs/cgroup/cgroup.controllers.
+	// This helps with running rootless mode + cgroup v2 + systemd but without hugetlb delegation.
```
mikebrow, line 241: should mention `silently` here... and / or..  log a `one time?` warning at init or at first collision that support has been disabled by configuration.

AkihiroSuda: Added `silently`. When the config is present, the user wouldn't want the warning to be printed.

mikebrow: it would be confusing... ok.

PR #1531, opened 2020-07-10
show runc options tag

mikebrow:
At least show users of / where runc.options should be added when they run containerd config default:

before
```
      [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
          runtime_type = "io.containerd.runc.v2"
          runtime_engine = ""
          runtime_root = ""
          base_runtime_spec = ""
```
after
```
      [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
          runtime_type = "io.containerd.runc.v2"
          runtime_engine = ""
          runtime_root = ""
          base_runtime_spec = ""
          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
```

### DIFF
--- a/pkg/config/config_unix.go
+++ b/pkg/config/config_unix.go
@@ -19,6 +19,7 @@
 package config
 
 import (
+	"github.com/BurntSushi/toml"
 	"github.com/containerd/containerd"
 	"github.com/containerd/cri/pkg/streaming"
 )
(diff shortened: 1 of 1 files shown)


mikebrow:
> SGTM
> 
> bit surprised that the TOML type is required (and not something more generic), but I'm not familiar with the code

Would be a good idea to consider refactoring how we do config.

PR #1534, opened 2020-07-22
Update selinux pkg and allow category range to be set in config

crosbymichael:
This allows an admin to set the upper bounds on the category range for selinux
labels.  This can be useful when handling allocation of PVs or other volume
types that need to be shared with selinux enabled on the hosts and volumes.


crosbymichael:
Windows is currently failing on other PRs fetching dependencies so this is unrelated to this change.  Verified it builds locally:


```
 GOOS=windows make
+ _output/containerd.exe
go build -o _output/containerd.exe \
        -tags '' \
        -ldflags '-X github.com/containerd/cri/vendor/github.com/containerd/containerd/version.Version=cde78726-TEST' \
        -gcflags '' \
        github.com/containerd/cri/cmd/containerd
+ binaries
```

vendor.conf
```diff
@@ -1,6 +1,7 @@
 # cri dependencies
 github.com/docker/docker                            4634ce647cf2ce2c6031129ccd109e557244986f
-github.com/opencontainers/selinux                   v1.5.1
+github.com/opencontainers/selinux                   v1.6.0
+github.com/willf/bitset								d5bec3311243426a3c6d1b7a795f24b17c686dbb # 1.1.10+ used by selinux pkg
```
dmcgowan, line 4: This looks like it is using tabs where the rest of the file separates with spaces. This should also sort below `tchap`

mikebrow:
will need to rebase after #1538 merges

crosbymichael:
@mikebrow @dmcgowan updated and rebased

PR #1535, opened 2020-07-22
Update doc cri plugin convention

lumjjb:
Signed-off-by: Brandon Lum <lumjjb@gmail.com>


crosbymichael:
LGTM

I don't think this needs integration tests to verify.  Also windows is currently failing on other PRs so this is un-related.

PR #1543, opened 2020-07-27
Allow GC to discard content after successful pull and unpack

ktock:
Related: https://github.com/containerd/containerd/issues/4304 https://github.com/containerd/containerd/pull/4414

This commit adds a config flag for allowing GC to clean layer contents up after unpacking these contents completed, which leads to deduplication of layer contents between the snapshotter and the content store.
This PR includes vendoring the latest containerd which includes the commit https://github.com/containerd/containerd/commit/d184a0a3430dc4a17a47cce37fb36126ac0c699a . (full diff: https://github.com/containerd/containerd/compare/v1.4.0-beta.0...d184a0a3430dc4a17a47cce37fb36126ac0c699a )

One thing I'm considering is the default value of this flag. In this patch, this "DiscardContent mode" is disabled by default but please tell me if it should be enabled by default.


pkg/config/config.go
```diff
@@ -80,6 +80,11 @@ type ContainerdConfig struct {
 	// related information) to snapshotters. These annotations are required by
 	// stargz snapshotter (https://github.com/containerd/stargz-snapshotter).
 	DisableSnapshotAnnotations bool `toml:"disable_snapshot_annotations" json:"disableSnapshotAnnotations"`
+
+	// DiscardContent is a boolean flag to specify whether to allow GC to clean
+	// layers up from the content store after successfully unpacking these
+	// contents to the snapshotter.
```
mikebrow, line 86: please add text for discard_content (and disable_snapshot_annotations while at it :-) to https://github.com/containerd/cri/blame/master/docs/config.md#L108

pkg/config/config.go
```diff
@@ -80,6 +80,11 @@ type ContainerdConfig struct {
 	// related information) to snapshotters. These annotations are required by
 	// stargz snapshotter (https://github.com/containerd/stargz-snapshotter).
 	DisableSnapshotAnnotations bool `toml:"disable_snapshot_annotations" json:"disableSnapshotAnnotations"`
+
+	// DiscardContent is a boolean flag to specify whether to allow GC to clean
```
mikebrow, line 84: suggest DiscardUnpackedLayers or DiscardUnpackedContainerImageLayers

mikebrow: s/to clean/to remove/

pkg/config/config.go
mikebrow: please indicate the default is false, layers are not removed from the content store during garbage collection.

mikebrow, line 86: s/contents/layers/

ktock:
@mikebrow Thanks for comments. Fixed the patch based on the comments. PTAL.

PR #1548, opened 2020-07-30
remove libseccomp cgo dependency

AkihiroSuda:
The CRI plugin was depending on libseccomp cgo dependency via libseccomp-golang via libcontainer.

https://github.com/seccomp/libseccomp-golang/blob/v0.9.1/seccomp_internal.go#L17


dims:
good one!

LGTM

PR #1551, opened 2020-08-05
vendor: kubernetes 1.19.0 rc.4 and dependencies

thaJeztah:


vendor.conf
```diff
@@ -51,9 +51,9 @@ github.com/syndtr/gocapability                      d98352740cb2c55f81556b63d4a1
 github.com/urfave/cli                               v1.22.1 # NOTE: urfave/cli must be <= v1.22.1 due to a regression: https://github.com/urfave/cli/issues/1092
 go.etcd.io/bbolt                                    v1.3.5
 go.opencensus.io                                    v0.22.0
-golang.org/x/net                                    f3200d17e092c607f615320ecaad13d87ad9a2b3
+golang.org/x/net                                    ab34263943818b32f575efc978a3d24e80b04bd7
 golang.org/x/sync                                   42b317875d0fa942474b76e1b46a6060d720ae6e
-golang.org/x/sys                                    9dae0f8f577553e0f21298e18926efc9644c281d
+golang.org/x/sys                                    ed371f2e16b4b305ee99df548828de367527b76b
```
thaJeztah, line 56: Note that these are in the "containerd dependencies" section. Wondering if they should stay in that section (and thus need to be updated there first), or if we need some "generic" section, as these are used in various places, not just containerd

vendor/golang.org/x/net/websocket/websocket.go
```diff
@@ -5,11 +5,11 @@
...
 // as specified in RFC 6455.
 //
-// This package currently lacks some features found in an alternative
-// and more actively maintained WebSocket package:
+// This package currently lacks some features found in alternative
+// and more actively maintained WebSocket packages:
 //
 //     https://godoc.org/github.com/gorilla/websocket
-//
+//     https://godoc.org/nhooyr.io/websocket
```
thaJeztah, line 12: I noticed this change in the comment, which relates to https://github.com/golang/go/issues/18152

It's used in k8s.io/apiserver/pkg/util/wsstream/stream.go, so perhaps that should be reviewed if the alternatives are better for how it's used.

vendor/k8s.io/apimachinery/go.mod
```diff
@@ -18,25 +18,22 @@ require (
...
+	golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4 // indirect
+	golang.org/x/text v0.3.3 // indirect
+	google.golang.org/protobuf v1.24.0 // indirect
 	gopkg.in/inf.v0 v0.9.1
 	gopkg.in/yaml.v2 v2.2.8
-	k8s.io/klog/v2 v2.1.0
+	k8s.io/klog/v2 v2.2.0
 	k8s.io/kube-openapi v0.0.0-20200427153329-656914f816f9
-	sigs.k8s.io/structured-merge-diff/v3 v3.0.0
+	sigs.k8s.io/structured-merge-diff/v3 v3.0.1-0.20200706213357-43c19bbb7fba
```
thaJeztah, line 37: I skipped the update to sigs.k8s.io/structured-merge-diff, as it would bring in a non-released version, and didn't bring changes in the vendored files

vendor/k8s.io/apimachinery/go.mod
```diff
@@ -2,13 +2,13 @@
...
 
-go 1.13
+go 1.15
 
 require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96
 	github.com/elazarl/goproxy v0.0.0-20180725130230-947c36da3153
-	github.com/evanphx/json-patch v4.2.0+incompatible
+	github.com/evanphx/json-patch v0.0.0-20190815234213-e83c0a1c26c8
```
thaJeztah, line 11: Perhaps something for @dims - I noticed kubernetes switched to this version to get a fix/feature  in. This commit is somewhere between v4.5.0 and v4.6.0:
https://github.com/evanphx/json-patch/compare/e83c0a1c26c884f8ec4fb743f6db92a941f605f1...v4.6.0

There's also a v4.7.0 release, that fixes some bugs:
https://github.com/evanphx/json-patch/compare/v4.6.0...v4.7.0


So, perhaps k/k can update to that version to get back on a released version of the dependency. I can try opening a PR to k/k (not sure if there's special magic needed to update dependencies there 😅), or perhaps someone else has time to do so

dims: @thaJeztah we are a bit sensitive to json-patch historically i think. Will run it by @liggitt. It is ok for containerd to switch first. We will just pick it up when we rev to containerd 1.4.0

liggitt: between v4.5.0 and v4.6.0, json-patch did the following things in this order:

https://github.com/evanphx/json-patch/compare/v4.5.0...v4.6.0

1. added a go.mod file which made semver revisions unaddressable until they renamed to `github.com/evanphx/json-patch/v5` (done in v5.0.0)
2. fixed an issue with significant performance implications
3. changed behavior to be RFC-compliant, but which is potentially breaking for API consumers

We are not yet comfortable picking up the behavior-breaking changes (especially not to backport to a patch release, see discussion in https://github.com/kubernetes/kubernetes/pull/91622#issuecomment-642781591), but did want the bugfix.

vendor.conf
mikebrow: we'll be merging the cri repo soon into containerd .. agree once something has two major dependencies it's time to move it to a more generic list

vendor/k8s.io/apimachinery/go.mod
thaJeztah: > It is ok for containerd to switch first. We will just pick it up when we rev to containerd 1.4.0

Heh, well, containerd (nor containerd/cri) currently vendor it / depend on it (it was actually a facepalm moment that I looked at the versions, then realised it's not used here) 😂 

> changed behavior to be RFC-compliant, but which is potentially breaking for API consumers

> We are not yet comfortable picking up the behavior-breaking changes (especially not to backport to a patch release, see discussion in https://github.com/kubernetes/kubernetes/pull/91622#issuecomment-642781591, but did want the bugfix.

Ahm, gotcha, I thought it was updated to get the fix, and this commit was "latest" at that point, and it wasn't updated after.

> We are not yet comfortable picking up the behavior-breaking changes

That makes perfect sense. So looking at the diff, it looks like https://github.com/evanphx/json-patch/commit/a5df74aab0efb032ca893c048330abcad0b78e6f is the offending commit. This actually sounds like the exact scenario for which go modules were designed. Given that this is a (potentially) breaking change, that warrants a major version bump, so it could (/should) be considered a mistake that it was merged in a minor release.

Perhaps we should suggest a revert of that commit, and a v4.8.0 tag to be created. Given that they have a v5.x version now, this gives consumers of the package the opportunity to opt-in to the new behavior when ready, and others to stay on v4.x until they're ready for the change.

I guess Kubernetes is the most prominent consumer of that package (https://grep.app/search?q=github.com/evanphx/json-patch), and still is on v4 (kubernetes-sigs/kind looks to have made the switch to v5, so should not be affected by a revert https://grep.app/search?q=github.com/evanphx/json-patch/v5)

thaJeztah: I went ahead and opened a pull request for consideration; https://github.com/evanphx/json-patch/pull/111

liggitt: The tagged v4.8.0 is functionally correct with no breaking changes, but still contains a go.mod file requiring it to be addressed as `v0.0.0-20200808040245-162e5629780b` in go.mod files

Opened https://github.com/evanphx/json-patch/pull/113 to drop the go.mod file in the 4.x stream for tidyness, though it would only affect usage by go module-based consumers.

thaJeztah:
Ah, yes, I just was fiddling with that, trying to update the dependency in kubernetes; looks like go1.13+ no longer allows that, and refuses to get the dependency.

tbh these breaking changes in Golang modules are kinda infuriating (imo even should warrant a major version bump of Golang itself)

liggitt:
I agree it is difficult... we accidentally tagged a version of client-go with a go.mod file as v12.0.0 not realizing that was an invalid tag.

It was officially documented as disallowed, but was not enforced until go1.13. The backwards compatibility scenarios go modules added (e.g. `v4.6.0+incompatible` to indicate `v4.6.0 but without a go.mod file or /v4 suffix`) made it really hard to understand when you were doing things correctly, and when you were doing things incorrectly and go modules were masking the issue for compatibility purposes.

AkihiroSuda:
Is this ready to be merged?

The json-patch issue doesn't seem directly related to containerd.

thaJeztah:
> The json-patch issue doesn't seem directly related to containerd.

Correct; I just noticed it in the diff, but not related to this repository (or this PR), so this should be ready for review

dims:
+1 LGTM

vendor/k8s.io/apimachinery/go.mod
liggitt: btw, json-patch v4.9.0 was just released that once again allows addressing as `v4.9.0+incompatible` via go modules

thaJeztah: Perfect! Thanks for the update

PR #1552, opened 2020-08-05
Add experimental NRI injection points

crosbymichael:
This allows development with container to be done for NRI without the need for
custom builds.

This is an experimental feature and is not enabled unless a user has a global
`/etc/nri/conf.json` config setup with plugins on the system.  No NRI code will
be executed if this config file does not exist.

### DIFF
--- a/pkg/server/container_start.go
+++ b/pkg/server/container_start.go
@@ -24,6 +24,8 @@ import (
 	containerdio "github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/log"
+	"github.com/containerd/nri"
+	v1 "github.com/containerd/nri/types/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
@@ -107,7 +109,7 @@ func (c *criService) StartContainer(ctx context.Context, r *runtime.StartContain
 			deferCtx, deferCancel := ctrdutil.DeferContext()
 			defer deferCancel()
 			// It's possible that task is deleted by event monitor.
-			if _, err := task.Delete(deferCtx, containerd.WithProcessKill); err != nil && !errdefs.IsNotFound(err) {
+			if _, err := task.Delete(deferCtx, WithNRISandboxDelete(sandboxID), containerd.WithProcessKill); err != nil && !errdefs.IsNotFound(err) {
 				log.G(ctx).WithError(err).Errorf("Failed to delete containerd task %q", id)
 			}
 		}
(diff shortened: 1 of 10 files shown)


pkg/server/events.go
```diff
@@ -296,6 +299,14 @@ func (em *eventMonitor) handleEvent(any interface{}) error {
...
 // handleContainerExit handles TaskExit event for container.
 func handleContainerExit(ctx context.Context, e *eventtypes.TaskExit, cntr containerstore.Container) error {
+	nric, err := nri.New()
+	if err != nil {
+		log.G(ctx).WithError(err).Error("unable to create nri client")
+	}
+	nriSB := &nri.Sandbox{
+		ID: cntr.SandboxID,
+	}
+
```
mikebrow, line 309: suggest a sandbox helper to do this block and the invoke

crosbymichael: What would a helper look like?

pkg/server/events.go
```diff
@@ -316,6 +327,12 @@ func handleContainerExit(ctx context.Context, e *eventtypes.TaskExit, cntr conta
 			return errors.Wrapf(err, "failed to load task for container")
 		}
 	} else {
+		if nric != nil {
+			if _, err := nric.InvokeWithSandbox(ctx, task, v1.Delete, nriSB); err != nil {
+				log.G(ctx).WithError(err).Errorf("Failed to delete nri for %q", cntr.ID)
+			}
+		}
+
```
mikebrow, line 335: instead of these code blocks what do you think of adding
```
containerd.With_NRI_Invoke_SB(
   &nri.Sandbox{ID: cntr.SandboxID,}..., 
```
to task.Delete?

pkg/server/container_start.go
```diff
@@ -119,6 +138,12 @@ func (c *criService) StartContainer(ctx context.Context, r *runtime.StartContain
 		return nil, errors.Wrap(err, "failed to wait for containerd task")
 	}
 
+	if nric != nil {
+		if _, err := nric.InvokeWithSandbox(ctx, task, v1.Create, nriSB); err != nil {
```
mikebrow, line 142: do you see an equivalence between sandbox and container here.. the naming is confusing given the concept of podsandbox has a different lifecycle to containers in the pod.

crosbymichael: talking with others, they want to make sure that both, the life cycle of sandboxes and containers are handled.

pkg/server/events.go
crosbymichael: ya, makes sense, I can add some opts to handle this more cleanly

pkg/server/container_start.go
mikebrow: fair.. and I definitely agree we need to manage both via. NRI.  We normally abstract away the concept of the pause container of the pod-sandbox as an internal implementation detail. CRUD events for pod would be at a higher level and include different config than it's sandbox pause container lifecycle events. I think we can do both and make sense of it, while allowing for some easy way to ignore pause container events.

crosbymichael: when we look at expanding out the sandbox from a simple pause container into a VM, this becomes more important.

crosbymichael:
I think @dmcgowan would feel better if we wait until 1.4 is out before pushing this forward more

dims:
+1 to wait until 1.5 opens up :)

crosbymichael:
Now that 1.4 is out, do we want to open this back up for review?

mikebrow:
> Now that 1.4 is out, do we want to open this back up for review?

sure ..

pkg/server/container_start.go
mikebrow: ok.. let's experiment away..

AkihiroSuda:
looks good but needs rebase

PR #1561, opened 2020-08-27
Handle KVM based runtimes with selinux

crosbymichael:
This handles kvm based runtimes for new and old versions of the selinux packages.  If the `container_kvm_t` label exists, then we use them, if not, we clear the process label. 

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -161,6 +161,12 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	}
 
 	meta.ProcessLabel = spec.Process.SelinuxLabel
+
+	// handle any KVM based runtime
+	if err := modifyProcessLabel(ociRuntime.Type, spec); err != nil {
+		return nil, err
+	}
+
 	if config.GetLinux().GetSecurityContext().GetPrivileged() {
 		// If privileged don't set the SELinux label but still record it on the container so
 		// the unused MCS label can be release later
--- a/pkg/seutil/seutil.go
+++ b/pkg/seutil/seutil.go
@@ -1,0 +1,71 @@
+/*
+   Copyright The containerd Authors.
+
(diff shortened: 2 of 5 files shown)


egernst:
Thanks @crosbymichael

Heads up @amshinde

pkg/seutil/seutil.go
```diff
@@ -0,0 +1,71 @@
...
+const typePath = "/etc/selinux/targeted/contexts/customizable_types"
+
+func init() {
+	seTypes = make(map[string]struct{})
+	if !selinux.GetEnabled() {
+		return
+	}
+	f, err := os.Open(typePath)
+	if err != nil {
+		return
```
dims, line 37: Do we want to log something?

crosbymichael: not in an init func

pkg/server/container_create.go
```diff
@@ -161,6 +161,12 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	}
 
 	meta.ProcessLabel = spec.Process.SelinuxLabel
+
+	// handle any KVM based runtime
+	if err := modifyProcessLabel(ociRuntime.Type, spec); err != nil {
+		return nil, err
```
mikebrow, line 167: is there an intent to fail here if there is a spec.Process.SelinuxLabel but selinux is not enabled? skimming over the code looks like there may be a case where ChangeToKVM could bubble up an error when we might be ok ignoring?

crosbymichael: let me double check

crosbymichael: I think this would be an error anyways but I'm updating the code to handle this case.

crosbymichael:
Humm, weird error coming from e2e.  I'll look into it, does not look related but who knows when you are dealing with selinux

dims:
LGTM

PR #1567, opened 2020-09-08
add oci.WithAllDevicesAllowed regardless of ociRuntime.PrivilegedWithoutHostDevices

### LINKED ISSUE #1566: privileged_without_host_devices=true prevents privileged containers with DIND
# Problem

I've noticed that when using `privileged_without_host_devices = true` and attempting to run a DIND (docker in docker) container inside a Kubernetes pod, it prevents the creation of privileged containers on the inside docker daemon.

We see the following error from the DIND daemon:

```
/ # docker run -it --rm --privileged ubuntu:18.04 bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:415: setting cgroup config for procHooks process caused \\\"failed to write \\\\\\\"a *:* rwm\\\\\\\" to \\\\\\\"/sys/fs/cgroup/devices/docker/d42eaf428371bbfbe053a73dcc5d8878e32e65a6ab346d6923cdac4e764685d4/devices.allow\\\\\\\": write /sys/fs/cgroup/devices/docker/d42eaf428371bbfbe053a73dcc5d8878e32e65a6ab346d6923cdac4e764685d4/devices.allow: operation not permitted\\\"\"": unknown.
```

This is due to the privileged Kubernetes container not having `oci.WithAllDevicesAllowed`:

`privileged=true` and `privileged_without_host_devices=true`:
```
/ # cat /sys/fs/cgroup/devices/devices.list
c *:* m
b *:* m
c 1:3 rwm
c 1:8 rwm
c 1:7 rwm
c 5:0 rwm
c 1:5 rwm
c 1:9 rwm
c 5:1 rwm
c 136:* rwm
c 5:2 rwm
c 10:200 rwm
```
versus

`privileged=true` and `privileged_without_host_devices=false`:
```
/ # cat /sys/fs/cgroup/devices/devices.list
a *:* rwm
```

This can be reproduced with the following pod spec:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: dind-test
  namespace: default
spec:
  containers:
    - name: client
      image: docker:stable-dind
      command:
        - sleep
        - "86400"
      env:
        - name: DOCKER_HOST
          value: tcp://127.0.0.1:2375
    - name: dind
      image: docker:stable-dind
      env:
        - name: DOCKER_TLS_CERTDIR
      securityContext:
        privileged: true
```
and then exec-ing into the client container:
```
$ kubectl exec -it test -c client -- sh
/ # docker run -it --rm --privileged ubuntu:18.04 bash
Unable to find image 'ubuntu:18.04' locally
18.04: Pulling from library/ubuntu
f08d8e2a3ba1: Pull complete 
3baa9cb2483b: Pull complete 
94e5ff4c0b15: Pull complete 
1860925334f9: Pull complete 
Digest: sha256:05a58ded9a2c792598e8f4aa8ffe300318eac6f294bf4f49a7abae7544918592
Status: Downloaded newer image for ubuntu:18.04
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:415: setting cgroup config for procHooks process caused \\\"failed to write \\\\\\\"a *:* rwm\\\\\\\" to \\\\\\\"/sys/fs/cgroup/devices/docker/47d81d930f0c55faa6e5551b4012eeb2ec14469592d4a93bdf2c2e0f916a6eaf/devices.allow\\\\\\\": write /sys/fs/cgroup/devices/docker/47d81d930f0c55faa6e5551b4012eeb2ec14469592d4a93bdf2c2e0f916a6eaf/devices.allow: operation not permitted\\\"\"": unknown.
/ # 
```

# Workaround

A workaround is to add a lifecycle to the privileged Kubernetes container to update the cgroup:
```
      lifecycle:
        postStart:
          exec:
            command:
              - /bin/sh
              - -c
              - echo "a *:* rwm" > /sys/fs/cgroup/devices/devices.allow
```
But to me this isn't ideal - it's a differing behaviour compared to that of privileged when `privileged_without_host_devices` is disabled. 

# Proposed Solution

Proposed solution would be to allow `oci.WithAllDevicesAllowed` regardless of whether `privileged_without_host_devices` is true or false. This would retain the behaviour of privileged prior to the introduction of `privileged_without_host_devices`. 

See diff:

```diff
diff --git a/pkg/server/container_create_unix.go b/pkg/server/container_create_unix.go
index 28863cb0..5441d15f 100644
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -193,9 +193,9 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
                if !sandboxConfig.GetLinux().GetSecurityContext().GetPrivileged() {
                        return nil, errors.New("no privileged container allowed in sandbox")
                }
-               specOpts = append(specOpts, oci.WithPrivileged)
+               specOpts = append(specOpts, oci.WithPrivileged, oci.WithAllDevicesAllowed)
                if !ociRuntime.PrivilegedWithoutHostDevices {
-                       specOpts = append(specOpts, oci.WithHostDevices, oci.WithAllDevicesAllowed)
+                       specOpts = append(specOpts, oci.WithHostDevices)
                } else {
                        // add requested devices by the config as host devices are not automatically added
                        specOpts = append(specOpts, customopts.WithDevices(c.os, config), customopts.WithCapabilities(securityContext))
```

awprice:
fixes containerd/containerd#6643

Signed-off-by: Alex Price <aprice@atlassian.com>

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -193,9 +193,9 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		if !sandboxConfig.GetLinux().GetSecurityContext().GetPrivileged() {
 			return nil, errors.New("no privileged container allowed in sandbox")
 		}
-		specOpts = append(specOpts, oci.WithPrivileged)
+		specOpts = append(specOpts, oci.WithPrivileged, oci.WithAllDevicesAllowed)
 		if !ociRuntime.PrivilegedWithoutHostDevices {
-			specOpts = append(specOpts, oci.WithHostDevices, oci.WithAllDevicesAllowed)
+			specOpts = append(specOpts, oci.WithHostDevices)
 		} else {
 			// add requested devices by the config as host devices are not automatically added
 			specOpts = append(specOpts, customopts.WithDevices(c.os, config), customopts.WithCapabilities(securityContext))
--- a/pkg/server/container_create_unix_test.go
+++ b/pkg/server/container_create_unix_test.go
@@ -1161,26 +1161,31 @@ func TestPrivilegedDevices(t *testing.T) {
 		privileged                   bool
 		privilegedWithoutHostDevices bool
 		expectHostDevices            bool
+		expectAllDevicesAllowed      bool
 	}{
 		"expect no host devices when privileged is false": {
 			privileged:                   false,
 			privilegedWithoutHostDevices: false,
 			expectHostDevices:            false,
+			expectAllDevicesAllowed:      false,
 		},
 		"expect no host devices when privileged is false and privilegedWithoutHostDevices is true": {
 			privileged:                   false,
 			privilegedWithoutHostDevices: true,
 			expectHostDevices:            false,
+			expectAllDevicesAllowed:      false,
 		},
 		"expect host devices when privileged is true": {
 			privileged:                   true,
 			privilegedWithoutHostDevices: false,
 			expectHostDevices:            true,
+			expectAllDevicesAllowed:      true,
 		},
 		"expect no host devices when privileged is true and privilegedWithoutHostDevices is true": {
 			privileged:                   true,
 			privilegedWithoutHostDevices: true,
 			expectHostDevices:            false,
+			expectAllDevicesAllowed:      true,
 		},
 	} {
 		t.Logf("TestCase %q", desc)
@@ -1209,6 +1214,10 @@ func TestPrivilegedDevices(t *testing.T) {
 		} else {
 			assert.Empty(t, spec.Linux.Devices)
 		}
+
+		assert.Len(t, spec.Linux.Resources.Devices, 1)
+		assert.Equal(t, spec.Linux.Resources.Devices[0].Allow, test.expectAllDevicesAllowed)
+		assert.Equal(t, spec.Linux.Resources.Devices[0].Access, "rwm")
 	}
 }
 


pkg/server/container_create_unix.go
```diff
@@ -193,9 +193,9 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 		if !sandboxConfig.GetLinux().GetSecurityContext().GetPrivileged() {
 			return nil, errors.New("no privileged container allowed in sandbox")
 		}
-		specOpts = append(specOpts, oci.WithPrivileged)
+		specOpts = append(specOpts, oci.WithPrivileged, oci.WithAllDevicesAllowed)
 		if !ociRuntime.PrivilegedWithoutHostDevices {
```
mikebrow, line 197: let's consider adding a new flag PrivilegedWithoutAllDevicesAllowed default false...

mikebrow (commented):
See comment

pkg/server/container_create_unix.go
awprice: I'm happy to add a new flag, my only concern is what behaviour do we want when both `PrivilegedWithoutHostDevices` and the new flag are true? i.e. what happens to `oci.WithAllDevicesAllowed`?

mikebrow: document that you can't have both.. and we can check for one or the other or neither in the config validate code..

mikebrow:
needs rebase

AkihiroSuda:
Closing, as we now have https://github.com/containerd/containerd/pull/5686

PR #1570, opened 2020-09-09
Set masked and readonly paths based on default Unix spec

### LINKED ISSUE #1569: Default values for masked and readonly paths are not used for container creation
The default values of masked and readonly paths are defined as follows. Those values are used when a sandbox is created.

https://github.com/containerd/cri/blob/35e623e6bf7512e8c82b8ac6052cb1d720189f28/vendor/github.com/containerd/containerd/oci/spec.go#L209-L226

When we create a new container without specifying masked/readonly paths, no masked/readonly paths are prepared at all, instead of preparing the default ones.

### How to reproduce

Using `sandbox-config.json` and `container-config.json` from https://github.com/containerd/cri/blob/master/docs/crictl.md#run-a-pod-sandbox-using-a-config-file

```
crictl run container-config.json sandbox-config.json
crictl ps
crictl exec CONTAINER_ID ls -l /sys/firmware
```

### Expected result
```
total 0
drwxr-xr-x    5 root     root             0 Sep  9 13:51 .
dr-xr-xr-x   13 root     root             0 Sep  9 13:51 ..
```

### Observed result
```
total 0
drwxr-xr-x    5 root     root             0 Sep  9 13:51 .
dr-xr-xr-x   13 root     root             0 Sep  9 13:51 ..
drwxr-xr-x    5 root     root             0 Sep  9 13:53 acpi
drwxr-xr-x    3 root     root             0 Sep  9 13:53 dmi
drwxr-xr-x    8 root     root             0 Sep  9 13:53 memmap
```

### Proposed fix

#1570

yoheiueda:
The default values of masked and readonly paths are defined
in populateDefaultUnixSpec, and are used when a sandbox is
created.  It is not, however, used for new containers.  If
a container definition does not contain a security context
specifying masked/readonly paths, a container created from
it does not have masked and readonly paths.

This patch applies the default values to masked and
readonly paths of a new container, when any specific values
are not specified.

Fixes #1569

Signed-off-by: Yohei Ueda <yohei@jp.ibm.com>

### DIFF
--- a/pkg/server/container_create_unix.go
+++ b/pkg/server/container_create_unix.go
@@ -182,11 +182,15 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 	if !c.config.DisableProcMount {
 		// Apply masked paths if specified.
 		// If the container is privileged, this will be cleared later on.
-		specOpts = append(specOpts, oci.WithMaskedPaths(securityContext.GetMaskedPaths()))
+		if maskedPaths := securityContext.GetMaskedPaths(); maskedPaths != nil {
+			specOpts = append(specOpts, oci.WithMaskedPaths(maskedPaths))
+		}
 
 		// Apply readonly paths if specified.
 		// If the container is privileged, this will be cleared later on.
-		specOpts = append(specOpts, oci.WithReadonlyPaths(securityContext.GetReadonlyPaths()))
+		if readonlyPaths := securityContext.GetReadonlyPaths(); readonlyPaths != nil {
+			specOpts = append(specOpts, oci.WithReadonlyPaths(readonlyPaths))
+		}
 	}
 
 	if securityContext.GetPrivileged() {
--- a/pkg/server/container_create_unix_test.go
+++ b/pkg/server/container_create_unix_test.go
@@ -959,12 +959,12 @@ func TestMaskedAndReadonlyPaths(t *testing.T) {
 			expectedReadonly: defaultSpec.Linux.ReadonlyPaths,
 			privileged:       false,
 		},
-		"should always apply CRI specified paths when disable_proc_mount = false": {
+		"should apply default if not specified when disable_proc_mount = false": {
 			disableProcMount: false,
 			masked:           nil,
 			readonly:         nil,
-			expectedMasked:   nil,
-			expectedReadonly: nil,
+			expectedMasked:   defaultSpec.Linux.MaskedPaths,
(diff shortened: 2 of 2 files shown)


pkg/server/container_create_unix.go
```diff
@@ -182,11 +182,15 @@ func (c *criService) containerSpec(id string, sandboxID string, sandboxPid uint3
 	if !c.config.DisableProcMount {
 		// Apply masked paths if specified.
 		// If the container is privileged, this will be cleared later on.
-		specOpts = append(specOpts, oci.WithMaskedPaths(securityContext.GetMaskedPaths()))
+		if maskedPaths := securityContext.GetMaskedPaths(); maskedPaths != nil {
+			specOpts = append(specOpts, oci.WithMaskedPaths(maskedPaths))
```
thaJeztah, line 186: Wondering if it would make sense to adjust `obi.WithMaskedPaths()` to treat `nil` as `don't modify` https://github.com/containerd/containerd/blob/62083eeed275db8c154dd87a5f84444e97ddb507/oci/spec_opts.go#L931-L934

That would be a change in behaviour for that function of course (and would no longer allow resetting previously set values, although I'm not sure if that's a realistic use). Option was original added as part of https://github.com/containerd/containerd/pull/2269

So let me /cc @justincormack  as well

yoheiueda: `oci.WithMaskedPaths()`  is also used to reset a previously set value here.
https://github.com/containerd/containerd/blob/2991bc3de7fd5438577a14570f0115aff38065fd/oci/spec_opts.go#L1130-L1131

I found only these two occurrences of `oci.WithMaskedPaths()` in conteinerd, so changing the behavior may be acceptable, but I am not sure whether it is used outside of containerd.

thaJeztah: Thanks! I recall a conversation we had on https://github.com/moby/moby/pull/36644 (not sure if it's recorded on GitHub), where we were discussing that `nil` / `unset` meant "keep defaults", and passing an empty slice would mean `no masked paths`.

I agree that changing the `oci.WithMaskedPaths()` behavior is risky; current change looks good for now, but perhaps a follow-up discussion would be good 🤔

mikebrow (approved):
LGTM
Good catch. I believe we have not seen this issue before because the paths are set by default in kubelet since 1.11. 

The example in the cri-tools docs needs more detail :-)

PR #1572, opened 2020-09-10
Limit size of additional annotation for avoiding unpack failure

ktock:
Related: https://github.com/containerd/stargz-snapshotter/issues/144

In containerd, there is a size limit for label size ([4096 chars](https://github.com/containerd/containerd/blob/c862000ab96364d71559afadd82f6d8155dc5463/labels/validate.go)).
If an image has many layers (> (4096-39)/72 > 56), `containerd.io/snapshot/cri.image-layers` will hit the limit of label size and the unpack will fail because the annotation will be passed to the snapshotter as a label.
This commit fixes this by limiting the size of the annotation.

### DIFF
--- a/pkg/server/image_pull.go
+++ b/pkg/server/image_pull.go
@@ -31,6 +31,7 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/errdefs"
 	containerdimages "github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/labels"
 	"github.com/containerd/containerd/log"
 	distribution "github.com/containerd/containerd/reference/docker"
 	"github.com/containerd/containerd/remotes/docker"
@@ -460,7 +461,7 @@ const (
 	targetDigestLabel = "containerd.io/snapshot/cri.layer-digest"
 	// targetImageLayersLabel is a label which contains layer digests contained in
 	// the target image and will be passed to snapshotters for preparing layers in
-	// parallel.
+	// parallel. Skipping some layers is allowed and only affects performance.
 	targetImageLayersLabel = "containerd.io/snapshot/cri.image-layers"
 )
 
@@ -478,15 +479,6 @@ func appendInfoHandlerWrapper(ref string) func(f containerdimages.Handler) conta
 			}
 			switch desc.MediaType {
 			case imagespec.MediaTypeImageManifest, containerdimages.MediaTypeDockerSchema2Manifest:
-				var layers string
-				for _, c := range children {
-					if containerdimages.IsLayerType(c.MediaType) {
-						layers += fmt.Sprintf("%s,", c.Digest.String())
-					}
-				}
-				if len(layers) >= 1 {
-					layers = layers[:len(layers)-1]
-				}
 				for i := range children {
 					c := &children[i]
 					if containerdimages.IsLayerType(c.MediaType) {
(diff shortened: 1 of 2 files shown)


jianzi123:
I wonder if we need to omit some labels or if we need to change the logic of the valid method.

ktock:
This takes the former approach (omitting the contents of the label). The [`Validate`](https://github.com/containerd/containerd/blob/c862000ab96364d71559afadd82f6d8155dc5463/labels/validate.go) method is used globally so changing the logic isn't easy. 
Maybe we should add validation [when inheriting annotations as labels](https://github.com/containerd/containerd/blob/bc4c3813997554d14449b34d336bca2513e84f96/unpacker.go#L127) and filters invalid annotations out. But let's keep it separated from this PR now.

mikebrow (commented):
after deciding what the plan is with respect to what happens if a layer goes over, should we have more labels etc., can you add a test bucket for this to match up expectations with output?

ktock:
@mikebrow Thanks for the comments.

> after deciding what the plan is with respect to what happens if a layer goes over

As mentioned in the above, I think we should add validation [when inheriting annotations as labels](https://github.com/containerd/containerd/blob/bc4c3813997554d14449b34d336bca2513e84f96/unpacker.go#L127) and filter invalid annotations out. How do you think about it? I'll open another PR (on containerd) if we need this.

> can you add a test bucket for this to match up expectations with output?

The label `containerd.io/snapshot/cri.image-layers` is used by stargz snapshotter to resolve layers in an image in parallel and incomplete (even if empty) list is allowed (but it possibly slows down the snapshot preparation). So in this label's case, if the label size of a layer goes over the limit, I think that simply omitting the content of the label is enough. I added a small unit test for this. Please tell me if I'm missing something or should add more tests, etc.

mikebrow:
> @mikebrow Thanks for the comments.
> 
> > after deciding what the plan is with respect to what happens if a layer goes over
> 
> As mentioned in the above, I think we should add validation [when inheriting annotations as labels](https://github.com/containerd/containerd/blob/bc4c3813997554d14449b34d336bca2513e84f96/unpacker.go#L127) and filter invalid annotations out. How do you think about it? I'll open another PR (on containerd) if we need this.
> 
> > can you add a test bucket for this to match up expectations with output?
> 
> The label `containerd.io/snapshot/cri.image-layers` is used by stargz snapshotter to resolve layers in an image in parallel and incomplete (even if empty) list is allowed (but it possibly slows down the snapshot preparation). So in this label's case, if the label size of a layer goes over the limit, I think that simply omitting the content of the label is enough. I added a small unit test for this. Please tell me if I'm missing something or should add more tests, etc.

I see so perf only not critical.  Just a warning or informational message should be fine then. Perhaps on a if first check.

pkg/server/image_pull.go
```diff
@@ -495,11 +487,33 @@ func appendInfoHandlerWrapper(ref string) func(f containerdimages.Handler) conta
...
+	var item string
+	for _, l := range descs {
+		if containerdimages.IsLayerType(l.MediaType) {
+			item = l.Digest.String()
+			if layers != "" {
+				item = "," + item
+			}
+			// This avoids the label hits the size limitation.
+			if err := validate(key, layers+item); err != nil {
+				log.G(ctx).WithError(err).WithField("label", key).Infof("%q is omitted in the layers list", l.Digest.String())
```
mikebrow, line 512: maybe debug?

mikebrow (commented):
looking good!  Just a nit on the log level..

mikebrow (approved):
LGTM
on green

pkg/server/image_pull.go
```diff
@@ -495,11 +487,33 @@ func appendInfoHandlerWrapper(ref string) func(f containerdimages.Handler) conta
 						}
 						c.Annotations[targetRefLabel] = ref
 						c.Annotations[targetDigestLabel] = c.Digest.String()
-						c.Annotations[targetImageLayersLabel] = layers
+						c.Annotations[targetImageLayersLabel] = getLayers(ctx, targetImageLayersLabel, children[i:], labels.Validate)
```
cpuguy83, line 490: This seems like an extra change that doesn't really fix the problem, except maybe now we store less redundant data.

cpuguy83: "This" being changing from all layers to just children.

ktock: Could you elaborate this? Doesn't calling `getLayers` on `children` avoids the problem of hitting the label limit (see L511) ?

cpuguy83: I was under the impression that the limit was per key, not total?

ktock: Currently, the limit is per key. If we need to have the "total" label size limit, what should be the limit (instead of 4096 chars per key)?

cpuguy83: The way we have this now the root gets all the layers anyway, so the way we are doing `children[:i]` does not really fix the label size problem.

It is handled by truncating the values in `getLayers`, however.

So, 1.4.2 changed the behavior even for values that do not hit the limit.

ktock: `getLayers` solves the label limit problem but `children[:i]` isn't for avoiding the label size limit. `containerd.io/snapshot/cri.image-layers` is currently a label for enabling snapshotter to pre-process(resolve) layers related to the target layer in parallel. So this is for optimization.

The reason why I do `children[:i]` here is otherwise `children[0:limit]` is always passed to the snapshotter and it can't get chance to pre-process `children[limit:]` layers.

However I agree with that we should find the better ways instead of `children[:i]`.

aojea:
@mikebrow  can you please take a look at this https://github.com/kubernetes-sigs/kind/issues/2518 and check if there was a regression?

ktock:
@aojea Please see https://github.com/containerd/containerd/pull/6187.

aojea:
> @aojea Please see [containerd/containerd#6187](https://github.com/containerd/containerd/pull/6187).

:heart:

PR #1577, opened 2020-09-15
Make script work for both python2.7 and 3.6

vteratipally:
Signed-off-by: Varsha Teratipally teratipally@google.com

### DIFF
--- a/cluster/gce/configure.sh
+++ b/cluster/gce/configure.sh
@@ -19,6 +19,22 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


vteratipally:
I have tested this by creating a gce instance with config and the containerd and docker services are running properly now.

Random-Liu:
The change is used to fix Kubernetes upstream containerd test.

The script change is GCE only and has been tested. I think it looks good to me.

/lgtm

estesp (approved):
LGTM

PR #1592, opened 2020-10-01
Align lint with containerd

dmcgowan:
Adds some stricter linting with fewer file exceptions

### DIFF
--- a/test/verify.sh
+++ b/test/verify.sh
@@ -26,4 +26,4 @@ cd "${ROOT}"
 
 make install.tools
 make verify
-GOOS=windows make verify
+#GOOS=windows make verify
--- a/test/verify.sh
+++ b/test/verify.sh
@@ -26,4 +26,4 @@ cd "${ROOT}"
 
 make install.tools
 make verify
-#GOOS=windows make verify
+GOOS=windows make verify
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,3 +18,5 @@ run:
   timeout: 3m
   skip-dirs:
     - docs
+  skip-files:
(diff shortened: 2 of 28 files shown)


test/verify.sh
```diff
@@ -26,4 +26,4 @@ cd "${ROOT}"
 
 make install.tools
 make verify
-GOOS=windows make verify
+#GOOS=windows make verify
```
dims, line 29: Did this `#` sneak in or we need this @dmcgowan ?

mikebrow: if we pull from here we may want to add it back in as a separate GOOS=windows ./test/verify.sh call on the pull-cri-containerd-verify ... I don't remember exactly why verify was moved from travis to prow way back when. There may be some dependency at the sig-node test grid for verify testing on windows..

mikebrow: containerd runs verify for each GOOS as a separate bucket.. and I agree we should go that route.  So I'm good with making this change we just need to open a PR to add windows verify on prow.

dmcgowan:
Added windows back, just disabled the integration and ignoring `*.pb.go` files

mikebrow (approved):
LGTM

dims:
LGTM

thanks @dmcgowan

test/verify.sh
mikebrow: fair ^. I think typically the Kubernetes guys like to have verify in pre-submit before running e2e buckets.. may be why they wanted a prow pre-submit. Course we run github actions.. and I don't think we can force that as a pre-submit before the prow bot picks up our `/test` requests.  We should consider if it's ok to run verify twice for this reason..

PR #1593, opened 2020-10-03
Add missing sandbox labels when invoking nri

moolen:
The sandbox labels are missing when the create event is invoked, as far as i can tell this is a bug.
This is a important thing to have in order to be able to differenciate the containers using their pod name, namespace or uid. 

When creating the pod `sandbox` the labels are passed correctly:
https://github.com/containerd/cri/blob/d620c30d7ecbc756f8d818237f9bccbbb7353e9e/pkg/server/sandbox_run.go#L291-L300


mikebrow (approved):
LGTM

crosbymichael:
LGTM

PR #1595, opened 2020-10-06
Fix unix lint

dmcgowan:
Aligns linting targets with containerd

Note: there were many files marked as `_unix.go` which were actually targeting Linux only and not compilable on Linux due to Linux-only dependencies or use of Linux specific syscalls.


mxpv (approved):
LGTM

Zyqsempai (approved):
LGTM

estesp (approved):
LGTM

PR #1598, opened 2020-10-13
Fix comment in RemovePodSandbox

dcantah:
Signed-off-by: Daniel Canter <dcanter@microsoft.com>


dcantah:
@mikebrow CI seems to be failing on something unrelated.

PR #1617, opened 2021-02-08
[release/1.4] further reduce uses of libcontainer and update runc v1.0.0-rc93

zhsj:
Backport https://github.com/containerd/containerd/pull/4717

cc @thaJeztah @AkihiroSuda


vendor.conf
```diff
@@ -47,7 +47,7 @@ github.com/prometheus/procfs                        v0.0.11
 github.com/russross/blackfriday/v2                  v2.0.1
 github.com/shurcooL/sanitized_anchor_name           v1.0.0
 github.com/sirupsen/logrus                          v1.6.0
-github.com/syndtr/gocapability                      d98352740cb2c55f81556b63d4a1ec64c5a319c2
+github.com/syndtr/gocapability                      42c35b4376354fd554efc7ad35e0b7f94e3a0ffb
```
AkihiroSuda, line 50: Bumping up this pkg causes catastrophe: https://github.com/containerd/containerd/pull/5017

So I suggest keeping vendor.conf as-is for v1.4 branch.
The runc binary itself can be bumped up without touching vendor.conf .

AkihiroSuda (changes requested):

vendor.conf
thaJeztah: I agree; if there's no strong reason to update the dependency, we should be conservative and keep it as-is

PR #1622, opened 2021-02-08
deepcopy image config to avoid env lost when append env

yadzhang:
I meet some env lost in our own environment occasionally recently. And this situation always happend when kubelet call CreateContainer for **many pods with the same image  at the same time**. 
I add some logs for generating env in container spec and discovery there are some problems with [container_create_unix.go#L150 ](https://github.com/containerd/cri/blob/release/1.4/pkg/server/container_create_unix.go#L150).

When call CreateContainer with the same image config at nearly the same time,  they get image config from imagestore but the **Envs slice in the image config is pointing to the same object**.  The Envs is [json.Unmarshal](https://github.com/containerd/cri/blob/release/1.4/pkg/store/image/image.go#L143) from the content db, so the cap and len of the Envs slice maybe not equal. So the operation for [appending container env into the image env](https://github.com/containerd/cri/blob/release/1.4/pkg/server/container_create_unix.go#L152) may cause the first few
 container envs of the first container is overlapped by the second container with some probability if the two container has the same image and created at the very near same time.

To resolve this, we can deepcopy image config directly after get image to avoid env or other field conflict.

### DIFF
--- a/pkg/server/container_create.go
+++ b/pkg/server/container_create.go
@@ -101,6 +101,11 @@ func (c *criService) CreateContainer(ctx context.Context, r *runtime.CreateConta
 		return nil, errors.Wrapf(err, "failed to get image from containerd %q", image.ID)
 	}
 
+	imageConfig := imagespec.ImageConfig{}
+	if err := util.DeepCopy(&imageConfig, &image.ImageSpec.Config); err != nil {
+		return nil, errors.Wrapf(err, "failed to clone image spec config %q", image.ID)
+	}
+
 	// Run container using the same runtime with sandbox.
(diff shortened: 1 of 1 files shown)


dims:
@yadzhang for containerd 1.5.x this repo will not be used as the code from here was folded into the main repo ( see https://github.com/containerd/containerd/blob/master/pkg/cri/server/container_create.go#L50 )

dmcgowan:
As @dims said, this branch in this repository is deprecated, please open in https://github.com/containerd/containerd

yadzhang:
Thanks. I raise a issue https://github.com/containerd/containerd/issues/5023 and relative pr is https://github.com/containerd/containerd/pull/5024

PR #1626, opened 2021-02-24
[release/1.4] Fix repeated sending signal

wzshiming:
Cherry pick of containerd/containerd#5021

Since our production environment still uses 1.3, we hope to add this patch


fuweid:
@wzshiming I don't think it is bug. We suggest that only backport bugfix. Thank you

PR #1627, opened 2021-02-24
[release/1.3] Fix repeated sending signal

wzshiming:
Cherry pick of containerd/containerd#5021

Since our production environment still uses 1.3, we hope to add this patch


wzshiming:
https://github.com/containerd/cri/pull/1626#issuecomment-785542433
I'm not sure if this is a bug, do you think it is necessary to cherry-pick it.

PR #1634, opened 2021-04-22
windows: Use GetFinalPathNameByHandle for ResolveSymbolicLink

kevpar:
This change splits the definition of pkg/cri/os.ResolveSymbolicLink by
platform (windows/!windows), and switches to an alternate implementation
for Windows. This change fixes the issue described in containerd/containerd#5405.

The previous implementation which just called filepath.EvalSymlinks has
historically had issues on Windows. One of these issues we were able to
fix in Go, but EvalSymlinks's behavior is not well understood on
Windows, and there could easily be more issues in the future, so it
seems more prudent to move to a separate implementation for Windows.

The new implementation uses the Windows GetFinalPathNameByHandle API,
which takes a handle to an open file or directory and some flags, and
returns the "real" name for the object. We first query for the "DOS"
name, which will return something like "\\?\C:\foo" for a local path, or
"\\?\UNC\server\foo" for a remote share path. This should work for all
cases except where the object is a disk that does not have a drive
letter assigned, in which case we will fallback to querying the
"volume GUID" path (e.g. \\?\Volume{1d7d818b-f43f-4c7a-90b5-68c0a9db2063}\).

The "\\?\" prefix on the paths returned from GetFinalPathNameByHandle
indicates to Windows to pass them through directly to the filesystem,
and therefore allows for some additional paths to be expressed that are
not supported by standard Windows path strings. I have tested this path
format with containerd and have not seen any issues. Functions that make
assumptions on what a Windows path can look like may have trouble with
them, but as ResolveSymbolicLink is only used to resolve mount source
paths, this doesn't seem like it will be an issue.

Also updated the golang.org/x/sys vendor to bring in
windows.GetFinalPathNameByHandle.

Signed-off-by: Kevin Parsons <kevpar@microsoft.com>


AkihiroSuda (changes requested):
Please get your patch merged in the master of `containerd/containerd` and then cherry-pick to v1.4. https://github.com/containerd/containerd/issues/5405#issuecomment-824696347

kevpar:
> Please get your patch merged in the master of `containerd/containerd` and then cherry-pick to v1.4. [containerd/containerd#5405 (comment)](https://github.com/containerd/containerd/issues/5405#issuecomment-824696347)

Thanks, I'll do that instead if it's preferred.

PR #1636, opened 2021-05-05
[release/1.4] windows: Use GetFinalPathNameByHandle for ResolveSymbolicLink

kevpar:
Backport of containerd/containerd#5411 for `release/1.4`. I also had to update some vendoring to bring in the right dependencies.

This change splits the definition of pkg/cri/os.ResolveSymbolicLink by
platform (windows/!windows), and switches to an alternate implementation
for Windows. This aims to fix the issue described in containerd/containerd#5405.

The previous implementation which just called filepath.EvalSymlinks has
historically had issues on Windows. One of these issues we were able to
fix in Go, but EvalSymlinks's behavior is not well specified on
Windows, and there could easily be more issues in the future, so it
seems prudent to move to a separate implementation for Windows.

The new implementation uses the Windows GetFinalPathNameByHandle API,
which takes a handle to an open file or directory and some flags, and
returns the "real" name for the object. See comments in the code for
details on the implementation.

I have tested this change with a variety of mounts and everything seems
to work as expected. Functions that make incorrect assumptions on what a
Windows path can look like may have some trouble with the \\?\ path
syntax. For instance EvalSymlinks fails when given a \\?\UNC\ path. For
this reason, the resolvePath implementation modifies the returned path
to translate to the more common form (\\?\UNC\server\share ->
\\server\share).

Signed-off-by: Kevin Parsons <kevpar@microsoft.com>

From containerd/containerd@b0d3b35


kevpar:
The CI failures look unrelated. Maybe just hitting rate limits on gcr.io? I think it's safe to ignore them.

estesp:
@mikebrow looks like `release/1.4` doesn't have the magic update to the CRI test image(s) stuff we did on the main branch. Getting the 403 authorization failure on the old image layer. Can you help us figure out which pieces need backported here?

kevpar:
@mikebrow any thoughts on what is needed here to help CI?

mikebrow:
> @mikebrow any thoughts on what is needed here to help CI?

yeah we have to figure out how to update the version of critest we are testing on in 1.4 because of the gcr changes

mikebrow:
https://github.com/containerd/containerd/blob/release/1.4/script/setup/install-critools#L24

^ that line of code needs to move to 

the version in this commit https://github.com/containerd/containerd/commit/da998c81ebbb520243e3da62e5a427a8a8abca30

or we need a downlevel version of critest that supports the k8s-staging-cri-tools test images

I can't remember if we had any breaking changes between us and critest over that time frame in critest.. I don't think so

so yeah try to move the commit up for critools/critest here and see what happens :-)

mikebrow:
> @mikebrow any thoughts on what is needed here to help CI?

> @mikebrow looks like `release/1.4` doesn't have the magic update to the CRI test image(s) stuff we did on the main branch. Getting the 403 authorization failure on the old image layer. Can you help us figure out which pieces need backported here?

thx almost forgot about this repo :-)

Let's merge this: https://github.com/containerd/cri/pull/1638 brings critest back on-line then you can rebase..

mikebrow:
kk you should be ok for critest.. after rebase on `release/1.4`

mikebrow:
FYI also updated vendor, and cleaned up the node e2e test over here ... https://github.com/containerd/cri/pull/1639  green lights!

fuweid:
@kevpar need rebase here~

kevpar:
@mikebrow thank you! I've rebased now.

pkg/os/os_windows_test.go
```diff
@@ -0,0 +1,258 @@
...
+		if err != nil {
+			return err
+		}
+		defer windows.CloseHandle(diskHandle)
+		h = diskHandle
+	}
+	// Formatting a disk directly in Windows is a pain, so we use FormatWritableLayerVhd to do it.
+	// It has a side effect of creating a sandbox directory on the formatted volume, but it's safe
+	// to just ignore that for our purposes here.
+	if err := computestorage.FormatWritableLayerVhd(context.Background(), h); err != nil {
```
mikebrow: nit... you can just return the result of this call :-)

pkg/os/os_windows.go
```diff
@@ -0,0 +1,180 @@
+/*
```
mikebrow, line 3: // +build windows

pkg/os/os_windows_test.go
```diff
@@ -0,0 +1,258 @@
+/*
```
mikebrow, line 3: // +build windows

mikebrow (commented):
just a couple nits

kevpar:
@mikebrow de-nit'd :)

pkg/os/os_unix.go
```diff
@@ -57,3 +60,15 @@ func Unmount(target string) error {
...
+func (RealOS) ResolveSymbolicLink(path string) (string, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		return "", err
+	}
+	if info.Mode()&os.ModeSymlink != os.ModeSymlink {
+		return path, nil
+	}
+	return filepath.EvalSymlinks(path)
+}
```
mikebrow: no newline at end of file.. :-)

mikebrow (approved):
LGTM
just a new line missing

pkg/os/os_windows.go
```diff
@@ -0,0 +1,182 @@
...
+// It works for both file and directory paths.
+//
+// We are not able to use builtin Go functionality for opening a directory path:
+// - os.Open on a directory returns a os.File where Fd() is a search handle from FindFirstFile.
+// - syscall.Open does not provide a way to specify FILE_FLAG_BACKUP_SEMANTICS, which is needed to
+//   open a directory.
+// We could use os.Open if the path is a file, but it's easier to just use the same code for both.
+// Therefore, we call windows.CreateFile directly.
+func openPath(path string) (windows.Handle, error) {
+	u16, err := windows.UTF16PtrFromString(path)
```
fuweid, line 40: ~~Is it possible to use `syscall.UTF16PtrFromString`? I see stdlib provides the function. https://golang.org/pkg/syscall/?GOOS=windows#UTF16PtrFromString.~~ 

`GetFinalPathNameByHandle` is unavailable in stdlib syscall.

pkg/os/os_windows.go
```diff
@@ -0,0 +1,182 @@
...
+
+	cVOLUME_NAME_DOS  = 0x0
+	cVOLUME_NAME_GUID = 0x1
+)
+
+var pool = sync.Pool{
+	New: func() interface{} {
+		// Size of buffer chosen somewhat arbitrarily to accommodate a large number of path strings.
+		// MAX_PATH (260) + size of volume GUID prefix (49) + null terminator = 310.
+		b := make([]uint16, 310)
```
fuweid, line 75: Suggest to add docs here https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd.

`Starting in Windows 10, version 1607, MAX_PATH limitations have been removed from common Win32 file and directory functions.`

fuweid: Maybe we can use `4096` :)

pkg/os/os_windows.go
```diff
@@ -0,0 +1,182 @@
...
+		return &b
+	},
+}
+
+// getFinalPathNameByHandle facilitates calling the Windows API GetFinalPathNameByHandle
+// with the given handle and flags. It transparently takes care of creating a buffer of the
+// correct size for the call.
+func getFinalPathNameByHandle(h windows.Handle, flags uint32) (string, error) {
+	b := *(pool.Get().(*[]uint16))
+	defer func() { pool.Put(&b) }()
```
fuweid, line 85: The `b` has been re-allocated. `defer` should be handled in some condition.


```go
--> in for loop
// If the buffer wasn't large enough, n will be the total size needed (including null terminator).		
// Resize and try again.		
if n > uint32(len(b) {
     b = make([]uint16, n)
     continue		
}
```

pkg/os/os_windows.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+	return h, nil
+}
+
+// GetFinalPathNameByHandle flags.
+//nolint:golint
+const (
+	cFILE_NAME_OPENED = 0x8
+
```
fuweid, line 66: remove the empty line.

fuweid (commented):
Left some comments~

pkg/os/os_windows.go
kevpar: Given this is a backport (of https://github.com/containerd/containerd/commit/b0d3b35), my assumption is we shouldn't diverge from what was put into containerd master. For this and the other comments you have left, do you feel these should be addressed here?

pkg/os/os_windows.go
kevpar: I'm not sure I understand your comment. Can you clarify what you're asking for?

pkg/os/os_windows.go
kevpar: I was trying to keep the `FILE_NAME` and `VOLUME_NAME` constants separate, since they are used for different arguments.

pkg/os/os_windows.go
fuweid: Sorry for the late reply. A `b` is from `pool.Get` so that we should put the `b` back into the `pool`. 
But the `b` has been changed when the size of original `b` is not large enough. In this case, it is not good to put the changed `b` into the `pool`.

pkg/os/os_windows.go
fuweid: I am fine with current change ~ just think 4096 looks better than 310

fuweid (approved):
LGTM

kevpar:
Thanks @fuweid!

